PR #67, opened 2018-09-12
Add macOS lwjgl dependencies

danielcompton:
Not sure if you want to sully your codebase with macOS support, but I was able to make the tests pass on macOS after adding these.


malcolmsparks:
Great - thanks Daniel!

PR #186, opened 2019-04-20
Update config examples to use valid ns alias

jorinvo:
Thanks for sharing this awesome project with the public!

Just came across a small typo in the documentation.

Looking forward to exploring Crux further :)


refset:
Thank you :)

PR #192, opened 2019-04-24
Use version from GitHub tag

jonpither:


hraberg (approved):
Checked with @megakorre about why there are two plugins now, and did ask if we couldn't simply print the version in the snippet that computes it necessary instead of depending on both plugins? Otherwise the print plugin could potentially be moved to the uberjar profile?

PR #193, opened 2019-04-24
184 change evict kafka modeling

megakorre:


hraberg (changes requested):
Looks good, matches what we discussed. Let's use this as an opportunity to get some tests in this area as well, both to check the actual replay doesn't block and that the tombstones don't interfere with the other parts, like the `entiy` API for example.

In theory `delete-objects` could now be removed from the `ObjectStore` unless I miss something? But guess there's no real harm in keeping it, assuming nothing unexpected uses it.

Of course, the lack of tests around eviction is on me, sorry about that.

megakorre:
pushed a test and code changes for bugs that the test detected

hraberg (approved):
Looks good, thanks a lot for the test.

Issue #194, opened 2019-04-24
Consider removing lein-project-version


jonpither:
Minor issue: since, #190, if circle can pull out the recent git tag, we don't need this plugin dependency.

jonpither:
Fixed by #195

PR #195, opened 2019-04-24
Remove plugins - #194

jonpither:


tasks/leiningen/project_version.clj
```diff
@@ -0,0 +1,20 @@
...
+  (:require [clojure.java.shell :as sh]))
+
+(defn project-version [project & args]
+  (println (:version project)))
+
+(defn- version-from-git []
+  (let [{:keys [exit out]} (sh/sh "git" "describe" "--tags" "--dirty" "--long")]
+    (assert (= 0 exit))
+    ;;"19.04-1.0.2-alpha-22-g9452e91b-dirty"
+    (let [[_ tag ahead sha dirty] (re-find #"(.*)\-(\d+)\-([0-9a-z]*)(\-dirty)?$" (.trim out))
```
hraberg: Have a strong feeling this will be a reflection warning for `.trim` here? Can hint out as `^String` or use `clojure.string/trim` to indirectly achieve the same thing.

hraberg (changes requested):
Nicer than than two random plugins. Assume its tested and good to go, as soon as the (assumed) reflection warning is fixed.

hraberg (approved):

PR #197, opened 2019-04-25
Fixes #173 API call for full-results?

### LINKED ISSUE #173: Support Full Query Results
A richer result is currently hidden in the meta of the result tuples. We should add a flag to the query map to optionally return these directly instead of the tuples. This flag could either be a boolean or an enum potentially.

This is useful when one wants to work directly with the documents, and can avoid having to pull out a lot of attributes one by one when one isn't actually filtering on them. This is already there in the meta data, but that's slightly obscure, and more importantly, doesn't work over HTTP, where you're likely to want this, as it can minimise chattiness.

jonpither:


test/crux/api_test.clj
```diff
@@ -96,6 +96,17 @@
...
 
+        (t/testing "query returning full results"
+          (let [db (.db f/*api*)]
+            (with-open [snapshot (.newSnapshot db)]
+              (let [result (.q db snapshot '{:find [e]
+                                             :where [[e :name "Ivan"]]
+                                             :full-results? true})]
+                (t/is (instance? LazySeq result))
+                (t/is (not (realized? result)))
+                (t/is (= '([{:var e, :value :ivan, :doc {:crux.db/id :ivan, :name "Ivan"}}]) result))
```
hraberg: For consistency with other APIs we probably want name spaced keys here, like `crux.query/var` or similar.

hraberg (changes requested):
Otherwise looks good. Can see the case for simpler keys as well.

hraberg (approved):
Cool, might prefer full keys rather than relying on double-colon in case someone moves the function to another namespace, but there's a test, so let's merge.

jonpither:
Fair enough @hraberg, will tidy this up in a subsequent commit

PR #203, opened 2019-04-30
Fixes #202 Enforce java.util.Date.

### LINKED ISSUE #202: Ensure Transaction Spec Checks for Date not inst?
We should avoid assuming Dates where we accept the `#inst` tag. This might be far reaching, but ultimately a fix is trivial.

Additionally it would useful to understand and document/avoid workflows that could lead to getting an `Instant` instead of a `Date`, as demonstrated in the screenshot below (@hraberg's guess is using Java 8 `java.time` "as default"):
image

cc @arrdem - thanks again for reporting

jonpither:


hraberg (approved):
Looks good as discussed.

PR #204, opened 2019-05-01
180 remove delete operation on evict

megakorre:
based on branch in https://github.com/juxt/crux/pull/193 to run against the latest evict tests. Will rebase against master after its merged


hraberg (approved):
As there's a new test in the other PR, I'm assuming this is indirectly covered and no unexpected behaviour arises from the removal of the tx log entries.

PR #221, opened 2019-05-14
Fixes #178 Initial work on exposing statistics

### LINKED ISSUE #178: Expose Attribute Statistics Map in API
They can be fetched like this, `(db/read-meta kv :crux.kv/stats)`. This will be a map of attribute to frequency. The frequency is the total across versions and cardinality many etc, so it cannot easily be used to deduce the number of entities or documents with an attribute directly.

jonpither:


src/crux/api.clj
```diff
@@ -178,6 +178,10 @@
     from and including the valid time of the db while respecting
     transaction time. Includes the documents.")
 
+  (stats
```
hraberg: Is this a leftover perhaps? Feels like the stats makes sense on the system level, not on the db.

src/crux/api/ICruxAPI.java
```diff
@@ -151,4 +151,11 @@
      * @return                 a lazy sequence of the transaction log.
      */
     public Iterable<List<?>> txLog(Closeable txLogContext, Long fromTxId, boolean withDocuments);
+
+    /**
+     * Return frequencies of indexed attributes.
+     *
+     * @return         Map containing Keyword freqencies.
+     */
+    public Map<?,?> stats();
```
hraberg: Guess this can be Keyword/Integer (or Long)?
What about calling this `attributeStats` or something similar?

hraberg (changes requested):

hraberg (approved):
Assuming sleep will go after #222.

PR #224, opened 2019-05-15
222 chaching key normalization error

megakorre:


megakorre:
based this on the branch of https://github.com/juxt/crux/pull/221 since the tests written there are helpful to test this error. This PR can wait until that is completed

hraberg (commented):
(Needs conflict resolving and ensure the wrong stats functions resurface.)
I'm not sure why we need `c/->id-buffer` for the cache keys, but assume the hint is in the PR name?

What does the lazy seqs which interleaves evictions with their actual underlying put/delete aim to do? In reality they might all be done in a tx below, so interleaving them might not matter, and would be making assumptions. What about chunking etc?

hraberg (approved):

PR #238, opened 2019-06-05
Content-hash not allowed as argument for put

johantonelli:


hraberg (changes requested):
Unless I'm mistaken, just removing the support for supplying the content hash will break on the ingest side as this assumes this has already been turned into a content hash when put on the wire? Do all tests still pass with this change?

src/crux/tx.clj
```diff
@@ -21,9 +21,8 @@
 
 (s/def ::id (s/conformer (comp str c/new-id)))
 (s/def :crux.db/id (s/and (complement string?) c/valid-id?))
-(s/def ::doc (s/and (s/or :doc (s/and (s/keys :req [:crux.db/id]) ::id)
-                          :content-hash (s/and (complement map?) ::id))
-                    (s/conformer second)))
+
+(s/def ::doc (s/and (s/keys :req [:crux.db/id]) ::id))
```
hraberg, line 25: This `s/and` can be removed.

johantonelli: Taking this out brakes it so I think the `s\and` is needed.

johantonelli:
All tests still pass with this change and I believe the `s/and` is required as there are two arguments.

src/crux/tx.clj
hraberg: Yes, sorry, my mistake, the `s/and` is needed, confused it with the original `s/or`. The `::id` spec also calls `s/conformer` so the message should still contain the hash, so this change might actually be good. But worth sanity checking this first by printing out the message going onto the wire before we merge.

hraberg: Also, checked, and there's currently no spec on the ingest side. This is a problem, but not necessarily part of this PR as that issue was already there, up to you and @jonpither.

hraberg: Actually, instead of just printing it out, you could simply add this to your test and see that the conformed message is indeed containing the hash and not the doc, as all data is already setup there to do this.

test/crux/api_test.clj
```diff
@@ -17,6 +17,13 @@
 
 (declare execute-sparql)
 
+(t/deftest test-content-hash-invalid
+  (let [valid-time (Date.)
+        content-ivan {:crux.db/id :ivan :name "Ivan"}
+        content-hash (c/new-id content-ivan)]
```
hraberg: See above, would be nice with a sanity check assertion that the confromed message contains the hash.

johantonelli:
Currently changing name of test as well as addressing the stacktrace that the test generates in `lein test` before merging.

PR #239, opened 2019-06-05
gh236/lru : fix key pooling

spacegangster:
tests are passing

### DIFF
--- a/src/crux/lru.clj
+++ b/src/crux/lru.clj
@@ -75,15 +75,14 @@
   (get-single-object [this snapshot k]
     (compute-if-absent
      cache
-     (c/->id-buffer k)
+     ; todo a fix without needless key copies (see https://github.com/juxt/crux/issues/236)
+     (-> k c/new-id c/safe-id c/->id-buffer)
      #(db/get-single-object object-store snapshot %)))
+
 
   (get-objects [this snapshot ks]
     (->> (for [k ks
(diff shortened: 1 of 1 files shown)


hraberg (changes requested):
I mean you can replace the entire `compute-if-absent` in `get-objects` with a call to `get-single-object`. No need for Java interop, you can use `db/get-single-object`.

spacegangster:
@hraberg comment is addressed now.
Tests are good on openjdk-11 and hotspot-10 and there's even *a* pass on hotspot-8

spacegangster:
@hraberg  I ran watdiv suite (is it the right one?) once on master branch and twice with our pooled buffers fix. I don’t see much variability.
On master
```
"Elapsed time: 268540.987686 msecs"
"Elapsed time: 117.944926 msecs"
```
On the branch
```
"Elapsed time: 279184.388337 msecs"
"Elapsed time: 107.121264 msecs"
...
"Elapsed time: 258216.086621 msecs"
"Elapsed time: 165.383238 msecs"
```

hraberg:
The WatDiv suite requires a environment variable to be set to run, the full suite takes a very long time (hours) to run, I usually run a subset. But as mentioned elsewhere, for this case you just need something that should show the difference, so almost any non trivial test might do.

hraberg (approved):

Issue #241, opened 2019-06-07
Events are not being checked against schema 


jonpither:
When transaction events are ingested off topics, they are not currently being checked against any specs.

Issue #208, opened 2019-05-10
Remove need for specifying ID in put and cas operations


jonpither:
This is a proposed ergonomic change - what do we think of removing the absolute need to supply the ID as an argument to the put and cas operations, when the ID can be easily derived from the supplied documents (currently an error is thrown is the supplied ID differs).

I believe this question has come up, and subjectively, it feels more intuitive to just hand Crux documents, rather than giving it IDs as well. There is also a little less duplication/boilerplate to getting on and using Crux.

Downside of this is a wider API, with overloaded function arguments (as we'd probably want to keep the existing case where you can supply the ID - although worth questioning this also?)

crimeminister:
I'm new to crux but for what it's worth I think DRYing up `put` and `cas` so that a document ID is only supplied once would be friendlier. I'm relatively agnostic as to whether the ID lives in the document or outside, but would appreciate if there was only the one way to do it. Otherwise I'm liable to start writing macros to ensure that I'm not using it one way in some places and the other way(s) in others. There's no way that ends well. 😄

hraberg:
One reason it works the way it does is to stay consistent with the arguments also for `delete` and `evict`, that is, the first argument to a transaction command is always an entity id. Not saying that we need to keep it like that, just providing some background.

PR #243, opened 2019-06-07
Split out event spec

jonpither:
Paves the way for #241 and #208.

### DIFF
--- a/src/crux/kafka.clj
+++ b/src/crux/kafka.clj
@@ -105,9 +105,9 @@
 
   (submit-tx [this tx-ops]
     (try
-      (let [conformed-tx-ops (tx/conform-tx-ops tx-ops)
-            tx-events (crux.tx.event/conform-tx-events conformed-tx-ops)
-            content-hash->doc (->> (for [doc (tx/tx-ops->docs conformed-tx-ops)]
+      (s/assert :crux.tx/tx-ops tx-ops)
+      (let [tx-events (crux.tx.event/tx-ops->tx-events tx-ops)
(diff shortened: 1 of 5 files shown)


jonpither:
A potential inconsistency is that users can submit txes for the same op using variable vector lengths, i.e. with `cas` you could omit the last argument `at-valid-time`, so the `cas` tx-op can be different lengths. 

Also, in `cas` the `old-doc` can be nillable - so sometimes a `nil` argument is used, and at other times the omission of the argument is used, for an optional argument. 

My main concern is consistency of argument nils vs absence, and tx lengths, from the perspective of reading the txes back out from the `tx-log` api.

If you wanted to make the vector length consistent, the fix would be to either 1) have a stricter API contract, 2) introduce a measure of conforming prior to txes going on the wire of padding out the tx-op with nils. 

This doesn't feel an important issue. but I'd favour the second option above. From a tx submission perspective, having variable length contracts has the measure of future-proofing, should we add more optional args to 'cas' for example.

jonpither:
On a different note, taking @hraberg your point about potentially raising up the API specs. Perhaps a good opening step is to move the `crux.tx/tx-op` and `crux.tx/tx-ops` to the `crux.api` namespace at the start of the file. Some renaming would be necessary, so `crux.tx/cas` would need to be either `crux.api/tx-cas` or `crux.api.tx/cas`. 

Since the refactoring in this PR, the above refactor ought be straightforward, as the spec in question is more purely about the data coming into the API.

hraberg:
Just want to point out that the `nil` or 'cas' isn't optional, but represent that there should be no existing version.

PR #246, opened 2019-06-09
Prevent crash when an id is a map

### LINKED ISSUE #245: Crash when using maps as IDs
Currently, maps are allowed as IDs, but `crux.tx/tx-ops->docs` will treat them as documents and throw an exception after failing to retrieve an id from them.

Example code which exhibits this behavior:

```clojure
(let [% {:crux.db/id {:user "Xwop1A7Xog4nD6AfhZaPgg"}, :name "Adam"}] 
  (tx-ops->docs [[:crux.tx/put (:crux.db/id %) % #inst "2019"]]))
;; IllegalArgumentException: Document's id does not match the operation id:  {:user "Xwop1A7Xog4nD6AfhZaPgg"}
```

AdamNiederer:
Resolves #245

### DIFF
--- a/src/crux/tx.clj
+++ b/src/crux/tx.clj
@@ -218,8 +218,8 @@
   (s/conform ::tx-ops tx-ops))
 
 (defn tx-ops->docs [tx-ops]
-  (vec (for [[op id :as tx-op] tx-ops
-             doc (filter map? tx-op)]
+  (vec (for [[op id & rest :as tx-op] tx-ops
+             doc (filter map? rest)]
          (if (and (satisfies? c/IdToBuffer id)
                   (= (c/new-id id) (c/new-id (get doc :crux.db/id))))
            doc


hraberg (commented):
Thanks for this @AdamNiederer. We'll have a think, as I mentioned on the issue the solution might be to remove support for maps as ids. We should hopefully get back to you here early next week after we've discussed it.

There's related work going on in #243 which @jonpither is doing to split the spec for submit vs. ingest, in which case we would already know if the argument should be a map or a hash, assuming we don't support map ids.

But if we would go down the path you started here, there are also `cas` operations which take two ids/maps as their arguments, which this PR would also need to take into account.

refset:
It looks like @jonpither already made pretty much the exact same change in a parallel commit just after this PR was forked: https://github.com/juxt/crux/commit/e0d12982871a5f1e9c5792e3c878578fef46b871#diff-4c615cb37646e08c75b2ec1e5a0b9f64R212

refset:
@AdamNiederer thank you for the feedback and input in any case :)

jonpither:
Reopening - I want to go through this PR, I can do now I've got my PR out the way :-)

jonpither:
Thanks for this PR. As @refset mentioned, it's been subsumed in a parallel PR. Thanks though - we've got a test around this now and it's good to be aware of the feature, of providing maps as IDs.

PR #255, opened 2019-06-13
gh236/lru – improve pooled buffers issue solution

spacegangster:


test/crux/fixtures.clj
```diff
@@ -1,28 +1,45 @@
...
+            [crux.memory :as mem])
+  (:import java.io.Closeable
            [java.util Properties UUID]
            org.apache.kafka.clients.admin.AdminClient
+           org.apache.kafka.clients.producer.KafkaProducer
            org.apache.kafka.clients.consumer.KafkaConsumer
-           org.apache.kafka.clients.producer.KafkaProducer))
+           [crux.api Crux ICruxAPI]))
+
+(defn random-person [] {:crux.db/id (UUID/randomUUID)
```
jonpither: I moved this people stuff to the bottom of this ns, make sure we don't duplicate

jonpither (changes requested):

src/crux/lru.clj
```diff
@@ -26,23 +27,20 @@
...
-          (let [v (f k)
-                stamp (.writeLock lock)]
-            (try
-              (.computeIfAbsent cache k (reify Function
-                                          (apply [_ k]
-                                            v)))
-              (finally
-                (.unlock lock stamp))))))
+      (compute-if-absent [_ k stored-key-fn f]
+        (let [v (.getOrDefault cache k ::not-found)] ; use ::not-found as values can be falsy
```
hraberg: Actually, this read also needs to use a lock I realised, sorry. Easiest way is to just delegate to `(get this k ::not-found)` which already does this, see the implementation of `ILookup` a few lines below. Calling `.valAt` explicitly (instead of `get`) would likely be faster.

hraberg (changes requested):

test/crux/fixtures.clj
spacegangster: sorry, I meant to see the diff here and mend it

hraberg (approved):

jonpither (approved):

PR #256, opened 2019-06-13
Fixes #254 move blocking call in db to sync

### LINKED ISSUE #254: Move blocking `db` call to sync
Currently calling `api/db` with the `transaction-time` argument blocks until the DB sees a transaction that is past this time. 

This issue is about potentially moving this blocking behavour to `sync` where `sync` could take a new `transaction-time` as a 3rd argument. 

The motivation for this is for aesthetics / intuitive API. `sync` is more where you'd expect the blocking behavour, `db` probably you wouldn't expect it there.

jonpither:
I will review the docstrings etc, but wanted to set out this PR to show the direction, and get feedback.

### DIFF
--- a/src/crux/api.clj
+++ b/src/crux/api.clj
@@ -99,12 +99,16 @@
     eid is an object that can be coerced into an entity id.
     Returns true if the entity was updated in this transaction.")
 
-  (sync [system ^Duration timeout]
-    "Blocks until the node has caught up indexing. Will throw an
-    exception on timeout. The returned date is the latest index
-    time when this node has caught up as of this call. This can be
-    used as the second parameter in (db valid-time, transaction-time)
-    for consistent reads.
+  (sync
+    [system ^Duration timeout]
+    [system ^Date transaction-time ^Duration timeout]
+    "If the transaction-time is supplied, blocks until indexing has
+    processed a tx with a greater-than transaction-time, otherwise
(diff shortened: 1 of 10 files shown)


hraberg (commented):
Looks good, but two questions:

1. We discussed renaming the method, did we decide against, or did we forget about it?
2. I feel that timeout should be the last argument, for reasons I cannot fully motivate.

jonpither:
Agree with 2.

I can't remember what we finalised on with 1. I just decided to make the change, in the view that's easy to finalise the details after, so I'm happy to rename the method. I've just lost my strong opinion on it :-)

hraberg:
I don't mind the name as such, but usually want to avoid things clashing with `clojure.core`. That said, once someone has an opinion they can state it and use query replace, so we can park that.

hraberg (changes requested):
We want to change argument order from duration, date to date, duration.

hraberg:
Also, to state the obvious, the name of this function is very old and not really related to this issue itself.

jonpither:
I think we'll stick with `sync` for now - we can always change in a subsequent release whilst we're in alpha.

Besides @malcolmsparks makes an eloquent case and impassioned defence for retaining `sync`.

jonpither:
I changed the arg order, good for merging, assuming we can rename `sync` at a later date (and we should discuss again).

PR #257, opened 2019-06-13
Fixes #208 Remove ID from put and cas outside of doc

### LINKED ISSUE #208: Remove need for specifying ID in put and cas operations

johantonelli:
I have also skimmed through the docs and removed mention of needing two instances of the ID.

Note that `put` and `cas` can no longer take an ID outside of the doc. `evict` and `delete` remain unchanged.

### DIFF
--- a/bench/src/crux_bench/watdiv.clj
+++ b/bench/src/crux_bench/watdiv.clj
@@ -402,7 +402,6 @@
               (crux/submit-tx
                 crux
                 [[:crux.tx/put
-                  ::watdiv-ingestion-status
                   {:crux.db/id ::watdiv-ingestion-status :done? false}]])]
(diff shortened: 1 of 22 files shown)


jonpither:
I've reviewed. I'm happy to merge this - having discussed with @hraberg prior.

PR #269, opened 2019-06-24
 DISCUSS ONLY. tests : add a standalone system bootstrap test

spacegangster:
This adds two tests for the standalone system init.
First of them passes, second fails with a weird error
```
ERROR in (test-api-standalone-system) (alpha.clj:1966)
expected: (api/start-standalone-system {:kv-backend "crux.kv.memdb.MemKv", :db-dir "data/db-dir-1"})
  actual: clojure.lang.ExceptionInfo: Spec assertion failed
nil - failed: string? in: [:db-dir] at: [:db-dir] spec: :crux.kv/db-dir
```

### DIFF
--- a/test/crux/bootstrap_standalone_test.clj
+++ b/test/crux/bootstrap_standalone_test.clj
@@ -1,0 +1,16 @@
+(ns crux.bootstrap-standalone-test
+  (:require [clojure.test :as t]
+            [crux.api :as api]))
+
+(t/deftest test-api-standalone-system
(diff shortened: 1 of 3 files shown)


test/crux/bootstrap_standalone_test.clj
```diff
@@ -0,0 +1,17 @@
+(ns crux.bootstrap-standalone-test
+  (:require [clojure.test :as t]
+            [crux.api :as api]))
+
```
jonpither, line 4: double spacing ..

spacegangster:
will be fixed by #271

Issue #247, opened 2019-06-09
Assertion failure when using `:args`


SevereOverfl0w:
```
(crux/q (crux/db system)
          '{:find [?x]
            :where [[?x :name]]
            :args [{:?x "Clojure"}]})
```

results in 

```
Syntax error (AssertionError) compiling at (core.clj:53:3).                                                                                                                                            
Assert failed: 04456e716c77746701                                                                                                                                                                      
(or (= id-size (.capacity entity)) (zero? (.capacity entity)))    
```

here's the pst:

```
Note: The following stack trace applies to the reader or compiler, your code was not executed.                                                                                                         
CompilerException Syntax error compiling at (io/dominic/koni/core.clj:53:3). #:clojure.error{:phase :compile-syntax-check, :line 53, :column 3, :source "io/dominic/koni/core.clj"}                    
        clojure.lang.Compiler.load (Compiler.java:7647)                                                                                                                                                
        user/eval66191 (NO_SOURCE_FILE:1)                                                                                                                                                              
        user/eval66191 (NO_SOURCE_FILE:1)                                                                                                                                                              
        clojure.lang.Compiler.eval (Compiler.java:7176)                                                                                                                                                
        clojure.lang.Compiler.eval (Compiler.java:7131)                                                                                                                                                
        clojure.core/eval (core.clj:3214)                                                                                                                                                              
        clojure.core/eval (core.clj:3210)                                                                                                                                                              
        clojure.main/repl/read-eval-print--9068/fn--9071 (main.clj:414)                                                                                                                                
        clojure.main/repl/read-eval-print--9068 (main.clj:414)                                                                                                                                         
        clojure.main/repl/fn--9077 (main.clj:435)                                                                                                                                                      
        clojure.main/repl (main.clj:435)                                                                                                                                                               
        clojure.main/repl (main.clj:345)                                                                                                                                                               
Caused by:                                                                                                                                                                                             
AssertionError Assert failed: 04456e716c77746701                                                                                                                                                       
(or (= id-size (.capacity entity)) (zero? (.capacity entity)))                                                                                                                                         
        crux.codec/encode-attribute+entity+content-hash+value-key-to (codec.clj:464)                                                                                                                   
        crux.codec/encode-attribute+entity+content-hash+value-key-to (codec.clj:453)                                                                                                                   
        crux.codec/encode-attribute+entity+content-hash+value-key-to (codec.clj:457)                                                                                                                   
[...]

PR #282, opened 2019-07-02
Update docs to use symbols in :args, relates to #247

johantonelli:
Updated the docs examples to use symbol instead of keywords in the interest of clearer examples.

### DIFF
--- a/docs/queries.adoc
+++ b/docs/queries.adoc
@@ -339,12 +339,12 @@ control the result order.
 
 [source,clj]
 ----
-{:find [time device-id temperature humidity]
- :where [[c :condition/time time]
-         [c :condition/device-id device-id]
-         [c :condition/temperature temperature]
-         [c :condition/humidity humidity]]
- :order-by [[time :desc] [device-id :asc]]}
+'{:find [time device-id temperature humidity]
+  :where [[c :condition/time time]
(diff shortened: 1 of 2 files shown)


docs/queries.adoc
```diff
@@ -382,7 +382,7 @@ queries to apply range filters using predicates.
          [c :condition/humidity humidity]]
  :order-by [[device-id :asc]]
  :limit 10
- :args [{:my-offset 990}]}
+ :args [{'my-offset 990}]}
```
refset, line 385: We should quote the outer vecs for `:where` `:find` and `:order-by` as well.

docs/queries.adoc
```diff
@@ -401,7 +401,7 @@ connected to a given entity via the `:follow` attribute.
 ----
 {:find [?e2]
  :where [(follow ?e1 ?e2)]
- :args [{:?e1 :1}]
+ :args [{'?e1 :1}]
```
refset: I don't think we need to quote this particular instance at all.

refset (changes requested):
@Antonelli712 a couple of small changes but otherwise looks good to go :+1:

johantonelli:
Changes made, and updated all quotes for continuity.

PR #283, opened 2019-07-02
Jp/pull out dev

jonpither:


hraberg (approved):
Looks good as discussed, there's a checked in pom.xml in crux-test, does it need to be there? Merging, but feel free to remove this after if not.

PR #290, opened 2019-07-04
Fixed broken link to webservice example

maacl:


refset:
Ah yes, we are still tidying up after the big reorganisation...thanks for helping! :)

PR #294, opened 2019-07-09
Add exclusions to /crux-dev/project.clj to resolve pedantic abort whe…

refset:
…n running 'lein repl'


hraberg (approved):
Unsure if this needs to be done this way, cannot reproduce it locally right now, might need to build all dependencies again to see it. Regardless, this should be harmless and it's only for `crux-dev` so merging for now to not hold up the release.

PR #298, opened 2019-07-11
Remove standalone system boilerplate

jonpither:


crux-core/src/crux/bootstrap.clj
```diff
@@ -115,13 +115,24 @@
...
+    (when (satisfies? tx-log backup/ISystemBackup)
+      (backup/write-checkpoint tx-log opts)))
 
   Closeable
   (close [_]
-    (some-> close-promise (deliver true))
-    (some-> node-thread (.join))))
+    (when close-fn (close-fn))))
+
+(defn start-crux-node [& bindings]
```
hraberg: Would prefer this to take a map and not do the last/partition stuff. Otherwise this PR looks good.

hraberg (changes requested):

crux-core/src/crux/bootstrap.clj
jonpither: Making a commit to address this.

PR #299, opened 2019-07-15
Introduce event-log consumer

jonpither:


hraberg (approved):
Would be interesting to see if this can evolve to take care of the Kafka case as well?

In theory I would like to split the generic part of Moberg from the tx log, but not end of the world, this kind of comes back to the idea of having the lower level protocol in place.

PR #300, opened 2019-07-15
Rename system->node

refset:
`lein build` passes all the way until `lein test crux.tx-test`.


crux-console/src/config.edn
```diff
@@ -1,8 +1,8 @@
-{:ig/system
- {:juxt.crux-ui.server.main/read-write {:system #ig/ref :juxt.crux.ig/system}
-  :juxt.crux-ui.server.main/home {:system #ig/ref :juxt.crux.ig/system}
-  :juxt.crux-ui.server.main/console {:system #ig/ref :juxt.crux.ig/system}
-  :juxt.crux.ig.system/standalone {:kv-backend "crux.kv.rocksdb.RocksKv"
+{:ig/node
```
jonpither: I'm not sure about this - you're building an integrant system? Does the console still work with this change? Actually, makes me think we need a test for the console that makes sure it can at least fire up.

refset: I changed `:ig/node` back to `:ig/system` and everything else seems to work okay. The console runs.

PR #308, opened 2019-07-24
WIP - Add test for uberjar

johantonelli:


hraberg (approved):
Looks good, various ways of doing this of course, but if it does the trick. Excellent thing to add a test for though.

There are wrappers for various String methods in `clojure.string` that might be a bit more idiomatic. But not an issue. Would sanity check that the `.contains` don't generate reflection warnings though.

PR #311, opened 2019-07-29
Support multiple nodes in crux-http-server via a node key at the root of URL paths

refset:
This now works by passing in a map of nodes (instead of a single node). The keys are used to direct requests according to the relevant URL paths (e.g. `/document/:myid` is now `/db1/document/:myid`, which is completely distinct from `/db2/document/:myid`).

A map of statuses for all nodes is now returned when sending a request to the root URL, which includes all of the known keys (so that nodes can be discovered).

### DIFF
--- a/crux-http-server/src/crux/http_server.clj
+++ b/crux-http-server/src/crux/http_server.clj
@@ -1,5 +1,5 @@
 (ns crux.http-server
-  "HTTP API for Crux.
+  "HTTP API for one or more Crux nodes.
 
   The optional SPARQL handler requires juxt.crux/rdf."
   (:require [clojure.edn :as edn]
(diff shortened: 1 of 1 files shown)


refset:
_**Context**_

This is desirable for the crux-console to be able to interact with a remote Kafka node at the same time as it's own local standalone node (which we would like to use for persisting query history and connection settings/auth)

Spinning up multiple endpoints on different ports is the current alternative, but that approach doesn't scale as easily and you may have to manage multiple port mappings in multiple places (proxies etc.)

Other use-cases this PR might help with include sharding and multi-tenancy.

hraberg:
Guess I'll get more context about this soon, but looks like your routing regexp would assume `//` as the second slash (or rather, the entire group of the node name + slash) isn't optional?

hraberg:
I'm tempted to close this, it's old, won't merge and the approach might not be the right one, as mentioned, one can use ports for the cases where this actually is an issue?

hraberg:
Actually, let me close this, can be reopened if necessary, just to clean out the PR list.

PR #313, opened 2019-08-01
Fix http-server CORS headers for 400s/500s requests

spacegangster:
This preserves CORS headers on requests resulted in exception

### DIFF
--- a/crux-http-server/src/crux/http_server.clj
+++ b/crux-http-server/src/crux/http_server.clj
@@ -321,9 +321,11 @@
               :or {server-port 3000 cors-access-control []}
               :as options}]
   (s/assert ::options options)
-  (let [server (j/run-jetty (-> (partial handler crux-node)
-                                (#(apply wrap-cors (into [%] cors-access-control)))
+  (let [wrap-cors' #(apply wrap-cors (cons % cors-access-control))
+        server (j/run-jetty (-> (partial handler crux-node)
+                                (wrap-exception-handling)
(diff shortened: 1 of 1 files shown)


crux-http-server/src/crux/http_server.clj
```diff
@@ -321,9 +321,11 @@
...
               :as options}]
   (s/assert ::options options)
-  (let [server (j/run-jetty (-> (partial handler crux-node)
-                                (#(apply wrap-cors (into [%] cors-access-control)))
+  (let [wrap-cors' #(apply wrap-cors (cons % cors-access-control))
+        server (j/run-jetty (-> (partial handler crux-node)
+                                (wrap-exception-handling)
                                 (p/wrap-params)
+                                (wrap-cors')
                                 (wrap-exception-handling))
```
spacegangster, line 329: This is a double duct-tape with `wrap-exception-handling`, but not a strong opinion here. Shouldn't be any perf hit though, so why not.

jonpither: hmm, should have one at the end no?

jonpither (changes requested):

crux-http-server/src/crux/http_server.clj
spacegangster: I added `(wrap-exception-handling)` before `(wrap-cors')`, otherwise exception responses don't get `Access-Control-Allow-Origin` header and get screwed by browser.

PR #320, opened 2019-08-13
Transaction Functions

hraberg:
See xtdb/xtdb#2019. This is very much a work in progress and can be seen as a conversation starter.


crux-core/src/crux/tx.clj
```diff
@@ -111,14 +112,67 @@
...
+(defn tx-command-fn [indexer kv object-store snapshot tx-log [op k args-v] transact-time tx-id]
+  (let [fn-id (c/new-id k)
+        db (q/db kv)
+        {:crux.db.fn/keys [body] :as fn-doc} (q/entity db fn-id)
+        {:crux.db.fn/keys [args] :as args-doc} (db/get-single-object object-store snapshot (c/new-id args-v))
+        args-id (:crux.db/id args-doc)
+        fn-result (try
+                    (let [tx-ops (apply (tx-fn-eval-cache body) db (eval args))]
+                      {:docs (tx-ops->docs tx-ops)
+                       :ops-result (vec (for [[op :as tx-op]  (tx-ops->tx-events tx-ops)]
```
refset: It looks like this allows recursive transaction function evaluation (which is great) -- do you think it needs a test to make sure?

hraberg: It is meant to handle functions returning transactions containing other functions. That said, it's not tested, agree a test would be good, did almost add one but it became a bit more complex than I felt was warranted for the basic functionality so backed away. Might have a new stab.

As we discussed, as the effects of the current transaction (as a whole) isn't visible until the end of it, complicated nested logic becomes a bit less useful, as you cannot see the effects of the earlier operations.

We could explore adding proper transaction support to the KV protocol so we can write as we go and still rollback, but that adds complexity of course. The current workaround is to submit several smaller transactions that build on each other.

hraberg: Added support for this and a simple test. It's not optimal and like other things complicate and has corner cases when it comes to eviction. See updated commit.

PR #327, opened 2019-08-27
Add bulk documents endpoints

spacegangster:
Adds `documents` for ICrux interface and `/documents` to HTTPServer


crux-core/src/crux/api/ICruxAPI.java
```diff
@@ -49,6 +49,16 @@
...
 
+    /**
+     *  Reads a document from the document store based on its
+     *  content hash.
+     *
+     * @param contentHashSet a set of objects that can be coerced into a content
+     * hashes.
+     * @return            a map from hashable objects to the corresponding documents.
+     */
+    public Map<String,?> documents(Object contentHashSet);
```
hraberg: As you are keen on keeping this a set, this should be `Set<?>` I think. That said, I would be fine relaxing this across the board to `Collection` - while a set is conceptually correct, it's not strictly necessary.

crux-http-server/src/crux/http_server.clj
```diff
@@ -260,6 +265,9 @@
     [#"^/document/.+$" [:get :post]]
     (document crux-node request)
 
+    [#"^/documents" [:post]]
```
hraberg, line 269: I'm not sure we should bother right now, but as briefly discussed, a set of documents is a special case, like `/document` with filter, and `/documents/abcd` which is a specific resource. There are various opinions about how to model this in REST idiomatically.

hraberg (approved):
I have some minor comments and reservations, but this should more or less be good to go, up to you.

spacegangster:
merged manually with squash

Issue #305, opened 2019-07-19
Make config nested sub-maps 


jonpither:
To marry up with the modules approach. This is instead of having a large, flattened map.

As part of this, also experiment to see if the API could work with a JSON config map - this may better suit people arriving at the Java API.

refset:
Also, it would be helpful to expose more of the provided and interpreted config information in the `status` API, perhaps including the versions of each module.

And resolve this while we're at it:
`Reflection warning, crux/bootstrap.clj:25:33 - reference to field trim can't be resolved.`

jonpither:
Am closing this issue in light of #340 . See comment https://github.com/juxt/crux/pull/340#issuecomment-543591512 for more information.

PR #340, opened 2019-09-26
Introduce Topologies

jonpither:
This PR introduces config based node topologies, making it easier for developers to extend Crux and to provide new sets of modules. This PR also contains refactorings to make properties namespaced keywords, and to encapsulate config defaults and validations within modules.

This PR also lays the groundwork for generating a more comprehensive Java API.


jonpither:
With this PR - I'm removing a lot of documentation inside the code. Currently thinking about where it should go.. inside the modules themselves? Or do we have a properties file with comments in. Could also have a doc-string asking the user to consult the relevant modules for docs. The price of having an unbundled architecture :-)

Expect this WIP PR to evolve to handle documentation.

jonpither:
Note for reviewer - before the standalone (probably soon to be deleted anyway) topology, reused config props for the main kv store (such as sync? and kv-backend). Due to complexity of this with the new setup, I've given standalone its own args, and broken apart the current reuse.

jonpither:
Remaining considerations:

- [ ] Use maps for module definitions
- [ ] Pull out topology merge into modules rather than happening implicitly
- [ ] Introduce naming of 'topology'
- [ ] Create-topics -> create-topics?

jonpither:
My current dilemma is that you can't change the topology from Java land, or from the Clojure API for that matter aside from the single `crux.node/topology` option which governs the whole topology. This brings into question the validity of PR/approach, although one could argue that this PR is a needed simplification, whilst not being the final destination that allows us to go BETA.

jonpither:
On a different note:

Example Java code, we could conceivably produce with the current topology change in this PR: 

```
new CruxNode().withTopology(new JDBCTopology().withKvStore(new CruxKvRocksDb())).build()
```

This could mirror the new topology stuff in CLJ land. 

The main issue, is what to then do for Clojure API users? Could we still have a thin Clojure wrapper around the above Java builder / OO styled code, or do we allow CLJ users to work with the resulting/undelying Clojure data-structure.

jonpither:
There is the also the challenge, that currently you can't create new modules in Java, without supplying the underlying Clojure data-structure. To rephrase, you can't extend Crux without using Clojure. This PR should reflect on this, before being merged.

danmason:
Added a fix for example-tests - it's not currently ideal, as it requires each node created for the different query tests to create it's own event-log-dir. In addition, running the tests multiple times leads to an 'incorrect doc state' error.

Would, ideally, use a memkv for this but trying to set that option currently causes a spec assertion error.

jonpither:
Note - we have removed the ability to draw config args down from the ENV - the preferred configuration approach for the CLI is to use explicit properties (those passed in directly, or those in a properties file referred to via `--properties-file`.

jonpither:
This PR has been merged. Some considerations for the future worth noting below:

This PR has modified the CLI interface so that one of the key arguments is `--properties-file` used for further configuration. One aspect that wasn't covered in the previous CLI or this new approach, is how to configure nested constructs such as the RocksDB Options, without building Crux nodes the standard programmatic way. This is an OK position for now - that if users want full configuration power, they should use the API and not the command line.

Another point of consideration is that currently there is no feature for printing out all configuration options for a given topology. This should be easy enough, and we should likely offer it for CLI users.

The node configuration now feels in a good state following this PR. One option previously considered was to have a nested property structure - a sub-map per module - and I'd say this is off the cards for now, and so I'm closing #305. The main benefit of the current approach is that aligns well with traditional properties file.

As part of the  forthcoming BETA program, we will be soliciting users feedback on Node usage and configuration, based on the work done in this PR.

PR #346, opened 2019-10-04
Fix typo in README.md

mkmik:


refset:
Hi @mkmik thanks for kicking the tyres on the Docker instructions! I wrote this line back in ~April and the idea was that someone can follow this instruction, multiple times if necessary, without having to think about what it's doing. Use of a single ampersand is probably not _ideal_ but it achieves what I wanted it to do because regardless of whether the directory already exists the other commands still proceed. I believe your PR will mean that someone re-running this command will get a `mkdir: cannot create directory ‘my-crux-data’: File exists` error and then have to adjust the command.

There's undoubtedly a more useful compromise solution that doesn't involve forking a background process. Or maybe the better answer is to simply split the instructions across multiple lines. What do you think?

mkmik:
Then why not `-p` ?

refset:
Ah yes, of course, that is a very convenient solution. Thank you. Feel free to amend the PR and I'll merge it :)

mkmik:
cool. PTAL

refset:
Thanks!

PR #355, opened 2019-10-17
Use thread-safe requiring-resolve instead of require

spacegangster:
This updates a few more places in our codebase to use thread-safe `requiring-resolve` rather than `require`. A couple of usages of `require` are still left and marked with a comment.


spacegangster:
all tests pass

hraberg (approved):

PR #358, opened 2019-10-17
Update README example to use crux/start-standalone-node

jtaylor-io:
Very small PR to reflect current crux api in example


jonpither:
Hi @iojtaylor thanks for this. We're in the process right now of cutting a release - the new API is `start-node` and `start-standalone-node` is no more. Inside of the the options you must specify the topology, i.e. `{:crux.node/topology :crux.standalone/topology`. We aim to release in the next few hrs.

PR #369, opened 2019-10-30
Fixes #368 protect modules from Node closed.

### LINKED ISSUE #368: (crux/status node) crashes JRE
```
(crux/status *node)
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x000000012d2fdc27, pid=79779, tid=0x000000000000fc03
#
# JRE version: Java(TM) SE Runtime Environment (8.0_141-b15) (build 1.8.0_141-b15)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.141-b15 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# C  [librocksdbjni-osx.jnilib+0x1ac27]  Java_org_rocksdb_RocksDB_getProperty+0x57
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/petrus/Projects/acct/hs_err_pid79779.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#

Process finished with exit code 134 (interrupted by signal 6: SIGABRT)

```

jonpither:


crux-core/src/crux/node.clj
```diff
@@ -100,27 +117,32 @@
...
   (write-checkpoint [this {:keys [crux.backup/checkpoint-directory] :as opts}]
+    (assert-node-open this)
     (kv/backup kv-store (io/file checkpoint-directory "kv-store"))
     (when (satisfies? tx-log backup/INodeBackup)
       (backup/write-checkpoint tx-log opts)))
 
   Closeable
   (close [_]
-    (when close-fn (close-fn))))
+    (when (and (not closed?) close-fn) (close-fn))
```
jarohen: `@closed?` ?

FWIW, don't know how prevalent this is in the rest of the Clojure world, but I tend to prefix atoms with `!` so I don't forget to do `@!closed?` - worth it here?

jarohen (changes requested):
just the `@` in the `close` fn from me

crux-core/src/crux/node.clj
jonpither: I'd say it's likely an orthogonal clean-up - I haven't seen this style before tbh.

jarohen:
(from @hraberg)
there may well be a race condition here between the functions asserting that the node's still open and actually reading from the node - another thread may close the node between the two calls.

there's a similar pattern in https://github.com/juxt/crux/blob/jp/protect-closed-nodes/crux-core/src/crux/lru.clj#L70 - functions accessing the node take out a read lock; the `close` function takes the 'write' lock.

happy to make the change either on this branch, or a new one?

jarohen:
we could add a `with-lock` type macro to crux.io and use this from both places?

jarohen:
we've added locks around the operations in `node.clj` but aware that there are a few other places where a similar issue could arise - particularly, around lazy queries, where the read lock is checked before the snapshot is created, but then the snapshot is returned, and usages of it are not checked

(commits to come in the morning)

hraberg (approved):
So as discussed, we're adding the lock around node, but in theory this is done at the wrong level, it would be better to protect the native resources closer to the source of the problems, but this has a few issues as @jarohen mentions, mainly around laziness. We also don't want to have to check this lock in the middle of a query for every low level operation. 

It's worth noting that RocksDB have a few issues open related to give the JVM user a exception instead of a crash: https://github.com/facebook/rocksdb/issues/5234

That said, I'm approving this for now, as it's better than it were and the proper change is bit more involved. To be continued.

crux-core/src/crux/io.clj
```diff
@@ -11,7 +11,8 @@
            java.nio.file.attribute.FileAttribute
            java.text.SimpleDateFormat
            java.time.Duration
-           [java.util Comparator Date IdentityHashMap PriorityQueue Properties]))
+           [java.util Comparator Date IdentityHashMap PriorityQueue Properties]
```
jonpither, line 14: Are these all still needed?

jonpither (commented):
LGTM, made a minor comment about potentially unused imports

jarohen (approved):

PR #373, opened 2019-11-01
Dan/generate topology api

danmason:
Adds a module to generate Java code from topology files (kafka.clj, standalone.clj etc) which can be used to easier create configuration options in Java, then adds these to the current API


danmason:
Worth mentioning - likely not worth merging this until #374 is merged (would allow me to change the types somewhat, remove some more of the Clojure specific detail within the API)

danmason:
As approach has somewhat changed with the new Java API, changes will need to be made to this to reflect the changes in the Java API. (If, indeed, we are going to generate the topology classes)

jonpither:
I'd be tempted to close this PR for now since it's not a mergable state and raise when the respective branch is brought back to date (and consider rebase / squash)

PR #374, opened 2019-11-01
Dan/java list spec fix

danmason:
Temporary fix allowing Java list types to be used within the Java API (in place of PersistentVector)

### DIFF
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -53,6 +53,15 @@
 (s/def ::tx-op (s/multi-spec tx-op first))
 (s/def ::tx-ops (s/coll-of ::tx-op :kind vector?))
 
+(defn- conform-tx-ops [tx-ops]
+  (->> tx-ops
+       (map
+        (fn [tx-op]
+          (map
+           #(when (instance? java.util.Map %) (into {} %))
+           tx-op)))
+       (mapv vec)))
+
 (defprotocol PCruxNode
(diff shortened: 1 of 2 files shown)


jarohen (changes requested):
cheers Dan :)

@hraberg and I have been discussing our plans for how we'd like the existing Clojure/Java API to change in the near future, conclusions as follows:

* We currently have a mixed Clojure/Java API, where the Clojure implementation implements the API as defined by the Java interface
* You're obviously adding a Java adapter on top of this
* We'd like to reduce the current Clojure/Java low-level API to just a low-level Clojure API, suitable for Clojure-based tools to integrate with, and then your Java adapter calls into this. 
* Your Java adapter will be making this conversion from Java types to Clojure types
* With this in mind, I'd like the lifetime of this fix to be the same as the lifetime of the Java part of the low-level API - i.e. we don't want this fix to linger beyond the time we get rid of the Clojure/Java dance. 

tl;dr: could this fix be made in the Java part of the low-level API instead?

danmason:
I've moved the conversion out of `api.clj` and into the java files, though not entirely certain on the placement - would it be worth adding a new class for this rather than placing it within `Crux.java`?

jarohen (approved):
on reflection, this was the right place for this functionality - I'd forgotten that there is no external Java implementation of the ICrux*API interfaces, so this is the outermost location to put this shim.

nice one @danmason :)

PR #378, opened 2019-11-06
Dan/java transactions api

danmason:
Adds 'Operations' into Java API, (**PutOperation, CasOperation, DeleteOperation** and **EvictOperation**). Allows easier construction of operations within Java using a builder pattern, and removes the need for users to have to use **Keyword.Intern** on all of their keys. 

Note: Likely should be merged *after* #374 as operations make use of Java list/map types.


crux-core/src/crux/api/CasOperation.java
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	public Builder putInOldMap(String key, Object val) {
+	    oldMap.put(Keyword.intern(key), val);
+	    return this;
+	}
+
+	public Builder putInOldMap(Map<String, Object> valueMap) {
+	    for (String key : valueMap.keySet()) {
+		oldMap.put(Keyword.intern(key), valueMap.get(key));
```
reutsharabani, line 83: indentation? (also the `for` loop and `if` in all PR, which makes me suspect its standard) :|.

Issue #381, opened 2019-11-14
Viewing Crux Kafka connector on the Confluent hub shouldn't crash the page


danmason:
This currently seems to be caused by the manifest file missing some information regarding the 'features' of the connector, which need to be added within the packaging process.

https://www.confluent.io/hub/juxt/kafka-connect-crux

danmason:
Currently, the features (in JSON) would look like this:
```
"features": {
    "confluent_control_center_integration": true,
    "delivery_guarantee": ["exactly_once"],
    "kafka_connect_api": true,
    "single_message_transforms": true,
    "supported_encodings": ["any"]
  },
```

PR #382, opened 2019-11-14
Add 'features' list to the connector's manifest

danmason:
Done to address #381, following the suggestion made by confluent.


jarohen (approved):
atm I don't have enough background on this to be able to give it an informed approval. given it can't break the Confluent hub further, happy to submit this a speculative fix. maybe there's a validation we can run against the generated JSON?

Issue #351, opened 2019-10-10
clojure.lang.Keyword cannot be cast to crux.codec.EntityTx


spacegangster:
Getting a 500 on this query
[link](https://console.crux.cloud/console?query=%7B%3Afind+%5Be+p%5D%0A+%3Awhere%0A+%5B%5Be+%3Acrux.db%2Fid+_%5D%0A++%5Be+%3Aticker%2Fprice+p%5D%0A++%5B%28>+p+p-min%29%5D%0A++%5B%28<%3D+p+p-max%29%5D%0A++%28is-traded-in-currency+e+%3Acurrency%2Fusd%29%5D%0A+%3Aargs+%5B%7Bp-min+10++p-max+80%7D%5D%0A+%3Arules%0A+%5B%5B%28is-traded-in-currency+ticker-id+curr-id%29%0A+++%5Bticker-id+%3Aticker%2Fmarket+m%5D%0A+++%5Bm+%3Ase%2Fcurrency+curr-id%5D%5D%5D%7D)

It runs fine if I remove rules.
Locally: rules work fine.
Both run 19.09-1.4.0-alpha on Java 8
___
Used data from EBS locally and reproduced the 500

[data.zip](https://github.com/juxt/crux/files/3713962/data.zip)

Full trace
```
{:type java.lang.ClassCastException,
    :message
    "clojure.lang.Keyword cannot be cast to crux.codec.EntityTx",
    :at
    [crux.query$bound_result_for_var invokeStatic "query.clj" 521]}],
  :trace
  [[crux.query$bound_result_for_var invokeStatic "query.clj" 521]
   [crux.query$bound_result_for_var invoke "query.clj" 515]
   [crux.query$build_pred_constraints$iter__14160__14164$fn__14165$fn__14166$pred_constraint__14173$iter__14178__14182$fn__14183$fn__14184
    invoke
    "query.clj"
    564]
   [crux.query$build_pred_constraints$iter__14160__14164$fn__14165$fn__14166$pred_constraint__14173$iter__14178__14182$fn__14183
    invoke
    "query.clj"
    562]
   [clojure.lang.LazySeq sval "LazySeq.java" 42]
   [clojure.lang.LazySeq seq "LazySeq.java" 51]
   [clojure.lang.Cons next "Cons.java" 39]
   [clojure.lang.RT next "RT.java" 709]
   [clojure.core$next__5371 invokeStatic "core.clj" 64]
[...]

PR #417, opened 2019-11-15
Additional safety check for #351

jonpither:


hraberg (approved):
As discussed, this might be premature as we haven't 100% understood the mechanics of how this happens, but if we handle one placeholder we should likely handle the other one as well.

PR #420, opened 2019-11-18
Fixes #418 by adding exception for order-by

### LINKED ISSUE #418: query order-by doesn't order unless sort variable returned
See test on branch/commit [here](https://github.com/juxt/crux/commit/1e20f8c280160ecb856d7036b45afa54b3794dee).
When `:order-by` is used in `crux.api/q` and the variable on which the sort is decided by is not returned in `:find`, the results returned are not ordered but still in a vector.

t-taylor:
Also removes deprecated test for previous behavior


hraberg (approved):

jarohen (approved):
cheers @t-taylor :)

Issue #360, opened 2019-10-18
Java API improvements


jonpither:
Improvements to the general usage of the Java API can be made. As it stands, we are currently creating a new layer of Java API on top of the existing API, aiming to make it's usage strongly typed and immutable.

- [X] Create skeleton classes for operations, topologies and queries to get a feel for the general usage.
- [x] Add 'toEdn' functionality to necessary classes, generating data structures necessary to be passed down to underlying API. 
- [x] Allow strings to be used in queries.
- [x] Documenting usage of the wrapper API, with examples reflecting the clojure examples in the documentation.
- [x] Add information about topology classes to 'configuration' in the docs..
- [x] Create some basic Wrapper API tests

The above are currently suitable for releasing a vertical slice of the Java API. Further changes which can be made to the API:

- [ ] Wrap all functionality of underlying API.
- [ ] Test all contents of wrapper API.
- [ ] Change the topology generation to reflect the new style of topology classes.
- [ ] Fully typed queries
- [ ] Add javadocs for everything in the new API.
- [ ] Intersperse contents of quickstart doc with rest of docs

kosiakk:
Documentation clearly demonstrates an intention to make Crux Java-friendly, which is awesome. Thank you for that!

Here are some of my impressions so far:
Current examples are mostly written in Clojure and they are somewhat easy to follow. But building your own examples/tests in Java, Scala, or Kotlin is quite tricky:
- Returned entity Map is immutable (it's not documented, it just fails at runtime). Using honest Clojure's IPeristentMap might be an option here?
- Normal lists and maps don't pass Spec validation with no explanation at all. At best I've got `vector?` message. My impression is that `submitTx` works only with Clojure types despite the signature.
- Snapshot argument is not ideomatic in Java, at all. I might accept `queryIterable` method, which returns a "box" object which implements both Closeable and Iterable. But even better would be to return just Iterable and rely on finalization queue to close the snapshot. Is it really that expensive to keep snapshot open a bit longer than needed?
- There must be some quick-start in Java or Kotlin. You'll get all the love of Clojure community anyway, but for a simple Java devs it might look quite alien. Let's maybe use a bit of ~~Latin~~ Java to be understood by the majority of developers of the JVM ecosystem.

I'm sure you are on the right path to success :)

danmason:
We're definitely on the same train of thought, here!

To address some of your points:
[...]

PR #425, opened 2019-11-19
Dan/java api v2 #360 #415

danmason:
Vertical slice of the wrapper Java API, currently able to start nodes, configure topologies, create and submit transactions and create and submit queries.


crux-core/src/crux/api/v2/CruxId.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public static CruxId cruxId(URI uri) {
+        return new CruxId(uri);
+    }
+
+    public static CruxId cruxId(URL url) {
+        return new CruxId(url);
+    }
+
```
kosiakk: https://opencrux.com/docs#transactions-valid-ids mentions `Keyword` as a valid id, which is great for meta-model (e.g. store multi-lingual translation of attribute name in the database, track per-attribute permissions, etc) .

Should we also add "type-safe" constructor for `Attribute` as well?
```suggestion

    public static CruxId cruxId(Attribute attribute) {
        return new CruxId(url);
    }
    
```

of course, `Object` already covers all the cases, this overload is only for the documentation / IDE help

crux-core/src/crux/api/v2/CruxNode.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @SuppressWarnings("unchecked")
+    public TxResult submitTx(Iterable<Operation> ops) {
+        PersistentVector txVector = PersistentVector.create();
+        for(Operation op: ops) {
+            txVector = txVector.cons(op.toEdn());
+        }
+
+        Map<Keyword,Object> result = node.submitTx(txVector);
+        Date txTime = (Date) result.get(Keyword.intern("crux.tx/tx-time"));
```
kosiakk: This is a nitpicking, but I'd replace all calls to `Keyword.intern` with private static fields for two reasons:
- make Java code a bit more understandable, showing that it is _the same_ Keyword every time
- just to save some cpu cycles on String parsing and hashmap lookup

From what I can tell, Clojure compiler does exactly this kind of per-class caching of Keywords.

crux-core/src/crux/api/v2/CruxNode.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public Database db(Date validTime) {
+        return database(node, validTime);
+    }
+
+    public Database db(Date validTime, Date transactionTime) {
+        return database(node, validTime, transactionTime);
+    }
+
+    public Document document(Object contentHash) {
```
kosiakk: It's not clear now, whether `CruxId` is acceptable here or not, or how to obtain an instance of `contentHash` from the public v2 API. If this is an internal method for Clojure query engine, it might be less useful in this Java-wrapper.

The `Document` itself does not offer `contentHash`, only the `id`. Maybe a new proposed `DocumentSnapshot` should hide the concept of content hashes from the public API, just like you did with `TxResult`.

crux-core/src/crux/api/v2/CruxNode.java
```diff
@@ -0,0 +1,75 @@
...
+        Map<Keyword,?> doc = node.document(contentHash);
+        return Document.document(doc);
+    }
+
+    public List<Document> history(CruxId id) {
```
kosiakk: I think, the spirit of this method is to return the history _lazily_: https://opencrux.com/docs#queries_history_api
So I'd change `List` to `Iterable` or `Iterator` which might be lazy, if we plan to return *full versions* of `Documents` with all their fields (resolve hash).

crux-core/src/crux/api/v2/CruxNode.java
```diff
@@ -0,0 +1,75 @@
...
+        List<Map<Keyword,?>> history = node.history(id.toEdn());
+        return history.stream()
+            .map(historyMap -> Document.document(historyMap))
```
kosiakk: This seems wrong to me!

According to the published documentation, Crux returns a stream of _hashes_, so each Map is like
```
{:crux.db/id ; sha1 hash of document id
  "c7e66f757f198e08a07a8ea6dfc84bc3ab1c6613",
  :crux.db/content-hash ; sha1 hash of document contents
  "6ca48d3bf05a16cd8d30e6b466f76d5cc281b561",
  :crux.db/valid-time #inst "2018-05-18T09:20:27.966-00:00",
  :crux.tx/tx-time #inst "2019-04-15T07:53:55.817-00:00",
  :crux.tx/tx-id 1555314835817}
```

It seems that `node.history` returns a Map<Keyword, Object>, but it is **not a `Document`**!

crux-core/src/crux/api/v2/CruxNode.java
```diff
@@ -0,0 +1,75 @@
...
+            .collect(Collectors.toList());
+    }
+
+    public Date sync(Duration timeout) {
+        return node.sync(timeout);
+    }
+
+    public void close() throws IOException {
```
kosiakk: Should `CruxNode` implement `AutoCloseable` then?

crux-core/src/crux/api/v2/CruxNode.java
kosiakk: It might return an instance of a new class `DocumentSnapshot`, which has times, transaction id, and a document itself (loaded lazily upon request from its contentHash)

crux-core/src/crux/api/v2/Database.java
```diff
@@ -0,0 +1,48 @@
...
+
+        Set<ResultTuple> resultSet = new HashSet<>();
+        for(List<?> result : queryResult){
+            resultSet.add(ResultTuple.resultTuple(symbols, result));
+        }
+
+        return resultSet;
+    }
+
+    public Collection<List<?>> query(String query) {
```
kosiakk: It seems odd that return type is different. Should it also be `Set<ResultTuple>` or `Collection<ResultTuple>` instead?

crux-core/src/crux/api/v2/CruxNode.java
danmason: Yeah, we were working through some of the API and doing just that! :) (Similar situation within the different operation classes, also)

crux-core/src/crux/api/v2/CruxId.java
danmason: We were considering removing Attribute/LogicVar and making a Util class from which you can get a Keyword/Symbol, though I do believe making an explicit type safe constructor for Keywords is a good idea.

crux-core/src/crux/api/v2/DeleteOperation.java
```diff
@@ -0,0 +1,32 @@
...
+    private DeleteOperation(CruxId deleteId, Date validTime) {
+        this.deleteId = deleteId;
+        this.validTime = validTime;
+    }
+
+    public DeleteOperation withValidTime(Date validTime) {
+        return new DeleteOperation(deleteId, validTime);
+    }
+
+    public static DeleteOperation deleteOp(CruxId deleteId) {
```
kosiakk: A general comment here - in Java it's not common to see many abbreviations like `op` or `attr` - autocomplete eliminates the need to type or remember such names. The call will look like 
`DeleteOperation.deleteOp( id )`, which is a bit verbose (even by Java standards). It's rare to use static imports for such methods, so some people might prefer `DeleteOperation.create(idToDelete)`

crux-core/src/crux/api/v2/Document.java
```diff
@@ -0,0 +1,74 @@
...
+        Map<Attribute, Object> initialDoc = Collections.singletonMap(Attribute.attr("crux.db/id"), id.toEdn());
+        return new Document(initialDoc);
+    }
+
+    static Document document(Map<Keyword, ?> map) {
+        Map<Attribute, Object> newDoc = new HashMap<>();
+        for(Keyword key : map.keySet()) {
+            newDoc.put(Attribute.attr(key), map.get(key));
+        }
+        return new Document(newDoc);
```
kosiakk: There is no need to use `HashMap` for internal state of the `Document` class! You might change it to `IPersistentMap` right way (and have a short-cut to simply assign it, if `PeristentMap` is already given by the core).

This is exactly why you do incapsulation in the first place :-) Then all your methods become simpler and much more efficient.

crux-core/src/crux/api/v2/Document.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public Document with(Attribute attr, Object value) {
+        Map<Attribute, Object> newDoc = new HashMap<>(this.document);
+        newDoc.put(attr, value);
+        return new Document(newDoc);
+    }
+
+    public Document with(String strKey, Object value) {
```
kosiakk:
```suggestion
    public Document with(String attr, Object value) {
```

kosiakk (changes requested):
I'd like to offer my help implementing some or all the suggestions! I'm already in close contact with Jon and Jeremy; please feel free to call me if you'd like to discuss anything

crux-core/src/crux/api/v2/CruxId.java
kosiakk: I'd vote for an explicit `Attribute` against very generic `Keyword`

crux-core/src/crux/api/v2/CruxNode.java
kosiakk: Hmm, I meant just the implementation - cache the results of calls to `Keyword.intern` per class:
```
class CurxNode { 
private static final TX_TIME = Keyword.intern("crux.tx/tx-time");
...
Date txTime = (Date) result.get(TX_TIME);
}
```
Internal calls to `Keyword.intern` or possible `Util.keyword` are irrelevant for the public API

crux-core/src/crux/api/v2/Database.java
danmason: We were going to remove this function for now, in favor of making users use and construct query objects rather than just passing in a string.

danmason:
Thank you for the assistance! :) James and I were going through this PR this morning, so that should be committed on here soon (and a few of these suggestions are actually in there already)! As soon as that's up on here, it would be worth going through and implementing/considering the rest :)

crux-core/src/crux/api/v2/EvictOperation.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    protected PersistentVector toEdn() {
+        PersistentVector outputVector = PersistentVector.create(Keyword.intern("crux.tx/evict"), evictId.toEdn());
+        if(validTime != null)
+            outputVector = outputVector.cons(validTime);
+        if(endValidTime != null)
+            outputVector = outputVector.cons(endValidTime);
+        if(keepLatest != null)
+            outputVector = outputVector.cons(keepLatest);
+        if(validTime != null)
```
kosiakk: It seems you meant `keepEarliest` here. You also might want to put explicit `false` if `keepLatest` is not specified, but `keetEarliest` is.

crux-core/src/crux/api/v2/EvictOperation.java
```diff
@@ -0,0 +1,58 @@
...
+        return new EvictOperation(evictId, validTime, endValidTime, keepLatest, keepEarliest);
+    }
+
+    public EvictOperation keepEarliest(Boolean keepEarliest) {
+        return new EvictOperation(evictId, validTime, endValidTime, keepLatest, keepEarliest);
+    }
+
+    @Override
+    protected PersistentVector toEdn() {
+        PersistentVector outputVector = PersistentVector.create(Keyword.intern("crux.tx/evict"), evictId.toEdn());
```
kosiakk: Oh, this looks very dangerous :fire:.
What happens if `validTime` is not specified, but `endValidTime` is? `endValidTime` will be used instead!
No-no, generic builder pattern is not really applicable to this class. 
Both timestamps can only come together in a single method call.

It could be something like 
```
EvictOperation.create(myId).validTimes(from, to).keepLatest()
```

danmason: Mhm, this was another one of the issues we were going through this morning. Evict has some planned changes, so at this stage we were going to remove the time ranges / keepearliest and keeplatest from this part of the API and re-add them down the line :)

crux-core/src/crux/api/v2/Database.java
kosiakk: Fair enough, I'll wait for the merge of this branch into `master` and then review the whole package again, if you don't mind :-)

danmason: Absolutely! There's a lot of great suggestions in here which I think we shall put in as soon as we commit our changes from earlier :)

jarohen:
Hey @kosiakk, thanks for all your suggestions and your offer of help!

We've addressed a good number of these as part of our code review (will reply to your individual comments) but for now, given it's an early pass at the Java API (albeit on a public repo - the nuances of open-source development!) I reckon it's best to let this sit for a few days, try out a few more use cases, write up some docs, and develop it further. Would obviously be great to get your thoughts and feedback through the process!

Cheers,

James

crux-core/src/crux/api/v2/EvictOperation.java
jarohen: yep, to confirm, have removed the time-based flags to evict in line with the upcoming upstream changes

crux-core/src/crux/api/v2/EvictOperation.java
jarohen: likewise, removed

crux-core/src/crux/api/v2/Document.java
jarohen: yep, agreed!

crux-core/src/crux/api/v2/Database.java
jarohen: yep, agreed, removed for now

crux-core/src/crux/api/v2/DeleteOperation.java
jarohen: yep, agreed with the abbrev's ;) 

re the static imports - I wasn't aware that this was that rare? intention behind it was to make the name of the method self-contained, although this may be a result of us coming from the Clojure world where such things are common :)

in the meantime, I've moved these under `TransactionOperation`, so they can be used either fully qualified or by a static import

crux-core/src/crux/api/v2/CruxNode.java
jarohen: absolutely!

crux-core/src/crux/api/v2/CruxNode.java
jarohen: agreed, this is unclear. have made this method private until we can clarify it

crux-core/src/crux/api/v2/CruxNode.java
jarohen: cheers, fixed

crux-core/src/crux/api/v2/Document.java
jarohen: sorted, cheers :)

crux-core/src/crux/api/v2/CruxNode.java
jarohen: fixed :)

crux-core/src/crux/api/v2/CruxNode.java
jarohen: yep, agreed :)

crux-core/src/crux/api/v2/CruxId.java
jarohen: reckon we should leave this one open for the time being - there are other places where we'll need Clojure keywords, and need to consider whether we want to expose these as an implementation detail or not (and, indeed, whether we can avoid exposing them as an implementation detail for the time being)

jarohen:
Have merged this into master under the `crux.api.alpha` package, going to iterate on it further before we present it as a public, stable Java API.

PR #429, opened 2019-11-20
Moves crux-console and crux-dataflow into their own repositories 

spacegangster:
under https://github.com/crux-labs account


jarohen (approved):
:fire:

PR #431, opened 2019-11-21
Further additions, improvements and fixes to the alpha Java API 

danmason:


crux-core/src/crux/api/alpha/CruxNode.java
```diff
@@ -93,17 +61,42 @@ private Document document(Object contentHash) {
...
-    public Iterable<Document> history(CruxId id) {
+    public Iterable<EntityTx> history(CruxId id) {
         List<Map<Keyword,Object>> history = node.history(id.toEdn());
         return history.stream()
-            .map(this::document)
+            .map(entity -> EntityTx.entityTx(entity))
             .collect(Collectors.toList());
     }
 
+    /**
```
jarohen: @danmason - can you omit these two (`hasSubmittedTx*`) for the time being? not convinced they're going to remain supported

crux-core/src/crux/api/alpha/NodeStatus.java
```diff
@@ -0,0 +1,47 @@
...
+    public final String version;
+    public final String revision;
+    public final String kvStore;
+    public final int estimatedNumberOfKeys;
+    public final long size;
+    public final long indexVersion;
+
+    @Override
+    public String toString() {
+        return "NodeStatus{" +
```
jarohen, line 17: IntelliJ'll generate you a String.format based toString if you ask it to, I think? preferable to string concatenation for both performance and (more subjective) readability

crux-core/src/crux/api/alpha/NodeStatus.java
```diff
@@ -0,0 +1,47 @@
+package crux.api.alpha;
+
+import clojure.lang.Keyword;
+
+import java.util.Map;
+
+public class NodeStatus {
+    public final String version;
```
jarohen, line 8: we should consider which of these we want to support, and whether we want to support them as a separate object or 'just' values on the node itself, in Java.

maybe not 100% either way, too - version, revision, indexversion etc are all static when the node's created; the others are dynamic

crux-core/src/crux/api/alpha/EntityTxWithDocument.java
```diff
@@ -0,0 +1,31 @@
+package crux.api.alpha;
+
+import clojure.lang.Keyword;
+
+import java.util.Map;
+
+public class EntityTxWithDocument {
```
jarohen, line 7: hmm :thinking:

danmason: Agreed on the hmm - was unsure if I should just edit what I had with EntityTx, or return some sort of tuple (as I did here, sort of)

crux-core/src/crux/api/alpha/TxResult.java
```diff
@@ -1,18 +1,33 @@
...
+
+import static crux.api.alpha.Util.keyword;
 
 public class TxResult {
+    private static final Keyword TX_TIME = keyword("crux.tx/tx-time");
+    private static final Keyword TX_ID = keyword("crux.tx/tx-id");
+
     public final Date txTime;
     public final long txId;
+    private Map<Keyword, ?> txResultMap;
```
jarohen: nit: `final`

crux-core/src/crux/api/alpha/CruxNode.java
danmason: Yeah, can do - remove, or make private?

crux-core/src/crux/api/alpha/NodeStatus.java
danmason: Might need to check that? I think this is what it does by default, so will need to have a look at my options

crux-core/src/crux/api/alpha/TransactionOperation.java
```diff
@@ -1,11 +1,26 @@
...
 public abstract class TransactionOperation {
+    static final Keyword TX_PUT = keyword("crux.tx/put");
+    static final Keyword TX_EVICT = keyword("crux.tx/evict");
+    static final Keyword TX_CAS = keyword("crux.tx/cas");
+    static final Keyword TX_DELETE = keyword("crux.tx/delete");
+
     TransactionOperation() {
     }
 
+    // May want to move these - are inherited by the classes
```
jarohen, line 23: the static constructors?

crux-core/src/crux/api/alpha/NodeStatus.java
danmason: Yeah - this was a sort of placeholder, one of the classes I wanted a bit more thinking on (same as EntityTx)

crux-core/src/crux/api/alpha/TxResult.java
danmason: Yeah - need to check a few places that probably need 'final'

crux-core/src/crux/api/alpha/TransactionOperation.java
danmason: Yep, in the generated Javadocs they're listed as inherited functions

jarohen:
@danmason - this is a good PR but I'm going to close it for now because I'd like to sort out the underlying Clojure API before checking the alpha Java API for completeness. when that's done, let's update and re-open this PR

danmason:
:+1: Good by me, better than it sitting open with no movement in our PR list :)

Issue #428, opened 2019-11-20
Temporarily removing the time-range parameters to the Evict op


jarohen:
We're going to temporarily remove the time-range parameters to the Evict operation, because the current semantics are inconsistent and buggy - this requires more thought before we can commit to it as a public API.

To be removed:
* valid-time/end-valid-time
* keep-earliest/keep-latest

To be kept:
* Evict all versions of a document, for all valid times

Related cards, out of scope of this one:
* #371 
* #375

Things to consider:
* How will this affect existing evict ops in live Kafka logs?

jarohen:
@hraberg @refset let's confirm the backward-compatibility behaviour we require for existing evict ops in live Kafka logs

hraberg:
This card doesn't currently deal with the race condition which would happen if someone tries to resurrect a document in quick succession of evicting it. We can split that work to a different card if necessary, as it still requires some analysis and discussion.

My basic idea is this:
1. Record evicted entity ids, these would simply be key-values stored by the evict op in a new internal index.
2. If a document arrives for an entity which has been marked as evicted, drop it instead of indexing and storing it, and emit a new tombstone to get rid of it again on Kafka. Log at debug or warn.
3. On the transaction side, potentially blow up the transaction in a general pre-condition if an operation involves an evicted event, alternatively log a warning and do nothing.
4. One could potentially introduce a new operation `:crux.tx/unevict` to allow writes to the entity id again at a later stage, but think that would be a separate card and discussion for sure. Can easily imagine opening up for various new race conditions.

hraberg:
Note, the reason we cannot simply allow a new put to resurrect it is that there's a race condition between the topics, as an eviction might have overwritten its document before it gets indexed, or vice versa for the tombstone.

The above solution actually doesn't care about if the events happen close in time, as it's designed to converge to the same result regardless.

jarohen:
> This card doesn't currently deal with the race condition which would happen if someone tries to resurrect a document in quick succession of evicting it. We can split that work to a different card if necessary

agree that we should split that out into a separate card and link it in the 'related' section here, if there's not already a card for it?

hraberg:
Pretty sure there's no other card, it was discussed on Slack a bit in the context of this card.

jarohen:
proposal re evicting/kafka log:
* if we spot an eviction op with time ranges on the Kafka log, we throw an exception and kill the indexer. the error message informs the user that they need to specify an env var flag to continue
[...]

PR #438, opened 2019-11-25
#428 - temporarily removing evict valid time ranges

jarohen:
This PR removes the time ranges and boolean flags from the evict tx-op, until we can define a consistent semantics for how they should behave (see #428 for more details and rationale)

N.B.:
* After this is released, there won't be any _new_ evict ops with time ranges on Kafka logs, but there may still be existing events
* With this in mind, we hard-stop the indexer if we encountered such an event during replay. The user can then manually intervene by setting the `CRUX_EVICT_TIME_RANGES` environment variable to `EVICT_ALL`, which tells the indexer to ignore the time ranges and evict the document for all valid times. (in fact, the error message points the user to this PR)
* We do intend to replace this behaviour in the future, but only once we've nailed down exactly how it should behave.
* Other evict docs + tests updated (and removed, where it didn't make sense to update the examples)

### DIFF
--- a/crux-core/src/crux/tx.clj
+++ b/crux-core/src/crux/tx.clj
@@ -97,7 +97,7 @@
             (c/->id-buffer new-v)]]}))
 
 (def evict-time-ranges-env-var "CRUX_EVICT_TIME_RANGES")
-(def ^:dynamic evict-all-on-legacy-time-ranges? (= (System/getenv evict-time-ranges-env-var) "EVICT_ALL"))
+(def ^:dynamic *evict-all-on-legacy-time-ranges?* (= (System/getenv evict-time-ranges-env-var) "EVICT_ALL"))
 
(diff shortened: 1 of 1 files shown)


jonpither (approved):

crux-core/src/crux/tx.clj
```diff
@@ -96,26 +96,34 @@
              tx-id)
             (c/->id-buffer new-v)]]}))
 
-(defn tx-command-evict [indexer kv object-store snapshot tx-log [op k start-valid-time end-valid-time keep-latest? keep-earliest?] transact-time tx-id]
+(def evict-time-ranges-env-var "CRUX_EVICT_TIME_RANGES")
+(def ^:dynamic evict-all-on-legacy-time-ranges? (= (System/getenv evict-time-ranges-env-var) "EVICT_ALL"))
```
hraberg: If this is dynamic it should have a `*evict...*` name.

hraberg (approved):

jarohen:
merged

PR #444, opened 2019-11-28
Put/delete range semantics, #326

### LINKED ISSUE #326: Consistent semantics for put/delete operations with valid time ranges
We'd like to amend the put/delete semantics so that the behaviour is consistent, without the user needing to understand the implementation details of the internal Crux indices to reason about the query results they're seeing.

We think of each of these operations as a function operating over a timeline - for this spec, we'll think of a the value of a single attribute of a single entity, over time. Each operation can be thought of as a function taking a timeline and returning a new timeline.

We have four such operations:
* 'put' with a start and end valid-time - this should override the value in the timeline between the two valid times. Particularly, after the end valid-time, the operation shouldn't change the value (as it does currently) - the previous value should be restored.
* 'put' with a single valid-time - this should put the value between the provided valid time and the next time the value changes or, if there isn't a subsequent value, indefinitely. (Due to the way the Crux index updates are currently implemented when a put has a start and an end valid-time, this may not take effect all the way until the next value change - we'd like to fix this)
* 'delete' with a start and end valid-time - behaves like a put, putting a 'nil' value within the range.
* 'delete' with a single valid-time - likewise, behaves like put with a single valid-time, putting a 'nil' value until the value next changes.

Worked example:
* Put value 25 at vt=25 - if we then query any time after vt=25, we should expect '25' to be returned
* Then, put value 27 at vt in `[27, 30)` - we then expect to see 27 at vt=28, but then return to 25 at vt>=30
* Delete at vt=28 - expect to see no value returned at vt=29, but 25 returned again at vt>=30
* Put 20 at vt=20, expect to see 20 for vt in `[20,25)`, then 25 for vt in `[25, 27)`, and so on.

Separately, 'evict' also used to take valid-time ranges - these were temporarily removed as part of #428, but we expect to reinstate these with similar, consistent semantics, in the near future.

---
(original PR spec follows)

In this newly added test I validated that the effect of using the `end-valid-time` argument with `put` means that, whilst corrections within the range happen as expected, the corrected document is still visible _after_ the `end-valid-time`: https://github.com/juxt/crux/commit/34a43c697704fc50d56d66956c27233a427ad76e

This might be desirable behaviour in some circumstances and may be even sensible to leave as the default option. However, I propose that there are other variations for using `end-valid-time` with `put` which may be useful (and potentially even more intuitive):

1) The equivalent of `delete` could be inserted at the end of `end-valid-time` i.e. the test I added should fail because there shouldn't be any version at `v4-valid-time #inst "2018-11-30"` due to the equivalent of a delete operation at `end-valid-time #inst "2018-11-29"`
2) If the `put` range sits within a larger range of a single version, then `put` could effectively re-`put` the initial version at `end-valid-time` i.e. if version A is valid `2018-11-26 -> 2018-11-30` and you `put` version B with the range `2018-11-27 -> 2018-11-28` (using start & end-valid-time arguments) then version A should still be valid at `2018-11-29`

These variations could be handled using additional arguments, like `open-ended?` and `put-within?`, respectively.

Additionally, a similar concern applies to how `end-valid-time` is used with the `delete` operation, which currently ends up leaving a gap in the valid time history, where it could instead have the equivalent effect of making a `put` of the last version (within the range) at `end-valid-time`.

Another option is to rename the current `end-valid-time` to `correct-until-valid-time`.

jarohen:
closes #326 

To get an idea of what the semantics have changed to, best to start with the `tx_test.clj` file, particularly `test-put-delete-range-semantics`.

Implementation-wise, we've changed the way the put/delete kvs are calculated - this now switches on whether the put/delete is 'from a vt', or 'for a vt range'.

For a vt range, we may have many existing documents between the start and end valid times - we need to overwrite each of these valid times with the new value. We also need to reinstate the previously visible value at the end of the range, so that the put stops taking effect at the end valid-time. There're a couple of cases here:
* If our previous value was at vt=28, we put vt between 25 and 30, we want to reinstate the value at vt=28 at the end of the range, vt=30.
* If the previous value was at vt=23, we still want to reinstate that value - we say that the implicit range for the previous one was from vt=23 until further notice; the range put then briefly interrupts that value, and then the vt=23 range continues afterwards.

'From a vt', we'd like the semantics to be that the put value holds until the next change. Importantly, though, with the above range change, we might well have added a number of otherwise invisible entries which aren't 'changes', but exist to overwrite previous values. 

As part of this change, we therefore have to read forward in the entity history for single-vt puts, overwriting values until we see a 'real change'.

TODO
- [x] CaS should behave like a compare then a put (with new put semantics)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [#428](https://github.com/juxt/crux/issues/428): Time ranges removed from 'evict' command, see [#PR438](https://github.com/juxt/crux/pull/438) for more details.
 * [#441](https://github.com/juxt/crux/issues/441): Fix two transactions updating the same entity in the same millisecond always returning the earlier of the two values - requires index rebuild.
+* [#326](https://github.com/juxt/crux/issues/326): Put/delete with start/end valid-time semantics made consistent
 
 ### Bug fixes
 
--- a/crux-core/src/crux/index.clj
+++ b/crux-core/src/crux/index.clj
@@ -594,8 +594,9 @@
             (enrich-entity-tx (kv/value i)))
         (entity-history-step i seek-k f-next f-next))))))
 
-(defn entity-history-seq-ascending [i eid ^Date from-valid-time ^Date transaction-time]
-  (let [seek-k (c/encode-entity+vt+tt+tx-id-key-to nil (c/->id-buffer eid) (Date. (dec (.getTime from-valid-time))))]
+(defn entity-history-seq-ascending [snapshot eid ^Date from-valid-time ^Date transaction-time]
+  (let [i (kv/new-iterator snapshot)
+        seek-k (c/encode-entity+vt+tt+tx-id-key-to nil (c/->id-buffer eid) (Date. (dec (.getTime from-valid-time))))]
     (->> (entity-history-step i seek-k #(when-let [k (kv/seek i seek-k)]
                                           (kv/prev i)) #(kv/prev i))
          (drop-while (fn [^EntityTx entity-tx]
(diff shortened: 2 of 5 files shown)


jonpither:
Please raise an issue to clearly document the semantics with worked examples for users

jonpither (approved):

Issue #362, opened 2019-10-19
The SHA1 hash of maps should not depend on the key ordering


asciiansi:
(original question below)

Currently, the content/value hashes for a map depend on the key ordering - with different orders (possibly arising from different concrete map implementations), we generate different content/value hashes for the 'same' map, which means:
* Entities with a map key (useful, but so far undocumented) aren't always returned/joined correctly.
* (Lesser impact, but) some documents that are essentially the same (varying only in key order) are stored under multiple content-hashes (sub-optimal, but no undesired behaviour)

---
(old proposed solution, for posterity, see #474 for what I ended up doing):

* The SHA1 hashes of maps should not depend on the key ordering - we'll take inspiration from java.util.Map's hashcodes here.
* Users that want to can preserve the old behaviour for document content-hashes
* Users that want to can preserve the old behaviour for map-keys (independently from the previous content-hash behaviour)
* Users that want to can migrate their Kafka streams via a tool.

---
(original question) 

Is it intended that map keys for a `crux.db/id` always must be in the same order, despite having the same content/hash? Querying for an entity using `(crux/entity db {:a 1 :b 1})` is not the same as `(crux/entity db {:b 1 :a 1})` for example. The query only seems to work for me using the same key order as the transacted entity.

I can semi-understand why this is case but nonetheless it's a bit surprising for keys and lead to some brief debugging. Specifically, it seems a bit strange as maps are by definition unordered and often manipulating programatically. 

Moreover, it's typical in a larger system when dealing with dynamically constructed keys to make use of `assoc`, `update`, etc. which becomes very dangerous and instead requires construction of a new map, creating even more garbage. I currently work around this simply by using a function every time I create keys dynamically, but there are a few cases where this is a bit annoying.

I find maps in general useful as keys since they allow a composite and additionally remove some overhead for reaching into an entity via a query using `crux/q`. I would prefer other options for key and would love to discuss further if you are willing along with use-cases and reasoning. For now, I'm mostly using maps to workaround dissatisfaction with other key types which aren't really suitable for my apps.

For context, I am using 100% RocksDB on both ends (though I prefer/will eventually switch to LMDB). Here are my deps:

```
juxt/crux-core                      {:mvn/version "19.09-1.5.0-alpha"}
juxt/crux-rocksdb                   {:mvn/version "19.09-1.5.0-alpha"}
```

At the very least, this should be documented given I found it surprising if it's not considered a bug. Thanks.

refset:
[...]

PR #450, opened 2019-12-02
Fix for #362

t-taylor:
Found by @danmason

When a map is created using clojure `{}` syntax with less than 10 elements, it creates a `clojure.lang.PersistentArrayMap` which preserves order. Otherwise it creates a `clojure.lang.PersistentHashMap` which is the typical definition of a hashmap (i.e. unordered). This fix adds a case for when `:crux.db/id` is the former type. This change is breaking as it changes the hash of existing documents.


crux-dev/dev/tmt/hash_map.clj
```diff
@@ -0,0 +1,91 @@
+(ns tmt.hash_map
```
danmason: I assume this namespace likely doesn't want to be in the PR, and in it's place something could be put within query test? (Could very well take the test from the branch on my fork)

Otherwise, seems like an entirely reasonable fix to me. May wish to have a slightly wider discussion of this first considering it is a breaking change.

danmason (changes requested):

crux-core/src/crux/codec.clj
```diff
@@ -274,6 +274,10 @@
         (id->buffer id to)
         (throw (IllegalArgumentException. (format "Not a %s hex, keyword, URL or an UUID string: %s" hash/id-hash-algorithm this))))))
 
+  PersistentArrayMap
```
jarohen: @t-taylor - reckon this could do with being `java.util.Map`, as discussed (would also mean removing the `IPersistentMap` implementation below.

Also, would it be worth us running with `(into (sorted-map) this)` instead? not convinced normal `{}` _has_ to preserve ordering...

crux-dev/dev/tmt/hash_map.clj
jarohen: yep, agreed, let's get this into something like `(ns crux.map-keys-test-362)` and have it run as part of the usual test suite, if it's reasonably quick to run?

jarohen: or, if you wouldn't consider it worth keeping, yeah, let's remove it - your call :smile:

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ### Bug fixes
 
 * [#PR363](https://github.com/juxt/crux/pull/363) Allow `full-results?` and other boolean flags in a vector-style query
+* [#362](https://github.com/juxt/crux/issues/362) Fixes 362 where hashes of small maps were dependent on oder *Breaking*
```
jarohen: thai-po ;)

jarohen: I'd consider this part of the breaking changes section, above (might even be worth explicitly calling this 'breaking changes'?

jarohen (changes requested):

CHANGELOG.md
t-taylor: Should we do this for all previous versions as well then? I was following the style of the log previously. I think it would make it stand out more, but since we only have a few items per version anyway I don't think it needs to be that displayed. Although I'm fine to be overruled here if you disagree

crux-dev/dev/tmt/hash_map.clj
t-taylor: I was just using this file to do quick tests to identify the bug better. I can remove it if you want but i just kept it there because it gives a more detailed view of the bug, I'll write an explicit test for the bug as well though.

crux-core/src/crux/codec.clj
t-taylor: Do we want to always alter the map? The only reason we need to touch the arraymap is to remove its ordered quality, but all the other types of maps don't need to be altered. Making it so we always alter the map would effect ingest a little but i agree it would make the code cleaner. Also it should probably be `(into (hash-map) this)` because that 100% doesn't preserve order.

jarohen: I'd rather go the other way and ensure that maps are always ordered consistently - there's nothing in the contract of Map that mandates it returns the same order every time, unless you explicitly use a sorted map. even in Java-land, there's a whole load of Map implementations that might return a different ordering.

(in practice, the same Map concrete class will _usually_ return the elements in the same order, but this isn't something I'd want to rely on, given it's not in the contract, and given the user might use different concrete classes over time)

crux-dev/dev/tmt/hash_map.clj
t-taylor: After talking with dan i'm going to remove it and just write a test

crux-core/src/crux/codec.clj
t-taylor: okay that make sense. I think I still disagree because I fell we should do the least amount of work to provide a consistent hash, and having to sort is unnecessary because of the definition of a hash map, i.e. it has no order so the hash of the data structure should only depend on elements rather than order, but I'll concede because i think its just the haskeller in me trying to find the least restrictive function.

jarohen: > the hash of the data structure should only depend on elements rather than order

agreed - we're not using hashes of the object, though - we're using the hash of the serialised form of the map, which doesn't have the same guarantees?

CHANGELOG.md
jarohen: let's stick with the convention for now and discuss whether to change it - could you move mine in while you're there?

t-taylor: will do

jarohen:
discussed on Slack, one for the standup this morning - need to consider the implications on the Kafka transaction log

jarohen:
closing for #474

Issue #365, opened 2019-10-25
Edn Serialisation is Likely Broken


hraberg:
See https://www.nitor.com/fi/uutiset-ja-blogi/pitfalls-and-bumps-clojures-extensible-data-notation-edn

Any places using `pr-str` in the code should be consolidated to use some helper in `crux.io` that deals with all this. The main culprits are the http-server and client. But also the Kafka EdnSerialiser or example.

Somewhat related to #151.

refset:
This may also impact hashing (we also need to identify all the places where hashes are generated, using nippy, for future analysis).

jarohen:
@hraberg I can't access that link - just me? if not, could you pop a rough outline of what's broken in the description?

cddr:
Ah we made that mistake in jackdaw (https://github.com/FundingCircle/jackdaw/pull/190)

hraberg:
The blog post above seems to have gone, but the points made seem to be the defaults anyway, so maybe this card isn't needed, as seen in the jackdaw commit above, they did remove the explicit bindings in their `edn2` serde. Relevant bindings are:

* `*print-length*` defaults to `nil`  but `100` in CIDER.
* `*print-level*` defaults to `nil`
* `*print-namespace-maps*` defaults to `false` but `true` in the REPL.

The jackdaw point about not emitting newline is also worth reflecting on, but think we use `pr-str` in all relevant cases, though would be worth sanity checking while bringing in a consistent helper function.

So raising this card might been false alarm after all. But this can easily be confusing if not handled in a consistent way.

hraberg:
As the original post seems to have disappeared, here's a discussion around it:
https://www.reddit.com/r/Clojure/comments/d967fd/pitfalls_and_bumps_in_clojures_extensible_data/

danmason:
Found a (non broken) link to the article: https://www.nitor.com/en/news-and-blogs/pitfalls-and-bumps-clojures-extensible-data-notation-edn

PR #451, opened 2019-12-03
Replacing usages of 'pr-str' with a helper function in crux.io #365

danmason:
There has been some amount of discussion on the EDN serialization card. Though the default bindings appear to be sufficient, I decided to replace the usages of `pr-str` with a function under `crux.io` called `prn-edn` and use explicit bindings for it. Happy to discuss whether or not you think this is necessary/would do it differently :)

### DIFF
--- a/crux-core/src/crux/io.clj
+++ b/crux-core/src/crux/io.clj
@@ -221,10 +221,7 @@
          (.unlock lock# stamp#)))))
 
 (defn prn-edn [xs]
-  (binding [*print-length* *print-length*
-            *print-level* *print-level*
-            *print-namespace-maps* *print-namespace-maps*]
-    (set! *print-length* nil)
-    (set! *print-level* nil)
-    (set! *print-namespace-maps* false)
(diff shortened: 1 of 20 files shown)


crux-core/src/crux/io.clj
```diff
@@ -219,3 +219,9 @@
        ~@body
        (finally
          (.unlock lock# stamp#)))))
+
+(defn prn-edn [xs]
```
hraberg: LGTM, though nitpick - this function should be called `pr-edn-str` or something similar - the `n` prefix implies that there would be a newline (see `pr` vs `prn` etc.) So a function called `prn-edn` sounds like it prints edn to standard out, followed by newline.

hraberg (changes requested):

crux-core/src/crux/io.clj
danmason: Alright, seems entirely reasonable! :)

Issue #447, opened 2019-11-29
We've published and documented an end-to-end Crux/GraphIQL example


jarohen:
* User should be able to run `docker-compose up -d` and get a GraphIQL webserver running on http://localhost:3000 (i.e. delete the haproxy stuff), serving the Star Trek data set
* Shouldn't contain any data that isn't publicly releasable
* A blog running the user through what's possible, and an overview of how they might be able to write their own GraphQL service over a Crux node (how to write the resolvers, etc)

Eventually we might release the GraphQL-to-Crux-adapter as a general library, but let's assume the user can manually write resolvers for the time being.

coconutpalm:
Status on this work?  (Thanks in advance.)

armincerf:
> Status on this work? (Thanks in advance.)

Have a look at Site https://github.com/juxt/site

Still a work in progress, currently it works well if you don't have many authz requirements and is in use in production, but there is a big piece of ongoing work in the 'actions' branch which adds comprehensive authn/authz abilities

There's also a juxt-site channel here https://clojurians.slack.com/archives/C02RDUVDEP2

jarohen:
I'll close this in the XT repo, given this is pretty much what Site's for now - it didn't exist when I wrote this :smile:

coconutpalm:
Thanks!

PR #452, opened 2019-12-04
Add Graphql webservice example under docs/example/ 

danmason:
(Issue #447)

Cut down and cleaned up to go into the repo, serving the star trek dataset.


jarohen:
as discussed, going to park this temporarily until the SPoC changes are more stable

jarohen:
thanks @danmason - another good PR, but again, let's wait for the underlying deployment work to become more stable, then update this and re-open the PR

Issue #453, opened 2019-12-04
When nil is given as an attribute value it should error, rather than acting as :crux.db/id


t-taylor:
Example:
```
(api/submit-tx node [[:crux.tx/put {:crux.db/id :this}]])

(api/q (api/db node) {:find ['e] :where [['_ nil 'e]]})
;; => #{[:this]}
```

This should either be documented or throw an error.

t-taylor:
https://github.com/juxt/crux/pull/455

PR #455, opened 2019-12-05
Fix for nil query attribute #453

t-taylor:
Updates `:triple` spec with a nil check for the value


jarohen (changes requested):
hey @t-taylor - as discussed, could you remove the other changes from this PR? seems to have some of the logging changes here too

t-taylor:
All log/dev files gone now, I had forked from the logging testing branch so I had to do some funky rebase/cherry-pick stuff

jarohen (approved):

PR #459, opened 2019-12-09
Fix #457 no-op when nil transaction function

### LINKED ISSUE #457: Crux should no-op if a transaction function returns nil
Currently, it throws a spec error, expecting a vector - if it returns nil, it should behave the same as if an empty vector were returned.

t-taylor:
With the latest crux release the test included in the pull request fails. However on master it passes. Something someone did changed the behavior. I think I know what caused it, but I'll write up this when I've done a bit more research. This will be written on the issue card.

fixes #457

### DIFF
--- a/crux-dev/dev/tmt/noopnil.clj
+++ b/crux-dev/dev/tmt/noopnil.clj
@@ -1,14 +1,12 @@
 (ns tmt.noopnil
-  (:require [crux.api]))
+  (:require [crux.api :as api]))
 
-(def node (crux.api/start-node {:crux.node/topology :crux.standalone/topology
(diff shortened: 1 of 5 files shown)


crux-dev/dev/tmt/noopnil.clj
```diff
@@ -0,0 +1,41 @@
+(ns tmt.noopnil
```
danmason, line 1: Might want to remove the crux-dev file :)

jarohen (approved):
yep, I agree - given that test passes it seems the problem's gone on master. 

once you've removed the dev file from the PR, feel free to merge the test - cheers @t-taylor :)

PR #460, opened 2019-12-10
Tests for leaking logs

### LINKED ISSUE #396: Verify that current default logging outputs are not leaking data
Aim of this card:
* We shouldn't leak personally sensitive information in log files (passwords, personal data, etc)
* We shouldn't leak information in logs that would allow an attacker to attack the system in any way

* Crux should never log any user-provided document data
* Crux shouldn't log any arguments from queries

t-taylor:
Runs all tx and query operations and verifies that no document values are leaked in logs.
#396 is partially resolved by this.


crux-test/project.clj
```diff
@@ -33,6 +33,7 @@
                  [org.clojure/test.check "0.10.0"]
                  [org.slf4j/slf4j-api "1.7.26"]
                  [ch.qos.logback/logback-classic "1.2.3"]
+                 [org.codehaus.janino/janino "2.6.1"]
```
jarohen: what's this?

crux-test/test/crux/logging_test.clj
```diff
@@ -0,0 +1,58 @@
...
+
+(defn- sync-submit-tx [node tx-ops]
+  (let [submitted-tx (api/submit-tx node tx-ops)]
+    (api/sync node (:crux.tx/tx-time submitted-tx) nil)
+    submitted-tx))
+
+(t/deftest test-submit-tx-log
+  (let [secret 33489857205]
+
+    ;; Put
```
jarohen: you can wrap these in a `(t/testing "put" ...)` block - means that test failures (if/when they happen) contain the extra context information

crux-test/test/logback-test.xml
```diff
@@ -1,22 +1,46 @@
...
+        </filter>
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+        </encoder>
+    </appender>
+    <root level="WARN">
+        <appender-ref ref="STDOUT" />
+    </root>
+
+    <logger name="crux" level="trace" />
```
jarohen: these loggers form a tree - i.e. `"crux"` is a parent logger of `"crux.node"` - unless you override it, `crux.node` will take on the level of `crux` (and they'll both take on the root logger level if neither is provided)

crux-test/test/logback-test.xml
```diff
@@ -1,22 +1,46 @@
...
-  <logger name="crux.hash" level="WARN" />
-  <logger name="crux.http-server" level="WARN" />
-  <logger name="crux.kafka" level="WARN" />
-  <logger name="crux.jdbc" level="WARN" />
-  <logger name="crux.kv.memdb" level="ERROR" />
-  <logger name="crux.rdf" level="ERROR" />
-  <logger name="crux.tx" level="WARN" />
-  <logger name="org.apache.kafka" level="ERROR" />
-  <logger name="org.apache.zookeeper" level="ERROR" />
+    <appender name="FILE" class="ch.qos.logback.core.FileAppender">
```
jarohen: rather than changing the log levels of all the tests, it might be better here to wrap the test code in `wrap-redefs` to override the behaviour of the log function. if memory serves, all of the various `log/*` calls eventually call a lower-level function (`log/log`, or something like that) - you can then store whatever's logged into an atom and search through it from there.

jarohen (changes requested):

crux-test/project.clj
t-taylor: Left over from trying to get it in memory, good catch

crux-test/test/logback-test.xml
jarohen: hey @t-taylor - this is still going to output log files for all the other tests that don't have a `remove-log-file` fixture

crux-core/src/crux/query.clj
```diff
@@ -1120,7 +1120,15 @@
...
          {:keys [find where args rules offset limit order-by full-results?]} q-conformed
          stats (idx/read-meta kv :crux.kv/stats)]
-     (log/debug :query (cio/pr-edn-str q))
+     ;; Hide args from logging outputs
+     (log/debug :query (cio/pr-edn-str
+                         (update q :args
+                                 (fn [args]
+                                   (mapv
+                                     (fn [argsmap]
+                                       (into #{} (map (fn [[k v]] [k :hidden])
```
jarohen: I'm not sure why this is a set?

jarohen (changes requested):

PR #474, opened 2019-12-13
Consistent map hashing, fixes #362

### LINKED ISSUE #362: The SHA1 hash of maps should not depend on the key ordering

jarohen:
A couple of changes from the reqmts in #362:
* SHA1 hashing maps without sorting them turned out to require hashing each value separately, rather than sorting-then-hashing, where you can just hash the whole map.
* because of how we actually use these hashes in practice, I don't think we need the backwards compatibility, nor the Kafka migration, nor the tool (although it'd no doubt be useful)

right. first, where do we use these hashes, and what properties do we need of them?

1. In the Kafka consumer, each message on the transaction log has a set of content-hashes, each message on the document log is keyed by the content-hash. The consumer waits for all the documents of a transaction to be in the object store before indexing the transaction
2/ In the KV indices, we look up AV and get EC (content-hash). we look up the content hash in the bitemp (EVtTtC) index to check whether it's current, and in the object store to get the underlying document - these three need to be consistent, and also consistent with the hashes in incoming queries/entity reqs. 

importantly, while the indexer does calculate the hashes of the eids/values, to insert into the indices, I don't think it re-calculates the document content-hash - it trusts the one that it gets in from Kafka.so, 

given we've got an index version bump going in anyway, we can assume that users are going to have to reindex their KV stores - which means that the new KV stores can be indexed using the consistent map-hashing. this means that the eid/value hashes in the updated AVEC/EVtTtC indices will match each other, and the query parameters. there'll be some old-style content-hashes in the 'C' fields, and some new-style, but old and new will both match with what's in the object store.

on Kafka, content-hashes don't need to match between transactions - so it's fine that we have old transactions with new in the same Kafka instance.

upshot is that we just need to ensure that, whenever we need to hash an object, we post-walk and sort all of the maps, so that they serialise consistently, and hence hash consistently. success :smile:

### DIFF
--- a/crux-core/src/crux/codec.clj
+++ b/crux-core/src/crux/codec.clj
@@ -4,12 +4,13 @@
             [crux.memory :as mem]
             [crux.morton :as morton]
             [taoensso.nippy :as nippy]
-            [crux.io :as cio])
-  (:import [clojure.lang IHashEq IPersistentMap Keyword]
+            [crux.io :as cio]
+            [clojure.walk :as walk])
+  (:import [clojure.lang IHashEq Keyword]
            [java.io Closeable Writer]
            [java.net MalformedURLException URI URL]
            [java.nio ByteOrder ByteBuffer]
-           [java.util Arrays Date UUID]
+           [java.util Arrays Date Map UUID]
            [org.agrona DirectBuffer ExpandableDirectByteBuffer MutableDirectBuffer]
            org.agrona.concurrent.UnsafeBuffer))
 
@@ -98,6 +99,9 @@
   (memoize
    (fn []
      (mem/->off-heap nil-id-bytes (mem/allocate-unpooled-buffer (count nil-id-bytes))))))
+
+(defn- sort-maps [o]
+  (->> o (walk/postwalk (fn [o] (cond->> o (instance? Map o) (into (sorted-map)))))))
 
 (defn id-function ^org.agrona.MutableDirectBuffer [^MutableDirectBuffer to bs]
   (.putByte to 0 (byte id-value-type-id))
@@ -196,7 +200,7 @@
   (value->buffer [this ^MutableDirectBuffer to]
     (if (satisfies? IdToBuffer this)
       (id->buffer this to)
-      (doto (id-function to (nippy/fast-freeze this))
+      (doto (id-function to (-> this sort-maps nippy/fast-freeze))
         (.putByte 0 (byte object-value-type-id))))))
 
 (defn ->value-buffer ^org.agrona.DirectBuffer [x]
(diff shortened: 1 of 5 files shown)


crux-core/src/crux/codec.clj
```diff
@@ -99,6 +100,9 @@
    (fn []
      (mem/->off-heap nil-id-bytes (mem/allocate-unpooled-buffer (count nil-id-bytes))))))
 
+(defn- sort-maps [o]
+  (->> o (walk/postwalk (fn [o] (cond->> o (map? o) (into (sorted-map)))))))
```
hraberg: This should probably be `instance? Map` as `map?` just checks for `IPeristentMap`? (Similar to how the change was done below for `id->buffer`.)

hraberg (changes requested):

hraberg (approved):

jarohen:
thanks @hraberg :smile:

refset:
I think one issue might be that users could want to run a CAS against an old document version which would require recalculating the old hash, unless they retrieved and used the hash via the history APIs.

Update: I am now satisfied that all is well.

PR #475, opened 2019-12-16
Fix crux.api protocol docstring indentation

SevereOverfl0w:
To match https://github.com/bbatsov/clojure-style-guide#docstring-indentation

Not doing so was wrecking havoc with editor display of docstrings.

### DIFF
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -66,25 +66,25 @@
     [node ^Date valid-time ^Date transaction-time]
     "Will return the latest value of the db currently known. Non-blocking.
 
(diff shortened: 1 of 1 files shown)


jarohen:
thanks @SevereOverfl0w - is this not two spaces relative to the surrounding block?

SevereOverfl0w:
You mean to the opening of the string? Not as far as I can tell from the behavior of tooling.

Tooling wants a consistent string, and knows nothing about the start index of the string.

jarohen:
hmm - TIL :smile: 

would be helpful if they did something like Kotlin's `trimIndent`, but guessing that's a _much_ wider change. will merge, cheers :)

PR #476, opened 2019-12-16
Adds a crux-docker module, with a docker instance for a HTTP node

### LINKED ISSUE #469: We've published a Crux HTTP docker image to the Docker Hub
We'd like users to be able to easily start up a Crux HTTP node via Docker.

Requirements:
* Ideally, no config set-up required by the user for a basic node (standalone Crux, HTTP on port 8080) beyond `docker run -p 8080:8080 juxt/crux` (or whatever the image's called)
* Further config to be supplied by an EDN file injected into the docker container, either as a volume or by extending the Dockerfile.
* Config file to be `/etc/crux.edn`, Crux data under `/var/lib/crux`, any required Crux source under `/usr/lib/crux`, so that we can re-use a similar deployment for other Linux package managers
* Docker instructions available on the Docker Hub page

danmason:
fixes #469

### DIFF
--- a/crux-docker/README.md
+++ b/crux-docker/README.md
@@ -2,26 +2,26 @@ If you wish to use Crux in Docker, you have two options:
 - Crux can be embedded within your JVM application - you don't need our Docker images for this, you can include Crux as a library and deploy it as you would any other JVM Docker application (see the [docs](https://opencrux.com/docs))
 - If you want to access Crux via HTTP, you want this Docker image, instructions below.
 
-The following assumes that you are copying the contents of this folder and intend to create your own node configuration. If you wish to see how to **interact** with a basic, non-configured setup, check out the overview on [docker hub](https://hub.docker.com/r/juxt/crux-http). 
+The following assumes that you are copying the contents of this folder and intend to create your own node configuration. If you wish to see how to **interact** with a basic, non-configured setup, check out the overview on [docker hub](https://hub.docker.com/r/juxt/crux-http).
 
---- 
+---
 ## Getting Started
 
-The only requirement for setting up this node is **docker**, so ensure that is installed. Firstly, if you have not already done so, 
+The only requirement for setting up this node is **docker**, so ensure that is installed. Firstly, if you have not already done so,
 download the most recent image of the http node from docker hub:
 
 ```bash
 docker pull juxt/crux-http
 ```
 
-To run the basic instance of the node (which runs using the **memdb**), run the script within `./bin/dev.sh`. This image makes use of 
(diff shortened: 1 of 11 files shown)


crux-docker/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+```clojure
+{:crux/node-opts {:crux.node/topology :crux.standalone/topology
+                  :crux.node/kv-store "crux.kv.memdb/kv"
+                  :crux.kv/db-dir "/var/lib/crux/db"
+                  :crux.standalone/event-log-dir "/var/lib/crux/events"
+                  :crux.standalone/event-log-kv-store "crux.kv.memdb/kv"}
+ :crux/server-opts 
+ {:cors-access-control
+   [:access-control-allow-origin [#".*"]
```
jarohen: map?

crux-docker/README.md
```diff
@@ -0,0 +1,70 @@
...
+    :access-control-allow-headers ["X-Requested-With"
```
jarohen: do we need all these headers on CORS requests? reckon this has been copied from somewhere? :wink:

crux-docker/README.md
```diff
@@ -0,0 +1,70 @@
...
+## Getting Started
+
+The only requirement for setting up this node is **docker**, so ensure that is installed. Firstly, if you have not already done so, 
+download the most recent image of the http node from docker hub:
+
+```bash
+docker pull juxt/crux-http
+```
+
+To run the basic instance of the node (which runs using the **memdb**), run the script within `./bin/dev.sh`. This image makes use of
```
jarohen: can we put the `docker run` command in here, so that users don't need to clone the repo for the most basic use case? (I don't think you even need to `docker pull` in that case, `docker run` might do it for you)
then, the first step of 'customising the node' becomes 'yeah, best clone this repo, cd crux-docker'

jarohen (changes requested):
nice one Dan - few suggestions for the docs there, once those are sorted we're good to merge :smile: 

guessing the Docker Hub README has to be updated over there?

crux-docker/README.md
jarohen: ah, I see - those instructions are on the Docker Hub page, guessing on the assumption that people coming here won't mind cloning the repo.

danmason: Yeah, exactly ^. The 'basic' version is on the docker hub.

crux-docker/README.md
danmason: You'd be right - the console ;) Wanted to try to cover as much ground as possible, though I suppose I'll make it as minimal as possible and get on to updating the docs with more info about it

PR #477, opened 2019-12-16
Evict failed CaS docs, #371

### LINKED ISSUE #371: Failed CAS operations can leave unevictable documents
Execute the following sequence on a clean instance (I'm running from source, revision 8ed4cf4aacc5f478644f21ed694706640b54e8c1):

```clojure
(crux/submit-tx node [[:crux.tx/put {:crux.db/id :doc-a, :value 0}]])
(crux/submit-tx node [[:crux.tx/cas {:crux.db/id :doc-a, :value 1}, {:crux.db/id :doc-a, :value 2}]])
(crux/submit-tx node [[:crux.tx/put {:crux.db/id :doc-a, :value 3}]])
(crux/submit-tx node [[:crux.tx/evict :doc-a]])
```

Now inspect your document log (I use the JDBC backend for easy access). The first and last revisions of `:doc-a` will have been replaced by tombstones. The two that were part of the CAS operation never got indexed, so they're still in there.

Half of the problem could perhaps be solved by not fully ingesting the first argument to CAS: we only ever care about the id and if it's not already in the log, the operation can't possibly succeed anyway.

The second argument seems like a much deeper problem, at least assuming that it's a design rather than implementation issue. I can speculate on two high-level approaches off the top of my head.

1. Somehow allow the eviction process to find all proposed revisions to a document, regardless of whether they were ultimately accepted and indexed.
2. Have some other garbage-collection-like mechanism to identify orphaned documents in the document log and redact them proactively.

jarohen:
fixes #371 - we evict docs from failed CaS txs immediately.

I considered keeping a log of which docs were part of the failed transactions and only evicting them when the corresponding entities were deleted, but then I found out that we filter failed CaS txs out of the tx-log, so there's no point keeping the documents around for posterity.

### DIFF
--- a/crux-core/src/crux/tx.clj
+++ b/crux-core/src/crux/tx.clj
@@ -34,6 +34,14 @@
         (or (nil? end)
             (neg? (compare % end)))))
 
+
+(defn tx-event->content-hashes [[op id & args]]
+  (->> args
+       (take (case op
+               :crux.tx/put 1
+               (:crux.tx/delete :crux.tx/fn :crux.tx/evict) 0
+               :crux.tx/cas 2))))
+
 (defn- put-delete-kvs [object-store snapshot k start-valid-time end-valid-time transact-time tx-id content-hash]
   (let [eid (c/new-id k)
         ->new-entity-tx (fn [vt]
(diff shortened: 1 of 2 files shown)


hraberg (changes requested):
I don't think we can just evict stuff like this. My preference is to find the things to evict via the lower-level prefix scans in `crux.index` and bypass the history API.

hraberg:
To be clear, a document in a failed transaction might exist already in the index, we cannot just evict it. Also, we don't want to use eviction unless we strictly have to, there are too many unresolved issues around it, like the race for example, which will eventually block an entity from even coming back into existence.

jarohen:
agreed, closing, wrong approach.

Issue #158, opened 2019-02-27
Graal Native Image Support


hraberg:
We want `crux.main.graal` to delegate to `crux.bootstrap.cli` like the normal `crux.main` namespace. This is currently blocked by this issue which stops Kafka clients from being used inside a native image:
https://github.com/oracle/graal/issues/532

We should track this issue and check new releases of Graal and see if we can move `crux.main.graal` to do the right thing.

hraberg:
Related to #155.

hraberg:
For reference, if someone else decided to look at this, there's a script to build the native image here:
`dev/build-graal-native-image.sh`

hraberg:
For reference:
https://github.com/oracle/graal/blob/master/substratevm/LIMITATIONS.md

hraberg:
Also, we currently use Clojure 1.9 in the Graal profile. This due to this Clojure bug:
https://dev.clojure.org/jira/browse/CLJ-1472

This bug is in 1.9 as well, but that version of spec doesn't use the `locking` macro, which is where the bug arises. When this ticket is fixed we can move the Graal profile to 1.10. The real fix will be in Clojure, but a potential fix might turn up in spec which circumvents it.

A potential alternative is to downgrade spec to the version in 1.9, but there might be other issues with that.

hraberg:
Also worth noting that Graal itself is currently a 1.8 based JVM.

hraberg:
As JNR doesn't work in the native image, we could consider cutting out JNI (and JNR) entirely and use the`org.graalvm.nativeimage.c` which is only available inside the native image to call C directly from Java.

This blog post and repository is the best example:
https://cornerwings.github.io/2018/07/graal-native-methods/
https://github.com/cornerwings/graal-native-interaction

It's somewhat documented by this test:
https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.tutorial/src/com/oracle/svm/tutorial/CInterfaceTutorial.java

The JavaDocs are here:
http://www.graalvm.org/sdk/javadoc/

The graal profile would need this dependency (or later):
`[org.graalvm.sdk/graal-sdk "1.0.0-rc12"]`

refset:
Graal has moved on slightly since it's latest big release so our use of it will have to be updated.

gklijs:
With quarkus it's possible to have a navive compiled kafka client, I'm not sure how easy this can be integrated with Clojure https://quarkus.io/guides/kafka-guide

PR #478, opened 2019-12-16
tx.clj refactorings

jarohen:
This PR's got a fair few different bits in it, sorry - might be worth reviewing the individual commit diffs. 

Particularly:
* I've dropped the log level of 'failed' transactions down to a debug - errors/warns are for logs that should alert and be actioned by developers
* `tx-command-*` have become a multimethod - it's a textbook case for one, and it means we can remove the `tx-command->op` map (which was essentially implementing multimethods) and the `declare` that was required because the tx-fn needed to recursively call the multimethod.
* subjective: I've refactored some of the functions in that namespace to take a single data structure rather than a collection (see https://stuartsierra.com/2015/08/10/clojure-donts-redundant-map).
* @hraberg - `in-range-pred`'s gone, you'll be pleased to know :smile:

### DIFF
--- a/crux-core/src/crux/moberg.clj
+++ b/crux-core/src/crux/moberg.clj
@@ -215,7 +215,7 @@
     (s/assert :crux.api/tx-ops tx-ops)
     (doseq [doc (tx/tx-ops->docs tx-ops)]
       (db/submit-doc this (str (c/new-id doc)) doc))
-    (let [tx-events (tx/tx-ops->tx-events tx-ops)
+    (let [tx-events (map tx/tx-op->tx-event tx-ops)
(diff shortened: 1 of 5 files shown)


hraberg:
Happy to see in-range gone :-) Might be worth knowing that the commands used to be multi-methods but it changed due to to some Graal issues way back in #158, but that's OK to ignore now.

I take your reasoning about the debug level, but the levels were higher so one had a chance to see this happening somewhere somehow. Arguably this would be better reported in another way (hooks, other topics)

hraberg (changes requested):
Looks good, as discussed, I would rather keep the log levels as were until we have a clear different way of seeing it. I can take the argument that the user cannot do much about it, but still.

jarohen:
@hraberg commit removed, will merge :)

PR #501, opened 2019-12-18
Fix performance issue caused by not releasing KV iterators in put/delete ops

jarohen:
We weren't releasing KV iterators back to the snapshot's pool after put/delete ops, so each put/delete in a transaction was creating a new iterator, which is expensive.

This became more of an issue after the put/delete semantics were changed - previously, puts/deletes (incorrectly) didn't read any of the indices before putting their KVs into the store.


jarohen:
paired with @hraberg, so will merge it

Issue #466, opened 2019-12-13
Users can await a tx using the map returned from submit-tx


jarohen:
* [x] Update `sync` zero-arg to sync to the latest transaction as of the function call (i.e. not chasing the producer)
* [x] Added 'await-tx' function on node, takes a transaction map and an optional timeout. 
* [x] Added 'await-tx-time' function on node, same as current `sync` to tx-time arity, deprecate `sync` arity
* [x] Updated tests to use new functions
* [x] Updated crux docs

alexanderkiel:
What would `await-tx` return? Could it return the outcome of the transaction we wait for? Or how do I test wether a transaction I just submitted was successful?

refset:
Hi @alexanderkiel the main purpose of `await-tx` will be to block the thread while ingestion takes place (essentially like `sync` currently does), and it will probably at least return the same map of `tx-time` and `tx-id` that was passed in. Returning the outcome of the transaction as well (i.e. `success`/`fail`) seems like a reasonable suggestion to me at least - thanks for chiming in!

> how do I test whether a transaction I just submitted was successful?

You can achieve this currently (though perhaps not as elegantly) using either `submitted-tx-updated-entity?` or `submitted-tx-corrected-entity?`: https://www.opencrux.com/docs#_submitted_tx_updated_entity

alexanderkiel:
Hi @refset. Thanks for pointing to `submitted-tx-updated-entity?`. In case the only way a transaction can fail is through `:crux.tx/cas` as it is currently in my understanding, using `submitted-tx-updated-entity?` on every CAS op should give me the information I need.

If you consider to implement a one-shot transaction failure detection, you have to also provide the reason for the failure. Here that would be the CAS op that failed. From looking at the code, I don't see that you have this information stored explicitly. So it might be actually favorable to have to test for each CAS update.

But still, given that a transaction can either fail or succeed, to me it feels a bit strange, that I can't simply ask Crux for this information.

refset:
> the only way a transaction can fail is through `:crux.tx/cas`

That's correct as things stand today (ignoring transaction functions, which are experimental and hence undocumented).

> given that a transaction can either fail or succeed, to me it feels a bit strange, that I can't simply ask Crux for this information

[...]

PR #502, opened 2019-12-18
tx.clj refactorings, take 2

jarohen:
(re-opening #478 because there's a bug somewhere in Moberg that causes tests to fail intermittently on this branch)

This PR's got a fair few different bits in it, sorry - might be worth reviewing the individual commit diffs. 

Particularly:
* `tx-command-*` have become a multimethod - it's a textbook case for one, and it means we can remove the `tx-command->op` map (which was essentially implementing multimethods) and the `declare` that was required because the tx-fn needed to recursively call the multimethod.
* subjective: I've refactored some of the functions in that namespace to take a single data structure rather than a collection (see https://stuartsierra.com/2015/08/10/clojure-donts-redundant-map).
* @hraberg - `in-range-pred`'s gone, you'll be pleased to know :smile:

### DIFF
--- a/crux-core/src/crux/moberg.clj
+++ b/crux-core/src/crux/moberg.clj
@@ -215,7 +215,7 @@
     (s/assert :crux.api/tx-ops tx-ops)
     (doseq [doc (tx/tx-ops->docs tx-ops)]
       (db/submit-doc this (str (c/new-id doc)) doc))
-    (let [tx-events (tx/tx-ops->tx-events tx-ops)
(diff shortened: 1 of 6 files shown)


jarohen:
@hraberg two more commits since you reviewed last - some of the tests were failing because 'sync' was only syncing to the tx-ms, rather than to the tx-id. I've temporarily added in a low-level `db/await-tx` function, used in the test fixture, which we'll be able to use when #466 comes around (don't want to block this release on getting all that implemented/documented/deprecated)

hraberg (approved):

PR #506, opened 2019-12-19
Being clearer about iterator lifetimes in EntityAsOfIndex

jarohen:
@hraberg - regarding your suggestion earlier, I've gone for a temporary fix of getting callers to `new-entity-as-of-idx` to provide (and hence, determine the lifetime of) the KV iterator - thoughts?

### DIFF
--- a/crux-core/src/crux/query.clj
+++ b/crux-core/src/crux/query.clj
@@ -1114,6 +1114,7 @@
      (if (instance? Throwable result)
        (throw result)
        result)))
+
   ([{:keys [object-store conform-cache kv valid-time transact-time] :as db} snapshot q]
    (let [conformed-query (normalize-and-conform-query conform-cache q)
          q (.q-normalized conformed-query)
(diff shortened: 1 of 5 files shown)


crux-core/src/crux/query.clj
```diff
@@ -1122,36 +1123,37 @@
...
-                      :let [bound-result-tuple (for [var find]
-                                                 (bound-result-for-var snapshot object-store var->bindings join-keys join-results var))]]
-                  (if full-results?
-                    (build-full-results db snapshot bound-result-tuple)
-                    (mapv #(.value ^BoundResult %) bound-result-tuple)))
-         order-by (cio/external-sort (order-by-comparator find order-by))
-         (or offset limit) dedupe
-         offset (drop offset)
-         limit (take limit))))))
+     (with-open [i (kv/new-iterator snapshot)]
```
hraberg: This should not use with-open here as this is what returns the lazy seq and this iterator needs to be open as long as the snapshot is, so we should just create it (like before) and let it live until the snapshot gets closed, like most iterators created by the query engine.

crux-core/src/crux/tx.clj
```diff
@@ -99,13 +99,14 @@
...
 
                          (->> (cons start-valid-time
-                                    (with-open [entity-as-of-idx (idx/new-entity-as-of-index snapshot start-valid-time transact-time)]
-                                      (when-let [visible-entity (some-> (idx/entity-at entity-as-of-idx eid)
-                                                                        (select-keys [:tx-time :tx-id :content-hash]))]
-                                        (->> (idx/entity-history-seq-ascending snapshot eid start-valid-time transact-time)
-                                             (remove #{start-valid-time})
-                                             (take-while #(= visible-entity (select-keys % [:tx-time :tx-id :content-hash])))
-                                             (map #(.vt ^EntityTx %))))))
+                                    (with-open [i (kv/new-iterator snapshot)]
```
hraberg, line 102: Similar, as this returns a lazy sequence you want to realise the sequence before the with-open block ends. Using with-open in this case is fine though. Worth noting that this issue was already there.

hraberg (changes requested):
Approach is OK, but a couple of changes needed as far as I can see, see line comments.

crux-core/src/crux/tx.clj
jarohen: hmm, interesting that it passed all the tests then :grimacing: anyway, mapv'd, cheers

crux-core/src/crux/query.clj
jarohen: thanks - yep, with-open removed and inlined binding. cheers

hraberg (approved):

Issue #505, opened 2019-12-19
Query engine should perform more consistently for simple range queries


refset:
The query engine should give more consistent performance when performing simple seeks, and ideally the first value returned by the query should align with the range (i.e. to be able to use `:limit 1` without having to use `:order-by` which materializes the entire set of results)

```
(for [i (range 5000)]
  (c/submit-tx node [[:crux.tx/put {:crux.db/id (java.util.UUID/randomUUID) :offer i}]]))


{:find [i],
 :where [[_ :offer i] [(test-pred i j)]],
 :args [{j test-val}],
 :limit 1}


[= 2]
#{[2]} ; slow

[< 10]
#{[9]} ; slow


[< 9223372036854775807]
#{[2690]} ; fast

[< -100]
#{} ; slow

[> -100]
#{[2690]} ; fast
```

hraberg:
This is related to #348 as if even if one is using this as documented above, there's no guarantee that Crux picks a useful join order based on the range constraints present in the query.

refset:
To summarise:

1. my test logic mistakenly used `:args`, which meant the range predicates didn't get recognised as range predicates - see the note in the committed test (also this needs documenting)
2. `<` range predicates never use reverse prefix seek iteration, which means you can't rely on them when using `limit` -- therefore you can either reverse your values before submitting it (i.e. `(- 0 i)`), or even store your data in both forwards AND reverse directions, so that it can be efficiently queried using `limit`. This is useful for pagination but is still not necessarily a complete pagination solution. See also https://github.com/xtdb/xtdb/discussions/1514#discussioncomment-758943
3. as things stand you should aim to stream range scans in and out of independent queries if you really need to guarantee efficient order/limit behaviours, rather than relying on the Datalog picking optimal sorted indexes (the Crux query engine has been designed around optimal joins, not sorting)

The main reason for 2) is that iterating in reverse order with prefix seeks has more limited (/inefficient) support within RocksDB:
https://github.com/facebook/rocksdb/wiki/Prefix-Seek-API-Changes#limitation
https://github.com/facebook/rocksdb/wiki/SeekForPrev
https://groups.google.com/g/rocksdb/c/Gq7b99-9Iq4

...however it seems LMDB can do this efficiently:
https://www.openldap.org/lists/openldap-technical/201502/msg00149.html

PR #508, opened 2019-12-20
Adding equals as a range predicate and showing intended literal range behaviour

hraberg:
(Issue #505)

Initial report likely used arguments to supply the range constraints, which would always result in using predicates, which will do the filtering purely in Clojure and have no way of knowing how to skip parts of the index. Test added showing it works as expected with literal constraints.

As a bonus, finally added `=` as a kind of range predicate. This predicate, unlike the ones depending on order, is also allowed on entity vars.

### DIFF
--- a/crux-core/src/crux/query.clj
+++ b/crux-core/src/crux/query.clj
@@ -55,7 +55,7 @@
 (s/def ::rule (s/and list? (s/cat :name (s/and symbol? (complement built-ins))
                                   :args (s/+ any?))))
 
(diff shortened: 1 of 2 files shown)


jonpither (approved):

jarohen (approved):
not 100% familiar with this area of the codebase as yet, but from what I do understand, looks good to me :)

jarohen:
rebased to run the CircleCI checks

PR #513, opened 2019-12-20
Ensuring we don't accidentally re-tag a previously pushed version of crux-http

jarohen:
* `build.sh` currently tags the most recently built docker image with the tag from the most recent release (which is likely not what we want when we're developing)
* `push.sh` pushes all the user's tags for `juxt/crux-http`, which may well override previous versions if they've been re-built

Separated out into an explicit `tag.sh` script, and only pushing the most recent tag.


danmason (approved):
Seems sensible to me :)

PR #517, opened 2019-12-23
Remove sync to no lag

jarohen:
First part of #466, removing sync-to-no-lag (i.e. without a provided tx-time)

* Removed from public API/docs
* Removed from HTTP server/remote client
* Updated tests that used it to pass the last submitted tx through, and sync to that instead.

Intention is to hold off on merging this til we've had chance to discuss it after Christmas - in the meantime, though, I'll base my multi-threaded Kafka branch on it, so that I don't have to make it backward compatible for this function.

There were a couple of bench tests that used `future` but then seemed to immediately deref them, I've removed the futures - shout if there was a reason for this that I haven't thought of?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,12 +70,8 @@ jobs:
       - run:
           name: Install datasets for ts-weather-test and ts-devices-test
           command: |
-            cd crux-test/
-            cd resources
-            mkdir ts
-            cd ts
-            mkdir data
-            cd data
+            mkdir -p crux-test/resources/ts/data
+            cd crux-test/resource/ts/data/
             wget https://timescaledata.blob.core.windows.net/datasets/weather_small.tar.gz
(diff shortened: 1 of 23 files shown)


jarohen:
discussion with @hraberg on Slack - this functionality was originally so that users could await a new node becoming 'reasonably up to date'. problem with this is that producers could continue to product in the background, so the node might never catch up.

tbd: whether we should continue to support this. one potential solution is that we could request the most recently available tx-id at the start of the call, and then await _that_ transaction?

jarohen:
closing this, let's see if we can preserve something similar to the original behaviour. 

> one potential solution is that we could request the most recently available tx-id at the start of the call, and then await that transaction?
I reckon this solution is the most sensible, and fulfills the underlying user intention, but without the potential of being affected by subsequent transactions being submitted?

PR #519, opened 2019-12-23
Multi-threaded Kafka consuming

### LINKED ISSUE #461: The Kafka consumer uses multiple threads to ingest documents if there are multiple doc-topic partitions

jarohen:
fixes #461

This branch depends on #517 (although it may not need to, we may be able to find a better semantics for 'sync to zero lag'), let's discuss both that PR and this one after Christmas :christmas_tree: 

TODO:
- [ ] write up more details on this PR
- [ ] benchmark
- [ ] there're some intermittent errors closing the embedded Kafka cluster, will investigate

### DIFF
--- a/crux-core/src/crux/db.clj
+++ b/crux-core/src/crux/db.clj
@@ -23,6 +23,11 @@
   (read-index-meta [this k]))
 ;; end::Indexer[]
 
+(defn swap-index-meta [indexer k f & args]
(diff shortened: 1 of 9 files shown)


jarohen:
this one doesn't have as much impact as I'd hoped, certainly with RocksDB as the underlying KV store, because Rocks seems to perform much better when documents are ingested in a single thread

jarohen:
closing this for the time being

PR #520, opened 2019-12-28
New 'entity' arity accepts a snapshot argument for >40% performance boost

refset:


jarohen (approved):
nice one @refset - guessing this means you tracked down whatever it was that caused a lack of speedup?

also, any thoughts while you were doing this about what we should be exposing to users re snapshots? what we should call it, etc?

hraberg (approved):

PR #523, opened 2019-12-31
Update doc with memory-only standalone config

straux:
The current configuration shown in the documentation forces the write on local storage. This update would help to set up memory-only nodes.


jarohen:
Thanks for this PR, @straux :smile: 

We're shortly going to make the memory-only kv-store the default for the standalone node, so that crux-core can be used without any other modules, and advise users on how to swap it out for RocksDB - your desired behaviour will be the default when #659 is merged.

PR #524, opened 2020-01-02
Batch index docs into the KV store

jarohen:
Currently, we're indexing docs one-at-a-time into the KVIndexer - RocksDB, in particular, appreciates putting documents in sorted batches (see 'What's the fastest way to load data into RocksDB?' at https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ)

The microbench, on my laptop, takes ~11s to load 100k docs individually, but ~6.2s to load 100k docs in batches of 100 (batches of 1000 didn't make a significant difference)

### DIFF
--- a/crux-core/src/crux/tx/polling.clj
+++ b/crux-core/src/crux/tx/polling.clj
@@ -21,7 +21,9 @@
                     (if-let [^Message last-message (reduce (fn [last-message ^Message m]
                                                              (case (get (.headers m) :crux.tx/sub-topic)
                                                                :docs
-                                                               (db/index-doc indexer (.key m) (.body m))
+                                                               ;; TODO we'll likely be able to make this more efficient
+                                                               ;; if it can insert multiple docs at once
+                                                               (db/index-docs indexer {(.key m) (.body m)})
                                                                :txs
                                                                (db/index-tx indexer
                                                                             (.body m)
--- a/crux-core/src/crux/tx/polling.clj
+++ b/crux-core/src/crux/tx/polling.clj
@@ -17,24 +17,25 @@
                                                       :time nil}}))
   (while @running?
     (let [idle? (with-open [context (consumer/new-event-log-context event-log-consumer)]
-                  (let [next-offset (get-in (db/read-index-meta indexer :crux.tx-log/consumer-state) [:crux.tx/event-log :next-offset])]
-                    (if-let [^Message last-message (reduce (fn [last-message ^Message m]
-                                                             (case (get (.headers m) :crux.tx/sub-topic)
-                                                               :docs
-                                                               ;; TODO we'll likely be able to make this more efficient
-                                                               ;; if it can insert multiple docs at once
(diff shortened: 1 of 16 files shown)


crux-core/src/crux/tx/polling.clj
```diff
@@ -21,7 +21,9 @@
                     (if-let [^Message last-message (reduce (fn [last-message ^Message m]
                                                              (case (get (.headers m) :crux.tx/sub-topic)
                                                                :docs
-                                                               (db/index-doc indexer (.key m) (.body m))
+                                                               ;; TODO we'll likely be able to make this more efficient
```
jonpither: This comment still valid?

crux-test/test/crux/tx_test.clj
```diff
@@ -638,14 +638,14 @@
                (api/entity (api/db *api*) :foo))))))
 
 (t/deftest map-ordering-362
-  (t/testing "cas is independent of map ordering"
+  #_(t/testing "cas is independent of map ordering"
```
jonpither: Why commented?

crux-test/test/crux/tx_test.clj
```diff
@@ -638,14 +638,14 @@
...
     (sync-submit-tx *api* [[:crux.tx/put {:crux.db/id :foo, :foo :bar}]])
     (sync-submit-tx *api* [[:crux.tx/cas {:foo :bar, :crux.db/id :foo} {:crux.db/id :foo, :foo :baz}]])
 
     (t/is (= {:crux.db/id :foo, :foo :baz}
              (api/entity (api/db *api*) :foo))))
 
-  (t/testing "entities with map keys can be retrieved regardless of ordering"
+  #_(t/testing "entities with map keys can be retrieved regardless of ordering"
```
jonpither: Also why commented out?

jonpither (changes requested):

jonpither:
LGTM and given the performance gains, I'm pleased to see this PR land :-)

crux-test/test/crux/tx_test.clj
jarohen: ah, because it was only the bottom part of the test that was breaking, so I commented the top two out to focus on the breaking part (so that I didn't get too many logs printed)

will uncomment :)

crux-core/src/crux/tx/polling.clj
jarohen: I think so - we only get one doc message at a time here (without a refactor) so we can't make full use of index-docs

jarohen: refactored :) although did mean changing Moberg slightly, don't know what the appetite for this would be...? anyway, build's green - if you're both happy, I'm ready to merge :smile:

jonpither (approved):

hraberg (approved):
Feels like the microbench isn't really needed as it just reuses parts of weather test and isn't really run as it's not a `-test` namespace?

jonpither:
Just needs a note on the CHANGELOG

PR #525, opened 2020-01-02
Decorating topologies 522

### LINKED ISSUE #522: Topology modules can decorate other modules
For #384, we'd like users to be able to drop-in other topologies that augment the base topologies:
* Modules can decorate other modules - e.g. a metrics module can wrap the indexer and collect stats on how many documents are being indexed
* Modules can continue to override other modules (e.g. for configuration/testing purposes)
* The Crux node can be 'just another module' - this allows modules to depend on it (e.g. an HTTP server)
* We'd like the user to continue to be able to start a Crux node using a plain data structure (i.e. no user code required to wire up the dependencies)
* We'd like to remain backwards compatible - existing calls to `node/start` should still run

e.g. for metrics, we'd like the user to be able to write:
```clojure
{:crux.node/topology [crux.kafka/topology crux.metrics/topology]
 :crux.metrics/opts {...}}
```
with a metrics topology that looks something like:
```clojure
(def metrics-topology
  {:crux.metrics/indexer {:start-fn (fn [{:keys [crux.node/indexer] :as deps} args]
                                      (reify
                                        db/Indexer
                                        ...
                                        Closeable
                                        ...))
                          :wraps :crux.node/indexer}

   :crux.metrics/query {:start-fn (fn [deps args]
                                    ...)
                        :wraps ...}})
```

Open question/thoughts: does this belong in Crux? Maybe none of the Crux modules (in core, at least) should care about how they're wired up, just that they accept what they need to accept, and close when requested?

PoC TODO:
* [X] given a list of topologies, return a module start order that takes overrides and decorators into account
* [ ] starting a Crux node using these new topologies (moving the topology code out of the node ns)
* [ ] updating the Crux node tests
* [ ] updating documentation

t-taylor:
Resolves #522 

Allows wrapping of modules using topolgies.

### DIFF
--- a/crux-core/src/crux/node.clj
+++ b/crux-core/src/crux/node.clj
@@ -301,3 +301,24 @@
                     :object-store object-store
                     :closed? (atom false)
                     :lock (StampedLock.)})))
+
+(def node-topology
+  {:crux/node {:start-fn (fn [{::keys [kv-store tx-log indexer object-store] :as deps} {::keys [node-opts] :as args}]
+                           (let [status-fn #(apply merge (map status/status-map (cons (crux-version) (vals deps)))) ]
+                             (map->CruxNode {:close-fn close-fn
+                                             :status-fn status-fn
+                                             :options node-opts
+                                             :kv-store kv-store
+                                             :tx-log tx-log
+                                             :indexer indexer
+                                             :object-store object-store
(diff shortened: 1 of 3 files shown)


jarohen:
as discussed, let's park this and investigate whether this would be more appropriate as explicit hooks

PR #529, opened 2020-01-06
Fix overlapping time ranges within a single transaction, #434

### LINKED ISSUE #434: Overlapping valid-time ranges for a single entity in a transaction should be merged
After we've clarified the put/delete valid-time range semantics #326, we should also clarify what happens when a user has two operations on the same entity-id within the same transaction.

Currently, each operation within a transaction effectively sees the database as at the start of the transaction, so any subsequent puts/deletes for the same entity id won't override earlier puts/delete KVs in the same transaction, as required.

If there are overlapping ranges for a single entity, the effects of the operations should be merged before writing into the KV store indices.

jarohen:
fixes #434 

Currently, if a user specifies overlapping time ranges for the same entity within a transaction, each operation only sees the database as at the start of the transaction - later operations do not correctly overwrite earlier operations within the transaction. Likewise, CaS/transaction functions currently compare against the start of the transaction.

This change ensures that operations can see (and hence correctly overwrite) earlier operations within the transaction, but without persisting those operations to the DB until the transaction commits.

N.B.:
* Might help to start with the [new test](https://github.com/juxt/crux/pull/529/files#diff-e9248c848b198b5586ceb964d0f2ede7R667)
* This [old test](https://github.com/juxt/crux/pull/529/files#diff-8d0564fb1283009c592ba069490c0f15R1424) shows that the CaS behaviour has changed for two CaS operations on the same entity in the same tx
* Related: in an isolated commit, I've also refactored the KVIndexer to pass around maps of values that commonly travel together (the 'Parameter Object' pattern), and given the keys/destructured locals names consistent with the rest of Crux. particularly, settling on `kv-store` (rather than `kv`) and `tx-time` (rather than `transact-time`) because of the keywords `:crux.node/kv-store` and `:crux.tx/tx-time` respectively. This also helps when there are more than ~4-5 positional args - don't have to explicitly pass all of these through, in a consistent order (or otherwise), every time. RFC!

### DIFF
--- a/crux-test/test/crux/query_test.clj
+++ b/crux-test/test/crux/query_test.clj
@@ -1421,7 +1421,7 @@
                                       :name "Ivan 3rd"}]
                                     [:crux.tx/cas
                                      {:crux.db/id :ivan
-                                      :name "Ivan 2nd"}
+                                      :name "Ivan 3rd"}
                                      {:crux.db/id :ivan
                                       :name "Ivan 4th"}]])
               updated? (api/submitted-tx-updated-entity? *api* submitted-tx :ivan)]
--- a/crux-core/src/crux/db.clj
+++ b/crux-core/src/crux/db.clj
@@ -17,7 +17,7 @@
 ;; tag::Indexer[]
 (defprotocol Indexer
   (index-docs [this docs])
-  (index-tx [this tx])
+  (index-tx [this tx tx-events])
   (docs-exist? [this content-hashes])
   (store-index-meta [this k v])
   (read-index-meta [this k]))
--- a/crux-core/src/crux/db.clj
+++ b/crux-core/src/crux/db.clj
@@ -17,7 +17,7 @@
 ;; tag::Indexer[]
 (defprotocol Indexer
   (index-docs [this docs])
-  (index-tx [this tx-events tx-time tx-id])
+  (index-tx [this tx])
   (docs-exist? [this content-hashes])
   (store-index-meta [this k v])
   (read-index-meta [this k]))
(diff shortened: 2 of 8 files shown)


crux-test/test/crux/query_test.clj
```diff
@@ -1421,7 +1421,7 @@
                                       :name "Ivan 3rd"}]
                                     [:crux.tx/cas
                                      {:crux.db/id :ivan
-                                      :name "Ivan 2nd"}
+                                      :name "Ivan 3rd"}
```
jarohen, line 1424: this change to an old test shows that the CaS behaviour has changed when you've got two CaS operations to the same entity in the same tx

refset (commented):
Suffice it to say I haven't dug into the full depth of `merge-histories` or the move from `kvs` to `etxs`, but it all appears sensible. To be explicit though, when you say "this change ensures that operations can see (and hence correctly overwrite) earlier operations within the transaction" - it only applies to the existing basic operations (as far I can tell). Transaction functions would still see the `db` value as-at (and as-of) the start of the transaction. This is good for now though - nice work!

hraberg (approved):
This is partly on trust, all tests pass, reusing the existing merge logic is a good choice. As mentioned elsewhere, I did consider the idea of simply not allowing overlaps and throw an exception, but let's go with this for now.

I don't like the `deps` field change of `KvIndexer` and would like the module for it to move to a better place (but this was already in `crux.node`), but that's my opinion, doesn't block merging.

But as discussed, it would be great to see some micro benchmark showing that the non-overlapping case isn't impacted performance-wise, and as a stretch goal, the overlapping one has "acceptable" speed.

Also noted that queries in tx-fns cannot see the new data. That was obviously already the case, but now might make the behaviour slightly inconsistent? One to reflect on.

hraberg (approved):

PR #532, opened 2020-01-06
Adding hooks

### LINKED ISSUE #422: Users can be notified when the indexer ingests documents/transactions
(original message below)

We'd like users to be able to receive notifications from the indexer (and other areas, in the future) so that modules can react to transactions flowing through the Crux node. Potential use cases to consider:
* Collecting metrics (#384)
* Subscribing to indexed transactions to update downstream consumers
* Re-publish the events to Kafka topics.

Potential solution - event bus:
* must be very light weight. it shouldn't be slow, but also not complex
* it should publish the events async, potentially in a thread per subscriber module (not subscription).
* events may be split by topics, or the filtering could simply be done by the listener. 
* the events need to adhere to specs - a few keys might be mandatory, like `crux.event-bus/topic` or similar.

----
A “hook” for processed txes would ease both user subscriptions to db updates and our work with #161 . It can be a hook, an event or a signal. In its most minimal form it's just a callback accepting tx meta `#:crux.tx{:tx-id 1, :tx-time 1}`, so users could leverage that when working with tx-log. 

Perhaps, an internal hook for accepted non-processed transactions could also ease development of new tx-log consumers, maybe even streamline the existing ones; but here I'm not so sure.

This issue is here to accumulate more thought and feedback from the core devs and the community.

t-taylor:
Resolves #422


crux-core/src/crux/api.clj
```diff
@@ -198,7 +200,9 @@
      (.sync this transaction-time timeout)))
 
   (attribute-stats [this]
-    (.attributeStats this)))
+    (.attributeStats this))
+  (add-query-hook! [this hook]
+    (.add_query_hook_BANG_ this hook)))
```
jarohen, line 205: in Java, this'd be `addQueryHook`

crux-core/src/crux/query.clj
```diff
@@ -1183,11 +1183,18 @@
   (newSnapshot [this]
     (lru/new-cached-snapshot (kv/new-snapshot (:kv this)) true))
 
+  ;; TODO query-hooks here
   (q [this q]
-    (crux.query/q this q))
+    (let [latter-fns (seq (map #(% {:query q}) query-hooks))
+          result (crux.query/q this q)]
+      (seq (map #(% {:result result :query q}) latter-fns))
```
jarohen, line 1190: reckon you're using `seq` here with the intention of forcing the sequence? (like the Haskell function, I think?)

in Clojure, `seq` only happens to force the _first_ element of a sequence (hence why your test passes) - you'd want something like `doseq`, `dorun`, or, better, just `run!` here (given you're only after the side effects)

(the one above, L1188, can remain lazy)

crux-test/test/crux/tx_test.clj
```diff
@@ -663,3 +663,17 @@
                (api/q (api/db *api*) '{:find [parent child]
                                        :where [[parent :crux.db/id _]
                                                [child :joins parent]]}))))))
+
+(t/deftest tx-doc-hooks-422
+
+  (t/testing "doc hooks"
+    (let [test-atom-doc (atom false)]
+      (db/add-doc-hook! (:indexer *api*) (fn [_] (fn [_] (swap! test-atom-doc not))))
```
jarohen, line 671: why the double fn here, out of interest?

crux-core/src/crux/query.clj
```diff
@@ -1183,11 +1183,18 @@
   (newSnapshot [this]
     (lru/new-cached-snapshot (kv/new-snapshot (:kv this)) true))
 
+  ;; TODO query-hooks here
```
jarohen, line 1186: TODO ^^ :smile: worth grepping for these before you PR

crux-test/test/crux/tx_test.clj
jarohen: oh, I see what's happening here - first function is the pre-hook, returns the function that you want to run after the transaction? interesting, I've not seen hooks implemented this way (and we probably shouldn't be inventing a new standard!)

for middleware, the hook usually accepts the underlying implementation as a parameter, and then has the opportunity to augment it both before and after the call. e.g.:

```clojure
(fn my-hook [f]
  (fn [params]
    ;; pre-hook work here, alter `params` if need be
    (let [ret (f params)]
      ;; post-hook work here, alter `ret` if need be
      ret)))
```

jarohen (changes requested):
thanks @t-taylor - couple of bits of feedback below. let me know if they're unclear, happy to pair?

that said, once it's clear, dw about actually making the changes I've requested - given the changes in #422 I'll close this PR

crux-core/src/crux/query.clj
t-taylor: `seq` does force the whole sequence, but I agree another function is better suited here

PR #534, opened 2020-01-06
Evict failed CaS docs #371

### LINKED ISSUE #371: Failed CAS operations can leave unevictable documents

jarohen:
fixes #371 

* using the `all-keys-in-prefix` as @hraberg suggested.
* ~This PR's based on #529 (because it would have been a hefty git conflict otherwise) - I'll merge them separately.~ done

### DIFF
--- a/crux-core/src/crux/tx.clj
+++ b/crux-core/src/crux/tx.clj
@@ -15,7 +15,7 @@
             crux.tx.event
             [crux.query :as q]
             [taoensso.nippy :as nippy])
-  (:import crux.codec.EntityTx
+  (:import [crux.codec EntityTx EntityValueContentHash]
            crux.tx.consumer.Message
            java.io.Closeable
            [java.util.concurrent ExecutorService Executors TimeoutException TimeUnit]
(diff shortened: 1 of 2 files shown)


hraberg (approved):
Note, I first missed the avec name style changes as they were done separately on master - not a problem, just confused me a bit as I thought there were missing commits here.

Issue #540, opened 2020-01-08
Provisioning the 'bench' infrastructure using CloudFormation


jarohen:
* [x] CloudFormation templates to provision the bench infrastructure, including IAM roles
* [x] Ensure we've removed any infrastructure created using the clicky-button UI

PR #544, opened 2020-01-09
Add aws crux instances for bench

### LINKED ISSUE #539: We can run a Docker task in ECS, and analyse the results in CloudWatch Insights
* [x] A 'hello-world'-like Clojure application that outputs a JSON log message and exits
* [x] Scripts to build a Docker image of that application, and deploy it to ECR
* [x] Script to run it as a one-off ECS task
* [x] Show that the log message appears in CloudWatch, and can be filtered/aggregated/etc using the Insights UI

t-taylor:
Resolves #539


jarohen (approved):
yep, this is all fine for now, thanks :smile: 

don't worry about merging this example, let's go for a fuller merge when you're done with #540/#541

PR #547, opened 2020-01-09
Greater than range bug

hraberg:
Reproduces and fixes issue occurring when using `>` queries against an empty database. Also includes a small refactoring to the `=` range predicate.

### DIFF
--- a/crux-test/test/crux/query_test.clj
+++ b/crux-test/test/crux/query_test.clj
@@ -2689,12 +2689,13 @@
          (api/q (api/db *api*)
                 '{:find [offset]
                   :where [[e :offset offset]
-                          [(> offset -9223372036854775808)]] ;; Long/MAX_VALUE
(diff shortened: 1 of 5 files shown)


jonpither:
CHANGELOG

jarohen (approved):
:+1: one minor nit - the value in your test is `LONG/MIN_VALUE` :wink:

crux-test/test/crux/query_test.clj
```diff
@@ -2683,3 +2683,18 @@
...
+
+(t/deftest test-greater-than-range-predicate-545
+  (t/is (empty?
+         (api/q (api/db *api*)
+                '{:find [offset]
+                  :where [[e :offset offset]
+                          [(> offset -9223372036854775808)]] ;; Long/MAX_VALUE
+                  :limit 1})))
+
+  (t/is (empty?
```
jarohen: separate test for this one? doesn't fit under `test-greater-than-range-predicate`?

Issue #527, opened 2020-01-03
Sync/await-tx then query over HTTP won't necessary hit the same node


jarohen:
If the Crux HTTP cluster is load-balanced, the sync/await (see #466) HTTP request will hit one node, wait for that node to become up-to-date, then the subsequent query HTTP request may hit a different node.

Let's say a user gets a `db` instance and then makes multiple calls (queries, history reqs, etc).

With the remote HTTP client, we aim that the API and behaviour (network instability aside) is as similar as possible to a local node. With a local node, users `await-tx` once, supplying a timeout - after waiting, they get a `db` instance and then make their calls. They know that once the `await-tx` returns, they're free to make whatever queries they like, without needing to wait again for the indexing to catch up.

With a remote node, though, if each of the subsequent requests hits a different node, they'll either have to accept the request failing, or await the transaction again, _for each request_ - i.e. the waiting has to be done on the individual request, rather than a single `await-tx` call. (This could be mitigated by things like session stickiness, but that's on the user to set up.)

Currently, if the `await-tx` call goes to one node, and a subsequent request goes to a node without that transaction indexed, the request will fail with a `NodeOutOfSyncException`.

Options:
1. We preserve this behaviour, bad luck.
2. We could remove sync/await-tx and friends, instead having this specified as part of the `db` call. For a remote connection, we could then opt to await for the specified timeout on _each_ call (which would still be subject to the errors above, but possibly less likely? possibly less predictable, too).
3. In addition to (2), to match the local behaviour as much as possible, we could experiment with something akin to a long-lived connection while the DB's open - i.e. encouraging using `open-db` rather than `db` for remote connections. Ironically, this would mean preserving await-tx on the HTTP API, so that the `db` call can await the transaction when it opens the connection.

refset:
My summary from the duplicate issue (now deleted - #615) :

> An await-tx combined with a non-sticky HTTP load-balancing doesn't make any guarantees that a subsequent query will be calculated against the awaited tx. This requires the user's client code to detect the synchronisation issue and retry the query.

Possible solutions:
* Add an await parameter to all read interfaces over HTTP (same as the proposal above)
* Allow the user to await a tx high-watermark maintained somewhere outside the Crux database

Related reading about read consistency, isolation levels, and "Online Event Processing": https://queue.acm.org/detail.cfm?id=3321612

[...]

PR #549, opened 2020-01-10
introducing await-tx/await-tx-time (#466 pt 1)

jarohen:
This is the non-breaking part of #466, adding in `await-tx`/`await-tx-time` - would like to merge this in advance of the breaking changes so that we can start to use this in tests, etc.

* adds `await-tx`/`await-tx-time` (non-breaking - keeping existing `(sync node tx-time)` for now)
* `sync` (with no args) awaits the latest submitted tx rather than trying to chase zero-lag
* internally, adds a `latest-submitted-tx` to TxLog protocol so that `sync` can find out what tx to await
* updated all the tests that used `(sync node tx-time)` to use `(await-tx node tx)` instead
* I deliberately haven't touched the HTTP sync endpoints (beyond preserving existing behaviour) because of #527 

coming up in part 2 (see #466) for more details:
* remove deprecated functionality
* update docs/tutorials
* timeout default change

### DIFF
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -117,18 +117,31 @@
   Returns true if the entity was updated in this transaction.")
 
   (sync
+    [node]
     [node ^Duration timeout]
-    [node ^Date transaction-time ^Duration timeout]
-    "If the transaction-time is supplied, blocks until indexing has
-  processed a tx with a greater-than transaction-time, otherwise
-  blocks until the node has caught up indexing the tx-log
-  backlog. Will throw an exception on timeout. The returned date is
-  the latest index time when this node has caught up as of this
-  call. This can be used as the second parameter in (db valid-time,
-  transaction-time) for consistent reads.
+    ^:deprecated [node ^Date transaction-time ^Duration timeout]
+    "Blocks until the node has caught up indexing to the latest tx available at
+  the time this method is called. Will throw an exception on timeout. The
+  returned date is the latest transaction time indexed by this node. This can be
+  used as the second parameter in (db valid-time, transaction-time) for
+  consistent reads.
 
   timeout – max time to wait, can be nil for the default.
   Returns the latest known transaction time.")
+
+  (await-tx-time
+    [node ^Date tx-time]
+    [node ^Date tx-time ^Duration timeout]
+    "Blocks until the node has indexed a transaction that is past the supplied
+  txTime. Will throw on timeout. The returned date is the latest index time when
+  this node has caught up as of this call.")
+
+  (await-tx
+    [node tx]
+    [node tx ^Duration timeout]
+    "Blocks until the node has indexed a transaction that is at or past the
+  supplied tx. Will throw on timeout. Returns the most recent tx indexed by the
+  node.")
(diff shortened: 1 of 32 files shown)


crux-core/src/crux/api.clj
```diff
@@ -117,19 +118,32 @@
...
+  (await-tx-time
+    [node ^Date tx-time]
+    [node ^Date tx-time ^Duration timeout]
+    "Blocks until the node has indexed a transaction that is past the supplied
+  txTime. Will throw on timeout. The returned date is the latest index time when
+  this node has caught up as of this call.")
+
+  (await-tx
+    [node tx]
+    [node tx ^Duration timeout]
```
hraberg, line 142: This map is usually called `submitted-tx` - not the best name, but to differentiate from an actual transaction itself. Consistency. We can rename it if we figure out a better name.

crux-core/src/crux/api/ICruxAPI.java
```diff
@@ -121,28 +121,44 @@
...
+     * the node.
      *
-     * @param transactionTime transaction time to sync past.
+     * @param tx Transaction to await, as returned from submitTx.
      * @param timeout max time to wait, can be null for the default.
-     * @return        the latest known transaction time.
+     * @return the latest known transaction.
      */
-    public Date sync(Date transactionTime, Duration timeout);
+    public Map<Keyword, ?> awaitTx(Map<Keyword,?> tx, Duration timeout);
```
hraberg, line 161: Same question about naming of `tx`, see comment above.

crux-kafka/src/crux/kafka.clj
```diff
@@ -152,33 +158,23 @@
...
-                                        next-offset (inc (.offset last-record-in-batch))
-                                        end-offset (get end-offsets partition)
-                                        lag (- end-offset next-offset)]]
-                              (do (when-not (zero? lag)
-                                    (log/debug "Falling behind" (str partition) "at:" next-offset "end:" end-offset))
-                                  [(topic-partition-meta-key partition)
-                                   {:next-offset next-offset
-                                    :time (Date. (.timestamp last-record-in-batch))
-                                    :lag lag}]))
+                                        next-offset (inc (.offset last-record-in-batch))]]
```
hraberg, line 166: Previously this lag was show in the status map. I know we discussed that this would become a calculated property, is the latest known offset also exposed into the status map somehow? Or is this "broken" until the new monitoring work catches up?

hraberg (approved):
Still some comments worth addressing, but my feeling is that it won't hold back merging.

Can't see it easily, but we discussed removing the default timeout, while keeping the config, is that in here?

crux-kafka/src/crux/kafka.clj
jarohen: yeah - I'm going to hold fire on reinstating this til metrics come in, unless there's something depending on it that'll completely break if it's not provided?

jarohen:
@hraberg thanks for the review. going to hold fire on the timeout default change for now (that'll be pt 2) as this part is non-breaking, can go into master and get us to start using it internally

PR #551, opened 2020-01-13
Adding an event bus, #422

### LINKED ISSUE #422: Users can be notified when the indexer ingests documents/transactions

jarohen:
(fixes #422)

raising as a draft PR for early feedback.

* crux.event-bus which keeps track of who's listening to what, filtered by event-type
* added an initial set of events to the indexer, for notifications about indexing docs and txs

TODO:
* [x] doc-strings/javadocs
* [x] spec-asserting the events
* [x] do we need more events for an initial pass? I'm leaning towards 'no for now', can be driven by ops monitoring use cases
* [ ] how should this be exposed in the public API?


crux-core/test/crux/event_bus_test.clj
```diff
@@ -0,0 +1,27 @@
...
+
+      (sut/listen bus {::sut/event-types #{:foo}} #(swap! !filtered-events conj %))
+      (sut/listen bus #(swap! !unfiltered-events conj %))
+
+      (sut/send bus {::sut/event-type :foo, :value 2})
+      (sut/send bus {::sut/event-type :bar, :value 1})
+
+      ;; just to ensure all the jobs are handled
+      ;; - we don't guarantee this if the node is shut down
+      (Thread/sleep 100))
```
hraberg: I know this should normally work, but could we get some promise and a deref with timeout here so we can remove the sleep?

hraberg (approved):

crux-core/test/crux/event_bus_test.clj
jarohen: ah, yes, of course - doh.

jonpither:
Interested to see how metrics will make use of this, i.e. timing query runs. Looks good to me, was mildly wondering for aesthetics if `crux.event-bus` should be `crux.bus`.

crux-core/src/crux/tx.clj
```diff
@@ -340,17 +350,20 @@
                              (->> (vals docs-to-remove)
                                   (map #(idx/doc-predicate-stats % true))))]
 
+      (db/put-objects object-store docs)
+
```
jonpither, line 354: Wearing my pedantic hat, possibly wouldn't have this blank line:-)

jonpither (approved):

PR #552, opened 2020-01-13
Adding time-series benchmark into crux-bench

danmason:
Currently resolves all associated cards, includes both `ts_weather` and `ts_devices` ran within the same main:
- Cloudformation file sets up entire stack, including creating the bench as a scheduled task (ran every weekday at midnight).
- Contains some scripts within /bin which download the dataset (for running locally), build and deploy the *docker* file (requiring the stack to be in place, and the caller to authenticated with Crux AWS account), and run the bench task manually.

As it stands, the bench is ran on the latest release of Crux, `20.01.1.6.2-alpha`, rather than the latest build from the Github. This is not included as part of this PR, as I wish to do this in a subsequent PR and create a proper card for it.


danmason:
Currently fails as `rocksdb-microbench` within `crux-test` relies on `ts-weather-test`, which we have removed.

crux-bench/time-series-bench/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM clojure:openjdk-11-lein-2.9.0
```
jarohen: let's split this up into a base image. aims:
* let's not download anything from Maven when the docker image runs
* let's minimise what we re-download from Maven when we're building the image
* let's see if we can remove the volume, means we won't depend on the local filesystem at all

crux-bench/time-series-bench/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+VOLUME /root/.m2
+
+WORKDIR /time-series-bench
+
+CMD ["lein", "run"]
+
+ADD project.clj .
+ADD src/ ./src/
+
+ADD https://timescaledata.blob.core.windows.net/datasets/devices_small.tar.gz ./resources/
```
jarohen: I thought the `ADD` command decompressed these automatically, evidently it doesn't for remote files. let's combine this with your download_dataset script - i.e. assume these have already been downloaded to resources

crux-bench/time-series-bench/bin/download_dataset.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/env bash
+wget https://timescaledata.blob.core.windows.net/datasets/devices_small.tar.gz
```
jarohen: can do this in fewer lines with curl piped to tar, and save yourself the cleanup

crux-bench/time-series-bench/bin/run_task.sh
```diff
@@ -0,0 +1,9 @@
+#! /bin/env bash
```
jarohen: nit: I've not seen this with a space between `#!` and `/bin/env`, usually `#!/bin/env`

crux-bench/time-series-bench/project.clj
```diff
@@ -0,0 +1,9 @@
+(defproject time-series-bench "0.1.0"
```
jarohen: nit: I know this isn't getting deployed to mvn, but let's give it a proper group-id (not `time-series-bench/time-series-bench` - I _hate_ that lein does this by default)

`juxt/crux-bench`?

crux-bench/time-series-bench/resources/cloudformation.yaml
```diff
@@ -0,0 +1,105 @@
...
+        - Name: 'containerInsights'
+          Value: 'enabled'
+
+  LogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: 'bench-loggroup'
+      RetentionInDays: 365
+
+
```
jarohen: nit: double line gap -> single

crux-bench/time-series-bench/resources/cloudformation.yaml
```diff
@@ -0,0 +1,105 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: An example stack.
+
+Resources:
+  # ECS Resources
+  ECSCluster:
+    Type: AWS::ECS::Cluster
+    Properties:
+      ClusterName: 'bench-cluster'
```
jarohen: nit: you don't need `-cluster` here (obvious it's a cluster) - `crux-bench`?

jarohen: likewise with log group, ECR repo, etc

crux-bench/time-series-bench/resources/cloudformation.yaml
```diff
@@ -0,0 +1,105 @@
...
+      Cpu: '2 vCPU'
+      Memory: '4GB'
+      Family: 'bench-family'
+      ExecutionRoleArn:
+        Fn::GetAtt: ["ECSTaskExecutionRole", "Arn"]
+      RequiresCompatibilities:
+        - 'FARGATE'
+      NetworkMode: 'awsvpc'
+      ContainerDefinitions:
+        -
```
jarohen: nit: `-` and `Name` can be on the same line (keeping it consistent with the rest of your file)

crux-bench/time-series-bench/src/time_series_bench/main.clj
```diff
@@ -0,0 +1,20 @@
+(ns time-series-bench.main
```
jarohen: let's include `crux` as part of the namespace - `crux.bench` is fine here

crux-bench/time-series-bench/src/time_series_bench/main.clj
```diff
@@ -0,0 +1,20 @@
...
+  (:require [time-series-bench.ts-weather :as weather]
+            [time-series-bench.ts-devices :as devices]
+            [clojure.data.json :as json]))
+
+(defn -main []
+  (let [start-time (System/currentTimeMillis)]
+    (println (json/write-str {:crux.bench/bench-type ::main
```
jarohen: I'd go with normal `log/info` here, already has namespace and time info - we're not planning to use these for metrics?

crux-bench/time-series-bench/src/time_series_bench/ts_devices.clj
```diff
@@ -0,0 +1,300 @@
...
+                          [#inst "2016-11-15T20:19:30.000-00:00" :device-info/demo000997 90.7]
+                          [#inst "2016-11-15T20:19:30.000-00:00" :device-info/demo000996 92.8]
+                          [#inst "2016-11-15T20:19:30.000-00:00" :device-info/demo000995 91.9]
+                          [#inst "2016-11-15T20:19:30.000-00:00" :device-info/demo000994 92.0]
+                          [#inst "2016-11-15T20:19:30.000-00:00" :device-info/demo000993 92.8]
+                          [#inst "2016-11-15T20:19:30.000-00:00" :device-info/demo000992 87.6]
+                          [#inst "2016-11-15T20:19:30.000-00:00" :device-info/demo000991 93.1]
+                          [#inst "2016-11-15T20:19:30.000-00:00" :device-info/demo000990 89.9]])]
+      (println (json/write-str {:crux.bench/bench-type ::recent-battery-readings
+                                ::query query
```
jarohen: these keys aren't specific to this namespace, more to `crux-bench` as a whole - if we can use a consistent name, we can compare them between benchmarks, as and when we have more

crux-bench/time-series-bench/src/time_series_bench/ts_devices.clj
```diff
@@ -0,0 +1,300 @@
...
+
+
+(defn run-queries [node]
+  (test-10-most-recent-battery-readings-for-charging-devices node)
+  (test-busiest-devices-1-min-avg-whose-battery-level-is-below-33-percent-and-is-not-charging node)
+  (test-min-max-battery-level-per-hour-for-pinto-or-focus-devices node))
+
+(defn bench []
+  (try
+    (with-open [embedded-kafka (ek/start-embedded-kafka
```
jarohen: I think we've got enough in common here to justify extracting this wiring out somewhere common

jarohen (changes requested):
looking good gents - might seem like a lot of comments but certainly nothing major :smile: let's run through these in a call - most of the feedback is prompts for me to go through in person, dw for now if they don't make sense

crux-bench/time-series-bench/Dockerfile
danmason: Haha - so did we. Seems sensible, stops it from downloading each time :+1:

crux-bench/time-series-bench/src/time_series_bench/main.clj
t-taylor: With Cloudwatch, you need to have json output on its own. If you use `log/info` it outputs INFO TIMESTAMP - <json> and cloudwatch then won't parse the output

crux-bench/time-series-bench/src/time_series_bench/ts_devices.clj
t-taylor: `clojure.data.json` strips namespaces anyway, so the `::` here is totally irrelevant, but point taken

crux-bench/time-series-bench/src/time_series_bench/main.clj
danmason: Although in this case, as these are basically logging messages rather than metrics we plan to use (unless we did want to use `::time-taken` as a final metric), Log/info would be appropriate.

t-taylor: I'd say the json format is just more rich/convenient in general. This way we can do `level="info"` rather than using `like` which just searches for a string, or `parse` where we'd have to input the logging format manually. Using the json format means that cloudwatch does all the parsing for us.

crux-bench/time-series-bench/resources/cloudformation.yaml
```diff
@@ -0,0 +1,103 @@
...
+          Image: '955308952094.dkr.ecr.eu-west-2.amazonaws.com/crux-bench'
+          Essential: true
+          LogConfiguration:
+            LogDriver: awslogs
+            Options:
+              awslogs-region: 'eu-west-2'
+              awslogs-group: !Ref LogGroup
+              awslogs-stream-prefix: ecs
+
+  TaskSchedule:
```
jarohen: update to include new CWE/Fargate configuration (https://github.com/aws/containers-roadmap/issues/92). also, public IP enabled

crux-bench/time-series-bench/resources/cloudformation.yaml
```diff
@@ -0,0 +1,103 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: An example stack.
```
jarohen: nit: ^^

crux-bench/time-series-bench/src/crux/bench/utils.clj
```diff
@@ -0,0 +1,41 @@
...
+                                      :crux.kv/db-dir "dev-storage/db-dir-1"
+                                      :crux.standalone/event-log-dir "dev-storage/eventlog-1"})]
+      (try
+        (let [start-time (System/currentTimeMillis)]
+          (api/sync node
+                    (:crux.tx/tx-time (:last-tx (ingest node)))
+                    (java.time.Duration/ofMinutes 20))
+          (output
+            {:bench-type :ingest
+             :bench-version version
```
jarohen: -> bench-ns

jarohen: add crux version as a dimension in this map

crux-bench/time-series-bench/src/crux/bench/utils.clj
```diff
@@ -0,0 +1,55 @@
...
+                            (cio/load-properties in)]
+                        version))))
+
+
+(defn output [mp]
+  (println (json/write-str (merge mp
+                                  {:crux-commit commit-hash
+                                   :crux-version crux-version}))))
+
+(defn bench [ingest queries bench-ns]
```
jarohen: subjective: this `bench` function is assuming quite a lot about the structure of a benchmark test - any way we can split this into its component parts so that each part is doing one thing?

maybe:
* `with-node`, starts kafka, passes node to callback, cleans up after
* `with-bench-timing`, takes callback, times fn, outputs `(merge result-of-fn {:ingest-time-ms ...})` in JSON format
* `queries` goes - it's 'just' part of the function passed to `with-node` (albeit that the queries will likely call `with-bench-timing`)

obv if you ask 5 developers to structure code into functions, you'll get 5 different suggestions - thoughts, though?

danmason: I'd certainly agree splitting out node startup + having with-bench-timing, unsure on queries

jarohen: in hindsight, this is best left as-is til we get a few more usage examples

jarohen:
morning @danmason, @t-taylor - have made a few minor changes (mostly formatting/nits) on the crux-bench branch on my fork, mostly documented out in the commit messages - could you review, make sure you're happy with them, and then pull those commits into this branch?

https://github.com/danmason/crux/compare/crux-bench...jarohen:crux-bench

~(got one more change I'd like to make, will let you know when I'm done)~

crux-bench/time-series-bench/src/crux/bench/utils.clj
```diff
@@ -0,0 +1,55 @@
...
+                                      :crux.kafka/bootstrap-servers "localhost:9092"
+                                      :crux.kv/db-dir "dev-storage/db-dir-1"
+                                      :crux.standalone/event-log-dir "dev-storage/eventlog-1"})]
+      (try
+        (let [start-time (System/currentTimeMillis)]
+          (api/await-tx node
+                        (:last-tx (ingest node))
+                        (java.time.Duration/ofMinutes 20))
+          (output
+            {:bench-type :ingest
```
jarohen: could you ensure these are consistent? some seem to be namespaced, some aren't

jarohen: values, too

jarohen:
all done with jarohen/crux-bench. just wanted to add a change s.t. you can choose which git branch in the docker image using env vars, e.g.
```
docker run -ti -e GIT_ORIGIN=jms -e GIT_BRANCH=crux-bench 955308952094.dkr.ecr.eu-west-2.amazonaws.com/crux-bench
```

for now, it's worth us manually changing the branch in Dockerfile.base, because otherwise it'll download all the dependencies for current master crux-bench

Issue #538, opened 2020-01-07
Transaction and document ingesting are no longer coupled


jarohen:

jonpither:
Fixed by https://github.com/juxt/crux/pull/557

PR #554, opened 2020-01-14
Untie tx and doc consumer #538

jonpither:
The PR takes a step on the journey of allowing us to swap out the doc-topic with the introduction of a new pluggable component, that would surface in a later PR (such as an S3 doc-store, or JDBC for docs, Kafka for logs). 

One consideration of this PR is all nodes will need to re-ingest documents - this will happen automatically with a node starts. This is because now there is a different consumer state for both doc and tx consumers. 

The new topic may be setup with a different topic partition configuration, so I don't think you can migrate the new doc consumer state from an existing tx consumer state.

### DIFF
--- a/crux-core/src/crux/codec.clj
+++ b/crux-core/src/crux/codec.clj
@@ -19,7 +19,7 @@
 ;; Indexes
 
 ;; NOTE: Must be updated when existing indexes change structure.
-(def ^:const index-version 5)
+(def ^:const index-version 6)
 (def ^:const index-version-size Long/BYTES)
 
 (def ^:const index-id-size Byte/BYTES)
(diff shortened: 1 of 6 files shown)


jonpither:
Moved to in-progress, want to discuss/implement the offset split migration.

jonpither:
From todays meeting, 

- [ ] Raise issue for analysing offsets
- [ ] Raise issue for moving offsets status to Kafka modules etc
- [ ] naming of `RemoteDocumentStore` -> `DocumentStore`
- [ ] Double check tx-fns

We don't need to bump the index version - this should be done when we refactor how the offsets are being stored.

jonpither:
This PR doesn't make the offsets any worse, but they are stored in a clunky way:

```
:crux.tx-log/consumer-state -> {:crux.kafka.topic-partition/tx-0 {:next-offset 0}}}
```
This could be optimised. Although #561 wants to do a way with separate stored offsets altogether (although harder to do once this PR is merged, as there are now two offsets to contend with). 

I would just note, that we should tackle offsets as part of the same release as this issue, because both require a node re-index.

jarohen:
merged with #557

PR #560, opened 2020-01-15
Adding tx-committed?

### LINKED ISSUE #555: Users can find out whether a given transaction committed or not
This function is mainly to be used by users who want to see whether their CaS passed or not. The existing functions were implemented before the 'failed tx' index existed, so it needed to check whether a given entity was updated at a given transaction time.

* [x] Add `tx-committed?` fn on the main API, accepting a submitted-tx, returns true if the transaction committed, false if it didn't (i.e. CaS failed, etc), and throws `NodeOutOfSyncException` if the node hasn't indexed the requested transaction.
* [x] Remove `submitted-tx-updated-entity?` and `submitted-tx-corrected-entity?`
* [x] Update tests/docs as required

danmason:
Fixes #555

### DIFF
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -104,6 +104,11 @@
   (status [node]
     "Returns the status of this node as a map.")
 
+  (tx-committed? [node submitted-tx]
+    "Checks if a submitted tx was successfully committed.
+     submitted-tx must be a map returned from `submit-tx`.
+     Returns true if the submitted transaction was committed.")
+
   (submitted-tx-updated-entity? [node submitted-tx eid]
     "Checks if a submitted tx did update an entity.
   submitted-tx must be a map returned from `submit-tx`
@@ -204,6 +209,9 @@
 
   (submitted-tx-corrected-entity? [this submitted-tx ^Date valid-time eid]
     (.hasSubmittedTxCorrectedEntity this submitted-tx valid-time eid))
+
+  (tx-committed? [this submitted-tx]
+    (.hasTxCommitted this submitted-tx))
 
   (sync
     ([this]
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -107,7 +107,9 @@
   (tx-committed? [node submitted-tx]
     "Checks if a submitted tx was successfully committed.
      submitted-tx must be a map returned from `submit-tx`.
-     Returns true if the submitted transaction was committed.")
+     Returns true if the submitted transaction was committed,
+     false if the transaction was not commi553e, and throws `NodeOutOfSyncException`
+     if the node has not yet indexed the transaction.")
 
   (sync
     [node]
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -108,7 +108,7 @@
     "Checks if a submitted tx was successfully committed.
      submitted-tx must be a map returned from `submit-tx`.
      Returns true if the submitted transaction was committed,
-     false if the transaction was not commi553e, and throws `NodeOutOfSyncException`
+     false if the transaction was not committed, and throws `NodeOutOfSyncException`
      if the node has not yet indexed the transaction.")
 
   (sync
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -109,19 +109,6 @@
      submitted-tx must be a map returned from `submit-tx`.
      Returns true if the submitted transaction was committed.")
 
-  (submitted-tx-updated-entity? [node submitted-tx eid]
-    "Checks if a submitted tx did update an entity.
-  submitted-tx must be a map returned from `submit-tx`
-  eid is an object that can be coerced into an entity id.
-  Returns true if the entity was updated in this transaction.")
-
-  (submitted-tx-corrected-entity? [node submitted-tx ^Date valid-time eid]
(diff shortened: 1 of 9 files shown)


crux-core/src/crux/node.clj
```diff
@@ -108,14 +108,19 @@
...
-    (.hasSubmittedTxCorrectedEntity this submitted-tx (:crux.tx/tx-time submitted-tx) eid))
-
-  (hasSubmittedTxCorrectedEntity [this submitted-tx valid-time eid]
+  (hasTxCommitted [this {:keys [crux.tx/tx-id
+                                crux.tx/tx-time] :as submitted-tx}]
     (cio/with-read-lock lock
       (ensure-node-open this)
-      (api/await-tx this submitted-tx)
-      (q/submitted-tx-updated-entity? kv-store object-store submitted-tx valid-time eid)))
+      (let [latest-tx-time (:crux.tx/tx-time (db/read-index-meta indexer :crux.tx/latest-completed-tx))]
```
jarohen: good, but compare tx-id instead - can have multiple transactions at the same ms

crux-http-server/src/crux/http_server.clj
```diff
@@ -254,6 +254,10 @@
 (defn- attribute-stats [^ICruxAPI crux-node]
   (success-response (.attributeStats crux-node)))
 
+(defn- tx-committed? [^ICruxAPI crux-node request]
+  (let [submitted-tx (body->edn request)]
+    (success-response (.hasTxCommitted crux-node submitted-tx))))
```
jarohen: yep - if this throws the NodeOutOfSyncException, though, we'll want to return a 400 ('bad request') status (unless there's a better 4xx error?)

crux-http-server/src/crux/http_server.clj
```diff
@@ -310,6 +314,9 @@
     [#"^/tx-log$" [:post]]
     (transact crux-node request)
 
+    [#"^/tx-committed$" [:post]]
```
jarohen: this'll want to be a `:get` HTTP request - it's not sending/mutating any data

crux-test/test/crux/query_test.clj
```diff
@@ -1371,9 +1371,8 @@
                           tx-id] :as submitted-tx}
           (api/submit-tx *api* [[:crux.tx/cas nil {:crux.db/id :ivan
                                                    :name "Ivan 1st"}]])]
-
-      (t/is (true? (api/submitted-tx-updated-entity? *api* submitted-tx :ivan)))
-      (t/is (false? (api/submitted-tx-updated-entity? *api* submitted-tx :petr)))
+      (api/await-tx *api* submitted-tx )
```
jarohen: nit: space

crux-core/src/crux/api.clj
```diff
@@ -104,18 +104,10 @@
...
-  (submitted-tx-corrected-entity? [node submitted-tx ^Date valid-time eid]
-    "Checks if a submitted tx did correct an entity as of valid time.
-  submitted-tx must be a map returned from `submit-tx`
-  valid-time valid time of the correction to check.
-  eid is an object that can be coerced into an entity id.
-  Returns true if the entity was updated in this transaction.")
+  (tx-committed? [node submitted-tx]
+    "Checks if a submitted tx was successfully committed.
+     submitted-tx must be a map returned from `submit-tx`.
+     Returns true if the submitted transaction was committed.")
```
jarohen: I'd specify the potential to throw `NodeOutOfSyncException` here - as a developer reading the docstring, I'd want to know that this was a case I should handle.

I might also be tempted to explicitly specify why this function might return false, just to distinguish it from the exception case?

crux-core/src/crux/api/ICruxAPI.java
```diff
@@ -98,27 +98,15 @@
...
-     * @param submittedTx  must be a map returned from {@link
+     * @param submittedTx must be a map returned from {@link
      * #submitTx(List txOps)}.
-     * @param validTime    valid time of correction to check.
-     * @param eid          an object that can be coerced into an entity id.
-     * @return             true if the entity was updated in this transaction.
+     * @return            true if the submitted transaction was committed.
      */
-    public boolean hasSubmittedTxCorrectedEntity(Map<Keyword,?> submittedTx, Date validTime, Object eid);
+    public boolean hasTxCommitted(Map<Keyword,?> submittedTx);
```
jarohen: I'd specify that this method `throws NodeOutOfSyncException` (and add it as `@throws NodeOutOfSyncException <...>` to the Javadoc) - it's an unchecked exception, so the user doesn't _have_ to handle it, but good to put it there as documentation

docs/clojure_api.adoc
```diff
@@ -84,28 +84,17 @@ toc::[]
...
+=== tx-committed?
 
 [source,clj]
 ----
-  (submitted-tx-updated-entity? [node submitted-tx eid]
-    "Checks if a submitted tx did update an entity.
-    submitted-tx must be a map returned from `submit-tx`
-    eid is an object that can be coerced into an entity id.
-    Returns true if the entity was updated in this transaction.")
+  (tx-committed? [node submitted-tx]
```
jarohen, line 91: same doc changes as above

jarohen (changes requested):
nice one Dan, well done :smile: 

few minor suggestions below, but happy to see this get merged once those are addressed

PR #563, opened 2020-01-16
Add TxLogIterator

### LINKED ISSUE #556: Encapsulating tx-log-context so that we don't have to explain it to users
We'd rather not explain the concept of `tx-log-context` to users if we don't have to.

Instead, we can have the existing `db/TxLog` protocol expose an `open-tx-log-iterator` function, which then behaves as a normal Java iterator.

```java 
public interface TxLogIterator extends Iterator<?>, Closeable { }
```

this might then be used as follows:
```clojure
;; calling from Clojure
(with-open [tx-log-iterator (db/open-tx-log-iterator tx-log ...)]
  (let [tx-events (iterator-seq tx-log-iterator)]
    ...))
```

```java
// calling from Java, using try-with-resources
try(TxLogIterator txLog = txLog.openTxLogIterator(...)) {
   while(txLog.hasNext()) {
     ...
   }
}
```

```kotlin
// calling from Kotlin
txLog.openTxLogIterator(...).use { txLog ->
  ...
}
```

related to #410, which we may be able to solve in a similar way

* [x] added `TxLogIterator` Java interface (above, easy)
* [x] `TxLog` protocol exposes `(open-tx-log-iterator [_ from-tx-id with-ops?])` function
* [x] existing implementors of TxLog implement this new function
* [x] existing users of tx-log use this new function
* [x] removed `tx-log` and `new-tx-log-context` functions from TxLog
* [x] update tests/docs

danmason:
(fixes #556)

Still a work in progress, currently still need to go through and update docs/tests. 

Would be worth considering/discussing the approach, and generally get some eyes on it.

### DIFF
--- a/crux-core/src/crux/api/ICruxIngestAPI.java
+++ b/crux-core/src/crux/api/ICruxIngestAPI.java
@@ -28,27 +28,4 @@ public interface ICruxIngestAPI extends Closeable {
      */
 
     public TxLogIterator openTxLogIterator (Long fromTxId, boolean withOps);
-
-    /**
-     * Returns a new transaction log context allowing for lazy reading
-     * of the transaction log in a try-with-resources block using
-     * {@link #txLog(Closeable txLogContext, Long fromTxId, boolean
-     * withDocuments)}.
-     *
-     * @return an implementation specific context.
-     */
-    public Closeable newTxLogContext();
-
-    /**
-     * Reads the transaction log lazily. Optionally includes
-     * operations, which allow the contents under the :crux.api/tx-ops
-     * key to be piped into {@link #submitTx(List txOps)} of another
-     * Crux instance.
-     *
-     * @param txLogContext     a context from {@link #newTxLogContext()}.
-     * @param fromTxId         optional transaction id to start from.
-     * @param withOps          should the operations with documents be included?
-     * @return                 a lazy sequence of the transaction log.
-     */
-    public Iterable<List<?>> txLog(Closeable txLogContext, Long fromTxId, boolean withOps);
 }
--- a/crux-core/src/crux/api/ICruxIngestAPI.java
+++ b/crux-core/src/crux/api/ICruxIngestAPI.java
@@ -27,5 +27,5 @@ public interface ICruxIngestAPI extends Closeable {
      * @return                 a lazy sequence of the transaction log.
      */
 
-    public TxLogIterator openTxLogIterator (Long fromTxId, boolean withOps);
+    public ITxLog openTxLog (Long fromTxId, boolean withOps);
 }
--- a/crux-core/src/crux/api/ICruxIngestAPI.java
+++ b/crux-core/src/crux/api/ICruxIngestAPI.java
@@ -18,6 +18,18 @@ public interface ICruxIngestAPI extends Closeable {
     public Map<Keyword,?> submitTx(List<List<?>> txOps);
 
     /**
+     * Reads the transaction log. Optionally includes  operations, which allow the contents
+     * under the :crux.api/tx-ops key to be piped into (submit-tx tx-ops) of another
+     * Crux instance.
+     *
+     * @param fromTxId         optional transaction id to start from.
+     * @param withOps          should the operations with documents be included?
+     * @return                 a lazy sequence of the transaction log.
+     */
+
+    public TxLogIterator openTxLogIterator (Long fromTxId, boolean withOps);
+
+    /**
      * Returns a new transaction log context allowing for lazy reading
      * of the transaction log in a try-with-resources block using
      * {@link #txLog(Closeable txLogContext, Long fromTxId, boolean
--- a/crux-core/src/crux/api/TxLogIterator.java
+++ b/crux-core/src/crux/api/TxLogIterator.java
@@ -5,5 +5,4 @@ import java.util.Iterator;
 
 @SuppressWarnings("rawtypes")
 public interface TxLogIterator extends Iterator, Closeable {
-
 }
--- a/crux-core/src/crux/api/TxLogIterator.java
+++ b/crux-core/src/crux/api/TxLogIterator.java
@@ -1,0 +1,9 @@
+package crux.api;
+
+import java.io.Closeable;
+import java.util.Iterator;
+
+@SuppressWarnings("rawtypes")
+public interface TxLogIterator extends Iterator, Closeable {
+
+}
--- a/crux-core/src/crux/api/TxLogIterator.java
+++ b/crux-core/src/crux/api/TxLogIterator.java
@@ -2,7 +2,8 @@ package crux.api;
 
 import java.io.Closeable;
 import java.util.Iterator;
+import java.util.Map;
+import clojure.lang.Keyword;
 
-@SuppressWarnings("rawtypes")
-public interface TxLogIterator extends Iterator, Closeable {
+public interface TxLogIterator extends Iterator<Map<Keyword, ?>>, Closeable {
 }
--- a/crux-core/src/crux/moberg.clj
+++ b/crux-core/src/crux/moberg.clj
@@ -220,6 +220,17 @@
       (delay {:crux.tx/tx-id (.id m)
               :crux.tx/tx-time (.time m)})))
 
+  (open-tx-log-iterator [this from-tx-id]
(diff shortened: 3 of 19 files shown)


crux-core/src/crux/api/TxLogIterator.java
```diff
@@ -0,0 +1,8 @@
+package crux.api;
+
+import java.io.Closeable;
+import java.util.Iterator;
+
+@SuppressWarnings("rawtypes")
+public interface TxLogIterator extends Iterator, Closeable {
```
jarohen: let's go for `Iterator<Map<Keyword, ?>>` here, rather than suppress rawtypes?

crux-core/src/crux/moberg.clj
```diff
@@ -220,19 +220,20 @@
...
-              :when (= :txs (get (.headers m) :crux.tx/sub-topic))]
-          {:crux.tx.event/tx-events (.body m)
-           :crux.tx/tx-id (.message-id m)
-           :crux.tx/tx-time (.message-time m)}))))
+  (open-tx-log-iterator [this from-tx-id]
+    (let [snapshot (kv/new-snapshot event-log-kv)
+          i (kv/new-iterator snapshot)]
+      (db/->closeable-tx-log-iterator
+       #(.close ^Closeable snapshot)
+       (lazy-seq
```
jarohen: in hindsight, let's move this lazy-seq call to `->closeable-tx-log-iterator` - it'll be clearer there that we're trying to avoid an NPE than here, and prevent new implementors from forgetting it

crux-core/src/crux/node.clj
```diff
@@ -123,26 +123,31 @@
...
-          (-> tx-log-entry
-              (dissoc :crux.tx.event/tx-events)
-              (assoc :crux.api/tx-ops
-                     (with-open [snapshot (kv/new-snapshot kv-store)]
-                       (->> tx-events
-                            (mapv #(tx/tx-event->tx-op % snapshot object-store))))))
-          tx-log-entry))))
+      (let [tx-log-iterator (db/open-tx-log-iterator tx-log from-tx-id)
+            snapshot (kv/new-snapshot kv-store)
+            filtered-txs (filter
```
jarohen: we can make this clearer to the reader with some threading - we're essentially piping one sequence through a few operations but the current code layout jumps around a little in terms of when the various operations happen:

```clojure
(let [tx-log (-> (iterator-seq tx-log-iterator)
                 (->> (filter #(nil? (...))))
                 (cond->> with-ops? (map (fn [...]
                                           ...))))]
  (db/->closeable-tx-log-iterator ...))
```

crux-http-server/src/crux/http_server.clj
```diff
@@ -159,6 +159,21 @@
       (.close ctx)
       (throw t))))
 
+(defn- streamed-iterator-response [^TxLogIterator iterator]
```
jarohen: there's enough commonality here with `streamed-edn-response` to justify re-using it, IMO
commonality is piped-input-stream, writing out an edn sequence, returning the response - so that should be common
difference is in how to get hold of the edn sequence, what to close - so that should be in the callers

I can put in a commit for this one :)

jarohen (changes requested):
nice one Dan - couple of minor changes and one little refactoring, otherwise we're good to go IMO

will get @hraberg to cast his eye over it too, given we paired up on it last week

jarohen (approved):
:ok_hand:

crux-core/src/crux/api/TxLogIterator.java
```diff
@@ -0,0 +1,9 @@
+package crux.api;
+
+import java.io.Closeable;
+import java.util.Iterator;
+import java.util.Map;
+import clojure.lang.Keyword;
+
+public interface TxLogIterator extends Iterator<Map<Keyword, ?>>, Closeable {
```
hraberg: This should be called ITxLogIterator for consistency.

crux-core/src/crux/api/ICruxIngestAPI.java
```diff
@@ -18,25 +18,14 @@
...
      * Crux instance.
      *
-     * @param txLogContext     a context from {@link #newTxLogContext()}.
      * @param fromTxId         optional transaction id to start from.
      * @param withOps          should the operations with documents be included?
      * @return                 a lazy sequence of the transaction log.
      */
-    public Iterable<List<?>> txLog(Closeable txLogContext, Long fromTxId, boolean withOps);
+
+    public TxLogIterator openTxLogIterator (Long fromTxId, boolean withOps);
```
hraberg: I think this can still be txLog? Like it will be an Iterable<List<?>> still except it will also be closeable.

crux-core/src/crux/api/TxLogIterator.java
hraberg: As mentioned below, I think this is actually an iterator of List<?> (each tx is a vector) but might be wrong?

crux-http-client/src/crux/remote_api_client.clj
```diff
@@ -188,23 +189,19 @@
...
+                       (when with-ops?
+                         (str "with-ops=" with-ops?))]
                       (remove nil?)
                       (str/join "&"))
           in (api-request-sync (cond-> (str url "/tx-log")
                                  (seq params) (str "?" params))
                                nil
                                {:method :get
                                 :as :stream})]
-      (register-stream-with-remote-stream! tx-log-context in)
```
hraberg, line 206: What happened to this client side resource management? Can't recall if it was actually doing anything useful TBH?

hraberg (commented):
Just adding a few comments as @jarohen asked me to have a peek. Still deferring to him for actual approval.

crux-core/src/crux/api/TxLogIterator.java
danmason: I definitely agree on the first one, though I'm fairly certain that it is an iterator of maps. For example, post running txLog you get back:
```
({:crux.tx/tx-id 0,
  :crux.tx/tx-time #inst "2020-01-23T10:14:55.340-00:00",
  :crux.api/tx-ops [[:crux.tx/put #:crux.db{:id :test}]]})
```

Which was a LazySeq of Maps.

crux-http-client/src/crux/remote_api_client.clj
danmason: I'm not entirely sure what it was for, removed it as you didn't have a tx-log-context anymore so wasn't sure if there was anything else you could've passed into it?

crux-core/src/crux/api/ICruxIngestAPI.java
danmason: Opinions on this, @jarohen?

crux-core/src/crux/api/TxLogIterator.java
jarohen: happy with `ITxLogIterator`

it's a map of `:crux.tx/tx-time`, `:crux.tx/tx-id`, and `:crux.tx/tx-events`, I think?

crux-core/src/crux/api/ICruxIngestAPI.java
jarohen: `openTxLog`? would like the `open` prefix to make it crystal clear that it needs to be closed

crux-http-client/src/crux/remote_api_client.clj
jarohen: pass!

crux-core/src/crux/api/TxLogIterator.java
hraberg: Ah, in that case ignore me. Who ever is closest to it picks a sensible Java type signature :-)

crux-core/src/crux/api/ICruxIngestAPI.java
hraberg: Like it, good balance.

crux-http-client/src/crux/remote_api_client.clj
hraberg: My point was more that these streams used to be tracked, potentially to be closed (client side), but can't really recall what that really did practically. There's the lazy query variant as well, if you look at how that works and why and if it uses this.

jarohen: yep - this needs to be closed in the close-fn, pushing a change now

PR #569, opened 2020-01-20
Replace `sync` in clojure api section of docs #466

danmason:
Small PR to remove `sync` from the Clojure API section of the docs and add in **await-tx** and **await-tx-time**

### DIFF
--- a/docs/clojure_api.adoc
+++ b/docs/clojure_api.adoc
@@ -120,6 +120,24 @@ toc::[]
   this node has caught up as of this call.")
 ----
 
+=== sync
+
+[source,clj]
+----
+ (sync
+    [node]
+    [node ^Duration timeout]
+    "Blocks until the node has caught up indexing to the latest tx available at
+  the time this method is called. Will throw an exception on timeout. The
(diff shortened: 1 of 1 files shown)


jarohen (changes requested):
cheers Dan :)

we're not removing _all_ of `sync` - just the arity with node, tx-time and timeout - could you reinstate the doc for the version with just node and timeout? (might need to remove the bits in the docstring about 'if the tx-time is supplied', although I think I've already updated that in the protocol definition itself)

danmason:
Ah, alright! My bad :) 

Kind of unrelated to this PR specifically, but I made a small comment on the await-tx card: "Note on remote-api-client / http-server: Both of these still use 'sync' (though http-server does call down to await-tx-time, if a transaction time is provided). Should probably also include await-tx and await-tx-time in both of these, and update the REST api docs thereafter"

Might be worth removing the await-tx-time behavior from the 'sync' within 'http-server', and make it a separate endpoint/function?

jarohen (approved):
cheers, will merge

PR #570, opened 2020-01-20
Adding ingest metrics

### LINKED ISSUE #495: Crux nodes expose metrics about what they're currently ingesting
Within the indexer, we'd like to expose:
* The current lag - the difference between the incoming tx-time and clock time
* How many documents/min we're currently ingesting
* The size of the documents we're ingesting, per minute

These stats should be exposed through:
* `dropwizard-metrics` with a `jmx` adapter and `console` adapter
* `prometheus`

t-taylor:
Resolves #495.

Currently uses the `metrics-clojure` lib to expose the gauges in `crux.metrics.gauges`.


crux-metrics/src/crux/metrics/bus.clj
```diff
@@ -0,0 +1,55 @@
...
+                {:crux.bus/event-types #{:crux.tx/indexed-tx}}
+                (fn [event]
+                  (let [meta-doc (get (:crux.metrics/indexing-tx @!metrics) (:crux.tx/submitted-tx event))
+                        start-time-ms (:start-time-ms meta-doc)
+                        end-time-ms (System/currentTimeMillis)]
+                    (swap! !metrics update :crux.metrics/indexing-tx dissoc (:crux.tx/submitted-tx event))
+                    (swap! !metrics update :crux.metrics/indexed-tx assoc (:crux.tx/submitted-tx event)
+                           {:start-time-ms start-time-ms
+                            :time-elapsed-ms (- end-time-ms start-time-ms)})
+                    (swap! !metrics update :crux.metrics/latest-latency-tx (- end-time-ms start-time-ms)))))
```
jarohen: I'm surprised this doesn't break? you're calling `update` with a value rather than a function

crux-metrics/src/crux/metrics/bus.clj
```diff
@@ -0,0 +1,55 @@
...
+                        :crux.metrics/indexed-tx {}
+                        :crux.metrics/indexing-docs {}
+                        :crux.metrics/indexed-docs {}
+                        :crux.metrics/latest-latency-docs -1
+                        :crux.metrics/latest-latency-tx -1
+                        :crux.metrics/latest-tx-id []})]
+    (bus/listen (:bus node)
+                {:crux.bus/event-types #{:crux.tx/indexing-docs}}
+                (fn [{:keys [doc-ids]}]
+                  (swap! !metrics update :crux.metrics/indexing-docs assoc
```
jarohen: nit: `(swap! a update k1 assoc k2 v)` === `(swap! a assoc-in [k1 k2] v)`

crux-metrics/src/crux/metrics/dropwizard.clj
```diff
@@ -0,0 +1,11 @@
+(ns crux.metrics.dropwizard
+  (:require [crux.metrics.bus :as met-bus]
+            [metrics.gauges :as gauges]))
+
+(defn cx-met->metrics [reg !metrics]
+  (map (fn [[fn-sym func]]
+         (gauges/gauge-fn reg (str fn-sym) #(func !metrics)))
+       (ns-publics 'crux.metrics.gauges)))
```
jarohen: using `ns-publics` here means that anything that gets added to `crux.metrics.gauges` is expected to be a gauge, but without there being anything in that namespace to suggest that might be the case - it might be clear in this PR, but the next person to refactor that ns, extracting a fn, say, might not spot it.

simpler - let's have crux.metrics.gauges export an explicit map, `(def gauges {...})`, say, from gauge-key -> gauge-fn, which this namespace can then iterate through

project.clj
```diff
@@ -9,7 +9,8 @@
               "crux-kafka-connect"
               "crux-kafka"
               "crux-uberjar"
-              "crux-test"])
+              "crux-test"
+              "crux-metrics"])
```
jarohen: these are ordered - `lein sub install` (for example) is called on each of them in turn. `crux-test` should depend on `crux-metrics` (as will `crux-uberjar`, now I think about it) so we'll want to build `crux-metrics` before both of them

jarohen (changes requested):
thanks Tom - a few pointers here:
* this could do with some tests to give us confidence that this is working as planned - particularly, I'd be surprised if things like that `update` bug didn't throw an exception as soon as it's run? let me know if it'd be helpful to pair on these tomorrow?
* I've specified a fair bit in the original card - can you make sure this PR addresses those requirements? entirely possible with card specs that the original specs don't make sense once you get into the implementation but, in that case, let's chat about it and update the card as necessary before the PR?
* the metrics atom, in this case, only ever accretes data in memory as more transactions are indexed - this is a memory leak (hence the 30m window in the card)
* let's pair tomorrow on how we expect users to use this

crux-metrics/src/crux/metrics/dropwizard.clj
t-taylor: I did that originally, but I didn't like that i was writing out data that should already be there. Although I do agree it can break easily.

jarohen:
thanks @t-taylor - a lot of back-n-forth on this one, so well done for getting it over the line :smile: 

probably goes without saying, but once there's a few more metrics in there (the easy bit, no doubt), this'll demo very nicely :smile:

PR #582, opened 2020-01-22
Add slack notifications to bench

### LINKED ISSUE #576: Each benchmark should report its results into Slack
When the benchmark is running in the bench environment (i.e. not in dev), it should report the stats for each benchmark into a `crux-bench` channel in Slack

danmason:
Slack messages are only sent if the SLACK_URL environment variable is set. This secret is set up on the AWS secret manager, and has been added to the container using CloudFormation.

resolves #576

### DIFF
--- a/crux-bench/src/crux/bench.clj
+++ b/crux-bench/src/crux/bench.clj
@@ -47,8 +47,10 @@
 
 (defn post-to-slack [message]
   (when (System/getenv "SLACK_URL")
-    (client/post (System/getenv "SLACK_URL")
-                 {:body (clojure.data.json/write-str {:text message})
+    (client/post (-> (System/getenv "SLACK_URL")
+                      (clojure.data.json/read-str)
+                      (get "slack-url"))
+                 {:body (json/write-str {:text message})
                   :content-type :json})))
 
 (defn format-and-post-results-to-slack [result]
--- a/crux-bench/src/crux/bench.clj
+++ b/crux-bench/src/crux/bench.clj
@@ -7,7 +7,8 @@
             [clojure.java.io :as io]
             [clojure.java.shell :as shell]
             [clojure.string :as string]
-            [clj-http.client :as client]))
+            [clj-http.client :as client]
+            [crux.fixtures :as f]))
 
 (def commit-hash
   (string/trim (:out (shell/sh "git" "rev-parse" "HEAD"))))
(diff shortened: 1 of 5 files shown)


danmason:
Note: may need to add permission to cloudformation so that the task can obtain secrets from the secret manager

crux-bench/src/crux/bench.clj
```diff
@@ -44,6 +45,29 @@
...
 
+(defn post-to-slack [message]
+  (when (System/getenv "SLACK_URL")
+    (client/post (-> (System/getenv "SLACK_URL")
+                      (clojure.data.json/read-str)
+                      (get "slack-url"))
+                 {:body (json/write-str {:text message})
+                  :content-type :json})))
+
+(defn format-and-post-results-to-slack [result]
```
jarohen: whenever you've got `and` in a function name, that's usually a clue that the function's doing more than one thing :wink:

crux-bench/src/crux/bench.clj
```diff
@@ -44,6 +45,29 @@
 (defmacro ^{:style/indent 1} run-bench [bench-type & body]
   `(run-bench* ~bench-type (fn [] ~@body)))
 
+(defn post-to-slack [message]
+  (when (System/getenv "SLACK_URL")
```
jarohen: nit: `when-let`

crux-bench/src/crux/bench.clj
```diff
@@ -44,6 +45,29 @@
...
+    (client/post (-> (System/getenv "SLACK_URL")
+                      (clojure.data.json/read-str)
```
jarohen: `clojure.data.json` isn't `require`d at the top?

crux-bench/src/crux/bench.clj
```diff
@@ -44,6 +45,29 @@
...
+                          (let [formatted-bench (assoc bench-map :time-taken (java.time.Duration/ofMillis bench-time))]
+                            (->> (dissoc formatted-bench :bench-ns :crux-commit :crux-version :time-taken-ms)
+                                 (keys)
+                                 (map (fn [key-name] (format "*%s*: %s" (name key-name) (formatted-bench key-name))))
+                                 (string/join "\n"))))
+                        result)
+        formatted-string (->> result-strings
+                              (string/join "\n\n")
+                              (format "*%s*\n========\n%s\n" *bench-ns*))]
+    (prn formatted-string)
```
danmason: shouldn't be `prn`ed

crux-bench/src/crux/bench.clj
```diff
@@ -54,24 +78,27 @@
     (log/infof "finished bench-ns '%s'." bench-ns)
 
     (doseq [result @*!bench-results*]
-      (println (json/write-str result)))))
+      (println (json/write-str result)))
+    (format-and-post-results-to-slack @*!bench-results*)))
```
jarohen: couple of things here:
* let's deref the atom once and let-bind it - it's not going to change in-between, here, but good practice regardless
* related to the `and` comment, above - something like `(post-to-slack (->> results (map ->slack-message) (str/join ...))`?

crux-bench/src/crux/bench.clj
```diff
@@ -54,24 +78,27 @@
...
-                              {:crux.kafka.embedded/zookeeper-data-dir "dev-storage/zookeeper"
-                               :crux.kafka.embedded/kafka-log-dir "dev-storage/kafka-log"
-                               :crux.kafka.embedded/kafka-port 9092})
-              node (api/start-node {:crux.node/topology 'crux.kafka/topology
-                                    :crux.node/kv-store 'crux.kv.rocksdb/kv
-                                    :crux.kafka/bootstrap-servers "localhost:9092"
-                                    :crux.kv/db-dir "dev-storage/db-dir-1"
-                                    :crux.standalone/event-log-dir "dev-storage/eventlog-1"})]
-    (f node)))
+  (try (with-open [embedded-kafka (ek/start-embedded-kafka
```
jarohen: can this use the new `crux.fixtures/with-tmp-dir` instead?

jarohen (changes requested):
few minor style suggestions, otherwise good - cheers :)

crux-bench/src/crux/bench.clj
danmason: how would you recommend using this? should I include crux-test, or a checkout to the file?

PR #583, opened 2020-01-22
Sorting out the dev process

### LINKED ISSUE #98: Devs can refresh their environment easily without bouncing a REPL
It can be quite easy to get the dev environment into a state where some things refuse to reload, or you need to know the exact order of namespaces to reload for a small change. This is mainly due to redefined records or protocols.

### LINKED ISSUE #487: Eval'ing the dev namespaces shouldn't be side-effecting
Some namespaces under `dev` don't load properly, some are misnamed (hyphen) and some also start various resources which randomly gets kicked off in the background after a `clojure.tools.namespace/refresh`. Let's fix all this.

jarohen:
resolves #487, resolves #98 

Moving shit around :smile: 

tl;dr:
* `crux-dev` is no more, develop from the root of the repo.
* `lein sub install` one-off, then `lein repl`
* Then, Integrant REPL commands to start/stop a dev node (`(go)`, `(halt!)`) - `user` namespace, not `dev`, and doesn't require a call to `(dev)` first
* `(reset)` stops the dev node, reloads changed namespaces in the right order, restarts the dev node
  * emacs users can use `cider-ns-refresh` for this, `C-c M-n r`, or `, s x` in Spacemacs
  * conjure users can use `ConjureRefresh`, see conjure docs for bindings
* Most of the time, you shouldn't need to bounce the REPL, but:
  * if you change any of the dependencies of any of the modules, that'll require another `lein sub install` and a REPL bounce
  * if you change any of the Java classes, that'll require a `lein sub javac` and a REPL bounce
  * otherwise, `(user/reset)` (or just `(reset)` if you're already in the `user` ns) should be sufficient.
* **important** please don't put any more side-effecting top-level code in dev namespaces - you'll break this reload ability and make me sad.

* You can run module tests from the root of the git repo without a `lein sub install`, because of the lein checkouts - all of the tests are in scope here, so things like `lein test :only crux.tx-test` should also work.
* `lein build` still does a `lein sub install`/`lein sub test` - so this is the one to use if you _don't_ want to use the checkouts

Also in this PR:
* tests all use actual temporary directories for db-dir/event-log-dir now, please don't just put them in '/data' or '/dev-storage' (you might want to look through the diff with whitespace-changes turned off)
* refactored space and repl walkthrough tests, mostly removing the inline `def`s
* root project has now become `juxt/crux-dev` so that if we accidentally include it in the deploy to Clojars, all our dev stuff doesn't just go under the `juxt/crux` artifact (as it has done in previous releases)

See https://github.com/jarohen/crux/tree/dev-process#developing-crux for new developer documentation.


hraberg (approved):
All 75 files looks excellent :-) No seriously, I take this is a good step to regain some sanity in dev land.

PR #589, opened 2020-01-23
'Node' is just another component

jarohen:
Refactored crux.node so that the node part is just another component - motivation is that we can then migrate the HTTP server (amongst other things) to be standard Crux modules.

I've refactored the topology code out to a separate namespace now, am aware that there wasn't much appetite for this when it was last discussed - I'm all in favour of larger namespaces for related functionality but, in this case, topology doesn't know about nodes, and the node doesn't know about the topology - it's now two very different areas of functionality. RFC?

also:
* making uberjar-test finish as soon as the server starts, for faster tests (it was previously always just waiting for the timeout)
* stopping node-test from creating directories within the repo

### DIFF
--- a/crux-uberjar/test/crux/uberjar_test.clj
+++ b/crux-uberjar/test/crux/uberjar_test.clj
@@ -9,10 +9,10 @@
   (:import (java.io File)))
 
 (def ^File working-directory
-  (.. (io/as-file (io/resource "crux/uberjar_test.clj"))
-      getParentFile
-      getParentFile
-      getParentFile))
+  (-> (io/as-file (io/resource "crux/uberjar_test.clj"))
+      (.getParentFile)
+      (.getParentFile)
+      (.getParentFile)))
 
 (defn build-uberjar []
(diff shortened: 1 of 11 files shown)


crux-uberjar/test/crux/uberjar_test.clj
```diff
@@ -5,13 +5,14 @@
...
+            [crux.io :as cio])
+  (:import (java.io File)))
 
-(def working-directory
-  (-> (io/as-file (io/resource "crux/uberjar_test.clj"))
-      .getParentFile
-      .getParentFile
-      .getParentFile))
+(def ^File working-directory
+  (.. (io/as-file (io/resource "crux/uberjar_test.clj"))
```
hraberg: First, parens around all threaded function calls. Second, I never use `..`. YMMV. There are other instances like this in this PR, but won't comment on each one of them.

hraberg (changes requested):

jonpither (approved):

Issue #573, opened 2020-01-22
CircleCI should build and deploy new bench images on every master build


jarohen:
- [x]  Dockerfile updated to include local source (i.e. not from Github) and `lein sub install` run at image build time
- [x]  CircleCI will need permission/credentials to deploy to ECR
- [x]  CircleCI to deploy the docker image to ECR with the tags `commit-<hash>` and `master`
- [x]  ECR lifecycle policy for the bench image to be latest 5 images on master, then `commit-<hash>` for 14 days
- [x]  Only run on every master build - we don't expose secrets to PR builds for security reasons, so it won't be able to deploy the image
- [x]  Overnight run to use image currently tagged `master`

Issue #585, opened 2020-01-23
Common docker images built and deployed for each tag


jarohen:
These docker images should be built by CI whenever we tag a release:
* standalone
* crux-builder (for users to use as a base for their own bespoke Crux docker images)

all of which:
* start an HTTP node
* can optionally have crux.properties/crux.edn/logback.xml files supplied at runtime to pass options

PR #590, opened 2020-01-24
CircleCI building and deploying new bench images

danmason:
Addressing #573

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
             - ~/.m2
           key: v1-dependencies-{{ checksum "project.clj" }}
 
-      - run: lein do version, check
+      - run: lein do version, sub test, check
 
       # one for when we can get CircleCI Clojars deploys working again
       # - run: ./.circleci/bin/deploy_clojars.sh
@@ -88,3 +88,6 @@ workflows:
       - deploy:
           requires:
             - build
+          filters:
+            branches:
+              only: master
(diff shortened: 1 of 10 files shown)


crux-bench/bin/download_dataset.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+cd crux-bench/
```
jarohen: you probably want `cd $(dirname "$0")/..` so that this works even if it's not run from the `crux-bench` dir, and a `cd -` at the end of the script to go back to the user's previous directory

.circleci/config.yml
```diff
@@ -41,61 +44,50 @@ jobs:
...
+            key: v1-dependencies-{{ checksum "project.clj" }}
+
+        - run: sh ./crux-bench/bin/download_dataset.sh
+        - run: lein sub -s crux-bench uberjar
+        - run:
+            command: |
+              echo 'export BUILD_ARG="--build-arg crux_commit=$CIRCLE_SHA1"' >> $BASH_ENV
+              echo 'export BUILD_TAGS="master,commit-$CIRCLE_SHA1"' >> $BASH_ENV
+
+        - setup_remote_docker
```
jarohen, line 75: guessing this activates configuration elsewhere in CircleCI?

danmason: Yeah, part of which is to set it up for deploying via docker

jarohen (approved):
nicely done :smile:  I'll definitely be using this in #585.

let's just have a quick demo of it working in Monday's standup and I'll be happy to merge it - cheers!

crux-bench/bin/download_dataset.sh
danmason: Quickly added that so that circle could use it, seems good to change it :+1:

PR #593, opened 2020-01-24
The remote API tests weren't actually running on the remote API

jarohen:
Because of the test fixture order in api-test, the remote API node was getting created, but then `with-node` overrode the binding for `fapi/*api*` with the local node, because it was later in the text fixture order... :grimacing: 

Anyway, the remote API client has hence gotten out of sync with the local API, here's bringing it up to date again.

### DIFF
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -138,7 +138,6 @@
   supplied tx. Will throw on timeout. Returns the most recent tx indexed by the
   node.")
 
-
   (latest-completed-tx [node]
     "Returns the latest transaction to have been indexed by this node.")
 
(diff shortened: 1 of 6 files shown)


crux-core/src/crux/api.clj
```diff
@@ -138,6 +138,13 @@
   supplied tx. Will throw on timeout. Returns the most recent tx indexed by the
   node.")
 
+
```
jonpither: Double blank line?

jonpither (approved):

crux-core/src/crux/api.clj
jarohen: derp. fixed, thanks :)

Issue #580, opened 2020-01-22
Remove deprecated functions


jarohen:
Once we've 'waited long enough' since 1.7.0 release, we can remove:
* the deprecated `(sync node tx-time <timeout>)` arity, see #466 
* the deprecated `(crux.http-server/start-http-server ...)` function, see #596

PR #596, opened 2020-01-27
HTTP server as a module

jarohen:
Now that node's a module, we can start the HTTP server using a module, which means we can specify it in the data-oriented topology config (e.g. EDN/properties files)

### DIFF
--- a/crux-core/src/crux/node.clj
+++ b/crux-core/src/crux/node.clj
@@ -89,9 +89,9 @@
   (status [this]
     (cio/with-read-lock lock
       (ensure-node-open this)
+      ;; we don't have status-fn set when other components use node as a dependency within the topology
       (if status-fn
         (status-fn)
-        ;; fallback for before status-fn's set
         (into {} (mapcat status/status-map) [indexer kv-store object-store tx-log]))))
 
   (attributeStats [this]
@@ -104,8 +104,7 @@
       (ensure-node-open this)
       @(db/submit-tx tx-log tx-ops)))
 
-  (hasTxCommitted [this {:keys [crux.tx/tx-id
-                                crux.tx/tx-time] :as submitted-tx}]
+  (hasTxCommitted [this {:keys [crux.tx/tx-id crux.tx/tx-time] :as submitted-tx}]
     (cio/with-read-lock lock
       (ensure-node-open this)
       (let [{latest-tx-id :crux.tx/tx-id
--- a/crux-core/src/crux/node.clj
+++ b/crux-core/src/crux/node.clj
@@ -89,7 +89,10 @@
   (status [this]
     (cio/with-read-lock lock
       (ensure-node-open this)
-      (status-fn)))
+      (if status-fn
+        (status-fn)
+        ;; fallback for before status-fn's set
+        (into {} (mapcat status/status-map) [indexer kv-store object-store tx-log]))))
 
   (attributeStats [this]
     (cio/with-read-lock lock
--- a/crux-core/src/crux/topology.clj
+++ b/crux-core/src/crux/topology.clj
@@ -47,9 +47,14 @@
   (let [topology (-> topology
                      (cond-> (not (vector? topology)) vector)
                      (->> (map #(s/conform ::module %))
-                          (apply merge)))]
-    (->> (merge topology (select-keys options (keys topology)))
-         (s/conform ::resolved-topology))))
+                          (apply merge))
+                     (as-> topology (merge topology (select-keys options (keys topology)))))
+        resolved-topology (s/conform ::resolved-topology topology)]
+    (when (s/invalid? resolved-topology)
+      (s/explain ::resolved-topology topology)
+      (throw (IllegalArgumentException. "invalid topology")))
+
+    resolved-topology))
 
 (defn- start-order [system]
   (let [g (reduce-kv (fn [g k c]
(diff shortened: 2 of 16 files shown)


crux-core/src/crux/node.clj
```diff
@@ -89,7 +89,10 @@
   (status [this]
     (cio/with-read-lock lock
       (ensure-node-open this)
-      (status-fn)))
+      (if status-fn
+        (status-fn)
+        ;; fallback for before status-fn's set
```
jonpither: No need for comment?

crux-core/src/crux/topology.clj
```diff
@@ -47,9 +47,14 @@
...
-                          (apply merge)))]
-    (->> (merge topology (select-keys options (keys topology)))
-         (s/conform ::resolved-topology))))
+                          (apply merge))
+                     (as-> topology (merge topology (select-keys options (keys topology)))))
+        resolved-topology (s/conform ::resolved-topology topology)]
+    (when (s/invalid? resolved-topology)
+      (s/explain ::resolved-topology topology)
+      (throw (IllegalArgumentException. "invalid topology")))
+
```
jonpither: blank line is incongruous? V minor

crux-docker/crux.edn
```diff
@@ -1,4 +1,4 @@
-{:crux/node-opts {:crux.node/topology ['crux.standalone/topology]
+{:crux/node-opts {:crux.node/topology '[crux.standalone/topology crux.http-server/module]
```
jonpither, line 1: not sure about module vs topology. A module can house multiple topologies?

crux-http-server/src/crux/http_server.clj
```diff
@@ -370,7 +370,7 @@
   (close [_]
     (.stop server)))
 
-(defn start-http-server
+(defn ^:deprecated start-http-server
```
jonpither, line 373: I was thinking about this - could we just remove the fn? What's the benefit in keeping - forcing users to update code during alpha isn't so bad?

crux-test/test/crux/examples_test.clj
```diff
@@ -32,17 +32,23 @@
...
-    ;; Testing http-server/http-client using the standalone node
-    (with-open [http-server ^Closeable (ex/example-start-http-server node)
-                remote-api ^Closeable (ex/example-start-http-client)]
+    ;; Testing example standalone node is closed properly
+    (t/is (nil? (ex/example-close-node node)))))
+
+(t/deftest test-example-http-node
+  (with-open [underlying-node ^Closeable (ex/example-start-standalone-http)
+              node ^Closeable (ex/example-start-http-client)]
+    ;; Testing example standalone node is created properly
```
jonpither: add a t/testing "something" rather than a comment?

crux-test/test/crux/examples_test.clj
```diff
@@ -32,17 +32,23 @@
     (t/is (= #{[:dbpedia.resource/Pablo-Picasso]} (ex/example-query node)))
     (t/is (not (empty? (ex/example-query-valid-time node))))
 
-    ;; Testing http-server/http-client using the standalone node
-    (with-open [http-server ^Closeable (ex/example-start-http-server node)
-                remote-api ^Closeable (ex/example-start-http-client)]
+    ;; Testing example standalone node is closed properly
```
jonpither: t/testing

jonpither (changes requested):

jonpither:
The module is being added by merging in a new topology - IIRC you can also add components by adding a config property? What's the preferred approach now?

crux-docker/crux.edn
jarohen: since #568 I've been calling these maps of components 'modules' (clarified definitions in that PR), but the standalone one existed beforehand, so I didn't want to break existing configurations - worth it?

crux-http-server/src/crux/http_server.clj
jarohen: one for a quick discussion, I reckon - I'd be tempted to keep this in for one overlap release so that people can migrate easier? not wedded to that, though.

crux-test/test/crux/examples_test.clj
jarohen: doh. thanks :smile:

jarohen: ah, this was existing code I moved about - but yeah, will update regardless

jarohen (commented):
> The module is being added by merging in a new topology - IIRC you can also add components by adding a config property? What's the preferred approach now?

you can't add components by adding config properties, you can only override components that existing topologies have specified, I think?

crux-core/src/crux/node.clj
jarohen: I wondered whether this was counter-intuitive? if not, happy to remove it

jarohen: have updated the comment to explain _why_ it's counter-intuitive - this one only said what it was doing - HTH?

crux-docker/crux.edn
jarohen: added #607 for discussion

crux-http-server/src/crux/http_server.clj
jarohen: gonna merge this for now, and add this discussion to #580

crux-core/src/crux/topology.clj
jarohen: happy to remove this for now, but generally speaking I'd like to see more of these in the codebase - helps the reader break the code down into logical units, and much easier to scan the code when your eye can lock onto blank lines

PR #598, opened 2020-01-27
Await waits indefinitely unless given a timeout, #578

### LINKED ISSUE #578: `await-tx` etc wait indefinitely by default, if the user doesn't supply a timeout
users can always override the 'default timeout' config to explicitly specify a default timeout

jarohen:
resolves #578 

main change is in cio/wait-while (needs to handle nils) and changing the node arg, but I've also made a change to pass Duration objects all the way down - safer than everywhere assuming millis. RFC?

### DIFF
--- a/crux-core/src/crux/config.clj
+++ b/crux-core/src/crux/config.clj
@@ -1,7 +1,8 @@
 (ns crux.config
   (:require [clojure.spec.alpha :as s]
             [clojure.java.io :as io])
-  (:import java.util.Properties))
+  (:import java.util.Properties
+           (java.time Duration)))
 
 (def property-types
   {::boolean [boolean? (fn [x]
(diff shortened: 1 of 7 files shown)


crux-core/src/crux/config.clj
```diff
@@ -12,10 +13,17 @@
...
+               (fn [d]
+                 (cond
+                   (instance? Duration d) d
+                   (nat-int? d) (Duration/ofMillis d)
+                   (string? d) (Duration/parse d)))]})
 
-(s/def ::type (s/and (s/conformer (fn [x] (or (property-types x) x)))
-                     (fn [x] (and (vector? x) (-> x first fn?) (some-> x second fn?)))))
+(s/def ::type
+  (s/and (s/conformer (fn [x] (or (property-types x) x)))
```
jonpither, line 25: why on new line

crux-core/src/crux/node.clj
```diff
@@ -194,9 +192,9 @@
                                :closed? (atom false)
                                :lock (StampedLock.)}))
    :deps #{::indexer ::kv-store ::bus ::object-store ::tx-log}
-   :args {:crux.tx-log/await-tx-timeout {:doc "Default timeout in milliseconds for waiting."
-                                         :default 10000
-                                         :crux.config/type :crux.config/nat-int}}})
+   :args {:crux.tx-log/await-tx-timeout {:doc "Default timeout for awaiting transactions being indexed."
+                                         :default nil
```
jonpither, line 199: think you can just remove default then?

jonpither (changes requested):

crux-core/src/crux/config.clj
jarohen: subjective, obv, but to keep it consistent with other multi-line defs?

crux-core/src/crux/node.clj
jarohen: yep - it's `crux.topology/parse-opts` that'd would complain - it checks whether the value's required, and then only runs the validate-fn if the value's non nil.

`(parse-opts {:crux.tx-log/await-tx-timeout {:default nil, :crux.config/type :crux.config/duration}} {})`
=> `{:crux.tx-log/await-tx-timeout nil}`

Issue #586, opened 2020-01-23
Common uberjars built and released for each release


jarohen:
We should build common uberjars for each release, so that users can 'just' download the uberjar and run `java -jar ...`
* standalone

uberjars should accept args supplying properties files, EDN configuration files, and logback.xml files

PR #599, opened 2020-01-28
Crux builder - building custom uberjars and docker images

danmason:
Part way to addressing #585 and #586 

What users will now be able to do:
* download a standalone uberjar, run `java -jar crux.jar` to start up a basic standalone Crux/HTTP node
* download a builder tarball, with which they can customise their artifact (using deps.edn/pom.xml) and then build either a custom uberjar or a custom Docker image

Not incorporating this into CircleCI for the time being, so that we can get an initial release out for the showcase this afternoon.


danmason:
To note - need to ensure this does, indeed, only run on *tags*, rather than any commit.

danmason:
Note: Requires us to set the following environment variables within CircleCI: 
- GITHUB_TOKEN
- DOCKER_USER
- DOCKER_PASS

PR #601, opened 2020-01-28
Add query metrics

### LINKED ISSUE #594: Crux exposes metrics on how long queries are taking
* Will need to add query events to the bus. let's include the query in the events but not any args (and probably a query id UUID, so that consumers can tie up started/stopped events) 
* DW timer so that we can get means/percentiles/counts etc.

t-taylor:
Resolves #594


jarohen:
hey Tom - some suggestions in t-taylor/crux#1

PR #602, opened 2020-01-28
Decouple Clojure + Java APIs #467

### LINKED ISSUE #467: The Clojure API shouldn't depend on the legacy Java API
Currently we flit between Clojure and Java in the node and related APIs - ideally the Clojure API wouldn't be aware that there's a Java wrapper around it.

We can then deprecate the legacy Java API in favour of the new `crux.api.alpha` package.

* [ ] Disentangle the Clojure API from the legacy Java API (protocols become the entry point), update tests
* [ ] Point crux.api.alpha at the Clojure protocols, update tests
* [ ] Ensure the legacy Java API still works for the time being

jarohen:
(see #467 for background/motivation)

draft PR for now, want to start seeing builds run through...
- [x] update docs

- we now implement `ICruxAPI` as a decorator over `PCruxNode`, rather than the other way around
- `crux.node.CruxNode` now implements `PCruxNode`, rather than `ICruxAPI`
- new `crux.java` namespace which is now the entry point for `Crux.java`
- added `fapi/*node*` - tests that use the Clojure API should use this, tests that call Java interop should continue to use `*api*` (I've left most of `api_test.clj` calling interop, to continue exercising the Java API)
  - `query_test.clj` looks like a big change - it's just `*api*` -> `*node*` and then correcting the indentation :grimacing:

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,6 @@ Start a standalone in-memory (i.e. not persisted anywhere) node:
 
 ``` clojure
 (require '[crux.api :as crux])
-(import '[crux.api ICruxAPI])
 
 (def my-node
   (crux/start-node
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -48,7 +48,7 @@
                                         :args-doc (s/? ::args-doc)))
 
 (s/def ::tx-op (s/multi-spec tx-op first))
-(s/def ::tx-ops (s/coll-of ::tx-op :kind vector?))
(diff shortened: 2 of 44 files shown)


jarohen:
decided against doing this. at the moment we have one node object which implements both the Clojure protocol and the Java interface - splitting them out meant that it needed to be two separate objects (one implementing the protocol, one implementing the interface) which means that Clojure people would need to explicitly get hold of the Java object for any methods that required it.

more thought required.

Issue #606, opened 2020-01-28
Store documents in JDBC Document Store


jonpither:
We should be able to use JDBC to store documents. This should still allow us to store txes in Kafka (or JDBC). As part of this we will also need a node local cache.

Implementation wise, the current Document Store / Object Store split as in https://github.com/juxt/crux/pull/559 could be collapsed (we have a single store with a cache).

PR #620, opened 2020-01-31
Document Store backed Object Store

jonpither:


jonpither:
An issue with this - if you use a DocumentStoreBackedObjectStore and with tx-fns, then the tx-fn docs will get submitted back to the DocumentStore, which might not be what you want if you're using Kafka (each node submitting tx-fn resulting docs back - same situation with tombstones, but less useful).

To solve this, we could make `put-objects` store docs locally-only? (keep it's own persistent store). 

For now though, it's an open question. I would like to tackle https://github.com/juxt/crux/issues/606 to gain more insight.

jonpither:
Interesting spike, but prefer to iterate the document-store and object-store as separate components.

PR #623, opened 2020-01-31
(take 2) fix #612, range inversion edge case

### LINKED ISSUE #612: Range constraints broken when value is before variable
The range constraints are getting reversed in the wrong way, see the `crux.query/range->inverse-range` map and the `crux.query-test/test-simple-numeric-range-search` test. Adding the following should pass, but it doesn't:

```clj
(t/is (= #{[:ivan]} (api/q (api/db *api*) '{:find [i]
                                                :where [[i :age age]
                                                        [(<= 21 age)]]})))
```
Unless I'm mistaken, the solution is to reverse `<` to `>` and `<=` to `>=` etc. in the map above. Like it's the operator arguments that get reversed, not the meaning of the operator itself.

jarohen:
- e.g. `<=` should invert to `>=`, not `>`

redoing #621 because I buggered up the merge

### DIFF
--- a/crux-core/src/crux/query.clj
+++ b/crux-core/src/crux/query.clj
@@ -146,10 +146,10 @@
                                           >= (comp not neg? compare)
                                           = =})
 
-(def ^:private range->inverse-range '{< >=
-                                      <= >
-                                      > <=
-                                      >= <
+(def ^:private range->inverse-range '{< >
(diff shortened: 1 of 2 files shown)


jarohen:
hey @hraberg, good to merge this?

PR #611, opened 2020-01-30
Adding Cloudwatch connectivity to crux-metrics

t-taylor:
#597 

Also
* All args are taken as a flat map to allow for java config format
* Now no dep on `metrics-clojure` the relevant code has been stripped out

PR #628, opened 2020-02-04
Adding dropwizard to prometheus

### LINKED ISSUE #625: Crux metrics component to start a Prometheus server
* using Dropwizard -> Prometheus bridge
* component to start a Prometheus HTTP server, configurable port
* investigate other standard exporters

t-taylor:
Resolves #625 

Based of pr #611

### DIFF
--- a/crux-metrics/src/crux/metrics/dropwizard/prometheus.clj
+++ b/crux-metrics/src/crux/metrics/dropwizard/prometheus.clj
@@ -3,46 +3,22 @@
             [iapetos.core :as prometheus]
             [iapetos.collector.jvm :as jvm]
             [iapetos.standalone :as server])
-  (:import [org.dhatim.dropwizard.prometheus PrometheusReporter]
-           [io.prometheus.client.exporter PushGateway]
+  (:import [org.dhatim.dropwizard.prometheus PrometheusReporter Pushgateway]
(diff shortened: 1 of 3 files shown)


jarohen (changes requested):
cheers Tom, looks pretty straightforward thanks to your previous work

I'd be tempted to just go with the pull-based metrics, to keep things simpler for our users - we don't seem to suit the use case for push metrics. what do you think?

crux-metrics/src/crux/metrics/dropwizard/prometheus.clj
```diff
@@ -0,0 +1,48 @@
...
+  (reify Closeable
+    (close [this]
+      (.stop reporter))))
+
+(defn prometheus-registry ^iapetos.registry.IapetosRegistry
+
+  [^MetricRegistry registry {::keys [jvm-metrics?]}]
+  (cond-> (-> (prometheus/collector-registry)
+              (prometheus/register (DropwizardExports. registry)))
+    jvm-metrics? (jvm/initialize)))
```
jarohen: might these get included anyway as part of the Dropwizard registry?

jarohen: not by default, let's let the prometheus reporter submit prometheus jvm metrics

Issue #575, opened 2020-01-22
Watdiv bench should be runnable against other databases


jarohen:
We've got RDF4J as a comparison database in crux-bench already, this is to port the others from crux-bench-watdiv to the new crux-bench framework

* Get watdiv up-n-running in crux-bench for the other comparison databases, ported from crux-bench-watdiv if need be (using the new crux-bench patterns)
* As a one-off, run the watdiv bench against the comparison database, storing the results in S3
* For the Crux overnight run, we'd like to see the comparison timings in the results, duplicating the comparison stats for each database:
```clojure
{:<database> {:crux-failures []
              :crux-errors []
              :crux-time-taken <duration>
              :<database>-time-taken <duration>}
 ...}
```
* I think this is the last of the crux-bench porting - let's use this opportunity to clear up `crux-bench-watdiv`, as well as any other watdiv code lying around the codebase

jarohen:
Reopened because there's a couple of other databases that we have legacy benchmarks for but haven't yet been ported to crux-bench

PR #632, opened 2020-02-05
Watdiv Benchmarking 

### LINKED ISSUE #562: Watdiv benchmark running nightly against Crux standalone
- port watdiv over to the new bench harness (with a suitable query timeout set)
- include comparisons against results from similar databases in output metrics (Slack and CW)

### LINKED ISSUE #627: Watdiv benchmark expected results saved
* Existing watdiv comparisons ported over to new bench harness, runnable as ad-hoc jobs through the same ECS setup 
* configurable thread count for how many queries can run in parallel
* Expected outputs saved in repo (to compare against our outputs)
* timings stored for each database for each query, along with metadata on the version used

danmason:
Addresses the various issues related to watdiv (#627, #575 and #562) - need to add some documentation for running various benchmarks within a subsequent PR.


crux-bench/src/crux/bench/main.clj
```diff
@@ -3,11 +3,25 @@
...
+            [crux.bench.watdiv :as watdiv]))
 
-(defn -main []
-  (bench/post-to-slack (format "*Starting Benchmark*, Crux Version: %s, Commit Hash: %s\n"
-                               bench/crux-version bench/commit-hash))
-  (bench/with-node [node]
-    (devices/run-devices-bench node)
-    (weather/run-weather-bench node)))
+(def cli-options
+  [["-nc" "--no-crux" "Don't run Crux benchmark"]
```
jarohen: nice idea, but this is a little unusual (to me, at least) - I can't think of another example of using flags to alter the behaviour of the application so much - we've essentially got three different applications here now. also, if you're going to use `-`, stick to a single character, `--` for long options, (as you've done). the `java` CLI is very much the annoying exception here rather than the rule!

I'd be tempted to have these as different Clojure main namespaces, which you can then specify using the docker CMD

crux-bench/src/crux/bench/watdiv.clj
```diff
@@ -0,0 +1,483 @@
...
+                       (reset! all-jobs-submitted true)
+                       (doseq [^java.util.concurrent.Future f job-features] (.get f))
+                       (reset! all-jobs-completed true)
+                       (.shutdownNow pool)
+                       {:num-tests num-tests :num-threads num-threads}
+                       (catch InterruptedException e
+                         (.shutdownNow pool)
+                         (throw e))))))
+
+;; Datomic bench ===
```
jarohen: this is a sign that you might want to consider a different namespace :smile: especially if you go with the 'different clojure mains' suggestion, above

crux-bench/src/crux/bench/watdiv.clj
```diff
@@ -0,0 +1,483 @@
...
+   "watdiv-stress-100/test.3.desc" "watdiv-stress-100/test.3.sparql"
+   "watdiv-stress-100/test.4.desc" "watdiv-stress-100/test.4.sparql"})
+
+(def query-timeout-ms 30000)
+
+(def watdiv-input-file (io/resource "watdiv.10M.nt"))
+
+(defn execute-stress-test
+  [conn query-fn test-name num-tests ^Long num-threads]
+  (bench/run-bench (keyword test-name)
```
jarohen: `run-bench` has `{:style/indent 1}` metadata on it to suggest to editors that it has one 'special' arg, and everything else is 2-space indented (because it's a block) - I'm guessing this is vim? :smile:

crux-bench/src/crux/bench/watdiv.clj
```diff
@@ -0,0 +1,483 @@
...
+                   (let [all-jobs-submitted (atom false)
+                         all-jobs-completed (atom false)
+                         pool (java.util.concurrent.Executors/newFixedThreadPool (inc num-threads))
+                         job-queue (java.util.concurrent.LinkedBlockingQueue. num-threads)
+                         job-features (mapv (fn [_]
+                                              (.submit
```
jarohen: beats me why does this code went to the effort of creating a fixed thread pool, and then implemented its own job management?! :man_facepalming:

crux-bench/src/crux/bench/watdiv.clj
```diff
@@ -0,0 +1,483 @@
...
+            [clojure.string :as string])
+  (:import [java.io Closeable File StringReader]
+           java.util.Date
+           org.eclipse.rdf4j.query.Binding
+           org.eclipse.rdf4j.repository.RepositoryConnection
+           org.eclipse.rdf4j.repository.sail.SailRepository
+           org.eclipse.rdf4j.rio.RDFFormat
+           org.eclipse.rdf4j.sail.nativerdf.NativeStore))
+
+(defn output-to-file [out output]
```
jarohen: best not to open the file, write a line to it, close the file every time - we can use a writer instead (one to pair on if you're not familiar with the pattern)

```clojure
(with-open [w (io/writer (io/file ...))]
  (doseq [m maps]
    (.write w (prn-str m))))
```
general pattern is ^^, but in this case you won't have a `doseq` - just write for each output line

crux-bench/src/crux/bench/watdiv.clj
```diff
@@ -0,0 +1,483 @@
...
+        (execute-stress-test
+         conn
+         (fn [conn q] (d/query {:query (sparql/sparql->datalog q)
+                                :timeout query-timeout-ms
+                                :args [(d/db conn)]}))
+         "stress-test-datomic"
+         test-count
+         thread-count)))))
+
+;; rdf bench ===
```
jarohen: likewise!

crux-bench/src/crux/bench/watdiv.clj
jarohen: yeah, let's pair up on a bit of a refactor here

jarohen (changes requested):
nice one gents :smile: 

bit of moving about to do, bit of pairing up on refactoring the harness, and I reckon we're good.

particularly, on the harness, I'm not sure whether we're making it clear which database is running; not sure what would get output into the analytics?

crux-bench/src/crux/bench/main.clj
danmason: considering we are already specifying the main directly anyway - yeah, I like that idea far more.

crux-bench/src/crux/bench/watdiv.clj
danmason: yeah, this is one of the relics of it from before - wasnt particularly sure what it was doing

crux-bench/src/crux/bench/watdiv.clj
danmason: Also, if there is still a file, the current method will just append result sets to the end - was wondering how id do this, will give using a writer a go :)

danmason:
- ~Do not post the query results to slack (figure that out)~
- Output information about watdiv ingest
- Run these on ECS to get representative results

jarohen:
For the other databases, let's save the results to `s3://crux-bench/<db>-<db-version>/<db>-<yyyymmdd>-<hhmmss>Z.edn` - will need to create the bucket, set up the perms, and save the file on completion

For Crux, let's pull down these files (hardcode a specific version to compare against) to use for comparison

crux-bench/bin/run_task.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 aws ecs run-task\
- --task-definition crux-bench\
+ --task-definition crux-bench-latest\
```
jarohen, line 3: rather than having multiple task definitions per version, could we specify the docker image version as a container override too?

jarohen: turns out it's not possible to override the image version. shame!

PR #633, opened 2020-02-05
Removing snapshots from the public API, #410

### LINKED ISSUE #410: We've clarified 'snapshots' in the DB API
Snapshots have been included in the public API for two main reasons - evict-safe reads, and query performance. Snapshots are resources that need to be closed - we close them internally where we can, but in the case of lazy queries, or when re-using a snapshot between several requests, we need the user to take responsibility for the lifecycle of the snapshot. Unfortunately, it's often unclear to a user what they're supposed to do with a snapshot, or even why they need them at all - we'd like to hide this from them if possible.

Using the following personas to judge potential solutions:

1. the 'first 5 minuter' - this person is at a REPL, doesn't know Crux, probably copy-pasting commands from a tutorial to see what happens. very little tolerance for complexity, we'd like to get them started as soon as possible, and not introduce too many new concepts (i.e. no 'trust us, you'll want this later', if possible)
2. the 'webapp implementor' - wants to write an HTTP handler for a request to write some data to Crux, and then respond with the results of the write. needs to understand that Crux is fundamentally async, but still a relatively simple use-case. data of a single tx/query easily fits into memory, but latency's relatively important here.
3. the 'bulk user' - has a lot of data to ingest/query. happy to accept some advanced Crux concepts in return for higher throughput/lower memory usage (i.e. laziness)

With those in mind, what might we want Crux to look like for each of these users (an 'ideal state', if you like)?

1 - this person is working at the REPL with a standalone node, so they probably don't need any 'sync' call
```clojure
(def node (start-node {...}))
(crux/submit-tx node [...])

(let [db (crux/db node)]
  (crux/q db '{...}) ;; eager, returns vector
  (crux/entity (crux/db node) :id))
```

```java
ICruxDatasource db = node.db();
db.query(query).forEach(row -> ...);
```

2 - this person needs a sync call to read their write, but we'd like them not to worry about closing resources, for example - everything can be eager.

```clojure
(let [tx (crux/submit-tx node [...])
      db (crux/db node tx <timeout>)]
  (crux/q db {...}) ; eager
  (crux/q db {...})
  (crux/history-descending db ...) ; also eager
  (crux/entity db :id))
```

3 - for this user, we introduce a resource to help with long running, lazy queries, and get them to own the lifetime of that resource:

```clojure
(let [tx (crux/submit-tx node [...])
      db (crux/db node tx <timeout>)]
  (with-open [res (crux/open-q db '{...})]
    ;; lazy now - we ensure we've finished processing within the with-open block
    (doseq [row (iterator-seq res)]
      ...))

  (with-open [history (crux/open-history-descending db ...)]
    (->> (iterator-seq history)
         (reduce (fn [acc el]
                   ...)
                 init))))
```

```java
ITransaction tx = node.submitTx(...);
ICruxDatasource db = node.db(tx, <timeout>);

try (Iterator res = node.openQuery(...)) {
  res.forEachRemaining(row -> ...);
}

try (Iterator history = node.openHistoryDescending(...)) {
  history.forEachRemaining(entry -> ...);
}
```

* We could also introduce a resource to share a snapshot between multiple calls.

jarohen:
resolves #410, see proposal over there. draft for now

N.B.
* The Java API for lazy functions returns `Stream`s, the Clojure API returns Clojure seqs, so the tests now need to use the Clojure API (it still tests the Java API).
* @refset the 'extra arity' for `entity` goes as part of this - we keep the performance benefit when the user's opened up a read-tx
* @danmason have updated the `ITxLog` pattern to use streams too
* it's possible to open up nested ReadTxs, but mainly because it was easier to do this than prevent it. closing a nested ReadTx is a no-op - it's the outermost ReadTx that closes the snapshot (JDBC does it, too - we wouldn't be the first!)
* ReadTx name suggestions welcome. ironically, 'snapshot' would probably be ideal, if we didn't use it throughout.

TODO:
- [X] api-test/tx-test/query-test
- [ ] remaining tests
- [X] replace ITxLog with same pattern
- [ ] docs

could probably do with reflecting on this for a bit, too. 

particularly, I'm not convinced about the user needing to open-read-tx, then open-history-asc 
- might just have another `history-asc` within the read-tx (in addition/as opposed to `open-history-asc` on the DB, would require a separate interface)
- although this same pattern wouldn't work for `q` - we already have an eager `q` function which'd conflict with the one on the read-tx interface.

### DIFF
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -5,8 +5,7 @@
             [crux.codec :as c]
             [crux.io :as cio]
             [clojure.tools.logging :as log])
-  (:import [crux.api Crux ICruxAPI ICruxIngestAPI
-            ICruxAsyncIngestAPI ICruxDatasource ITxLog]
+  (:import (crux.api Crux ICruxAPI ICruxIngestAPI ICruxAsyncIngestAPI ICruxDatasource)
            java.io.Closeable
            java.util.Date
            java.time.Duration
@@ -157,16 +156,16 @@
   Returns a map with details about the submitted transaction,
   including tx-time and tx-id.")
 
-  (open-tx-log ^crux.api.ITxLog [this from-tx-id with-ops?]
+  (open-tx-log ^java.io.Closeable [this from-tx-id with-ops?]
     "Reads the transaction log. Optionally includes
   operations, which allow the contents under the :crux.api/tx-ops
   key to be piped into (submit-tx tx-ops) of another
   Crux instance.
 
-  from-tx-id      optional transaction id to start from.
-  with-ops?       should the operations with documents be included?
+  from-tx-id  optional transaction id to start from.
+  with-ops?   the operations with documents be included?
 
-  Returns an iterator of the TxLog"))
+  Returns a closeable seq of the transaction log"))
 
 (extend-protocol PCruxNode
   ICruxAPI
@@ -233,8 +232,9 @@
   (submit-tx [this tx-ops]
     (.submitTx this (conform-tx-ops tx-ops)))
 
-  (open-tx-log ^crux.api.ITxLog [this from-tx-id with-ops?]
-    (.openTxLog this from-tx-id with-ops?)))
+  (open-tx-log ^java.io.Closeable [this from-tx-id with-ops?]
+    (-> (.openTxLog this from-tx-id with-ops?)
+        (cio/stream->closeable-seq))))
 
 ;; TODO docs
 (defprotocol PCruxDatasource
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -3,12 +3,14 @@
   (:refer-clojure :exclude [sync])
   (:require [clojure.spec.alpha :as s]
             [crux.codec :as c]
+            [crux.io :as cio]
             [clojure.tools.logging :as log])
   (:import [crux.api Crux ICruxAPI ICruxIngestAPI
             ICruxAsyncIngestAPI ICruxDatasource ITxLog]
            java.io.Closeable
            java.util.Date
-           java.time.Duration))
+           java.time.Duration
+           (java.util.stream Stream)))
 
 (s/def :crux.db/id (s/and (complement string?) c/valid-id?))
 (s/def :crux.db/evicted? boolean?)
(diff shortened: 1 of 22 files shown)


danmason:
A big :+1: from me in terms of using Streams. :)

Looking good to me so far! Docs will need to be clear about the fact we're using Streams, but I think this is a welcome change.

crux-core/src/crux/api.clj
```diff
@@ -155,16 +156,16 @@
...
+  (open-tx-log ^java.io.Closeable [this from-tx-id with-ops?]
     "Reads the transaction log. Optionally includes
   operations, which allow the contents under the :crux.api/tx-ops
   key to be piped into (submit-tx tx-ops) of another
   Crux instance.
 
-  from-tx-id      optional transaction id to start from.
-  with-ops?       should the operations with documents be included?
+  from-tx-id  optional transaction id to start from.
+  with-ops?   the operations with documents be included?
```
danmason, line 166: nit: 'include operations with documents?'.
Unsure on that one, just reads a bit strange to me.

jarohen: yep, unintentional delete, derp - thanks

hraberg:
Overall looks good, in line with what we discussed! A few minor comments, not really related to the core PR:

* Consistency, use of `[]`for imports / require etc.
* The change to use the Clojure API for `crux.api-test` has the drawback of not properly exercising the Java API (which would, at least earlier, exercises the Clojure one indirectly so you get two tests for one), so we might lose coverage unless that's done somewhere else. Are a bunch of boring corner cases and class cast exceptions it can run into.

jarohen:
cheers @hraberg 
> Overall looks good, in line with what we discussed! A few minor comments, not really related to the core PR:
> 
> * Consistency, use of `[]`for imports / require etc.

Yep, will do. I prefer `()` for imports, because the first element is different to the rest (and formatters align it as such - aligning subsequent elements under the second rather than the first, as vectors do). That said, consistency is arguably more important, especially within a namespace.

> * The change to use the Clojure API for `crux.api-test` has the drawback of not properly exercising the Java API (which would, at least earlier, exercises the Clojure one indirectly so you get two tests for one), so we might lose coverage unless that's done somewhere else. Are a bunch of boring corner cases and class cast exceptions it can run into.

The Clojure API is now (well, already was) exercising the Java API via the `(extend-protocol PCruxNode ICruxAPI ...)` - e.g. calls to `api/submit-tx` necessary go via `(.submitTx ...)` - in fact, it's already picked up on a few cast exceptions in dev :smile:

jarohen:
> A big +1 from me in terms of using Streams. :)
> 
> Looking good to me so far! Docs will need to be clear about the fact we're using Streams, but I think this is a welcome change.

cheers @danmason :smile:

jarohen:
worth me starting fresh on this, I reckon

PR #637, opened 2020-02-07
Freezing CruxId's in Documents (Java Alpha API)

### LINKED ISSUE #636: Alpha Java API: CruxId cannot be used in Document
New alpha API (`20.01-1.6.2-alpha`) fails with an exception `Unfreezable type: class crux.api.alpha.CruxId` when I try to put an instance of `CruxId` into a `Document` to represent a foreign-key reference. Unfortunately `CruxId` is not transparent, so it's not even possible to properly unbox it in client code.

I guess, a quick patch would be to delegate encoding of this type to its inner field `id`. A drawback would be that such values will be *decoded* into a different type (actual Keyword, UUID, or a Map in our case), which might be quite confusing.

danmason:
(Resolves #636)

Makes the `.toEdn()` function public on `CruxId`, and uses it to delegate freezing to the underlying type of CruxId. Thawing reconstructs a CruxId from these types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Bug fixes
 * [#565](https://github.com/juxt/crux/pull/565): query predicate at zero join depth can now stop tuples from being returned
 * [#507](https://github.com/juxt/crux/issues/507): ranges in rules revert to a predicate if neither argument is a logic var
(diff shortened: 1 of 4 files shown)


jarohen:
Hey @danmason - as discussed, this'll cause issues for Clojure clients reading data submitted through the Java API, because Clojure doesn't have any awareness of `CruxId` objects. I doubt we'll want to serialise these all the way down to disk, so we'll need to carefully consider the right place for this conversion

PR #638, opened 2020-02-09
Docs : missing part for start node and whitespace

spacegangster:
mostly whitespace

### DIFF
--- a/docs/src/docs/examples.clj
+++ b/docs/src/docs/examples.clj
@@ -12,7 +12,7 @@
 
 ;; tag::start-standalone-node[]
 (defn start-standalone-node ^crux.api.ICruxAPI [storage-dir]
-  (crux/start-node {:crux.node/topology 'crux.standalone/topology
+  (crux/start-node {:crux.node/topology '[crux.standalone/topology]
                     :crux.node/kv-store 'crux.kv.memdb/kv
                     :crux.kv/db-dir (str (io/file storage-dir "db"))
                     :crux.standalone/event-log-kv-store 'crux.kv.memdb/kv
--- a/docs/src/docs/examples.clj
+++ b/docs/src/docs/examples.clj
@@ -12,7 +12,7 @@
 
 ;; tag::start-standalone-node[]
 (defn start-standalone-node ^crux.api.ICruxAPI [storage-dir]
-  (crux/start-node {:crux.node/topology '[crux.standalone/topology]
+  (crux/start-node {:crux.node/topology 'crux.standalone/topology
                     :crux.node/kv-store 'crux.kv.memdb/kv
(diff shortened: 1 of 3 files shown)


docs/src/docs/examples.clj
```diff
@@ -12,11 +12,12 @@
 
 ;; tag::start-standalone-node[]
 (defn start-standalone-node ^crux.api.ICruxAPI [storage-dir]
-  (crux/start-node {:crux.node/topology '[crux.standalone/topology]
```
spacegangster: so the vector form doesn't conform the actual spec

docs/src/docs/examples.clj
```diff
@@ -12,11 +12,12 @@
...
+  (crux/start-node {:crux.node/topology 'crux.standalone/topology
                     :crux.node/kv-store 'crux.kv.memdb/kv
                     :crux.kv/db-dir (str (io/file storage-dir "db"))
                     :crux.standalone/event-log-kv-store 'crux.kv.memdb/kv
                     :crux.standalone/event-log-dir (str (io/file storage-dir "event-log"))}))
+(def node (start-standalone-node "crux-store"))
```
spacegangster: so you can just copy and paste

spacegangster:
another thought was, from a canary developer's perspective, would be cool to have a ready scenario for a canary deployment, so to prioritise rocks-db usage. I could _try_ to bring that into docs with a PR if you agree.

docs/src/docs/examples.clj
danmason: This was bought in by something currently awaiting release - #589 (The docs currently point to master, so things like this pop up from time to time. We want to address that, see #610)

spacegangster: Thanks! Yeah, I forgot that docs could point to unreleased changes.
#610 would be awesome

docs/src/docs/examples.clj
jarohen: I'd come down on the side of not including these top-level side-effects in our examples, particularly given this opens a closable resource - it's ok for REPL usage but outside of this we should be encouraging users to use `with-open`, or a component-like framework that ensures resources are closed properly.

(also, for internal dev experience, the tests evaluate this namespace to ensure the examples still work - opening a node here would likely affect being able to run the tests multiple times)

spacegangster: @jarohen good, then two options are left – discard the line or wrap into a `comment` form

spacegangster: @jarohen I've left a commented out suggestion, but feel free to drop it

jarohen:
thanks @spacegangster :smile:

PR #639, opened 2020-02-10
Corrected NPE when writing node checkpoint with event-log-kv-store

neuromantik33:


jarohen:
thanks @neuromantik33 :smile:

PR #645, opened 2020-02-12
Add rocksdb metrics

### LINKED ISSUE #495: Crux nodes expose metrics about what they're currently ingesting

### LINKED ISSUE #496: Crux exposes basic RocksDB metrics through the status endpoint
Including `num-snapshots` if possible (requested by @refset)

t-taylor:
Fixes #496 

- [x] Configuration parameter for how often Rocks sends us stats
- [x] Can we get away with a lower 'stats level' (on Statistics, i.e. lower overhead) and still get the tickers?
- [x] Update RocksDB docs with instructions about how to enable metrics
- [x] There's also some gauges that could be useful (register them in `crux.kv.rocksdb.metrics`), via db.getLongProperty - anything in [here](https://github.com/facebook/rocksdb/blob/master/include/rocksdb/db.h#L593) that's numeric?
- [x] Don't think we need with-kv-metrics now? tbh, we could consider dropping estimate-num-keys in the status map in general
- [x] Tidy up previous RocksDB metric parsing


refset (approved):
I had a quick look through the other info available and nothing else jumped out as a priority: https://github.com/facebook/rocksdb/blob/c6f75516b7a13a0de4d197e0157b3191d28a4c18/include/rocksdb/db.h#L588-L804

Equally for the raw stats excerpt mentioned previously (and copied here for posterity):
```
** Compaction Stats [default] **
Level    Files   Size     Score Read(GB)  Rn(GB) Rnp1(GB) Write(GB) Wnew(GB) Moved(GB) W-Amp Rd(MB/s) Wr(MB/s) Comp(sec) Comp(cnt) Avg(sec) KeyIn KeyDrop
----------------------------------------------------------------------------------------------------------------------------------------------------------
  L0      2/0   16.29 KB   0.5      0.0     0.0      0.0       0.0      0.0       0.0   1.0      0.0      3.4         0         1    0.002       0      0
 Sum      2/0   16.29 KB   0.0      0.0     0.0      0.0       0.0      0.0       0.0   1.0      0.0      3.4         0         1    0.002       0      0
 Int      0/0    0.00 KB   0.0      0.0     0.0      0.0       0.0      0.0       0.0   1.0      0.0      3.4         0         1    0.002       0      0
Uptime(secs): 0.0 total, 0.0 interval
Flush(GB): cumulative 0.000, interval 0.000
AddFile(GB): cumulative 0.000, interval 0.000
AddFile(Total Files): cumulative 0, interval 0
AddFile(L0 Files): cumulative 0, interval 0
AddFile(Keys): cumulative 0, interval 0
Cumulative compaction: 0.00 GB write, 0.19 MB/s write, 0.00 GB read, 0.00 MB/s read, 0.0 seconds
Interval compaction: 0.00 GB write, 0.19 MB/s write, 0.00 GB read, 0.00 MB/s read, 0.0 seconds
Stalls(count): 0 level0_slowdown, 0 level0_slowdown_with_compaction, 0 level0_numfiles, 0 level0_numfiles_with_compaction, 0 stop for pending_compaction_bytes, 0 slowdown for pending_compaction_bytes, 0 memtable_compaction, 0 memtable_slowdown, interval 0 total count
** File Read Latency Histogram By Level [default] **
** DB Stats **
Uptime(secs): 0.0 total, 0.0 interval
Cumulative writes: 2000 writes, 2000 keys, 2000 commit groups, 1.0 writes per commit group, ingest: 0.00 GB, 1.27 MB/s
Cumulative WAL: 2000 writes, 0 syncs, 2000.00 writes per sync, written: 0.00 GB, 1.27 MB/s
Cumulative stall: 00:00:0.000 H:M:S, 0.0 percent
Interval writes: 2000 writes, 2000 keys, 2000 commit groups, 1.0 writes per commit group, ingest: 0.05 MB, 1.27 MB/s
Interval WAL: 2000 writes, 0 syncs, 2000.00 writes per sync, written: 0.00 MB, 1.27 MB/s
Interval stall: 00:00:0.000 H:M:S, 0.0 percent
```

Nice job!

crux-metrics/src/crux/metrics/rocksdb.clj
```diff
@@ -0,0 +1,89 @@
...
+             :read-speed #'crux.metrics.rocksdb/read-speed
+             :db-uptime #'crux.metrics.rocksdb/db-uptime
+             :file-size #'crux.metrics.rocksdb/file-size
+             :total-keys-dropped #'crux.metrics.rocksdb/total-keys-dropped
+             :total-keys-in #'crux.metrics.rocksdb/total-keys-in
+             :num-snapshots #'crux.metrics.rocksdb/num-snapshots})
+
+(defn assign-gauges [registry {:keys [crux.node/kv-store]}]
+  (run! (fn [[k v]]
+          (dropwizard/gauge registry ["rocksdb" (name k)] #(v (:db (:kv kv-store))))) gauges))
```
jarohen: formatting nit: `gauges` is the second param to `run!` but it's hidden behind the first one, not easy to spot - new line?

crux-metrics/src/crux/metrics/rocksdb.clj
```diff
@@ -0,0 +1,89 @@
...
+      (string/split #"Uptime")
+      (second)
+      (string/split #"\s+")
+      (nth 3)
+      (read-string)))
+
+(defn cumulative-writes
+"Secs"
+  [kv]
+  (-> (.getProperty kv "rocksdb.stats")
```
jarohen: we're using this more than a few times, reckon it's time to extract out a function.
also, these functions are combining the side effect (reading the raw stats) and the pure transformation - if you split these two parts out, you could write a test for the parsing logic

jarohen: it's also not ideal that we have to pull the stats from Rocks and parse for each individual gauge - have you got a sense of how expensive this operation is?

jarohen (commented):
thanks @t-taylor - did you try using the `Statistics` classes btw? on the face of it, it seems like these allow you to get stats without parsing this property string. also seems like it might be possible to get Rocks to call us back with stats (so we could use meters, etc, which would open up rate metrics) rather than a cumulative gauge?

Javadoc here: https://javadoc.io/doc/org.rocksdb/rocksdbjni/6.2.2/index.html, particularly the `Statistics` and `StatisticsCollector` classes, as entry points

as discussed, it'd also be worth considering the pros/cons of where to put this logic - it likely isn't the responsibility of crux-metrics to collect stats for non-core modules (we'll probably want third-party modules to be able to supply metrics in the future too). what options would you suggest?

jarohen (changes requested):
will put TODOs on the PR

crux-metrics/src/crux/metrics/rocksdb.clj
```diff
@@ -0,0 +1,89 @@
...
+      (string/split #"Cumulative writes: ")
+      (second)
+      (string/split #"\s+")
+      (first)
+      (read-string)))
+
+(defn num-snapshots
+  [kv]
+  (read-string (.getProperty kv "rocksdb.num-snapshots")))
```
jarohen: `getLongProperty` saves this `read-string` step

jarohen (commented):
cheers @t-taylor, good PR :smile: 

I'd like to ponder on the kv -> topology change a little more before merging this in, will also get some feedback from the others

PR #646, opened 2020-02-12
Change crux.node & crux.standalone to use memdb by default

danmason:


jarohen:
all good @danmason - have incorporated these commits into #659

PR #647, opened 2020-02-12
API iteration (removing snapshot parameters, `db` -> `at`), #410

### LINKED ISSUE #410: We've clarified 'snapshots' in the DB API

jarohen:
(eventually resolves #410)

Still a fair way to go on this one, but wanted to get the draft PR up to get some eyes on it before I go much further.

* [x] Update #410 with the aims of this new spike
* [ ] HTTP server/client API
* [ ] Docstrings/Javadoc
* [ ] Docs/tutorials

* I've kept it backwards-compatible for now, so the API tests are passing (for local nodes, I haven't updated the HTTP server/client yet). indeed, for however long we're preserving backwards compatibility, I'd like to keep these tests about. a lot of these API tests indirectly call through to the new API, so I'm not _too_ worried about not extensively testing the new API directly yet.
* There's a couple of names that could do with being clearer, particularly `IFixedInstantQueryAPI` :nauseated_face:.
* I've expanded `q` to `query` - we don't shorten much else in the public API (even less so in the Java API), auto-complete should get this most of the time, and even if it doesn't, I doubt it takes significantly longer to type. discuss?
* `BitemporalInstant` (containing vt, tt, tid) has namespaced keys, despite being a record, because this is how users currently access these fields. I've also experimented with moving a couple of other records that way - particularly Node - because most of its components come from a namespaced map, with the aim of trying to always refer to the components in the same way, no matter what map (or record) they're in. That said, namespaced keys in Clojure records aren't ideal, RFC. 
* I'm also not a fan of users needing to know which namespace each of the different keys are in - maybe `:crux/valid-time` is sufficient, given valid-time only has one meaning throughout Crux?

### DIFF
--- a/crux-bench/test/crux/bench/rocksdb_microbench_test.clj
+++ b/crux-bench/test/crux/bench/rocksdb_microbench_test.clj
@@ -7,7 +7,8 @@
             [crux.kv :as kv]
             [crux.fixtures :as f]
             [crux.bench :as bench]
-            [clojure.java.io :as io]))
+            [clojure.java.io :as io]
+            [crux.node :as n]))
 
 (defn with-rocksdb-node* [f]
   (f/with-tmp-dir "dev-storage" [data-dir]
@@ -34,4 +35,4 @@
                (db/index-docs (:indexer node) (->> doc-batch (into {} (map (juxt c/new-id identity)))))))
 
             (with-open [snapshot (kv/new-snapshot (:kv-store node))]
-              (t/is (= first-doc (db/get-single-object (:object-store node) snapshot (c/new-id first-doc)))))))))))
+              (t/is (= first-doc (db/get-single-object (::n/object-store node) snapshot (c/new-id first-doc)))))))))))
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -138,7 +138,10 @@
   "Provides API access to Crux."
 
   (at [node] [node bitemp-inst] [node bitemp-inst timeout])
-  (open-snapshot-at [node] [node bitemp-inst] [node bitemp-inst timeout])
+  (open-snapshot-at
+    ^crux.api.ISnapshot [node]
+    ^crux.api.ISnapshot [node bitemp-inst]
+    ^crux.api.ISnapshot [node bitemp-inst timeout])
 
   (db
     [node]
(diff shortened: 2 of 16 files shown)


danmason:
Seems like a very positive step towards our API reworking! :+1: Will have to look a bit deeper / talk about at how you are going about implementing classes at some point :)

jarohen:
killing in favour of #681

PR #651, opened 2020-02-13
Crux-bench send email

### LINKED ISSUE #650: Sending the daily bench results by email
(from @jonpither)

Once the bench run's finished, could we send the bench results by email to crux-bench-at-juxt.pro? Can be the same format as what's output to Slack

danmason:
Need to set up the crux-bench account with AWS SES, but this PR makes use of the SES SDK to send a set of emails whenever we run the crux bench


jarohen (changes requested):
thanks @danmason - can you set up crux-bench with SES (and/or let me know what access you need in order to do it), ensure it works e2e, and then re-request review?

PR #652, opened 2020-02-14
Bump memory allocated to container/JVM heap on crux-bench

danmason:
Attempting to fix the OutOfMemory error on the AWS container by allocating it the same amount of memory as used previously in the Kubernetes setup: https://github.com/juxt/crux/blob/master/crux-bench-watdiv/kubernetes_templates/app_deployment.yml

### DIFF
--- a/crux-bench/cloudformation.yaml
+++ b/crux-bench/cloudformation.yaml
@@ -132,7 +132,7 @@ Resources:
       - LogGroup
     Properties:
       Cpu: '2 vCPU'
-      Memory: '4GB'
+      Memory: '7GB'
       Family: 'crux-bench'
       ExecutionRoleArn:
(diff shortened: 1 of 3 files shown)


jarohen (changes requested):
thanks @danmason 

I reckon this still might have some issues, though - @hraberg reliably informs me that direct allocation can allocate as much memory again as is specified in Xmx, so moving to 4/7 may well still OOM. shall we go to 3/8, say? alternative is that we set `-XX:MaxDirectMemorySize` instead.

wdyt, @hraberg ?

danmason:
Fairly happy to change that to 3/8, though will let you discuss if needs be.

jarohen (approved):

PR #658, opened 2020-02-14
Introduce queue

jonpither:
Note that now the Kafka consumers will fatally exit upon an exception, as per Standalone/JDBC, as part of the unification.

I've also made the Moberg lazy-seq be realised (see `MobergQueue`).


hraberg:
Looks promising, consolidating this with the changes in #659 I guess isn't too hard, as that basically splits half of Moberg into a "Queue"? Nice twist to separate out the topics themselves from the protocol and simplify it, essentially moving them to the instance itself. That is, a queue might be a topic, it might be something else, it doesn't matter as long as the basic semantics are supported.

jarohen:
derp, sorry

PR #659, opened 2020-02-17
Replacing Moberg

### LINKED ISSUE #322: Replace Moberg with local event log
We should consider removing Moberg and just use JDBC with either H2 or SQLite for the standalone mode. Could help solve #280 and maybe also #321 (assuming this is a timing issue in related to the log). Would remove a moving piece and some code to maintain.

A drawback is that the JDBC log is still very fresh, and not tuned or tested in anger. It also has no indexes yet for example.

### LINKED ISSUE #619: Updating the standalone module to not depend on anything outside of core
The standalone module currently depends on RocksDB - we should make this only depend on core modules (memkv)

I suspect this might also involve a doc update
I also suspect that this might resolve #523

### LINKED ISSUE #653: Duplicate documents submitted to standalone node cause 'incorrect doc state' error
Repro:
* on standalone node (Kafka is unaffected)
* submit transaction T1 containing doc with content-hash C1
* submit loads more transactions (more than the Moberg batch size)
* submit another transaction T2 containing doc with content-hash C1
* if T1 hasn't been indexed by the time T2 is submitted (can simulate either with a bulk ingest, or by starting a second node and reindexing), you'll get an error like `ERROR crux.tx - Put, incorrect doc state for: #crux/id 91ef073d59b7ba3f52dbb6f6f58181dcfc2f44a4 tx id: 1619646077277186`

cheers @kosiakk for the report :smile: does this repro look like it'd affect your use case?

jarohen:
resolves #322, resolves #653, and hopefully most of the other bugs labelled 'moberg'

I've used @jonpither's tx-log/document-store separation to create a `StandaloneTxLog` and `StandaloneDocumentStore`, both just using the local kv-store/object-store. 
* main change is that documents now go into a document-store (cheers Jon for making this possible), so we don't need to worry about trying to replicate Kafka's compaction
* it creates a new WAL-like index in the event-log kv-store to store the tx-events when they're submitted.
* defaulting the standalone kv-store to use memdb so that the base topology doesn't depend on anything outside of crux-core (resolves #619)

TODO:
* [x] Ensure all the API functions for TxLog and DocumentStore are implemented and correct
* [x] Ensure that we continue indexing if we use a persistent kv-store and the node shuts down halfway through catching up on the tx backlog
* [x] Doc updates as required (particularly, don't need `event-log-kv` etc any more)


jarohen:
@jonpither @hraberg just got the doc updates to do on this one now, reckon the rest of the code is good to review :smile:

jarohen:
also cc @danmason, @johantonelli - we're removing all the `:crux.standalone/event-*` configuration params. I've done all the occurrences in the repo but I suspect there are a few elsewhere too, in tutorials/blogs?

crux-core/src/crux/backup.clj
```diff
@@ -12,13 +12,11 @@
 
 (s/def ::node-options
   (s/keys :req-un [:crux.kv/db-dir]
-          :gen    [:backup-dir]
-          :opt-un [:crux.standalone/event-log-dir]))
+          :gen    [:backup-dir]))
```
jonpither: spacing here is inconsistent with L19, prefer L19

jonpither (approved):

crux-core/src/crux/backup.clj
jarohen: likewise - have been tidying up where I've seen it but missed this one, cheers

jarohen:
@refset's pointed out a spanner in the works here that I should have considered - can't rebuild a Crux index when both the event log and the indices are in the same KV store. more thought required.

jarohen:
right, all done, separate event-log reinstated. 

just to confirm (@danmason, @johantonelli) no change to tutorials required in the end, sorry for the noise

and thanks again @refset for flagging it up :smile:

hraberg (approved):
Looks good at a glance. Some changes seem to remove the event-log-dir, despite it being "necessary" to separate them in the persisted case, but guess this is dealt with properly somehow indirectly (by having two different KV stores?)

I'm also wondering about Docker/Uberjar use-cases, seems unlikely you would want to use the MemKv there?

hraberg:
For some context, this moves it somewhat back to -way- back where we had a version of the TxLog that persisted into the kv store directly, but that version was only used for tests, as it, as @refset mentions, requires separation to actually work for rebuilding the index.

jarohen:
thanks for the review @hraberg :smile: 
> Looks good at a glance. Some changes seem to remove the event-log-dir, despite it being "necessary" to separate them in the persisted case, but guess this is dealt with properly somehow indirectly (by having two different KV stores?)

looks like I might not have reinstated enough of the event-log-dir changes, sorry :flushed: 

> I'm also wondering about Docker/Uberjar use-cases, seems unlikely you would want to use the MemKv there?

I was considering having both 'standalone' and 'rocks' configurations for these as part of #660, making the simplest case memory-only for first-time users just trying Crux (should be one line of configuration), with a simple upgrade to Rocks?

> For some context, this moves it somewhat back to -way- back where we had a version of the TxLog that persisted into the kv store directly, but that version was only used for tests, as it, as @refset mentions, requires separation to actually work for rebuilding the index.

yep - although there's still the separation here between writing to the event-log vs writing to the tx indices, so it's possible to rebuild the 'main' kv store while retaining the event-log - that what you mean?

PR #661, opened 2020-02-18
Add all node configurations to benchmarking

### LINKED ISSUE #574: Bench should run on multiple Crux configurations
We'd like to run each bench against multiple Crux configurations:

* tx-log: both standalone and embedded-kafka
* kv-store: memdb, rocksdb

Minus lmdb for now until https://github.com/juxt/crux/issues/280 is resolved

* for each configuration, we should add dimensions for the tx-log/kv-store, so that we can see which configuration it ran against

### LINKED ISSUE #579: Add a new chapter to the space tutorial with await-tx, await-tx-time

t-taylor:
Fixes #574

Adds rocksdb & memdb with embedded kafka & standalone

### DIFF
--- a/crux-bench/src/crux/bench.clj
+++ b/crux-bench/src/crux/bench.clj
@@ -29,10 +29,8 @@
 (def ^:dynamic ^:private *bench-dimensions* {})
 (def ^:dynamic ^:private *!bench-results*)
 
-(def !crux-node-type (atom ""))
-
 (defn with-dimensions* [dims f]
-  (binding [*bench-dimensions* (merge *bench-dimensions* dims {:crux-node-type @!crux-node-type})]
+  (binding [*bench-dimensions* (merge *bench-dimensions* dims)]
     (f)))
 
 (defmacro with-dimensions [dims & body]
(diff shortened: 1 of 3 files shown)


crux-bench/src/crux/bench.clj
```diff
@@ -27,8 +27,10 @@
 (def ^:dynamic ^:private *bench-dimensions* {})
 (def ^:dynamic ^:private *!bench-results*)
 
+(def !crux-node-type (atom ""))
+
 (defn with-dimensions* [dims f]
-  (binding [*bench-dimensions* (merge *bench-dimensions* dims)]
+  (binding [*bench-dimensions* (merge *bench-dimensions* dims {:crux-node-type @!crux-node-type})]
```
jarohen: this will bind the crux-node-type every time with-dimensions is called - you could call with-dimensions from your own function?

crux-bench/src/crux/bench.clj
```diff
@@ -104,22 +106,145 @@
 (defmacro with-bench-ns [bench-ns & body]
   `(with-bench-ns* ~bench-ns (fn [] ~@body)))
 
-(defn with-node* [f]
+(defprotocol GetNode
+  (give [this]))
+
+(defn nodes [data-dir] {;"standalone-lmdb"
```
jarohen: few pointers about this:
* aware that embedded kafka might throw a spanner in the works here, but is there any way we can represent this more as data so that we don't have to create a new protocol + reify it?
* guessing you've considered this, but I've missed it - is there any way we can avoid the permutation explosion here?
* formatting nit: if a `def`/`defn` is more than one line, can we have it down a line?

jarohen (changes requested):
thanks Tom - comments below. happy to pair on Monday after the standup if it'd help?

PR #677, opened 2020-02-21
Add load balanced sample webservice 

### LINKED ISSUE #674: 'Hello world' Clojure service available on AWS
- [x] provide a http server that provides some aggregations of data
- [x] Dockerise
- [x] Run on AWS ECS as a cluster of instances
- [x] Set up ELB to manage webservice cluster usage

danmason:
(#674) The 'hello world' example, with the associated cloudformation file and dockerfile.


examples/soak/bin/run_task.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
jarohen, line 1: what's this doing? assuming soak's deployed as a service, not a task?

examples/soak/setup_ecs_cluster.yml
```diff
@@ -0,0 +1,242 @@
...
+  # Remove any of the following ingress rules that are not needed.
+  FargateContainerSecurityGroup:
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      GroupDescription: Access to the Fargate containers
+      VpcId:
+        Fn::ImportValue:
+          !Join [':', [!Ref 'VpcStackName', 'VPCId']]
+
+  # Load balancers for getting traffic to containers.
```
jarohen: I don't mind comments sticking about (I normally wouldn't have comments in CF templates explaining what the AWS services are, assumption is that people will look these up in AWS docs if need be) but let's make sure the comments that do stick about still apply (e.g. we're not using a private load balancer)

examples/soak/setup_ecs_cluster.yml
```diff
@@ -0,0 +1,242 @@
...
+                - 'ecr:BatchCheckLayerAvailability'
+                - 'ecr:GetDownloadUrlForLayer'
+                - 'ecr:BatchGetImage'
+
+                # Allow the ECS tasks to upload logs to CloudWatch
+                - 'logs:CreateLogStream'
+                - 'logs:PutLogEvents'
+              Resource: '*'
+
+  SoakTask:
```
jarohen: I'd call this `SoakTaskDef` - 'task' could be an ECS task

examples/soak/setup_ecs_cluster.yml
```diff
@@ -0,0 +1,242 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
jarohen: there're a few resources in here that'd be per-application, and a few that are per-cluster - e.g. the load balancer isn't application-specific (the forwarding rule is, though) - let's figure out what'll change together, and that's likely how best to split the CF up too

examples/soak/setup_vpc.yml
```diff
@@ -0,0 +1,160 @@
...
+    Properties:
+      SubnetId: !Ref PublicSubnetOne
+      RouteTableId: !Ref PublicRouteTable
+  PublicSubnetTwoRouteTableAssociation:
+    Type: AWS::EC2::SubnetRouteTableAssociation
+    Properties:
+      SubnetId: !Ref PublicSubnetTwo
+      RouteTableId: !Ref PublicRouteTable
+
+  # Setup networking resources for the private subnets. Containers
```
jarohen: I suspect, when you do your cost analysis, you might find this costs quite a lot, relative to what it's doing :)

examples/soak/src/crux/soak/main.clj
```diff
@@ -0,0 +1,12 @@
+(ns crux.soak.main
+  (:require [ring.util.response :as ring]
```
jarohen: I'd usually alias this as `resp` - there's plenty of other namespaces in ring

jarohen (changes requested):
Thanks gents - looks like it's got all the right bits, but not necessarily the right places in CF - let's pair :)

How're you planning to redeploy a new version of the service?

examples/soak/bin/run_task.sh
danmason: It is - should kill this, I think we were considering keeping it to manually trigger the ingest task though if we do want a script to run that we should include it in that PR

PR #678, opened 2020-02-21
Jp/doc backed obj store

jonpither:


jarohen:
suspect this one's gone stale, let's reopen when we pick it up again

PR #684, opened 2020-02-25
JDBC doc store, and using open-tx-log to consume transactions

### LINKED ISSUE #606: Store documents in JDBC Document Store

jarohen:
(resolves #606)

* Adds JDBC document store
* Now that we have document stores, we can also re-use `open-tx-log` to chew through the transaction logs - don't need each tx-log to specify a separate tx consumer. 
* It calls 'fetch docs' to index the docs for a transaction just-in-time (although we could get ahead by indexing the next 10 txs worth of docs, say)
* c.kafka.consumer and c.kafka weren't doing much on their own, so I pulled them back together (subjective, though)
* Kafka is now the exception to the rule with having a document indexer - fetch-docs on the Kafka doc store is implemented as polling waiting for the doc indexer to have put the docs in the object store. the others (standalone and JDBC) implement it as a single tx indexer that pulls docs from the doc store as required
* refactored Kafka tests - removed the ones that were covered by api_test, kept the ones that were testing Kafka specific functionality (i.e. compaction + ingest client)


crux-core/src/crux/tx/consumer.clj
```diff
@@ -1,101 +1,58 @@
...
+                :let [tx (select-keys tx [::tx/tx-time ::tx/tx-id])]]
+          (log/debug "indexing" tx)
+          (db/index-docs indexer (db/fetch-docs document-store
+                                                (->> tx-events
+                                                     (into #{} (comp (mapcat tx/tx-event->doc-hashes)
+                                                                     (map c/new-id))))))
+
+          (db/index-tx indexer tx tx-events)
+
+          (log/debug "indexed" tx)
```
jonpither: Remove blank lines + debug line?

crux-core/src/crux/tx/consumer.clj
```diff
@@ -1,101 +1,58 @@
...
+
+          (when (Thread/interrupted)
+            (throw (InterruptedException.)))))
+
+      (when (Thread/interrupted)
+        (throw (InterruptedException.)))
+
+      (Thread/sleep 100))
```
jonpither: Thread sleep?

jonpither (changes requested):

crux-test/test/crux/kafka_test.clj
```diff
@@ -4,165 +4,111 @@
...
-    (doseq [[k v] (->> docs
-                       (reverse)
-                       (reduce (fn [docs doc]
-                                 (if (contains? (set (map (comp str first) docs)) (str (first doc)))
-                                   docs
-                                   (conj docs doc))) [])
-                       reverse)]
-      @(.send producer (ProducerRecord. topic k v)))))
+          (let [
+
```
jonpither: blank lines? bit sparse here.

crux-core/src/crux/tx/consumer.clj
jarohen: note: make config

jonpither (approved):

PR #690, opened 2020-02-26
Aggregates bench results into one email

### LINKED ISSUE #683: Bench should send one email per run, rather than per node config
After #661, crux-bench sends out an email for every Crux configuration, rather than one email per run. 

Likely involves refactoring `with-nodes` to return the results, and moving the slack/email code higher up, as discussed on Slack

t-taylor:
Fixes #683

### DIFF
--- a/crux-bench/src/crux/bench/main.clj
+++ b/crux-bench/src/crux/bench/main.clj
@@ -18,7 +18,10 @@
                weather-results (doto (weather/run-weather-bench node)
                                  (bench/post-slack-results :ts-weather))
 
-               watdiv-results (doto (watdiv-crux/run-watdiv-bench node {:test-count 100})
+               raw-watdiv-results (watdiv-crux/run-watdiv-bench node {:test-count 100})
+
+               watdiv-results (doto [(first raw-watdiv-results)
+                                     (watdiv-crux/summarise-query-results (rest raw-watdiv-results))]
                                 (bench/post-slack-results :watdiv-crux))]
            [devices-results
             weather-results
--- a/crux-bench/src/crux/bench/main.clj
+++ b/crux-bench/src/crux/bench/main.clj
@@ -12,20 +12,16 @@
                                bench/crux-version bench/commit-hash))
 
   (->> (bench/with-nodes [node]
-         (let [devices-results (doto (devices/run-devices-bench node)
-                                 (bench/post-slack-results :ts-devices))
+         [(doto (devices/run-devices-bench node)
+            (bench/post-slack-results :ts-devices))
 
-               weather-results (doto (weather/run-weather-bench node)
-                                 (bench/post-slack-results :ts-weather))
+          (doto (weather/run-weather-bench node)
+            (bench/post-slack-results :ts-weather))
 
-               raw-watdiv-results (watdiv-crux/run-watdiv-bench node {:test-count 100})
-
-               watdiv-results (doto [(first raw-watdiv-results)
-                                     (watdiv-crux/summarise-query-results (rest raw-watdiv-results))]
-                                (bench/post-slack-results :watdiv-crux))]
-           [devices-results
-            weather-results
-            watdiv-results]))
+          (let [raw-watdiv-results (watdiv-crux/run-watdiv-bench node {:test-count 100})]
+            (doto [(first raw-watdiv-results)
+                   (watdiv-crux/summarise-query-results (rest raw-watdiv-results))]
+              (bench/post-slack-results :watdiv-crux)))])
 
        (reduce (fn [acc [node-type result]]
                  (str acc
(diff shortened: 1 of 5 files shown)


crux-bench/src/crux/bench/main.clj
```diff
@@ -11,15 +11,21 @@
...
+         (let [devices-results (devices/run-devices-bench node)
+               weather-results (weather/run-weather-bench node)
+               watdiv-results (watdiv-crux/run-watdiv-bench node {:test-count 100})
+               result-messages [(bench/results->slack-message devices-results :ts-devices)
+                                (bench/results->slack-message weather-results :ts-weather)
+                                (bench/results->slack-message
+                                  [(first watdiv-results) (watdiv-crux/->query-result (rest watdiv-results))]
+                                  :watdiv-crux)]]))
+
+       (reduce (fn [acc [node-type result]]
```
jarohen: you're not using `acc` at all here, so I'd be surprised if this included all the results?

tbh, I suspect you don't need a `reduce` anyway - `reduce` is usually used when later elements need to access the results of iterating through earlier elements, whereas in this case you can probably combine them at the end?

crux-bench/src/crux/bench/main.clj
```diff
@@ -11,15 +11,21 @@
...
-          weather-results (weather/run-weather-bench node)
-          watdiv-results (watdiv-crux/run-watdiv-bench node {:test-count 100})
-          result-messages [(bench/results->slack-message devices-results :ts-devices)
-                           (bench/results->slack-message weather-results :ts-weather)
-                           (bench/results->slack-message
-                            [(first watdiv-results) (watdiv-crux/->query-result (rest watdiv-results))]
-                            :watdiv-crux)]]
-      (bench/send-email-via-ses (string/join "\n" result-messages))))
+  (->> (bench/with-nodes [node]
+         (let [devices-results (devices/run-devices-bench node)
```
jarohen: it looks like this `let` isn't yielding any value, so I'd be surprised if this worked?

jarohen: let's also try to split out the part that runs the bench vs the part that outputs to slack/email, it seems they're a little convoluted here. maybe the `let` could return the results out of the with-nodes, then we can process the results outside?

jarohen (changes requested):
hey gents - I'd be surprised if this one worked, I'm afraid. how're you testing this locally?

t-taylor:
> hey gents - I'd be surprised if this one worked, I'm afraid. how're you testing this locally?

Sorry James, it was meant to still be a draft at this stage. Github changes it back from draft to pr if you've written text in the pr message and then switched branch, which caught me out. I've refactored out the posting to slack per feedback now however

crux-bench/src/crux/bench/main.clj
```diff
@@ -11,15 +11,24 @@
...
-          weather-results (weather/run-weather-bench node)
-          watdiv-results (watdiv-crux/run-watdiv-bench node {:test-count 100})
-          result-messages [(bench/results->slack-message devices-results :ts-devices)
-                           (bench/results->slack-message weather-results :ts-weather)
-                           (bench/results->slack-message
-                            [(first watdiv-results) (watdiv-crux/->query-result (rest watdiv-results))]
-                            :watdiv-crux)]]
-      (bench/send-email-via-ses (string/join "\n" result-messages))))
+  (->> (bench/with-nodes [node]
+         (let [devices-results (doto (devices/run-devices-bench node)
```
jarohen: I'd inline these three in the vector below - you're only using the bindings once and the names don't really add any clarity

crux-bench/src/crux/bench/main.clj
```diff
@@ -11,15 +11,24 @@
...
+               weather-results (doto (weather/run-weather-bench node)
+                                 (bench/post-slack-results :ts-weather))
+
+               watdiv-results (doto (watdiv-crux/run-watdiv-bench node {:test-count 100})
+                                (bench/post-slack-results :watdiv-crux))]
+           [devices-results
+            weather-results
+            watdiv-results]))
+
+       (reduce (fn [acc [node-type result]]
```
jarohen: you don't need the power of a `reduce` here, map + string/join is sufficient. I'd try to steer clear of `reduce` unless later values need the accumulator - as a reader, I need to reason about how the elements might interact with each other; with a `map`, I _know_ they can't.

jarohen (changes requested):
cheers - couple of stylistic points below but I like the way the bench runs are now completely decoupled from where we're outputting the results

crux-bench/src/crux/bench/watdiv_crux.clj
```diff
@@ -75,12 +75,7 @@
                                (bench/run-bench (format "query-%d" idx)
                                                 {:result-count (count (crux/q (crux/db node) (sparql/sparql->datalog q)))}))))))))))]
 
-    (-> [(first watdiv-results)
-         (->query-result (rest watdiv-results))]
```
jarohen, line 79: this returned the raw results so that the raw results were put into CloudWatch, the summarised results to Slack - let's preserve this behaviour

jarohen (changes requested):

PR #691, opened 2020-02-26
Remove nodes using memdb from crux-bench.

danmason:
These memdb nodes would fail, as they would run out of memory fairly quickly (which was to be expected). I haved removed these from the bench run entirely, as we likely do not want to benchmark them in such a way.

### DIFF
--- a/crux-bench/src/crux/bench.clj
+++ b/crux-bench/src/crux/bench.clj
@@ -109,44 +109,12 @@
 (defn nodes
   [data-dir]
 
-  {#_"standalone-lmdb"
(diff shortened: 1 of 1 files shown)


jarohen (changes requested):
thanks Dan - standalone-* also uses memdb by default, let's remove those too

danmason:
How do you mean? The remaining `standalone-` nodes are using either rocksdb or lmdb?

jarohen:
ah, sorry you're right - don't know why I didn't spot it. will merge

PR #692, opened 2020-02-26
Generate javadoc on release

### LINKED ISSUE #530: The Javadocs should be generated and deployed on every release
We should generate the Javadoc on every release, in CI.

danmason:
(Fix #530)

Generates javadocs for both the normal crux.api java files and the crux.api.alpha java files, then uploads them to S3 under the 'crux-doc' bucket.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -157,11 +157,15 @@ jobs:
           at: ~/
       - run: javadoc -d crux-javadoc/ -cp `lein classpath` -sourcepath crux-core/src crux.api
       - aws-s3/sync:
+          arguments: |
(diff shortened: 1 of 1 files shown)


jarohen (changes requested):
looking good :) just the version in the path as mentioned on Slack, then should be good to go.

how's this getting permission to sync to S3, btw - we got some kind of CircleCI AWS user?

danmason:
Yeah, we have an IAM user called 'circleci-user', we also use it for deployment to AWS.

jarohen (approved):

jarohen:
merged

PR #695, opened 2020-02-28
Soak ingesting weather data

danmason:
An ingesting node, which reads weather data from twenty different locations every five minutes, and a 5 day forecast for each location every midnight.

### DIFF
--- a/examples/soak/cloudformation/ecs-ingest.yml
+++ b/examples/soak/cloudformation/ecs-ingest.yml
@@ -1,0 +1,198 @@
+AWSTemplateFormatVersion: '2010-09-09'
+
+Parameters:
+  EcsClusterStackName:
+    Type: String
+    Default: soak-ecs-cluster
+
+Resources:
+  LogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: 'crux-soak-ingest'
+      RetentionInDays: 365
+
+  SoakIngestECSTaskExecutionRole:
+    Type: AWS::IAM::Role
+    Properties:
+      AssumeRolePolicyDocument:
+        Statement:
+        - Effect: Allow
+          Principal:
+            Service: [ecs-tasks.amazonaws.com]
+          Action: ['sts:AssumeRole']
+      Path: /
+      Policies:
+        - PolicyName: AmazonECSTaskExecutionRolePolicy
+          PolicyDocument:
+            Statement:
+            - Effect: Allow
+              Action:
+                # Allow the ECS Tasks to download images from ECR
+                - 'ecr:GetAuthorizationToken'
+                - 'ecr:BatchCheckLayerAvailability'
+                - 'ecr:GetDownloadUrlForLayer'
+                - 'ecr:BatchGetImage'
+
+                # Allow the ECS tasks to upload logs to CloudWatch
+                - 'logs:CreateLogStream'
+                - 'logs:PutLogEvents'
+
+                # Allow the ECS tasks to get secret values from the secrets manager
+                - 'secretsmanager:GetSecretValue'
+              Resource: '*'
+
+  SoakIngestCurrentTaskDef:
+    Type: AWS::ECS::TaskDefinition
+    DependsOn:
+      - LogGroup
+    Properties:
+      Cpu: '0.5 vCPU'
+      Memory: '2GB'
+      Family: 'crux-soak-ingest-current'
+      ExecutionRoleArn:
+        Fn::GetAtt: ["SoakIngestECSTaskExecutionRole", "Arn"]
+      RequiresCompatibilities:
+        - 'FARGATE'
+      NetworkMode: 'awsvpc'
+      ContainerDefinitions:
+        - Name: 'crux-soak'
+          Image: '955308952094.dkr.ecr.eu-west-2.amazonaws.com/crux-soak:latest'
+          PortMappings:
+            - ContainerPort: 8080
+          Command:
+            - "-m"
+            - "crux.soak.ingest"
+            - "current"
+          Secrets:
+            - Name: "SOAK_SECRETS"
+              ValueFrom:
+                "arn:aws:secretsmanager:eu-west-2:955308952094:secret:soak/soak-secrets-OHMj3T"
+          LogConfiguration:
+            LogDriver: awslogs
+            Options:
+              awslogs-region: 'eu-west-2'
+              awslogs-group: !Ref LogGroup
+              awslogs-stream-prefix: ecs
+
+  SoakIngestForecastTaskDef:
+    Type: AWS::ECS::TaskDefinition
+    DependsOn:
+      - LogGroup
+    Properties:
+      Cpu: '0.5 vCPU'
+      Memory: '2GB'
+      Family: 'crux-soak-ingest-forecast'
+      ExecutionRoleArn:
+        Fn::GetAtt: ["SoakIngestECSTaskExecutionRole", "Arn"]
+      RequiresCompatibilities:
+        - 'FARGATE'
+      NetworkMode: 'awsvpc'
+      ContainerDefinitions:
+        - Name: 'crux-soak'
+          Image: '955308952094.dkr.ecr.eu-west-2.amazonaws.com/crux-soak:latest'
+          PortMappings:
+            - ContainerPort: 8080
+          Command:
+            - "-m"
+            - "crux.soak.ingest"
+            - "forecast"
+          Secrets:
+            - Name: "SOAK_SECRETS"
+              ValueFrom:
+                "arn:aws:secretsmanager:eu-west-2:955308952094:secret:soak/soak-secrets-OHMj3T"
+          LogConfiguration:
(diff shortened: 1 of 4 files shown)


examples/soak/src/crux/soak/ingest.clj
```diff
@@ -0,0 +1,88 @@
...
+                      (slurp)
+                      (string/split-lines)))
+(def soak-secrets
+  (-> (System/getenv "SOAK_SECRETS")
+      (json/read-str)))
+
+(defn get-weather-now-at-id [id]
+  (let [response (http/get current-endpoint
+                           {:query-params {"id" id "appid" (get soak-secrets "WEATHER_API_TOKEN")}
+                            :content-type :json})]
```
jarohen: `:content-type` is for the type of the content you're sending up (or not, in this case) - you want to send `:accept :json`
if you also specify `:as :json` it'll coerce the response for you - it may even keywordize the keys (details will be in the docs, no doubt)

jarohen: may need to include 'cheshire' as a dependency for it to do that automatically, if memory serves

examples/soak/src/crux/soak/ingest.clj
```diff
@@ -0,0 +1,88 @@
...
+    (:body response)))
+
+(defn parse-weather-map [weather-map]
+  (-> (select-keys weather-map ["id" "name" "main" "wind" "weather" "dt_txt"])
+      (walk/keywordize-keys)))
+
+(defn weather-put-from-id [id]
+  (let [doc (-> (get-weather-now-at-id id)
```
jarohen: I wouldn't worry about let-binding `doc` and `end-valid-time` - you're only using them once, and the names don't really add any more clarity (it'd be clear enough as is)

examples/soak/src/crux/soak/ingest.clj
```diff
@@ -0,0 +1,88 @@
...
+                     (dissoc :id))
+             date-string (clojure.string/replace (:dt_txt doc) #" " "T")
+             start-valid-time (inst/read-instant-date date-string)
+             end-valid-time (-> (.toInstant start-valid-time)
+                                (.plus (Duration/ofHours 3))
+                                (Date/from))]
+         [:crux.tx/put (dissoc doc :dt_txt) start-valid-time end-valid-time]))
+     forecast-map)))
+
+(def kafka-properties-map {"ssl.endpoint.identification.algorithm" "https"
```
jarohen: formatting nit: when you've got a multi-line value in `def`, bring it down a line

(not strictly the same as [this](https://github.com/bbatsov/clojure-style-guide#oneline-short-fn) in the Clojure style guide, but it's the same principle)

examples/soak/src/crux/soak/ingest.clj
```diff
@@ -0,0 +1,88 @@
...
+(defn weather-put-forecast-from-id [id]
+  (let [forecast-edn (-> (get-weather-forecast-at-id id)
+                         (json/read-str)
+                         (get "list"))
+        forecast-map (map parse-weather-map forecast-edn)]
+    (mapv
+     (fn [edn]
+       (let [doc (-> (assoc edn :crux.db/id (keyword (str id "-forecast")))
+                     (dissoc :id))
+             date-string (clojure.string/replace (:dt_txt doc) #" " "T")
```
jarohen: looks like [one of their samples](https://samples.openweathermap.org/data/2.5/weather?q=London,uk&appid=b6907d289e10d714a6e88b30761fae22) has a `dt` key, which appears to be a timestamp - if this is present in the requests you're using too, you can pass it straight to the `java.util.Date` constructor

examples/soak/src/crux/soak/ingest.clj
```diff
@@ -0,0 +1,88 @@
...
+                           (.plus (Duration/ofHours 1))
+                           (Date/from))]
+    [:crux.tx/put doc start-valid-time end-valid-time]))
+
+(defn weather-put-forecast-from-id [id]
+  (let [forecast-edn (-> (get-weather-forecast-at-id id)
+                         (json/read-str)
+                         (get "list"))
+        forecast-map (map parse-weather-map forecast-edn)]
+    (mapv
```
jarohen: I'd be tempted to include both `(map parse-weather-map)` and your `mapv` below in a `->>` - at the moment, when you're tracing through the data flow in this function, your eye has to jump around from the forecast-map let-binding to where it's used, then to the `mapv`, whereas it's really just piping the forecast through a few transformations

jarohen: separately, I'd use `map` over `mapv` here when you don't absolutely need it to be eagerly evaluated

examples/soak/src/crux/soak/ingest.clj
```diff
@@ -0,0 +1,88 @@
...
+            [crux.api :as crux]
+            [clojure.string :as string]
+            [clojure.instant :as inst]
+            [clojure.java.io :as io])
+  (:import java.util.Date
+           java.time.Duration))
+
+(def current-endpoint "http://api.openweathermap.org/data/2.5/weather")
+(def forecast-endpoint "http://api.openweathermap.org/data/2.5/forecast")
+(def location-ids (-> (io/resource "locationids.txt")
```
jarohen: I'd advise storing these as a map from a human-readable name to the location-id - it's easy to `(-> ... slurp read-string vals)` on a map to get the location ids, but a PITA to remember which location-id was which when someone asks you to change one!

(as you can probably tell, I've done exactly this in the past and then been burned 3 months later by ... someone asking me to change one)

examples/soak/src/crux/soak/ingest.clj
```diff
@@ -0,0 +1,88 @@
...
+
+(defn get-weather-now-at-id [id]
+  (let [response (http/get current-endpoint
+                           {:query-params {"id" id "appid" (get soak-secrets "WEATHER_API_TOKEN")}
+                            :content-type :json})]
+    (:body response)))
+
+(defn get-weather-forecast-at-id [id]
+  (let [response (http/get forecast-endpoint
+                           {:query-params {"id" id "appid" (get soak-secrets "WEATHER_API_TOKEN")}
```
jarohen: nit: commas in single-line maps please :)

examples/soak/src/crux/soak/ingest.clj
```diff
@@ -0,0 +1,88 @@
...
+
+(def kafka-properties-map {"ssl.endpoint.identification.algorithm" "https"
+                           "sasl.mechanism" "PLAIN"
+                           "sasl.jaas.config" (format
+                                               "org.apache.kafka.common.security.plain.PlainLoginModule required username=\"%s\" password=\"%s\";"
+                                               (get soak-secrets "CONFLUENT_API_TOKEN")
+                                               (get soak-secrets "CONFLUENT_API_SECRET"))
+                           "security.protocol" "SASL_SSL"})
+
+(defn -main [& args]
```
jarohen, line 86: need a `shutdown-agents` in here, like bench?

examples/soak/src/crux/soak/ingest.clj
```diff
@@ -0,0 +1,88 @@
...
+  (let [response (http/get forecast-endpoint
+                           {:query-params {"id" id "appid" (get soak-secrets "WEATHER_API_TOKEN")}
+                            :content-type :json})]
+    (:body response)))
+
+(defn parse-weather-map [weather-map]
+  (-> (select-keys weather-map ["id" "name" "main" "wind" "weather" "dt_txt"])
+      (walk/keywordize-keys)))
+
+(defn weather-put-from-id [id]
```
jarohen: your functions here are both fetching data and transforming data - best to split these two out.

a general pattern I find works well in most cases is to split it down (broadly speaking) into:

```clojure
(defn pull-data [...]
  ...)

(defn transform-data [...]
  ...)

(defn push-data [...]
  ...)

(defn run-the-thing []
  (push-data (-> (pull-data ...)
                 (transform-data))))
```
then `transform-data` can be easily checked at the REPL/in tests. (you've already split the push-data part out :) )

(Zach Tellman talks about this pattern in his Elements of Clojure book - it's a good read, highly recommend - I'd be surprised if there wasn't a copy in the office somewhere)

examples/soak/cloudformation/ecs-ingest.yml
```diff
@@ -0,0 +1,198 @@
...
+              awslogs-region: 'eu-west-2'
+              awslogs-group: !Ref LogGroup
+              awslogs-stream-prefix: ecs
+
+  SoakIngestForecastTaskDef:
+    Type: AWS::ECS::TaskDefinition
+    DependsOn:
+      - LogGroup
+    Properties:
+      Cpu: '2 vCPU'
```
jarohen: doubt we need 2 vCPUs here - anything happening in parallel?
likewise, 8GB is high when we're not doing any memory-intensive processing - 2GB is probably sufficient?

jarohen (changes requested):
thanks Dan - overall well done and looks like it'll do the job. as always, some style suggestions below :)

PR #698, opened 2020-03-02
Fix #697, sorting args before populating virtual index

### LINKED ISSUE #697: Unsorted args in query doesn't yield enough results
If args are supplied in an unsorted order, the query returns incorrect results.

transaction:
```clojure
[[:crux.tx/put {:crux.db/id :foo-some-bar-nil
                :bar nil
                :foo true}]
 [:crux.tx/put {:crux.db/id :foo-nil-bar-some
                :bar true
                :foo nil}]
 [:crux.tx/put {:crux.db/id :foo-some-bar-some
                :foo true
                :bar true}]]
```

query:
```clojure
{:find [e]
 :where [[e :foo f]
         [e :bar g]]

 :args [{f true, g true}
        {f true, g nil}
        {f nil, g true}]}
```

in this query, the chosen join order is `[g f]`, and the `true` values of `g` are not grouped together. 
when creating the sorted index, this then calls `partition-all`, which creates 3 partitions - a later sort (`new-sorted-virtual-index`) brings them together in the tree, but doesn't combine them (because the `partition-all` happens before the sort), so joining on this index doesn't yield all the results.

jarohen:
When we create a virtual index from args, the query engine assumed that values were grouped together in the prefix tree - this works fine for indices pulled from the kv-store, but not so much for values from user-provided args.

This PR sorts user-provided args before populating the virtual indices, so that the `partition-by` in `build-nested-index` groups together values before generating the sorted arg indices

### DIFF
--- a/crux-core/src/crux/index.clj
+++ b/crux-core/src/crux/index.clj
@@ -970,7 +970,7 @@
 
 (defn- build-nested-index [tuples [range-constraints & next-range-constraints]]
   (-> (new-sorted-virtual-index
-       (for [prefix (partition-by first tuples)
+       (for [prefix (vals (group-by first tuples))
              :let [value (ffirst prefix)]]
          [(c/->value-buffer value)
           (RelationNestedIndexState.
--- a/crux-test/test/crux/query_test.clj
+++ b/crux-test/test/crux/query_test.clj
@@ -2802,3 +2802,18 @@
       (with-open [snapshot (api/new-snapshot db)]
         (t/is (= [[20] [25] [30]]
                  (sort (api/q db snapshot '{:find [a] :where [[_ :age a]]}))))))))
+
+(t/deftest test-unsorted-args-697
+  (f/transact! *api* [{:crux.db/id :foo-some-bar-nil, :bar nil, :foo true}
+                      {:crux.db/id :foo-nil-bar-some, :bar true, :foo nil}
+                      {:crux.db/id :foo-some-bar-some, :foo true, :bar true}])
+
+  (t/is (= #{[:foo-some-bar-nil] [:foo-nil-bar-some] [:foo-some-bar-some]}
+           (api/q (api/db *api*)
+                  '{:find [e]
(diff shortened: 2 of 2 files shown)


dotemacs:
Hello @jarohen 

The way the test is written now: 
https://github.com/juxt/crux/pull/698/files#diff-8d0564fb1283009c592ba069490c0f15R2811

This passes with the current release`20.02-1.7.0-alpha`.

The issue is more apparent when you put the argument with no `nil` values as the last argument, then the issue shows up.

See this modification, in particular the last `t/is` sexpression/test clause.

```clojure
(t/deftest test-unsorted-args-697
  (f/transact! *api* [{:crux.db/id :foo-some-bar-nil, :bar nil, :foo true}
                      {:crux.db/id :foo-nil-bar-some, :bar true, :foo nil}
                      {:crux.db/id :foo-some-bar-some, :foo true, :bar true}])

  (let [query1 (api/q (api/db *api*)
                      '{:find [e]
                        :where [[e :foo f]
                                [e :bar g]]

                        :args [{f true, g true}
                               {f true, g nil}
                               {f nil, g true}]})
        query2 (api/q (api/db *api*)
                      '{:find [e]
                        :where [[e :foo f]
                                [e :bar g]]

                        :args [{f true, g nil}
                               {f nil, g true}
                               {f true, g true}]})]

    (t/is (= query1 #{[:foo-some-bar-nil] [:foo-nil-bar-some] [:foo-some-bar-some]}))
    (t/is (= query2 #{[:foo-some-bar-nil] [:foo-nil-bar-some] [:foo-some-bar-some]}))
    (t/is (= query1 query2))))
```

jarohen:
cheers @dotemacs - as discussed on Slack, from our CI I'm confident that we've isolated the issue into the failing test on this PR, and that the proposed fix then fixes that test :smile:

jarohen:
discussed with @hraberg, he's happy with the fix, will merge :smile:

PR #704, opened 2020-03-03
Run watdiv against datomic

### LINKED ISSUE #575: Watdiv bench should be runnable against other databases

danmason:
(resolves #575)

Runs watdiv against in memory database, though currently with a `system/exit` on the end to ensure it closes properly,

### DIFF
--- a/crux-bench/src/crux/bench/watdiv_datomic.clj
+++ b/crux-bench/src/crux/bench/watdiv_datomic.clj
@@ -300,9 +300,7 @@
          (vec))))
 
 (defn with-datomic [f]
-  (let [uri (str "datomic:free://"
-                 (or (System/getenv "DATOMIC_TRANSACTOR_URI") "datomic")
-                 ":4334/bench?password=password")]
+  (let [uri (str "datomic:mem://bench")]
     (try
         (d/delete-database uri)
         (d/create-database uri)
@@ -354,4 +352,5 @@
                         (->> (run-watdiv-bench {:test-count 100})
                              (filter :query-idx)
                              (sort-by :query-idx)))
-    (bench/save-to-s3 {:database "datomic" :version "0.9.5697"} output-file)))
+    (bench/save-to-s3 {:database "datomic" :version "0.9.5697"} output-file))
+  (shutdown-agents))
--- a/crux-bench/src/crux/bench/watdiv_datomic.clj
+++ b/crux-bench/src/crux/bench/watdiv_datomic.clj
@@ -304,11 +304,15 @@
     (try
         (d/delete-database uri)
         (d/create-database uri)
-        (with-open [conn ^Closeable (d/connect uri)]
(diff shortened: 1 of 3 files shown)


crux-bench/src/crux/bench/watdiv_datomic.clj
```diff
@@ -300,17 +300,17 @@
...
+  (let [uri (str "datomic:mem://bench")]
     (try
         (d/delete-database uri)
         (d/create-database uri)
-        (with-open [conn ^Closeable (d/connect uri)]
+        (let [conn (d/connect uri)]
           @(d/transact conn datomic-watdiv-schema)
-          (f conn))
+          (f conn)
+          (d/release conn))
```
jarohen: should `release` be in the `finally` block?

crux-bench/src/crux/bench/watdiv_datomic.clj
```diff
@@ -354,4 +354,7 @@
                         (->> (run-watdiv-bench {:test-count 100})
                              (filter :query-idx)
                              (sort-by :query-idx)))
-    (bench/save-to-s3 {:database "datomic" :version "0.9.5697"} output-file)))
+    (bench/save-to-s3 {:database "datomic" :version "0.9.5697"} output-file))
+  (shutdown-agents)
+  ;; TODO: Queries with timeouts don't close properly (currently), so this system exit is necessary. Post a fix, it should be removed.
```
jarohen, line 361: is this still the case with the `release` -> `finally` change above?

jarohen (changes requested):
one small change here and this is good to go, cheers - could you also ensure its rebased on master?

crux-bench/src/crux/bench/watdiv_datomic.clj
danmason: Potentially, though would need to move the `let`

crux-bench/src/crux/bench/watdiv_datomic.clj
danmason: Yes, I imagine that it would be - The issue seems to be primarily with queries, if I were to remove the querying the program exits just fine.

PR #708, opened 2020-03-05
Sorted maps microbench

jarohen:
Three fairly independent commits here:
* Add sorted-maps microbench - hopefully as a template for further microbenches. FWIW I've been testing this with an easy REPL call to `(run-sorted-maps-microbench (user/crux-node))`, once you're using the integrant-based dev workflow
* (regression) we weren't including bench-ns in the CloudWatch insights messages - these were being manually added back for the slack messages (which are sent later) - fixed by making bench-ns an explicit dimension and using this dimension in the slack messages
* refactor `with-nodes` s.t. callers can choose which nodes to call - microbenches may well only want to run under one of the node configurations


crux-bench/src/crux/bench.clj
```diff
@@ -109,59 +107,61 @@
...
-        (for [[k v] (nodes data-dir)]
-          (with-open [node (api/start-node v)]
-            (with-dimensions {:crux-node-type k}
-              (log/infof "Running bench on %s node." k)
-              (post-to-slack (str "running on node: " k))
-              [k (f node)])))))))
-
-(defmacro with-nodes [[node-binding] & body]
-  `(with-nodes* (fn [~node-binding] ~@body)))
+       (for [[k ->node] nodes]
```
danmason: A nitpick - and not necessarily one with your code: there has to be a better/clearer varname for the node's name than just 'k' (I realize it is the 'key' but I don't think that's particularly descriptive or helpful)

crux-bench/src/crux/bench/main.clj
```diff
@@ -1,25 +1,34 @@
...
+    (-> (bench/results->slack-message) (bench/post-to-slack))))
+
 (defn -main []
   (bench/post-to-slack (format "*Starting Benchmark*, Crux Version: %s, Commit Hash: %s\n"
                                bench/crux-version bench/commit-hash))
 
-  (let [bench-results (bench/with-nodes [node]
-                        [(doto (devices/run-devices-bench node)
-                           (bench/post-slack-results :ts-devices))
+  (let [bench-results (concat (bench/with-nodes [node (select-keys bench/nodes ["standalone-rocksdb"])]
```
danmason: Think this should be  `(select-keys bench/nodes ["standalone-rocksdb", "kafka-rocksdb"])`.

crux-bench/src/crux/bench/sorted_maps_microbench.clj
```diff
@@ -0,0 +1,28 @@
+(ns crux.bench.sorted-maps-microbench
```
danmason, line 1: Might be worth adding a `-main` to this which selects a default node to run it on (So you can quickly run it using `lein run -m crux.bench.sorted_maps_microbench`)?

danmason (changes requested):
Looking good - couple of minor changes I have suggest, but in particular the main needs to also run on the embedded-kafka node.

crux-bench/src/crux/bench.clj
jarohen: `node-type`?

crux-bench/src/crux/bench/main.clj
danmason: By default, makes sense (to me) to run it on all of the nodes (as we do on our nightly bench).

jarohen: yep - didn't need it initially but will do shortly, will add

crux-bench/src/crux/bench/sorted_maps_microbench.clj
jarohen: I'm trying to get away from requiring `lein run` for these - the feedback loop time's much quicker from the REPL - `(run-sorted-map-microbench (user/crux-node))` gets you the results much faster

crux-bench/src/crux/bench.clj
danmason: Yeah, I think that's far better.

crux-bench/src/crux/bench/sorted_maps_microbench.clj
danmason: Seems fair. On top of that, I would suggest adding a note about it (and the general method to run it) to the `crux-bench` README.

danmason (approved):

PR #710, opened 2020-03-05
Start up kafka containers alongside bench

### LINKED ISSUE #663: Crux bench should run with full Kafka instance

danmason:
(Resolves #663) 

Starts up a zookeeper and kafka container on the bench tasks, connecting to them within the `kafka-rocksdb` node. (Increases the memory/cpu usage of the benchmark, each of the new containers using 1vCPU and 4GB RAM).

### DIFF
--- a/crux-bench/cloudformation.yaml
+++ b/crux-bench/cloudformation.yaml
@@ -133,8 +133,8 @@ Resources:
       - ECRRepository
       - LogGroup
     Properties:
-      Cpu: '2 vCPU'
-      Memory: '8GB'
+      Cpu: '4 vCPU'
+      Memory: '12GB'
       Family: 'crux-bench'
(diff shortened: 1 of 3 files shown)


crux-bench/cloudformation.yaml
```diff
@@ -144,7 +144,40 @@ Resources:
         - 'FARGATE'
       NetworkMode: 'awsvpc'
       ContainerDefinitions:
+        - Name: 'zookeeper-container'
+          Cpu: 1024
```
jarohen, line 148: this seems high - can you confirm what zookeeper/kafka are actually using during a bench run?

jarohen:
thanks Dan, looking good - let's just confirm the right resources to use and this is good to merge

PR #711, opened 2020-03-06
Bundle of ingest performance

jarohen:
details in individual commits, but main change is that taking a call to `satisfies?` call off a common `value->buffer` path cuts time for index-docs by 40%


jarohen:
going to resubmit just the value->buffer change

PR #716, opened 2020-03-06
Not creating unnecessary sorted maps, #685

### LINKED ISSUE #685: Sorting maps adding a 10-20% performance overhead on ingest
See comments from @kosiakk on https://github.com/juxt/crux/commit/f768fd4573881b89a21c10b78b46520ba08a536a.

Background: sorting maps was introduced so that maps consistently hash to the same value regardless of the key ordering within the map implementation. It needs to be consistent for anything that uses indices rather than documents - e.g. using maps as keys/values, but also CaS (which compares document hashes in the bitemp indices)

jarohen:
(resolves #685)

Not creating extra sorted maps when hashing - we just serialise them in order. 

Does involve changing Nippy behaviour, but we set a flag s.t. we only alter the behaviour when it's us that's serialising the maps.

I've also corrected the behaviour for sets here - these will have been vulnerable to the same consistent hashing issue from way back.

### DIFF
--- a/crux-bench/src/crux/bench/sorted_maps_microbench.clj
+++ b/crux-bench/src/crux/bench/sorted_maps_microbench.clj
@@ -4,12 +4,12 @@
             [crux.bench :as bench]))
 
 (defn submit-batches [node]
-  (for [doc-batch (->> (for [n (range 15000)]
+  (for [doc-batch (->> (for [n (range 25000)]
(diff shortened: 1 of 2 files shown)


jonpither (approved):

jonpither:
Notice that you're leaving sort-maps around.

PR #723, opened 2020-03-09
Add HTML formatting for the nightly benchmark email.

### LINKED ISSUE #713: Bench email should be formatted like the Slack messages
Tough to read it when it's just a pr-str! :wink: 

doesn't have to be exactly the same, but splitting out by ns/type, highlighting time taken, and monospacing the rest would be useful

danmason:
(Resolves #713)

### DIFF
--- a/crux-bench/src/crux/bench.clj
+++ b/crux-bench/src/crux/bench.clj
@@ -90,6 +90,28 @@
                (map result->slack-message)
                (string/join "\n"))))
 
+(defn- result->html [{:keys [time-taken-ms bench-type] :as bench-map}]
+  (format "<p> <b>%s</b> (%s): <code>%s</code></p>"
+          (name bench-type)
+          (java.time.Duration/ofMillis time-taken-ms)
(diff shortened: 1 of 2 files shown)


crux-bench/src/crux/bench.clj
```diff
@@ -117,8 +139,8 @@
       :crux.standalone/event-log-dir (str (io/file data-dir "eventlog/rocksdb"))
       :crux.standalone/event-log-kv-store 'crux.kv.rocksdb/kv})
 
-   "kafka-rocksdb"
-   (fn [data-dir]
+   #_"kafka-rocksdb"
```
jonpither: why commented out?

crux-bench/src/crux/bench.clj
```diff
@@ -117,8 +139,8 @@
...
+   #_(fn [data-dir]
```
jonpither: Why commented again?

jonpither (changes requested):

crux-bench/src/crux/bench.clj
danmason: Whoops - was running it locally and forgot to uncomment, thanks jon.

Issue #724, opened 2020-03-10
lazy-id-hash showing up in query profiling


jonpither:
`lazy-id-hash` should no longer be called, as it is swapped out for the real hash algorithm on the first call made to `crux.hash/id-hash`.

Profiling shows significant CPU time is spent inside of `lazy-id-hash`. This is either a problem with the locking code, whereby the swapping out doesn't take real effect, or the latency in the swapping out upon initialization of the namespace by multiple threads, is overshadowing/confusing the intended/main code-path in the profiler.

jonpither:
@hraberg and I were able to prove this issue happens when direct linking is configured. There is a PR open that shows how this specific bug/issue could be fixed, but we will consider first, what would be involved in officially supporting direct linking. We would want a proper analysis and measurement of the performance gains. I'll close this issue for now.

PR #725, opened 2020-03-10
#724 Init the hashing-fn directly when hash ns loads

jonpither:


jonpither:
We're not officially supporting direct linking, so will close this PR. Should we want to support it, it is worth reconsidering. See #724 for more information.

PR #727, opened 2020-03-12
Auto-Compaction tool

jonpither:
This new module will compact documents older than certain valid-time 'water-mark'.

Documents are pruned when new documents arrive via `put` and `cas`. The compaction operation happens as part of the main tx ingestion process.

This is a basis on which to eventually build configurable policies, such as:

1. reduce granularity of valid time histories older than 12 months to 1-day
2. compact transaction time histories older than 3 months (useful for a correction-heavy scenario)


jonpither:
We should tie it to an index version that the compacter can understand.

jarohen:
closing while we investigate other methods of bringing the index sizes down

jonpither:
For those visiting this PR as something to potentially make use of, rather than wire up the compactor via a decorator, we would now use a persistent bus listener, see how we do it with Lucene: https://github.com/juxt/crux/blob/master/crux-lucene/src/crux/lucene.clj

PR #730, opened 2020-03-17
Make console into a Crux module #718

### LINKED ISSUE #718: Console to be started as a Crux module
Ideally we'd like users to be able to start the Crux console the same way as the HTTP server is now started - as a module that depends on the Crux node.

* Should no longer rely on the HTTP server having an open CORS policy - means we either need to have the console augment the HTTP server routes (maybe using a similar pattern to listeners calling through to the event bus, adding themselves to the bus's state?), or it should have its own HTTP service with only the endpoints the console needs. (currently prefer the latter)
* Documentation to be updated with instructions for how users can start up a console on their own Crux nodes

danmason:
(Resolves #718 )

A few things of note:
- This adds back in crux-console from the current version at crux-labs, without it's commit history.
- I have added some of the code / the handler from http-server into a file under console, augmenting the routes in the handler with "/crux/". The `crux.console` namespace makes use of a handler combining this handler and the ui handler from `crux-ui-server.main` to start both the node and the console on the same port. 

Aside from that, was relatively straightforward to make this change. Before lein installing crux-console, one must first do a lein build (and have yarn and node installed locally), so I added this to the master build/deploy steps on circleci.


jarohen:
as discussed, more console work incoming, will close this one for now

PR #731, opened 2020-03-17
Make console artifacts

### LINKED ISSUE #719: Console Docker/Uberjar configuration
Let's add another Docker/Uberjar configuration that includes the Console (depends on #718, I suspect). 

* This should be an additional configuration (i.e. should be 'standalone' and 'standalone-console')
* Given this is probably only for first-5-minute users, let's only add this with standalone config (i.e. don't worry about creating all the permutations with Rocks/Kafka/etc)

danmason:
Resolves #719


jarohen:
as discussed, more console work incoming, will close this one for now

Issue #693, opened 2020-02-27
How should we back up soak nodes?


jarohen:

jonpither:
See #407 (older issues #6 #159 #200 )

jarohen:
a: use #407

PR #733, opened 2020-03-23
Add webpage for crux-soak

### LINKED ISSUE #670: Soak renders frequently updating data on its webpage
- [x] Set up a scheduled Fargate task to run ingest
- [x] Render data on soak webpage

Idea for UX:
* dropdown of all the locations we know about. user selects a location
* per location: dropdown for dates we know about for that location, descending order, defaults to today (will have dates in the future, too) - shows the forecast history for that location/date, and the actual weather, if we have it

danmason:
NOTE: Somewhat relies on #728, otherwise unable to poll the tx-log from confluent cloud

Adds the webpage for crux soak and resolves #670 

Still some work on soak to do, particularly #693 , and also potentially implementing some of the changes to be brought in soon (this could be a powerful use case for the compaction module, as well as bringing in the console as a module to poke at the underlying data).

Some screenshots of the webpage as it currently stands are attached below:

**Homepage**

**Weather at current date/time at a location**

**Weather forecasts for a location in a few days time**


danmason:
To note: May also want to add what time the currently displayed forecasts are for, and what time the 'current' weather is from.

PR #734, opened 2020-03-23
Compaction using daily intervals

jonpither:


jarohen:
closing while we investigate other methods of bringing the index sizes down

Issue #514, opened 2019-12-21
Users can query Crux using SQL


jarohen:
See PR #735 for more discussion.

(from @malcolmsparks on Slack)
SQL has now become a protocol to allow the integration of a vast ecosystem of tools for:

- Reporting, management information (MIS)
- Business Intelligence (BI)
- Dashboards
- Data extract-transform-load (ETL)
- Systems integration

There are some SQL standards defined by ANSI, such as SQL-92, SQL:2003, SQL:2011. Historically, the conformance of SQL implementations vary so much that the common denominator is quite small. Therefore, the most rudimentary implementation of SQL can yield interoperability with a significant fraction of the industry toolset.

We'd likely want to allow users to:
* transact a fixed relation (a 'table', in SQL)
* query these fixed relations from Crux using a SQL (or subset thereof) query

jarohen:
Cassandra/Calcite docs (thanks @refset)
* https://calcite.apache.org/docs/cassandra_adapter.html
* https://github.com/apache/calcite/tree/master/cassandra/src/main/java/org/apache/calcite/adapter/cassandra

jonpither:
See also for an initial Crux/Calcite spike: https://github.com/juxt/edge/tree/calcite

jarohen:
Things to think about:
* What do we expose over the network?
* Schema generation? Can it be dynamic?
* Taking control of joins
* Bitemp SQL - 'as-of?'
* Prior art (Cassandra?)
* Evaluate Presto vs Calcite
* DML/DDL

jonpither:
From @SevereOverfl0w:

https://prestodb.io/docs/current/connector/mongodb.html#table-definition-label seems worth evaluating.

In particular: "MongoDB [connector plugin] maintains table definitions on the special collection where mongodb.schema-collection configuration value specifies."

That is, they store table definitions in MongoDB.

jonpither:
Fixed by #735

PR #735, opened 2020-03-23
SQL Module #514

jonpither:
See #514 for more background. This PR uses Calcite to provide a SQL layer on top of Crux.

Remaining TODOs:
- [x] Ensure using args for literals doesn't preclude range searches
- [x] SQL extension permissive for newlines etc
- [x] Some Microbenches
- [x] Query Args (timeouts)
- [x] Use new open-q
- [x] PreparedStatements with calcs
- [x] UUIDs

Not in scope for this PR:

- LEFT OUTER JOIN handled by Crux instead of Calcite (see #835 )
- Make use RexToLixTranslator, would need more rules, i.e. FilterToProjectRule
- Potential use of tx-id for consistency (current Crux API doesn't - @refset) 
- Build up queries based on projected fields
- Aggregations (TPCH 001)
- All data-types Crux supports (i.e. URLs)
- Caching Table schemas (looked up ~twice for each query)
- Consideration for Map IDs?

Test suits to investigate:
  - https://www.itl.nist.gov/div897/ctg/sql-testing/sqlman60.htm
  - https://github.com/dremio/dremio-oss/blob/f1067683783c99255ecc47fabc1892bfe7edfe14/sabot/kernel/src/test/java/com/dremio/TestDecimalQueries.java#L257

RexToLixTranslator usage:
  - https://github.com/apache/beam/blob/master/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamCalcRel.java#L159

UUID Examples
  - https://phauer.com/2016/uuids-hibernate-mysql/


jonpither:
I have looked into providing bitemp support in SQL. The best outcome would be to follow the [SQL 2011](https://en.wikipedia.org/wiki/SQL:2011) spec, so you could do `SELECT * FROM PERSONS FOR SYSTEM_TIME AS OF '2020...`. 

Calcite does support the concept of a `TemporalTable`, but this would require a tighter integration with Calcite that what we have now - it doesn't work with our current approach of leveraging `ProjectableFilterableTable`. It's unclear also how `TemporalTable` could help us offer bitemporal support (application time / business time / valid time) - presumably we would need to extend the Calcite parser.

Pragmatic options have been considered 

1) Adding tx-time/valid-time as connection-string args, 
2) expose columns for querying against, 
3) Add a custom function to Calcite.

I'm not a fan of 1, as users typically pull SQL connection strings out of config, before setting up connection pools when a system starts. Encouraging users to manipulate the SQL connection strings feels error-prone and unsatisfactory.

2 is doable, but it would confuse Calcite. One example. is that if we declare new columns, then we would also get them returned when doing a `select *`.

3 I spiked to look like:

```
SELECT * FROM PERSON WHERE VALID_TIME() = '2012-01-01 
```

This can be made to work - our code can strip out of the VALID_TIME filter after we have gleaned the information we need, so that the main query doesn't get impacted by it. 

Both 2 & 3 have the problem of compromising the relational algebra. For example with 3, what happens here:

```
SELECT * FROM person WHERE VALID_TIME() < '...'
```

It can quickly become confusing. 

I currently prefer 3 over 2 because 3 is custom function that our users would use knowing the edge-case practicalities, by way of a documented disclaimer.

Offering bitemporal support isn't a showstopper for giving our users query capability - so we could target at a later date.

jonpither:
Responding to a comment by @hraberg to use `full-results?`. I'm opting against it - having tried - because `full-results?` gives me other attributes on docs that may not be columns. I can't avoid making syms for all columns anyway (the way the current code works), so `full-results?` doesn't buy me much to begin with. Should the code get refactored (as is likely at some future point), we should consider `full-results?`.

crux-calcite/README.adoc
```diff
@@ -0,0 +1,118 @@
...
+----
+
+== Table Definitions
+
+To define a table, transact a document into Crux:
+
+``` clojure
+{:crux.db/id :crux.sql.schema/person
+ :crux.sql.table/name "person"
+ :crux.sql.table/query {:find ['id 'name 'homeworld]
```
jarohen: possible/better to quote the whole map form? (likewise the `columns`?)

crux-calcite/README.adoc
```diff
@@ -0,0 +1,118 @@
...
+----
+
+Get a connection and query as such:
+
+[source,clojure]
+----
+(require '[crux.calcite])
+
+(defn query [q]
+  (with-open [conn (crux.calcite/jdbc-connection)]
```
jarohen: presumably I have to pass a node here, otherwise there's no way Calcite knows which node I want to query?

crux-calcite/README.adoc
```diff
@@ -0,0 +1,118 @@
...
+SQL driver] to your project and use the connection string:
+
+[source,properties]
+----
+jdbc:avatica:remote:url=http://<server-hostname>:1501;serialization=protobuf
+----
+
+`server-hostname` is the hostname of the Crux node. The port by
+default is `1501` but you can swap it out by configuring the Crux
+`crux.calcite.port` configuration property.
```
jarohen: this a JVM property? any way we could configure this in the topology map with the other Crux config, or is this another victim of Calcite requiring its own JSON config all the way down?

crux-calcite/project.clj
```diff
@@ -0,0 +1,28 @@
+(defproject juxt/crux-calcite "derived-from-git"
```
jarohen: for discoverability, would it be worth naming this module `crux-sql`, with a `crux.sql.calcite` ns if we're considering multiple SQL implementations down the line? don't think I'd've known that Calcite was the module to include for SQL support before this project...

crux-calcite/project.clj
```diff
@@ -0,0 +1,28 @@
...
+  :description "SQL for Crux using Apache Calcite"
+  :url "https://github.com/juxt/crux"
+  :license {:name "The MIT License"
+            :url "http://opensource.org/licenses/MIT"}
+  :dependencies [[org.clojure/clojure "1.10.1"]
+                 [juxt/crux-core "derived-from-git"]
+                 [com.google.guava/guava "28.2-jre"]
```
jarohen: these for dependency resolution conflicts? if so, ok to mark them as such?

also, if these aren't dependency conflicts when crux-calcite is used alone, could they go in the root `project.clj`, to keep the explicit versions out of downstream apps?

crux-calcite/project.clj
```diff
@@ -0,0 +1,28 @@
...
+                 [com.google.code.findbugs/jsr305 "3.0.2"]
+                 [com.google.protobuf/protobuf-java "3.9.2"]
+                 [org.eclipse.jetty/jetty-http "9.4.22.v20191022"]
+                 [org.apache.calcite/calcite-core "1.21.0" :exclusions [org.apache.commons/commons-lang3
```
jarohen: rather than exclusions, could we include these at specific versions further down in the project.clj? with exclusions, we're explicitly trusting the other library that pulls in Jackson (whichever one that is?) to be the most up to date

crux-calcite/src/crux/calcite.clj
```diff
@@ -0,0 +1,180 @@
...
+           java.util.Properties
+           org.apache.calcite.avatica.jdbc.JdbcMeta
+           [org.apache.calcite.avatica.remote Driver LocalService]
+           [org.apache.calcite.avatica.server HttpServer HttpServer$Builder]
+           [org.apache.calcite.rel.type RelDataTypeFactory RelDataTypeFactory$Builder]
+           [org.apache.calcite.rex RexCall RexInputRef RexLiteral RexNode]
+           org.apache.calcite.sql.SqlKind
+           org.apache.calcite.sql.type.SqlTypeName))
+
+(defonce !crux-nodes (atom {}))
```
jarohen: worth this being a WeakHashMap so that we're not the one hanging on to the node reference, should the user forget to explicitly close a node?

crux-calcite/src/crux/calcite.clj
```diff
@@ -0,0 +1,180 @@
...
+  (operand->v [this schema]
+    (.getValue2 this)))
+
+(defn- ->operands [schema ^RexCall filter*]
+  (->> (.getOperands filter*)
+       (map #(operand->v % schema))))
+
+(defn- ->crux-where-clauses
+  [schema ^RexNode filter*]
+  (condp = (.getKind filter*)
```
jarohen: might these be easier to read (by being closer in structure to what you'd write by hand) with syntax-quoting, rather than explicit calls to `list`/`apply list`?
e.g. ``` `[(~'or ~@(mapcat ...))]```

crux-calcite/src/crux/calcite.clj
```diff
@@ -0,0 +1,180 @@
...
+    SqlKind/AND
+    (mapcat (partial ->crux-where-clauses schema) (.-operands ^RexCall filter*))
+    SqlKind/OR
+    [(apply list 'or (mapcat (partial ->crux-where-clauses schema) (.-operands ^RexCall filter*)))]
+    SqlKind/INPUT_REF
+    [[(list '= (operand->v filter* schema) true)]]
+    SqlKind/NOT
+    [(apply list 'not (mapcat (partial ->crux-where-clauses schema) (.-operands ^RexCall filter*)))]
+    SqlKind/GREATER_THAN
+    [[(list '> (->operands schema filter*))]]
```
jarohen: does this one need to be `apply list` too?

crux-calcite/src/crux/calcite.clj
```diff
@@ -0,0 +1,180 @@
...
+                                        :boolean SqlTypeName/BOOLEAN
+                                        :double SqlTypeName/DOUBLE
+                                        :datetime SqlTypeName/DATE})
+
+(defn- ^java.util.List perform-query [node q]
+  (->> (crux/q (crux/db node) q)
+       (mapv to-array)))
+
+(defn- ^String ->column-name [c]
+  (string/replace (string/upper-case (str c)) #"^\?" ""))
```
jarohen: placeholder comment for 'do we need to do this?' discussion :)

crux-calcite/src/crux/calcite.clj
```diff
@@ -0,0 +1,180 @@
...
+  (proxy
+      [org.apache.calcite.schema.impl.AbstractTable
+       org.apache.calcite.schema.ProjectableFilterableTable]
+      []
+      (getRowType [^RelDataTypeFactory type-factory]
+        (let [field-info  (RelDataTypeFactory$Builder. type-factory)]
+          (doseq [c (:find query)]
+            (let  [col-name (->column-name c)
+                   col-type ^SqlTypeName (column-types->sql-types (columns c))]
+              (assert col-type (str "Unrecognized column: " c))
```
jarohen: this is an error I'd want to throw in production (because it's checking user input), and we can't rely on assertions being turned on - could this be an explicit `(or ... (throw (IllegalArgumentException. ...))`?

crux-calcite/src/crux/calcite.clj
```diff
@@ -0,0 +1,180 @@
...
+         (map (comp (partial crux/entity db) first)))))
+
+(defn create-schema [parent-schema name operands]
+  (let [node (@!crux-nodes (get operands "CRUX_NODE"))]
+    (assert node)
+    (proxy [org.apache.calcite.schema.impl.AbstractSchema] []
+      (getTableMap []
+        (into {}
+              (for [table-schema (lookup-schema node)]
+                (do (s/valid? ::table table-schema)
```
jarohen: `s/valid?` isn't side-effecting, think this is a no-op?

crux-calcite/src/crux/calcite.clj
```diff
@@ -0,0 +1,180 @@
...
+                    [(string/upper-case (:crux.sql.table/name table-schema)) (make-table node table-schema)])))))))
+
+(def ^:private model {:version "1.0",
```
jarohen: I know it's 'how we do it elsewhere in Crux', but it always reads weird to me when multi-line defs are started on the same line - we'd never dream of doing this for a `defn`, and it hides arguably the most important part of the def (`model`) in the middle of the line, which makes the file harder to scan. one for the conventions chat :)

crux-test/test/crux/calcite_test.clj
```diff
@@ -0,0 +1,192 @@
...
+  (t/testing "arbitrary sql function"
+    (t/is (= #{{:name "Iva"}}
+             (set (query "SELECT SUBSTRING(NAME,1,3) AS NAME FROM PERSON WHERE NAME = 'Ivan'")))))
+
+  (t/testing "namespaced keywords"
+    (f/transact! *api* (f/people [{:crux.db/id :human/ivan :name "Ivan" :homeworld "Earth" :alive true}]))
+    (t/is (= [{:id ":human/ivan", :name "Ivan"}] (query "SELECT ID,NAME FROM PERSON WHERE ID = CRUXID('human/ivan')"))))
+
+  (t/testing "numeric values"
+    (fapi/delete-all-entities)
```
jarohen: the fixtures deliberately start up a fresh Crux node for every test - could we split this test out into two if it doesn't depend on the state from the first part? otherwise, you have to check every `t/testing` block to figure out the state of the database

crux-test/test/crux/fixtures/calcite.clj
```diff
@@ -0,0 +1,26 @@
+(ns crux.fixtures.calcite
```
jarohen: given this is a new fixture, worth putting it in `crux-test/src` instead, so that it can be used from tests in the individual modules?

crux-test/test/crux/fixtures/calcite.clj
```diff
@@ -0,0 +1,26 @@
...
+  (with-open [conn (DriverManager/getConnection "jdbc:avatica:remote:url=http://localhost:1501;serialization=protobuf")]
+    (binding [*conn* conn]
+      (f))))
+
+(defn with-calcite-module [f]
+  (fapi/with-opts (-> fapi/*opts*
+                      (update ::n/topology conj cal/module))
+    f))
+
+(defn query [q]
```
jarohen: I know it's only a test fixture, but could we either eagerly eval this resultset, or offer some way of the user closing the resources? (maybe a `with-query`, if we're expecting sufficiently large result sets?)

crux-test/test/crux/fixtures/tpch.clj
```diff
@@ -0,0 +1,57 @@
...
+
+(defn tpch-table->docs [^TpchTable t]
+  ;; first happens to be customers (;; 150000)
+  (map (partial tpch-entity->doc t) (seq (.createGenerator ^TpchTable t 0.005 1 1))))
+
+(comment
+  (first (tpch-tables->crux-sql-schemas))
+  {:crux.db/id :crux.sql.schema/customer,
+   :crux.sql.table/name "customer",
+   :crux.sql.table/columns [{:crux.sql.column/attribute :c_custkey, :crux.sql.column/name "c_custkey", :crux.sql.column/type :varchar}
```
jarohen: this one out of date?

crux-test/test/crux/tpch_test.clj
```diff
@@ -0,0 +1,462 @@
...
+;; http://www.tpc.org/
+
+(defn- with-tpch-schema [f]
+  (f/transact! *api* (tf/tpch-tables->crux-sql-schemas))
+  (f))
+
+(defn- with-tpch-dataset [f]
+  (doseq [^TpchTable t (TpchTable/getTables)]
+    ;; Using a scale-factor of .005 and no take, takes ~3 mins combined for all tests
+    (f/transact! *api* (take 10 (tf/tpch-table->docs t))))
```
jarohen: I wonder if this test belongs in crux-bench? or maybe we put the full test in crux-bench and leave a subset here

crux-calcite/README.adoc
jarohen: in fact, can't see this elsewhere in the PR?

jarohen (dismissed):
thanks Jon :) worth a demo to take us through it in more detail?

crux-test/test/crux/fixtures/calcite.clj
jonpither: I'd like to do this in a subsequent PR - which would also involve moving the calcite test into the calcite/sql module and making full use of test/src. I want to properly scratch this itch in one go.

crux-calcite/project.clj
jonpither: That's a good point WRT discoverability. If we wanted to support Presto later, I was thinking we'd have a crux-presto module and keep them separate. But your discoverability argument is winning me over.

crux-calcite/README.adoc
jonpither: Will fix

crux-test/test/crux/tpch_test.clj
jonpither: Currently all tpch tests runs with a very small amount of docs (5 per tpch type). We could up the type and perhaps move to bench, but I judge we're OK for this PR as is? I've found the tpch tests are quick enough.

crux-calcite/src/crux/calcite.clj
jonpither: On the README if we're supporting this functionality

crux-calcite/src/crux/calcite.clj
jonpither: throw an exception

crux-test/test/crux/tpch_test.clj
jonpither: We'll revisit taking advantage of TCPH in future.

jonpither:
We ought test lazyiness

crux-calcite/project.clj
jonpither: Will fix. There are dep conflicts between Calcite Vs Avatica, so there will be some noise here. I'll make it more explicit though.

crux-calcite/project.clj
jonpither: Changes made

crux-calcite/src/crux/calcite.clj
jonpither: Changed - no longer an atom though

crux-calcite/src/crux/calcite.clj
jonpither: Will skip this one

crux-calcite/src/crux/calcite.clj
jonpither: Done

crux-calcite/src/crux/calcite.clj
jonpither: Added to the readme

crux-calcite/project.clj
jonpither: Done

jonpither:
Verified it's lazy

crux-calcite/project.clj
jonpither: Done.

crux-sql/src/crux/calcite.clj
```diff
@@ -0,0 +1,183 @@
...
+
+(s/def :crux.sql.table/name string?)
+(s/def :crux.sql.table/columns (s/map-of symbol? column-types->sql-types))
+(s/def ::table (s/keys :req [:crux.db/id :crux.sql.table/name :crux.sql.table/columns]))
+
+(defn- lookup-schema [node]
+  (let [db (crux/db node)]
+    (->> (crux/q db '{:find [e]
+                      :where [[e :crux.sql.table/name]]})
+         (map (comp (partial crux/entity db) first)))))
```
hraberg: This entity call can be replaced with full-results.

crux-sql/src/crux/calcite.clj
```diff
@@ -0,0 +1,183 @@
...
+
+(defn- ->operands [schema ^RexCall filter*]
+  (->> (.getOperands filter*)
+       (map #(operand->v % schema))))
+
+(defn- ->crux-where-clauses
+  [schema ^RexNode filter*]
+  (condp = (.getKind filter*)
+    SqlKind/EQUALS
+    [[(apply list '= (->operands schema filter*))]]
```
hraberg: Are there ever literals generated? In most cases this will be faster than pulling out two values and comparing them using Clojure. (that is, do we ever generate `[e :name "Ivan"]`) ?

crux-sql/src/crux/calcite.clj
```diff
@@ -0,0 +1,183 @@
...
+    SqlKind/GREATER_THAN_OR_EQUAL
+    [[(apply list '>= (->operands schema filter*))]]
+    SqlKind/LESS_THAN
+    [[(apply list '< (->operands schema filter*))]]
+    SqlKind/LESS_THAN_OR_EQUAL
+    [[(apply list '<= (->operands schema filter*))]]
+    SqlKind/LIKE
+    [[(apply list 'like (->operands schema filter*))]]
+    SqlKind/IS_NULL
+    [[(apply list '= (->operands schema filter*))]]
```
hraberg: Does this imply that one of the operands is `nil`?

crux-sql/src/crux/calcite.clj
```diff
@@ -0,0 +1,183 @@
...
+    SqlKind/IS_NOT_NULL
+    [[(list 'not (apply list '= (->operands schema filter*)))]]))
```
hraberg: Same as above.

crux-sql/src/crux/calcite.clj
```diff
@@ -0,0 +1,183 @@
...
+  (->> (.getOperands filter*)
+       (map #(operand->v % schema))))
+
+(defn- ->crux-where-clauses
+  [schema ^RexNode filter*]
+  (condp = (.getKind filter*)
+    SqlKind/EQUALS
+    [[(apply list '= (->operands schema filter*))]]
+    SqlKind/NOT_EQUALS
+    [(list 'not [(apply list '= (->operands schema filter*))])]
```
hraberg: Depending on what one is doing, `not=` is likely faster than `(not (= ...))` as the latter will generate a sub query containing a predicate, and negate it, while `not=` will just be a predicate call.

crux-sql/src/crux/calcite.clj
hraberg: And same issue regarding using `not` (which will generate a sub-query) here applies.

crux-sql/src/crux/calcite.clj
```diff
@@ -0,0 +1,183 @@
...
+    [(apply list 'not (mapcat (partial ->crux-where-clauses schema) (.-operands ^RexCall filter*)))]
+    SqlKind/GREATER_THAN
+    [[(apply list '> (->operands schema filter*))]]
+    SqlKind/GREATER_THAN_OR_EQUAL
+    [[(apply list '>= (->operands schema filter*))]]
+    SqlKind/LESS_THAN
+    [[(apply list '< (->operands schema filter*))]]
+    SqlKind/LESS_THAN_OR_EQUAL
+    [[(apply list '<= (->operands schema filter*))]]
+    SqlKind/LIKE
```
hraberg: I see below that this is supplied as an argument. An alternative is to expose a public function in a namespace and refer to it with its fully qualified name. If supplied as argument, you might want to give it a gensymed or mangled name to not clash with anything else (unlikely, but possible).

crux-sql/src/crux/calcite.clj
```diff
@@ -0,0 +1,183 @@
...
+  [schema filters projects]
+  (try
+    (let [{:keys [crux.sql.table/columns crux.sql.table/query]} schema
+          {:keys [find where]} query]
+      {:find (if (seq projects) (mapv find projects) find)
+       :where (vec
+               (concat
+                (mapcat (partial ->crux-where-clauses schema) filters)
+                where))
+       :args [{:like #(org.apache.calcite.runtime.SqlFunctions/like %1 %2)}]})
```
hraberg: See comment above regarding this, basically wrapping this static Java method in a namespaced function instead.
On another note, can there not be other arguments that needs to be bound / merged here?

dev/user.clj
```diff
@@ -23,7 +24,8 @@
 
 (def standalone-config
   {:node {:crux.node/topology ['crux.standalone/topology
-                               'crux.kv.rocksdb/kv-store]
+                               'crux.kv.rocksdb/kv-store
+                               'crux.calcite/module]
```
hraberg: Should this really be added by default?

crux-sql/src/crux/calcite.clj
```diff
@@ -0,0 +1,183 @@
...
+    (catch Throwable e
+      (log/error e)
+      (throw e))))
+
+(def ^:private column-types->sql-types {:varchar SqlTypeName/VARCHAR
```
hraberg: Should things like `:float` be supported? Doesn't seem like an exhaustive list. Also, what guides the naming of the keyword versions, are they Java or SQL? For example why`:datetime` and not `:date`, `:integer` and not `:int`? My guess is that either Calcite or JDBC has a mapping. JDBC spec can be found here, type mapping is towards the end: https://download.oracle.com/otndocs/jcp/jdbc-4_2-mrel2-spec/index.html Similar info: http://www.h2database.com/html/datatypes.html

crux-test/test/crux/tpch_test.clj
```diff
@@ -0,0 +1,462 @@
...
+                    "  l_returnflag,\n"
+                    "  l_linestatus\n"
+                    "\n"
+                    "order by\n"
+                    "  l_returnflag,\n"
+                    "  l_linestatus"))))
+
+;; Skipping 2: Calcite: @Disabled("Infinite planning")
+
+(t/deftest test-003
```
hraberg: What does this (and similar tests) actually test? Not saying this isn't the right way, just not clear with no check of the result returned something or not and an `(is true)`.

hraberg (commented):
Some comments, OK with merging as such, but not explicitly approving.

crux-test/test/crux/tpch_test.clj
jonpither: It tests that the queries run - they do not if we don't support any of the needed SQL.

dev/user.clj
jonpither: Doesn't need to be - given it's a dev user ns, I thought it would be useful. Ok to remove though?

crux-sql/src/crux/calcite.clj
jonpither: Fixed.

dev/user.clj
jonpither: Have removed.

crux-sql/src/crux/calcite.clj
jonpither: Note, the code above facilitates SELECT * FROM X WHERE "FOO" = "FOO"

crux-sql/src/crux/calcite.clj
jonpither: This generates the clause `[(= surname)]`, which appears to work :-)

crux-sql/src/crux/calcite.clj
jonpither: Have changed

crux-sql/src/crux/calcite.clj
jonpither: Using a fn with a fully qualified name as suggested

crux-sql/src/crux/calcite.clj
jonpither: We're using a literal triple for this

crux-sql/src/crux/calcite.clj
jonpither: Still can't explain it, but changed it anyway to use the more explicit nil? predicate, and using a triple for IS_NOT_NULL

crux-sql/src/crux/calcite.clj
jonpither: Using a literal - would like to take you through

crux-sql/src/crux/calcite.clj
jonpither: We support java.sql.Types as in the README now.

jonpither:
Writing thoughts here on cardinality many. Here's one example to ponder:

Given we have a schema as such:

```
(f/transact! *api* [{:crux.db/id :crux.sql.schema/person
                         :crux.sql.table/name "person"
                         :crux.sql.table/query '{:find [?id ?name ?role]
                                                 :where [[?id :name ?name]
                                                              [?id :role ?role]}
                         :crux.sql.table/columns '{?id :keyword, ?name :varchar, ?role :varchar}}])
```

And we insert this data:

```
  (f/transact! *api* [{:crux.db/id :ivan :name "Ivan" :role "dev"}
                             {:crux.db/id :malcolm :name "Malcolm" :role ["manager" "dev"]}]))
```

If we do this query:

```
  (let [q "SELECT * FROM PERSON"]
    (t/is (= ["Ivan" "Malcolm" "Malcolm"] (sort (map :name (query q))))))
```

This is because the cardinality-many explodes out into multple tuples, just as it does when you do a datalog query.

Firstly, is this a problem? The user could add a `GROUP BY` into their SQL if they needed to collapse the rows. It's arguably correct to return multiple tuples per entity. 

Thoughts? If we think it is a problem... we could perhaps collapse the cardinality-many attributes into an array - perhaps as a function the user could use?

crux-sql/src/crux/CruxCalcitePrepareImpl.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.regex.Pattern;
+import java.time.ZonedDateTime;
+import java.util.Date;
+import java.util.Map;
+import java.util.HashMap;
+
+public class CruxCalcitePrepareImpl extends CalcitePrepareImpl {
+    private static final Pattern PATTERN  = Pattern.compile("((TRANSACTIONTIME|VALIDTIME)\\s\\(\\'(.*?)\\'\\))");
+
+    public CruxCalcitePrepareImpl() {
```
jarohen: don't need this constructor?

jonpither:
Blocked waiting on #907. Would also benefit from #905 and #906 .

jonpither:
Current thoughts on this PR. Currently users are expected to write table-backing-queries that 'project' out fields to create virtual SQL columns.

The issue with is, is that adding fields as filters into a datalog queries for projection reasons only, is quite an inefficient thing to do. I.e. the below query isn't performant.

```
'{:find [l_orderkey, l_partkey, l_suppkey, l_linenumber],
                                             :where [[e :l_orderkey l_orderkey]
                                                     [e :l_partkey l_partkey]
                                                     [e :l_suppkey l_suppkey]
                                                     [e :l_linenumber l_linenumber]
                                                     [e :l_quantity l_quantity]
                                                     [e :l_extendedprice l_extendedprice]
                                                     [e :l_discount l_discount]
                                                     [e :l_tax l_tax]
                                                     [e :l_returnflag l_returnflag]
                                                     [e :l_linestatus l_linestatus]
                                                     [e :l_shipdate l_shipdate]
                                                     [e :l_commitdate l_commitdate]]
                                             :timeout 100000}
```

We would expect our users to either use queries in conjunction with the `entity` api if they want to pull out lots of fields. They will also be able to use the upcoming #849 EQL pull API.

The above query isn't performant because each field is taking part in the nested query acting as individual filters, even though we only want to project the fields.

In #907 I have spiked adding a new kind of pred-fn that is a speed-up for 'leaf-vars'. This gives in the region of a 40% speed up. There are a couple of implications for using this:

1) We would need to adapt the query planner to take advantage of the pred-fn for leaf-vars, or we either recommend users to explicitly use it in their SQL table-backing queries, or we go back to generating SQL queries for the users behalf out of specified columns.

2) Ideally this is hacking around a deeper fix that could be done in the Crux index layer to optimise for leaf-vars, this is commented on in https://github.com/juxt/crux/issues/905.

Taking a step back, I had considered - and spiked - introducing a pull-like approach to projecting out fields, and not including them as filters at all. This is non-trivial we need to make calcs work inside the datalog fns (for filtering) but then also in the projection layer. 

This could be worth a more considered spike.

crux-test/src/crux/fixtures/api.clj
```diff
@@ -0,0 +1,32 @@
+(ns crux.fixtures.api
```
jarohen: think this already exists as `crux.fixtures`?

crux-sql/src/crux/calcite.clj
```diff
@@ -0,0 +1,487 @@
...
+    x))
+
+(defn- vectorize-value [v]
+  (if (or (not (vector? v)) (= 1 (count v))) (vector v) v))
+
+(defn enrich-filter [schema ^RexNode filter]
+  (log/debug "Enriching with filter" filter)
+  (let [[filters clauses exprs args] (->vars+clauses+exprs schema (->ast filter schema))]
+    (-> schema
+        (update-in [:crux.sql.table/query :where] (comp vec concat) (vectorize-value (->where filters)))
```
jarohen: nit: can do this with `(comp vec concat)` replaced with `(fnil into [])` (or just `into` if you know it's present) - saves going to a lazy-seq and back again

jarohen (approved):
Looking good @jonpither :smile:

crux-test/src/crux/fixtures/api.clj
jonpither: have deleted - it crept back in

crux-sql/src/crux/CruxCalcitePrepareImpl.java
jonpither: yep

crux-sql/src/crux/calcite.clj
jonpither: went with `fnil`

hraberg (approved):

PR #741, opened 2020-03-25
Add status endpoint to the webservice, and use for health check.

danmason:
Health check was failing, as it would expect a response code of 200
but was checking the `/` path, which would return a redirect (code
307) to index.html. This commit fixes that issue, and adds a status endpoint.

### DIFF
--- a/examples/soak/src/crux/soak/main.clj
+++ b/examples/soak/src/crux/soak/main.clj
@@ -169,7 +169,7 @@
                           (some-fn {:homepage #(homepage-handler % {:crux-node crux-node})
                                     :weather #(weather-handler % {:crux-node crux-node})
                                     :status #(status-handler % {:crux-node crux-node})}
-                                   (fn [handler] (when (ifn? handler) handler))
+                                   identity
                                    (constantly (constantly (resp/not-found "Not found"))))))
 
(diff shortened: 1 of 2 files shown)


examples/soak/src/crux/soak/main.clj
```diff
@@ -154,10 +156,15 @@
                             (map first))]
     (resp/response (render-homepage location-names))))
 
+(defn status-handler [req {:keys [crux-node]}]
+  (let [status (api/status crux-node)]
```
jarohen, line 159: anything we can check in this status response to check everything's a-ok?

danmason: Oh whoops I meant to add a check completely slipped my mind - I was thinking of more less taking it straight from http-server, which checks for `:crux.zk/zk-active?`? (Out of interest, what is that, exactly?)

examples/soak/src/crux/soak/main.clj
```diff
@@ -154,12 +155,23 @@
...
+    {:status (if
+               (or (not (contains? status :crux.zk/zk-active?))
```
jarohen: nit: predicate can go on the same line as the `if`

jarohen (changes requested):

PR #742, opened 2020-03-25
Add `open-q` function, sans snapshot, for lazy queries

jarohen:
Trying to cut down the API changes into small manageable chunks :) 

Adding `api/open-q` which returns a closeable seq, doesn't need a snapshot.

RFC:
* Calling it `openQuery` in Java - `openQ` didn't seem idiomatic.
* Clojure users see a Closeable seq, to be used within a `with-open` block; Java users see a Stream, to be used within a try-with-resources.
* I deliberately haven't tackled the 're-using snapshots between queries' in this PR
* Have also updated the new `open-tx-log` API to use the same pattern - it was using closeable iterators previously

### DIFF
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -263,7 +263,7 @@
 
   (q
     [db query]
-    [db snapshot query]
+    ^:deprecated [db snapshot query]
     "q[uery] a Crux db.
   query param is a datalog query in map, vector or string form.
   First signature will evaluate eagerly and will return a set or vector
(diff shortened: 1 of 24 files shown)


danmason:
In general terms of the approach, I'm particularly a fan of the 'Clojure users see a Closeable seq' and 'Java users see a Stream' - seems like the best of both worlds to me.

Also, calling it openQuery in Java seems good to me - I have, at points, felt unsure about our query function being 'q' within the API - not sure how most other Clojure APIs are in that respect but I honestly think just 'query' would be better.

jonpither (approved):

crux-core/src/crux/api.clj
```diff
@@ -231,8 +234,8 @@
   (submit-tx [this tx-ops]
     (.submitTx this (conform-tx-ops tx-ops)))
 
-  (open-tx-log ^crux.api.ITxLog [this after-tx-id with-ops?]
-    (.openTxLog this after-tx-id with-ops?)))
+  (open-tx-log ^java.io.Closeable [this after-tx-id with-ops?]
```
danmason, line 237: So this PR also affects the tx-log API?

jarohen: yep - same pattern - and I'm aiming to migrate the history API this way too, up next.

I'll tack a note on the PR

PR #746, opened 2020-03-26
Make extra dependencies optional in crux metrics - move out reporters to their own namespaces

### LINKED ISSUE #701: crux-metrics shouldn't bring in so many dependencies
Users may not require (for example) CloudWatch or Prometheus - let's not bring these in as mandatory dependencies

* figure out what's required for crux-metrics (chances are, just dropwizard core?) and what's optional
* mark the optional dependencies as 'provided'
* ensure everything else still works when the optional dependencies aren't on the classpath
* document up which dependencies users have to add for each metrics configuration

danmason:
Resolves #701


jarohen (changes requested):
thanks Dan :) couple of small changes:

* Could you add a note to the README about _why_ we need the user to add these dependencies?
* Could you use the dependencies specified in project.clj in the docs, rather than a separate file, to reduce the risk of them getting out of date? (aside, but the dependencies you've put in deps.edn wouldn't be valid deps.edn dependencies - they'd need `:mvn/version`)

PR #748, opened 2020-03-27
Replace spec/assert in places where it is necessary for data integrity to check the data.

### LINKED ISSUE #686: We can disable spec asserts without compromising data integrity
Spec asserts currently account for a non-trivial amount of ingest time (in particular). Let's aim for being able to turn spec asserts off in production usage.

* check which specific asserts are taking the time
* ensure that the spec asserts that are required for correct behaviour are migrated to not rely on asserts being turned on
* turn spec asserts off by default in prod, on in dev

danmason:
Resolves #686 

Notes from the commit messages:
- I have replaced calls to (s/assert) in functions which are triggered by data entered by a user (submitted transactions, submitted queries and data submitted in calls to the REST API). I make use of `s/explain-data`, as spec's assert does the same, and throw an **ex-info** of the same format as that thrown by (s/assert).

### DIFF
--- a/crux-core/src/crux/query.clj
+++ b/crux-core/src/crux/query.clj
@@ -1087,13 +1087,17 @@
                            q)
                          identity
                          (fn [q]
-                           (let [q (normalize-query (s/assert ::query q))]
-                             (->ConformedQuery q (s/conform ::query q)))))]
+                           (if-let [s-error (s/explain-data ::query q)]
+                             (throw (ex-info "Spec assertion failed:" s-error))
+                             (let [q (normalize-query q)]
+                               (->ConformedQuery q (s/conform ::query q))))))]
     (if args
-      (do (s/assert ::args args)
+      (if-let [s-error (s/explain-data  ::args args)]
+        (throw (ex-info "Spec assertion failed:" s-error))
+        (do
           (-> conformed-query
               (assoc-in [:q-normalized :args] args)
-              (assoc-in [:q-conformed :args] args)))
+              (assoc-in [:q-conformed :args] args))))
       conformed-query)))
 
 ;; TODO: Move future here to a bounded thread pool.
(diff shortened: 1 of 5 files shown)


crux-core/src/crux/query.clj
```diff
@@ -1087,13 +1087,17 @@
                            q)
                          identity
                          (fn [q]
-                           (let [q (normalize-query (s/assert ::query q))]
-                             (->ConformedQuery q (s/conform ::query q)))))]
+                           (if-let [s-error (s/explain-data ::query q)]
```
jarohen: couple of things with the general pattern here:

* could you use `s/valid?` for the initial check, so that the initial check's quick for the normal case, where the query's valid? don't mind it using slower functions once we know the query's invalid, but trying to keep the fast-path as fast as possible.
* if memory serves there's an easier way to get the error data out than the dance you're having to do here with `with-out-str` - could you have a look through the other functions spec offers?
* I try to use a pattern of 
```
(when invalid? (throw ...))

(process-valid-query ...)
```
rather than
```
(if invalid? 
  (throw ...)
  (process-valid-query ...))
```
to avoid the extra nesting for the happy path - more important if there are more validity checks, obviously

crux-http-server/src/crux/http_server.clj
```diff
@@ -164,17 +164,22 @@
                                   :opt-un [::valid-time
                                            ::transact-time])))
 
+(defn- check-and-return-body [spec body-edn]
```
jarohen: I'd be tempted to call this using `doto` to avoid the `-and-` in the function name
e.g.
```clojure
(doto (body->edn request)
  (validate-or-throw ::query-map))
```

makes a guarantee to the reader that the validate function absolutely cannot alter the value it's being passed

jarohen (changes requested):
hey Dan - thanks - comments below

jarohen (approved):

PR #749, opened 2020-03-27
Storing attributes in a dictionary to save space in content indices, #715

### LINKED ISSUE #715: Store attributes in content indices using a dictionary rather than a full hash
We currently allocate 20 bytes for attributes in the two content indices, allowing for 2^160, ~10^48 potential attributes - we hash the attribute to ensure uniqueness.

This could instead be implemented by a dictionary of attributes maintained by the indexer (and stored in a new index).

We could use 4 byte ints (2^32, 4B) for these attributes and save ourselves 16 bytes in each 80 byte key, and accept that there'll be a slight overhead in translating between ints and attributes in indexing and querying.

(potentially 2 byte shorts if we're prepared to accept that 'users won't ever have more than 65536 attributes', although this seems a little small for a hard upper limit)

jarohen:
resolves #715

### DIFF
--- a/crux-bench/src/crux/bench.clj
+++ b/crux-bench/src/crux/bench.clj
@@ -8,7 +8,8 @@
             [clojure.string :as string]
             [clj-http.client :as client]
             [crux.fixtures :as f]
-            [crux.bus :as bus])
+            [crux.bus :as bus]
+            [crux.kv :as kv])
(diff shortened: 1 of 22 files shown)


jarohen:
this only saved 0.5-1% disk space in practice (compared to 80->60 byte keys), so we decided to park this one, given we have other disk space optimisations in the pipeline.

Issue #465, opened 2019-12-12
Optimising the Crux disk-space requirements


jarohen:
Potential solutions:
* Allowing users to save disk space by compacting older data (accepting that it will no longer be accessible by temporal queries)
* Splitting out storage, keeping current storage local to the node, historical storage shared between nodes on a higher-latency store (NFS, say)

## Chat 2020-05-12

Let's have another round of iteration on this. 

* (#754) Had a pass at removing the C from the AEVC and AVEC indices. Saved ~60-70% disk space on certain benchmarks but it ran slow on watdiv queries, seemingly because the query engine had to go to the object store to resolve each of the values in the tuples
* (#821) Adding a separate CAV index, so that the query engine doesn't go to the object store. Didn't save as much disk space (~35-40%) as removing C, and ingest was slower, but queries were much more acceptable.
* Next round - let's try re-jigging the indices, changing the denormalisation we perform to balance disk space, ingest speed and query speed.

hraberg:
A couple of other approaches:
* Sanity check storage using different compaction flags for Rocks.
* Replace hashes in the index with numeric ids. This requires a mapping step and indexes that tracks hashes to and from their numeric ids. If one wants to really optimise for space one can go for ints (configurable). Various schemes to compress the integers to variable length would complicate the query engine a lot. This extra mapping needs a large in-memory cache. One need to test to see the performance impact, which might even be positive due to the smaller data on disk, but my guess is that tracking and checking for ids would slow things down.
* Mapping the attributes to their own id space, similar to above, but they could likely be ints or even shorts (configurable).

hraberg:
Some statistics after ingesting watdiv.10M.nt, Rocks is 1.1G on disk:

```
| :id | :entries | :key-size | :value-size | :total-size |                                     :name | :size-percent |       :size-entry |
|-----+----------+-----------+-------------+-------------+-------------------------------------------+---------------+-------------------|
|   1 | 11438042 | 968276200 |           0 |   968276200 |                             avec-index-id |            36 | 84.65401683260124 |
|   2 | 11438042 | 968276200 |           0 |   968276200 |                             aecv-index-id |            36 | 84.65401683260124 |
|   0 |   521585 |  11474870 |   617033332 |   628508202 |                content-hash->doc-index-id |            23 | 1204.996696607456 |
|   3 |   521585 |  23992910 |    10953285 |    34946195 | entity+vt+tt+tx-id->content-hash-index-id |             1 |              67.0 |
[...]

PR #759, opened 2020-03-31
Cutting down the content indices - AVEC -> AVE, AECV -> AE

### LINKED ISSUE #754: Removing content-hashes from the content indices to reduce index size
Currently, for each distinct AVE/AEV triple, we store a KV in the content index for each content hash that contains the triple. If a document doesn't change much between versions, this is a lot of near-duplicate entries - let's investigate what's required to remove this.

We include the 'C' in the content indices so that we can check whether the 'current' content hash of the document contains that value for the given attribute - we'll likely need to source this information from the object store (which obviously may impact query times.)

jarohen:
First pass at cutting down the content indices to remove between-content-hash duplication. resolves #754.

* This includes #749 as both changes touched the same index code - will merge the other PR first. first four commits are the attribute dictionary, the remainder are this PR. 
* We naively replace the content-hash lookups with calls into the object-store, which we expect to perform horribly in the query engine - not intending to merge this PR until we've both measured the impact and (almost certainly necessary) found ways to make this faster. Let's discuss potential solutions on the call tomorrow (2020-04-01).
* The buffer allocations are also a little suspicious in parts, no doubt we could re-use some. Holding fire on this until we've discussed ways to make it faster, as some may go away depending on the performance optimisations.
* We've had to introduce an EC index, both to check which documents have been indexed (for the check we perform before index-tx) and to check which documents need to be evicted.
* We'd also like to measure the space saving involved before merging this PR - will involve merging @danmason's bench work.

### DIFF
--- a/crux-core/src/crux/codec.clj
+++ b/crux-core/src/crux/codec.clj
@@ -31,7 +31,7 @@
 
 ; two main indexes for querying
 (def ^:const ^:private ave-index-id 1)
-(def ^:const ^:private aecv-index-id 2)
+(def ^:const ^:private ae-index-id 2)
 
 ; how they work
 (comment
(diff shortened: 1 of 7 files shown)


jarohen:
closing this in favour of some other investigation, see #465

PR #763, opened 2020-04-02
Expose a count of indexed av-pairs

danmason:
(Resolves part of #760)


jarohen (changes requested):
@danmason - can we include this as part of the existing events, rather than creating a new event?

jarohen:
combining with #764, see #767

PR #764, opened 2020-04-02
Expose a 'bytes-indexed' metric to crux-metrics

danmason:
Further work on #760


jarohen:
combining with #763, see #767

PR #766, opened 2020-04-02
Add metric filter for 'time-taken-ms' and 'op-count'

danmason:


jarohen:
as discussed, closing in favour of doing the calculations in Clojure
we needed metrics filters to allow dimensions, but they don't (yet)

PR #767, opened 2020-04-02
KV and bytes-indexed metrics, #760

### LINKED ISSUE #760: Statistics on AV-pairs and kB indexed are available through crux-metrics

jarohen:
combining @danmason's PRs #763 and #764, they both touch the same code

resolves #760

### DIFF
--- a/crux-core/src/crux/tx.clj
+++ b/crux-core/src/crux/tx.clj
@@ -27,11 +27,10 @@
 (s/def ::tx-time date?)
 (s/def ::submitted-tx (s/keys :req [::tx-id ::tx-time]))
 (s/def ::committed? boolean?)
-
+(s/def ::av-count nat-int?)
 (s/def ::doc-ids (s/coll-of #(instance? crux.codec.Id %) :kind set?))
 (defmethod bus/event-spec ::indexing-docs [_] (s/keys :req-un [::doc-ids]))
-(defmethod bus/event-spec ::indexed-docs [_] (s/keys :req-un [::doc-ids]))
-
+(defmethod bus/event-spec ::indexed-docs [_] (s/keys :req-un [::doc-ids ::av-count]))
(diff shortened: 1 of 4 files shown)


crux-core/src/crux/tx.clj
```diff
@@ -383,7 +383,15 @@
 
       (db/put-objects object-store docs)
 
-      (bus/send bus {::bus/event-type ::indexed-docs, :doc-ids (set (keys docs))})
+      (bus/send bus {::bus/event-type ::indexed-docs,
+                     :doc-ids (set (keys docs))
+                     :av-count (->> (vals docs)
```
jarohen: we don't need to call `set` here, and I reckon it might be simpler to combine the AVs first and then count?

crux-core/src/crux/tx.clj
```diff
@@ -383,7 +383,15 @@
...
+                                    (set)
+                                    (map count)
+                                    (reduce +))
+                     :bytes-indexed (->> (vals docs)
+                                         (map #(count (nippy/fast-freeze %)))
```
jarohen: it'll be expensive to re-serialise these just for the bus - let's see if we can re-use the existing serialisation

jarohen (commented):
hey @danmason - thanks :) meta point - best to include any changes required to make tests pass in the same commit as your implementation changes (i.e. considering changing the code and the tests as a single change) - as far as possible it's good to have the tests passing after each commit

jarohen:
using the output from `idx/doc-idx-keys` as the way of counting bytes indexed. it won't be completely accurate because 
* there's also a value in the KV (albeit an empty one)
* duplicate docs don't cost us any more on disk (although they are sorta reindexed, so I'm happy including them in this metric)
* evicted docs are excluded (although hopefully this will be insignificant compared to put docs - and, again, they are reindexed, if you squint)

PR #769, opened 2020-04-06
Bench extract comparison metrics and include further metrics

### LINKED ISSUE #761: Bench metrics extracted from Cloudwatch and included in the daily report
Within the email/slack messages, we should show:

* For each time-taken-ms
  * a percentage increase/decrease (bonus points for conditional colouring) against the most recent (master) run
  * the maximum time-taken-ms in the last week ('week high')
  * the minimum time-taken-ms in the last week ('week low')

* include `:av-count` and `:bytes-indexed` in ingest benches (might need to use a custom bus listener for `:bytes-indexed`)
* when rendering, for each ingest with an `:op-count` key, derive `:ops-per-second`
* likewise `:avs-per-second` and `:bytes-indexed-per-second`

danmason:
Resolves #761

### DIFF
--- a/crux-bench/src/crux/bench.clj
+++ b/crux-bench/src/crux/bench.clj
@@ -110,12 +110,12 @@
                                       doc-count av-count bytes-indexed] :as bench-map}]
   (cond-> (format "*%s* (%s, *%s%%*. 7D Min: %s, 7D Max: %s): `%s`"
                   (name bench-type)
-                  (java.time.Duration/ofMillis time-taken-ms)
+                  (Duration/ofMillis time-taken-ms)
                   (if (neg? percentage-difference-since-last-run)
                     (format "%.2f" percentage-difference-since-last-run)
                     (format "+%.2f" percentage-difference-since-last-run))
-                  (java.time.Duration/ofMillis minimum-time-taken-this-week)
-                  (java.time.Duration/ofMillis maximum-time-taken-this-week)
+                  (Duration/ofMillis minimum-time-taken-this-week)
+                  (Duration/ofMillis maximum-time-taken-this-week)
                   (let [time-taken-seconds (/ time-taken-ms 1000)]
                     (pr-str (dissoc bench-map :bench-ns :bench-type :crux-node-type :crux-commit :crux-version :time-taken-ms
                                     :percentage-difference-since-last-run :minimum-time-taken-this-week :maximum-time-taken-this-week))))
@@ -141,12 +141,12 @@
                              doc-count av-count bytes-indexed] :as bench-map}]
   (cond-> (format "<p> <b>%s</b> (%s, %s. 7D Min: %s, 7D Max: %s): <code>%s</code></p>"
                   (name bench-type)
-                  (java.time.Duration/ofMillis time-taken-ms)
+                  (Duration/ofMillis time-taken-ms)
                   (if (neg? percentage-difference-since-last-run)
                     (format "<b style=\"color: green\">%.2f%%</b>" percentage-difference-since-last-run)
                     (format "<b style=\"color: red\">+%.2f%%</b>" percentage-difference-since-last-run))
-                  (java.time.Duration/ofMillis minimum-time-taken-this-week)
-                  (java.time.Duration/ofMillis maximum-time-taken-this-week)
+                  (Duration/ofMillis minimum-time-taken-this-week)
+                  (Duration/ofMillis maximum-time-taken-this-week)
                   (pr-str (dissoc bench-map :bench-ns :bench-type :crux-node-type :crux-commit :crux-version :time-taken-ms
                                   :percentage-difference-since-last-run :minimum-time-taken-this-week :maximum-time-taken-this-week)))
     (= bench-type :ingest) (str " "
--- a/crux-bench/src/crux/bench.clj
+++ b/crux-bench/src/crux/bench.clj
@@ -108,20 +108,25 @@
 (defn- result->slack-message [{:keys [time-taken-ms bench-type percentage-difference-since-last-run
                                       minimum-time-taken-this-week maximum-time-taken-this-week
                                       doc-count av-count bytes-indexed] :as bench-map}]
-  (format "*%s* (%s, *%s%%*. 7D Min: %s, 7D Max: %s): `%s`"
-          (name bench-type)
-          (java.time.Duration/ofMillis time-taken-ms)
-          (if (neg? percentage-difference-since-last-run)
-            (format "%.2f" percentage-difference-since-last-run)
-            (format "+%.2f" percentage-difference-since-last-run))
-          (java.time.Duration/ofMillis minimum-time-taken-this-week)
-          (java.time.Duration/ofMillis maximum-time-taken-this-week)
-          (let [time-taken-seconds (/ time-taken-ms 1000)]
-            (pr-str (cond-> (dissoc bench-map :bench-ns :bench-type :crux-node-type :crux-commit :crux-version :time-taken-ms
-                                    :percentage-difference-since-last-run :minimum-time-taken-this-week :maximum-time-taken-this-week)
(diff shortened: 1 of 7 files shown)


crux-bench/src/crux/bench.clj
```diff
@@ -53,6 +59,25 @@
...
   `(with-timing* (fn [] ~@body)))
 
+(defn with-additional-index-metrics* [bus f]
+  (let [!av-count (atom 0)
+        !bytes-indexed (atom 0)
+        !doc-count (atom 0)]
+    (bus/listen
+     bus {:crux.bus/event-types #{:crux.tx/indexed-docs}}
+     (fn [{:keys [doc-ids av-count bytes-indexed]}]
+       (swap! !av-count + av-count)
```
jarohen: given both updates and reads to this mutable state are always done together, I'd move this to be one `!index-metrics` atom. 

in the general case, with multiple atoms, you might get a situation where some of the atoms is updated but the others haven't, and then they're all read by a different thread in this intermediate state - what you want is for readers to either see the before state or the after state (an 'atomic' update)

long shot in this case, granted, but best to get into these patterns :smile:

crux-bench/src/crux/bench.clj
```diff
@@ -263,6 +321,47 @@
...
+       result
+       :percentage-difference-since-last-run (-> time-taken-ms
+                                                 (- (first times-taken))
+                                                 (/ (first times-taken))
+                                                 (* 100)
+                                                 (float))
+       :minimum-time-taken-this-week (apply min (conj times-taken time-taken-ms))
+       :maximum-time-taken-this-week (apply max (conj times-taken time-taken-ms))))))
+
+(defmacro with-comparison-times [& body]
```
jarohen: you don't need a macro here - we're using macros for the others because we're changing the execution order of the forms:

```clojure
(with-node [node]
  (run-bench ...))
```
if `with-node` wasn't a macro, Clojure'd try to evaluate `[node]` and `(run-bench ...)`, and then pass them to `with-node` - `[node]` wouldn't compile, and `(run-bench ...)` would run outside of the context of `with-node`

here, you've got a good ol' function that's taking in data and returning data - normal Clojure execution rules are fine :)

crux-bench/src/crux/bench.clj
```diff
@@ -263,6 +321,47 @@
...
+                                                                     (.toInstant)
+                                                                     (.minus (Duration/ofDays 7))
+                                                                     (.toEpochMilli)))
+                                                 (.withEndTime (.getTime (Date.)))))
+                     (.getQueryId))
+        get-query-request (-> (GetQueryResultsRequest.)
+                              (.withQueryId query-id))
+        _ (while (not= "Complete" (-> (.getQueryResults log-client get-query-request)
+                                      (.getStatus)))
+            (Thread/sleep 10))
```
jarohen: 100 reqs/s is probably a bit high :) 10's probably fine, and I'd likely say 1 if we weren't mapping them.

given the requests are asynchronous, maybe we could kick them all off and then wait for all of them?

crux-bench/src/crux/bench.clj
```diff
@@ -263,6 +321,47 @@
     (catch SdkClientException e
       (log/warn (format "AWS credentials not found! File %s not loaded" key)))))
 
+(defn with-comparison-times* [{:keys [crux-node-type bench-type bench-ns time-taken-ms] :as result}]
+  (let [log-client (AWSLogsClientBuilder/defaultClient)
```
jarohen: this function seems to be doing the 'pull' and the 'transform', in Zach's terminology - worth a split?

crux-bench/src/crux/bench/ts_devices.clj
```diff
@@ -58,21 +58,20 @@
...
-                                                                                              [:crux.tx/put reading-doc time])))})
-                                                      {:op-count (count info-tx-ops)}))))]
-      (crux/await-tx node last-tx (Duration/ofMinutes 20))
-      {:op-count op-count
-       :node-size-bytes (bench/node-size-in-bytes node)})))
+   (bench/with-additional-index-metrics node
+     (let [info-tx-ops (vec (for [info-doc info-docs]
+                              [:crux.tx/put info-doc]))
+           _ (crux/submit-tx node info-tx-ops)
+           {:keys [last-tx]} (with-readings-docs
```
jarohen: given `last-tx` is the only thing in this map now, we don't really need the map

crux-bench/src/crux/bench/ts_weather.clj
```diff
@@ -47,23 +47,21 @@
...
-                                                      {:op-count (count location-tx-ops)
-                                                       :last-tx nil}))))]
-      (api/await-tx node last-tx (Duration/ofMinutes 20))
-      {:op-count op-count
-       :node-size-bytes (bench/node-size-in-bytes node)})))
+    (bench/with-additional-index-metrics node
+      (let [location-tx-ops (vec (for [location-doc location-docs]
+                                   [:crux.tx/put location-doc]))
+            _ (api/submit-tx node location-tx-ops)
+            {:keys [last-tx]} (with-condition-docs
```
jarohen: likewise

crux-bench/src/crux/bench/watdiv_crux.clj
```diff
@@ -39,22 +41,30 @@
...
-                      crux-correct (->> both-completed (filter #(= (:db-result-count %) (:result-count %))))
-                      correct-idxs (into #{} (map :query-idx) crux-correct)]
-                  [db-name
-                   {:crux-failures (->> both-completed (map :query-idx) (remove correct-idxs) sort vec)
-                    :crux-errors (->> watdiv-results-with-db (filter :db-result-count) (remove :result-count) (map :query-idx) sort vec)
-
-                    :crux-time-taken (->> crux-correct (map :time-taken-ms) (reduce +) (Duration/ofMillis) (str))
-                    :db-time-taken (->> crux-correct (map :db-time-taken-ms) (reduce +) (Duration/ofMillis) (str))}]))
-              (into {}))))
+  (let [base-map (select-keys (first watdiv-query-results) [:bench-ns :crux-node-type])
```
jarohen, line 67: just a gut feeling, but this function seems like it's doing a lot - 'transform' and 'push', maybe?

crux-bench/src/crux/bench.clj
```diff
@@ -78,11 +103,28 @@
...
 
-(defn- result->slack-message [{:keys [time-taken-ms bench-type] :as bench-map}]
-  (format "*%s* (%s): `%s`"
-          (name bench-type)
-          (java.time.Duration/ofMillis time-taken-ms)
-          (pr-str (dissoc bench-map :bench-ns :bench-type :crux-node-type :crux-commit :crux-version :time-taken-ms))))
+(defn- result->slack-message [{:keys [time-taken-ms bench-type percentage-difference-since-last-run
+                                      minimum-time-taken-this-week maximum-time-taken-this-week
+                                      doc-count av-count bytes-indexed] :as bench-map}]
+  (cond-> (format "*%s* (%s, *%s%%*. 7D Min: %s, 7D Max: %s): `%s`"
```
jarohen: it's probably clear enough in this case, but I prefer not having functions like `str` (particularly, functions that take multiple parameters of the same type) within an arrow - makes the reader (well, me, at least) explicitly consider what the desired behaviour is

jarohen: we could make this clearer using a pattern like this
```clojure
(->> (concat [(format ...)]
             (when ingest?
               (map ...)))
     (string/join "\n"))
```
where the order is immediately obvious?

jarohen (changes requested):
hey Dan, chunky PR, nice work :) 

fancy pairing up on this once I've run through your other PR?

crux-bench/src/crux/bench/watdiv_crux.clj
danmason: Attempt to refactor at a later point.

PR #770, opened 2020-04-06
Include watdiv ingest time other databases

### LINKED ISSUE #751: Bench includes ingest times for watdiv for the comparison DBs in the daily reports

danmason:
Resolves #751

### DIFF
--- a/crux-bench/src/crux/bench/watdiv_crux.clj
+++ b/crux-bench/src/crux/bench/watdiv_crux.clj
@@ -32,9 +32,9 @@
                      (crux/await-tx node last-tx)
                      {:entity-count entity-count
                       :node-size-bytes (bench/node-size-in-bytes node)
-                      :neo4j-time-taken (get-in parsed-db-results [:neo4j :ingest])
-                      :rdf4j-time-taken (get-in parsed-db-results [:rdf4j :ingest])
-                      :datomic-time-taken (get-in parsed-db-results [:datomic :ingest])})))
+                      :neo4j-time-taken-ms (get-in parsed-db-results [:neo4j :ingest])
+                      :rdf4j-time-taken-ms (get-in parsed-db-results [:rdf4j :ingest])
+                      :datomic-time-taken-ms (get-in parsed-db-results [:datomic :ingest])})))
 
 (def db-query-results
   {:rdf4j (get-in parsed-db-results [:rdf4j :queries])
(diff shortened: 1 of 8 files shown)


danmason:
This failed on AWS - it couldn't do `(json/write-str)` on duration. Will convert it to a string.

crux-bench/src/crux/bench/watdiv_datomic.clj
```diff
@@ -351,11 +351,12 @@
...
 
 (defn -main []
-  (let [output-file (io/file "datomic-results.edn")]
-    (bench/save-to-file output-file
-                        (->> (run-watdiv-bench {:test-count 100})
-                             (filter :query-idx)
-                             (sort-by :query-idx)))
+  (let [output-file (io/file "datomic-results.edn")
+        watdiv-results (run-watdiv-bench {:test-count 100})]
+    (bench/save-to-file output-file (conj (->> (rest watdiv-results)
```
jarohen: nit: this is correct, but when I was reading it I had to double-check whether you wanted to put the `(first watdiv-results)` on the end or the beginning of this seq. let's take away the ambiguity with `(cons (first ...) (->> (rest ...) ...))`?

crux-bench/src/crux/bench/watdiv_crux.clj
```diff
@@ -7,28 +7,41 @@
...
-
-(defn parse-results [resource]
-  (with-open [rdr (io/reader resource)]
-    (->> (line-seq rdr)
-         (map read-string)
-         (filter :result-count)
-         (map (juxt :query-idx identity))
-         (into {}))))
+                      :node-size-bytes (bench/node-size-in-bytes node)
+                      :neo4j-time-taken (get-in parsed-db-results [:neo4j :ingest])
```
jarohen: units, maybe?

jarohen: oh, I see, these are Durations. given we're still in raw-data territory here, (indeed, this is what's getting sent to CloudWatch?) let's try migrate the conversion to strings until we're rendering the results?

jarohen (changes requested):
cheers Dan :) few little bits to update but nothing major, nicely done

crux-bench/src/crux/bench/watdiv_crux.clj
danmason: In which case, should move the conversion to a Duration in general.

Issue #528, opened 2020-01-03
CaS then evict may not be replayable after #362, migrating to 1.6.0


jarohen:
Previously, we used to compare documents using just the content-hash, which was preserved after an eviction (even though the content of the document was removed).

When we put the #362 change in to make the content hashes of maps consistent (i.e. not dependent on their ordering) we had to put a change in to make CaS compare the document content, because the hash of the document when it was first added to Crux may well not have been calculated using the same method as the content-hash in the CaS operation.

The extent of the impact of this is: 
* Existing cluster on <1.6.0 migrating to 1.6.0 or later
* CaS op on doc A, evict doc A in a subsequent transaction, wait for Kafka compaction, replay the log (e.g. on a new node) - the CaS operation (and all the operations in the same tx) will fail on the new node where it would have succeeded on the existing node

jarohen:
Closing this one - if you do run into it, please get in touch via our usual support channels.

Cheers!

PR #777, opened 2020-04-08
Match operation, #500

### LINKED ISSUE #500: Splitting `cas` into `match` and `put`
Technically speaking, `crux.tx/cas` isn't a primitive, it can be thought of as two parts - the compare, and the set. We'd like to split these out, so that users can (for example) compare one document and then set another, or set the same document at a different valid time.

* [x] Adding a `:crux.tx/match` operation, `[:crux.tx/match <eid> <doc> <valid-time?>]`, which rolls back the transaction if the document is incorrect
* [x] Refactoring CaS in the indexer, in terms of match and put.
* [ ] Updating documentation/tutorials to refer to match instead of CaS 

to be discussed:
* should we remove CaS? it will, at least, remain on existing Kafka logs, so it will need to remain as a tx-event, but we may be able to remove it as a tx-op (so users will no longer be able to submit new CaS ops)

jarohen:
resolves xtdb/xtdb#500

TODO:
* [ ] Update tutorials
* [x] Decide whether we're removing CaS, or leaving it as backwards compatibility *leaving it for one release*

### DIFF
--- a/crux-core/src/crux/tx.clj
+++ b/crux-core/src/crux/tx.clj
@@ -76,6 +76,7 @@
         ::put (zipmap [:eid :content-hash :start-valid-time :end-valid-time] args)
         ::delete (zipmap [:eid :start-valid-time :end-valid-time] args)
         ::cas (zipmap [:eid :old-content-hash :new-content-hash :valid-time] args)
+        ::match (zipmap [:eid :content-hash :valid-time] args)
         ::evict (let [[eid & args] args
                       [start-valid-time end-valid-time] (filter inst? args)
                       [keep-latest? keep-earliest?] (filter boolean? args)]
@@ -232,6 +233,16 @@
 
 (defmethod index-tx-event :crux.tx/delete [[op k start-valid-time end-valid-time] tx deps]
   {:etxs (put-delete-etxs k start-valid-time end-valid-time nil tx deps)})
+
+(defmethod index-tx-event :crux.tx/match [[op k v at-valid-time :as match-op]
+                                          {:crux.tx/keys [tx-time tx-id] :as tx}
+                                          {:keys [history] :as deps}]
+  {:pre-commit-fn #(let [{:keys [content-hash] :as entity} (entity-at history
+                                                                      (c/new-id k)
+                                                                      (or at-valid-time tx-time)
+                                                                      tx-time)]
+                     (or (= (c/new-id content-hash) (c/new-id v))
+                         (log/debug "crux.tx/match failure:" (cio/pr-edn-str match-op) "was:" (c/new-id content-hash))))})
 
 (defmethod index-tx-event :crux.tx/cas [[op k old-v new-v at-valid-time :as cas-op]
                                         {:crux.tx/keys [tx-time tx-id] :as tx}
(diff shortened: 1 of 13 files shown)


jonpither:
As mentioned - I'd like to review the overall status of transaction functions and predicates before committing to this - to avoid possible churn. See xtdb/xtdb#2019

jarohen:
as discussed, going for a `match` operation for now, let's discuss tx-fns and/or a more general `pre` operation in more detail once we've moved into beta :)

jonpither (approved):

docs/img/crux-node-1.svg
```diff
@@ -485,16 +485,11 @@
...
-         x="149.2623"
-         y="31.105722"
-         style="font-weight:bold;font-size:7.76111126px;line-height:1.25;text-align:start;text-anchor:start;stroke-width:0.18165512"
-         id="tspan2155-0-6-3-7-0-5-4-7-4-4-3">Put/</tspan><tspan
          sodipodi:role="line"
          x="149.2623"
          y="40.80711"
          style="font-weight:bold;font-size:7.76111126px;line-height:1.25;text-align:start;text-anchor:start;stroke-width:0.18165512"
-         id="tspan7694-9-0">CAS/</tspan><tspan
+         id="tspan7694-9-0">Put/</tspan><tspan
```
hraberg, line 492: Hard to tell without looking at the SVG, but should this maybe be "Match"?

crux-core/src/crux/tx.clj
```diff
@@ -233,6 +234,16 @@
 (defmethod index-tx-event :crux.tx/delete [[op k start-valid-time end-valid-time] tx deps]
   {:etxs (put-delete-etxs k start-valid-time end-valid-time nil tx deps)})
 
+(defmethod index-tx-event :crux.tx/match [[op k v at-valid-time :as match-op]
+                                          {:crux.tx/keys [tx-time tx-id] :as tx}
+                                          {:keys [history] :as deps}]
+  {:pre-commit-fn #(let [{:keys [content-hash] :as entity} (entity-at history
```
hraberg, line 240: Feels like it make sense to share the logic between CAS and match here? That is, CAS delegates to the same logic match uses. My impression is that all CAS tests have been removed (or rather, converted), so feels better to have indirect coverage and same behaviour for both.

hraberg (approved):
Looks good, but wasn't sure we were planning to directly remove CAS from docs and everything? Like this kind of makes sense, but guess it's open to debate. Doesn't block anything though!

Can see a case for both keeping CAS as well as deprecating it.

docs/img/crux-node-1.svg
jarohen: if memory serves, it was originally 3 lines, put, cas, delete - I replaced cas with put and removed the original put, because they were labels for a line below

crux-core/src/crux/tx.clj
jarohen: mm. :thinking: I didn't because of the additional object store check in the original CaS (needed because of the sorted-maps rehash legacy issue), which isn't required in the new match, and also causes other issues, like #528. not sure though?

jarohen: looked over it this morning - we've still got the test-can-cas-entity (amongst others) in tx-test. if/when we remove CaS as a user-visible tx-op, we can use a similar pattern to when we removed the evict time ranges - creating tests that manually inserted legacy events in the tx-log.

PR #779, opened 2020-04-09
Adding manual KV compaction for Rocks

jarohen:
Compacting the KV store after we ingest so that we get more reliable disk space usage stats in bench.

Bench shows pre/post compaction disk usage stats, as well as how many raw bytes we've thrown at it.

Have added support for compaction to Rocks but have left the others as no-ops for now, can't find an equivalent in LMDB?


jonpither:
LGTM

PR #783, opened 2020-04-14
Add `open-db` to share a snapshot between DB queries/history reqs/etc

jarohen:
Easy case is unchanged: `(crux/q (crux/db *api*) '{...})`

If you want to share a snapshot between different calls to query/history/entity/etc, it's pretty much the same, but call `open-db` (most likely in a `with-open`)

```clojure
(with-open [db (crux/open-db *api*)]
  (t/is (= ... (crux/entity db ...)))
  (t/is (= ... (crux/q db '{...}))))
```

Have deliberately kept the change surface as small as possible, but happy to talk about bigger changes if you feel they'd be more suitable?

As with the other #410 PRs, this one is backwards compatible - the backwards compatibility will be removed in a later PR.

This PR is based on top of #782, I'll merge that before merging this PR

### DIFF
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -63,18 +63,36 @@
 (defprotocol PCruxNode
   "Provides API access to Crux."
   (db
-    [node]
-    [node ^Date valid-time]
-    [node ^Date valid-time ^Date transaction-time]
-    "Will return the latest value of the db currently known. Non-blocking.
+    ^crux.api.ICruxDatasource [node]
+    ^crux.api.ICruxDatasource [node ^Date valid-time]
+    ^crux.api.ICruxDatasource [node ^Date valid-time ^Date transaction-time]
+    "When a valid time is specified then returned db value contains only those
+  documents whose valid time is before the specified time.
 
-  When a valid time is specified then returned db value contains only those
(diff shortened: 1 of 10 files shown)


crux-core/src/crux/api.clj
```diff
@@ -172,62 +190,48 @@
...
      (defonce warn-on-deprecated-sync
        (log/warn "(sync tx-time <timeout?>) is deprecated, replace with either (await-tx-time tx-time <timeout?>) or, preferably, (await-tx tx <timeout?>)"))
      (.awaitTxTime this tx-time timeout)))
 
   (await-tx
-    ([this submitted-tx]
-     (await-tx this submitted-tx nil))
-    ([this submitted-tx timeout]
-     (.awaitTx this submitted-tx timeout)))
+    ([this submitted-tx] (await-tx this submitted-tx nil))
```
jonpither, line 224: These seem like purely stylistic / convention changes...

jonpither:
When would we remove the deprecated fns?

jarohen:
@jonpither plan is to remove them after 1.8.2, so that we can release 1.9.0 shortly afterwards with the breaking changes, but still give users 1.8.2 where both versions work so that they can migrate incrementally

jonpither (approved):

hraberg (approved):

Issue #788, opened 2020-04-16
Users can see the state of an entity


jarohen:
- Accessible through `/_entity/<url-encoded-eid>` in the crux-http-server, decoded via same logic as crux.codec/id-edn-reader
- Negotiates response content type, application/edn, application/transit+json, text/html (SPA/JSON out of scope gavel-emoji)
- Accepts valid-time/transaction-time as query parameters
- If the entity doesn’t exist, the endpoint returns a 404

PR #800, opened 2020-04-17
Crux Data Browser - Entity State

danmason:
Adds `_entity`, as per #788

### DIFF
--- a/crux-http-server/src/crux/http_server.clj
+++ b/crux-http-server/src/crux/http_server.clj
@@ -8,6 +8,7 @@
             [clojure.spec.alpha :as s]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
+            [clojure.instant :as instant]
             [crux.codec :as c]
             [crux.io :as cio]
             [crux.tx :as tx]
@@ -205,7 +206,11 @@
 (defn- entity-state [^ICruxAPI crux-node request]
   (let [[_ encoded-eid] (re-find #"^/_entity/(.+)$" (req/path-info request))]
     (let [eid (c/id-edn-reader encoded-eid)
-          db (db-for-request crux-node {})]
+          query-params (:query-params request)
+          db (db-for-request crux-node {:valid-time (some-> (get query-params "valid-time")
+                                                            (instant/read-instant-date))
+                                        :transaction-time (some-> (get query-params "transaction-time")
+                                                                  (instant/read-instant-date))})]
       {:status 200
        :body (.entity db eid)})))
 
--- a/crux-http-server/src/crux/http_server.clj
+++ b/crux-http-server/src/crux/http_server.clj
@@ -455,10 +455,10 @@
 
                                                             (-> (partial data-browser-handler node)
                                                                 (p/wrap-params)
-                                                                (wrap-format
-                                                                 (assoc-in m/default-options [:formats "text/html"] (mfc/map->Format
-                                                                                                                     {:name "text/html"
-                                                                                                                      :encoder [html-encoder]})))
+                                                                (wrap-format (assoc-in m/default-options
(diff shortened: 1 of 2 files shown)


danmason:
Failing build - Might need to move the muuntaja middleware into the `entity-state` itself, if possible

crux-http-server/src/crux/http_server.clj
```diff
@@ -393,10 +412,40 @@
      (log/info "HTTP server started on port: " server-port)
      (->HTTPServer server options))))
 
+(defn- edn->html [edn]
+  (format
```
jarohen: what will this do with nested structures?

crux-http-server/src/crux/http_server.clj
```diff
@@ -393,10 +412,40 @@
...
+(def html-formatter
+  (mfc/map->Format
+    {:name "text/html"
+     :encoder [html-encoder]}))
+
 (def module
   {::server {:start-fn (fn [{:keys [crux.node/node]} {::keys [port] :as options}]
                          (let [server (j/run-jetty (-> (partial handler node)
                                                        (p/wrap-params)
+                                                       (wrap-format
```
jarohen: this will wrap all the endpoints with this HTML formatter - that what we want?

crux-http-server/src/crux/http_server.clj
```diff
@@ -199,6 +203,18 @@
     (-> (success-response (.entity db eid))
         (add-last-modified tx-time))))
 
+(defn- entity-state [^ICruxAPI crux-node request]
+  (let [[_ encoded-eid] (re-find #"^/_entity/(.+)$" (req/path-info request))]
```
jarohen, line 208: I'd say the time's come to start with a proper routing library - we've outgrown manually parsing out each individual value

crux-http-server/src/crux/http_server.clj
```diff
@@ -199,6 +203,18 @@
...
+    (let [eid (c/id-edn-reader encoded-eid)
+          query-params (:query-params request)
+          db (db-for-request crux-node {:valid-time (some-> (get query-params "valid-time")
+                                                            (instant/read-instant-date))
+                                        :transact-time (some-> (get query-params "transaction-time")
+                                                               (instant/read-instant-date))})
+          entity-map (.entity db eid)]
```
jarohen: advise using the Clojure API here, `(crux/entity db eid)`

crux-http-server/src/crux/http_server.clj
```diff
@@ -393,10 +412,40 @@
...
+
+(defn- html-encoder [_]
+  (reify
+    mfc/EncodeToBytes
+    (encode-to-bytes [_ data charset]
+      (.getBytes
+       ^String (edn->html data)
+       ^String charset))
+    mfc/EncodeToOutputStream
+    (encode-to-output-stream [_ data charset]
```
jarohen, line 444: the reason for a separate `encode-to-output-stream` (in addition to `encode-to-bytes`) is likely that the former's to be used for large streaming response bodies (which we will likely need for queries), the latter being sufficient for when the response fits in memory.

probably more of an issue when you implement `/_query`, so maybe this is fine for now

jarohen (changes requested):
thanks Dan - few things to think about below. 

also, can't see a build for this commit?

crux-http-server/src/crux/http_server.clj
jarohen: as discussed, bringing this in as a separate change

crux-http-server/src/crux/http_server.clj
jarohen: yep, let's leave this for query

PR #801, opened 2020-04-17
Making sure we don't keep a reference to a seq in open-q (et al)

jarohen:
@jonpither flagged that `open-q` in its current form is going to hold on to the head of the seq
I'd thought that because we were taking an iterator over the query results and that because the iterator was (by its nature) stateful, it'd release elements from the sequence as it iterated through
and, in truth, it does - callers using the Java API will get a fully lazy sequence.
the problem comes, though, when the Clojure API calls iterator-seq on this iterator, and then uses that seq in a call to with-open:

```clojure
(with-open [res (crux/open-q ...)]
  (doseq [row res]
    ...))
```

which translates to

```clojure
(let [res (crux/open-q)]
  (try
    (doseq [row res]
      ...)
    (finally
      (.close res))))
```

which, then, keeps a reference to res; the implementation of the iterator-seq will hang on to the elements. sad times.

only implementing a select few interfaces now - means that the Seqable implementation will behave mutably. users can explicitly call `seq` to take a cached snapshot, which will then (if you keep a reference to the head) be able to be iterated consistently, multiple times (with the usual Clojure lazy-seq memory usage caveats)

I haven't had to change any of the tests for this, so usage remains the same as before

### DIFF
--- a/crux-core/src/crux/io.clj
+++ b/crux-core/src/crux/io.clj
@@ -4,7 +4,7 @@
             [clojure.spec.alpha :as s]
             [clojure.tools.logging :as log]
             [taoensso.nippy :as nippy])
-  (:import [clojure.lang ISeq Sequential IPending]
+  (:import [clojure.lang Counted Sequential IPending IReduceInit Seqable]
            [java.io Closeable DataInputStream DataOutputStream File IOException Reader]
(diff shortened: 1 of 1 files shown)


jarohen:
see #807

PR #808, opened 2020-04-21
Add _query endpoint, and ability to find linked entities

### LINKED ISSUE #789: Users can navigate to linked entities
- Users can navigate to other linked entity documents from an entity
- Users can navigate back to previous entities results using standard browser navigation
- Preserves selected valid-time/transaction-time

### LINKED ISSUE #791: Users can run queries against the node
- Users can enter any query into crux, and get a set of results
- `GET /_query` renders a simple form if no query
  - Depending on the Accept header, this also returns edn/transit/html/csv/tsv
  - TBC: Accepts `?q=...&args=...` query-param initially, or maybe `?find=e,w&where=...`
  - Users can select a valid time and a transaction time to query against
- Failed queries output clear syntax error to user
- Queries can be paginated via limit/offset (for now), and are subject to a server-mandated max-count
  - Links included to next/previous pages in response headers

danmason:
Moves the edn->html rendering back into handlers, such that there's no need to send around extra necessary data for rendering via meta.


crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,172 @@
       ((resolve 'crux.sparql.protocol/sparql-query) crux-node request)
       nil)))
 
+(defn link-all-entities
```
danmason, line 372: Used an atom here to avoid having to pass around `links` through the recursive function - happy to change if it is a particular problem.

jarohen: as discussed, let's pass it down through the recursive calls

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
...
+  [crux-node vt tt path result]
+  (let [resolved-links (fn recur-on-result [result links]
+                         (if (and (c/valid-id? result)
+                                  (api/entity (api/db crux-node vt tt) result))
+                           (let [query-params (cond-> "?"
+                                                vt (str "valid-time=" vt "&")
```
jarohen: even if the user doesn't specify vt and tt, you still want to link to the entity at the vt and tt that this entity was fetched at, so that they're consistent - if you pass `db` into here instead you can take `(crux/valid-time db)` and `(crux/transaction-time db)`.

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
       ((resolve 'crux.sparql.protocol/sparql-query) crux-node request)
       nil)))
 
+(defn link-all-entities
+  [crux-node vt tt path result]
+  (let [resolved-links (fn recur-on-result [result links]
```
jarohen: if you need to name the function, it's worth using `letfn`

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
...
+                         (if (and (c/valid-id? result)
+                                  (api/entity (api/db crux-node vt tt) result))
+                           (let [query-params (cond-> "?"
+                                                vt (str "valid-time=" vt "&")
+                                                tt (str "transact-time=" tt))]
+                             (assoc links result (str path "/" result query-params)))
+                           (cond
+                             (map? result) (reduce-kv #(merge %1 (recur-on-result %3 %1)) links result)
+                             (sequential? result) (reduce #(merge %1 (recur-on-result %2 %1)) links result)
```
jarohen: don't think you need to use `reduce` or `reduce-kv` here - if you don't need to use the accumulated value it's often clearer to use `map`/`mapcat` and friends.

separately, if you need to dive into `%3` it's probably time for a `(fn [your param names] ...)`

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
...
+        (for [row rows]
+          [:tr
+           (for [header headers]
+             (let [cell-value (get row header)]
+               [:td
+                (if-let [href (get links cell-value)]
+                  [:a {:href href}
+                   (str cell-value)]
+                  (str cell-value))]))])]]
+      [:div "No results found"])]])
```
jarohen: I'm not sure about this 'no results found' - we've tried to make the rest of the HTML as computer-readable as possible, but this kind of thing would need to be explicitly coded against.

Obvious counter-point is that computers will likely go for JSON/EDN instead, so likely not an issue.

@malcolmsparks - is the preference that the JavaScript-less HTML is easily machine-parsable or still vaguely pleasant to a human?

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
...
+  [{:keys [find full-results?] :as query} results]
+  (if (and full-results? (= 1 (count (first results))))
+    (->> results
+         (mapv #(map keys %))
+         (flatten)
+         (into #{}))
+    find))
+
+(defn resolve-rows
+  [{:keys [find full-results?] :as query} results]
```
jarohen: I'm not clear on the `full-results?` behaviour here (or indeed what we want it to be!) - maybe one for a quick catchup?

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
...
+           :else param)]
+     (c/read-edn-string-with-readers formatted-param))))
+
+(defn- build-query
+  [{:strs [find where args order-by limit offset full-results]}]
+  (cond-> {}
+    :find (assoc :find (coerce-param find))
+    :where (assoc :where (coerce-param where :wrap))
+    args (assoc :args (coerce-param args))
+    order-by (assoc :order-by (coerce-param args :wrap))
```
jarohen: should this be `order-by` - copy pasta?

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
...
+         (cond
+           (vector? param) (str "[" (reduce str param) "]")
+           wrap? (str "[" param "]")
+           :else param)]
+     (c/read-edn-string-with-readers formatted-param))))
+
+(defn- build-query
+  [{:strs [find where args order-by limit offset full-results]}]
+  (cond-> {}
+    :find (assoc :find (coerce-param find))
```
jarohen: re the `:find` as the `cond->` test - intentionally a keyword? (as in, always truthy?)

if so, clearer to unconditionally include `find` and `where` in the `{}`?

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
...
+    limit (assoc :limit (Integer/parseInt limit))
+    offset (assoc :offset (Integer/parseInt offset))
+    full-results (assoc :full-results? true)))
+
+(defn resolve-headers
+  [{:keys [find full-results?] :as query} results]
+  (if (and full-results? (= 1 (count (first results))))
+    (->> results
+         (mapv #(map keys %))
+         (flatten)
```
jarohen: `mapcat` is your friend here

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
...
+                 entity-map))})))
+
+(defn- coerce-param
+  ([param]
+   (coerce-param param nil))
+  ([param wrap?]
+   (let [formatted-param
+         (cond
+           (vector? param) (str "[" (reduce str param) "]")
+           wrap? (str "[" param "]")
```
jarohen: why `wrap?`?

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
...
+                   (html5 (entity->html linked-entities entity-map)))
+                 entity-map))})))
+
+(defn- coerce-param
+  ([param]
+   (coerce-param param nil))
+  ([param wrap?]
+   (let [formatted-param
+         (cond
+           (vector? param) (str "[" (reduce str param) "]")
```
jarohen: bit dirty to enclose these in the `[` and `]` chars and then parsing, worth mapping `c/read-edn-...` over it instead?

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
...
+(defn resolve-headers
+  [{:keys [find full-results?] :as query} results]
+  (if (and full-results? (= 1 (count (first results))))
+    (->> results
+         (mapv #(map keys %))
+         (flatten)
+         (into #{}))
+    find))
+
+(defn resolve-rows
```
jarohen: seems like this belongs with `resolve-headers`, let's refactor?

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
...
+  [{:keys [find full-results?] :as query} results]
+  (if (and full-results? (= 1 (count (first results))))
+    (map first results)
+    (map #(zipmap find %) results)))
+
+(defn link-top-level-entities
+  [crux-node vt tt path rows]
+  (reduce
```
jarohen: see `reduce`/`reduce-kv` comment above

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
...
+(defn- coerce-param
+  ([param]
+   (coerce-param param nil))
+  ([param wrap?]
+   (let [formatted-param
+         (cond
+           (vector? param) (str "[" (reduce str param) "]")
+           wrap? (str "[" param "]")
+           :else param)]
+     (c/read-edn-string-with-readers formatted-param))))
```
jarohen: seems strange that we're giving the user the choice of EDN/JSON/etc response types, but mandating EDN query params - haven't come across this before. thinking about it, whenever I've had something sufficiently complicated to warrant EDN (i.e. more than numbers, bools, dates, etc) in query params, I've usually moved it to a request body instead, at which point you've got request-body coercion using the `content-type` header - but this goes against the 'hackable query URL' principle.

again, likely one for @malcolmsparks's opinion

crux-http-server/src/crux/http_server.clj
```diff
@@ -380,14 +368,167 @@
...
+  (let [query-params (:query-params request)
+        query (build-query query-params)
+        valid-time (some-> (get query-params "valid-time")
+                           (instant/read-instant-date))
+        transaction-time (some-> (get query-params "transaction-time")
+                                 (instant/read-instant-date))
+        db (db-for-request crux-node {:valid-time valid-time
+                                      :transact-time transaction-time})
+        results (api/q db query)]
+    {:status (if (some? results) 200 404)
```
jarohen: suspect this `some?` isn't doing what you think it is - `q` will always return at least an empty set, so `some?` will always return true.

that said, I don't think we want to return 404 on no-results either - it's not like 'entity' where the entity requested in the URL is not found - the the query was valid, it just didn't have any results.

crux-http-server/src/crux/http_server.clj
jarohen: wider point - this seems quite low-level string wrangling - time to consider a library?

crux-http-server/src/crux/http_server.clj
```diff
@@ -416,35 +557,18 @@
...
-   (html
-    (edn->html edn))))
-
 (defn- html-encoder [_]
   (reify
     mfc/EncodeToBytes
     (encode-to-bytes [_ data charset]
       (.getBytes
-       ^String (entity->html data)
+       ^String (pr-str data)
```
jarohen: I know we're not using this now, because we're manually specifying formats for each handler, but `pr-str` doesn't seem like a reasonable default - shall we `(throw (UnsupportedOperationException.))` instead, to be really clear that we don't expect to get here?

crux-http-server/src/crux/http_server.clj
```diff
@@ -477,3 +601,9 @@
              :args {::port {:crux.config/type :crux.config/nat-int
                             :doc "Port to start the HTTP server on"
                             :default default-server-port}}}})
+
+
+#_(comment
```
jarohen: you can use `user.clj` for this, happy to demo

jarohen (changes requested):
thanks gents :) feedback below. reckon it's worth us having a pairing (trioing? :) ) session on a few of them.

more generally:

* don't know if it's intended, but there's a few points missing from the `/_query` card - particularly, pagination?
* noticed a couple of likely bugs - given we're going to be using these endpoints as the public Crux API, can we either get some tests around them, or maybe make use of our existing remote client tests?
* meta point - in hindsight I suspect this might have been easier to keep track of as distinct changes, rather than mixing query + linked entities?

crux-http-server/src/crux/http_server.clj
danmason: It still gets here, we're still using html-encoder (to encode the HTML strings returned from the respective handlers)

crux-http-server/src/crux/http_server.clj
jarohen: from @malcolmsparks: fine to keep EDN-formatted QPs for now, especially as we have the raw `q=` param too

crux-http-server/src/crux/http_server.clj
jarohen: from @malcolmsparks: prefer that it's vaguely pleasant to humans, because machines will go for JSON/EDN, so this is fine

PR #814, opened 2020-04-22
S3 doc-store, #810

### LINKED ISSUE #810: S3 can be used as a document store

jarohen:
resolves #810

TODO:
- [x] Config - discuss. There's a plethora of S3 options (where to get creds from, object encryption, etc) - should we/how should we expose these to users? Don't want to replicate the S3 API, but also don't want to prevent users using it because we don't expose the `<x>` config option.
- [x] Tests - discuss. Could include this as part of API test, but would probably take a long time for not much benefit (and would need config/creds). Maybe we should have doc-store specific acceptance tests?
- [x] README
- [ ] docs, blogs etc. (e.g. walk-through creating a file based doc store)


crux-s3/project.clj
```diff
@@ -0,0 +1,19 @@
+(defproject juxt/crux-bench "derived-from-git"
```
jonpither: should be crux-s3

jonpither (changes requested):

jonpither (approved):

crux-s3/src/crux/s3.clj
```diff
@@ -0,0 +1,81 @@
...
+         (into {} (keep (fn [[id ^CompletableFuture resp]]
+                          (when-let [doc (.get resp)]
+                            [id doc])))))))
+
+(s/def ::bucket string?)
+(s/def ::prefix string?)
+
+(def s3-doc-store
+  {::configurator {:start-fn (fn [deps args]
+                               (reify S3Configurator))}
```
hraberg, line 65: This could/should simply be `S3Configurator.` ctor.

hraberg (approved):

PR #815, opened 2020-04-23
Index store

hraberg:
This PR aims to introduce the `IndexStore`, which here can be created from the `Indexer` via `open-index-store`. This protocol can be seen as the read side of the index. Refactoring here removes all direct references to `crux.kv` from the `crux.query` namespace and instead go via the new protocol to access and create the indexes.

This PR **does not** aim to provide a clean and good protocol here yet, it just aims to establish a boundary. I expect several later PRs to hone the design of the actual protocol and tease apart the query engine a bit.

Introducing this protocol is the foundation for many other pieces of work related to the indexes and query engine.

_Note: there's no issue for this work yet, but there's a note related to it on the Roadmap board._

A few of things to note:

1. `KvIndexStore` also implements `KvSnapshot` as it simply wraps one, this is to ease of use in places where the object store currently expects a snapshot.
2. the methods on the new protocol are simply delegating to existing code in `crux.index` - nothing has really been moved around or added.
3. a concept of opening up nested index stores is added simply to keep in line with the current implementation which nests cached snapshots for `or` and `not` sub queries. This was originally done way back to allow reuse (instead of just accumulate) iterators between sub queries.
4. I've also decided to add the `tx-failed?` method, to avoid digging straight into the kv store to detect this. Similarly, I've changed a few uses  of `read-meta` to go via the `read-index-meta` method to hide the kv store.
5. the concept of sending down the snapshot into the object store is kept, but is messy. I did change the `RocksKvSnapshot` to actually use the read options (which contain snapshot info) in the `get-value` method used by the object store.

In spirit of this protocol, there are more places that should be hidden behind protocols to make the query engine more modular, but one thing at a time. There are still assumptions here and there that everything is always stored in the kv store.

But more  immediate is to fix this protocol itself in subsequent PRs. That's harder than just introducing it, as it would require rewriting and redesigning parts of the query engine internals, especially lower levels. In addition to this, `crux.index` should be split into the parts related to `KvIndexStore` and the higher levels that are not.

### DIFF
--- a/crux-core/src/crux/object_store.clj
+++ b/crux-core/src/crux/object_store.clj
@@ -19,20 +19,26 @@
 (defrecord KvObjectStore [kv]
   db/ObjectStore
   (get-single-object [this snapshot k]
-    (let [doc-key (c/->id-buffer k)
-          seek-k (c/encode-doc-key-to (.get i/seek-buffer-tl) doc-key)]
-      (some->> (kv/get-value snapshot seek-k)
-               (DirectBufferInputStream.)
-               (DataInputStream.)
-               (nippy/thaw-from-in!))))
+    (if (satisfies? kv/KvSnapshot snapshot)
+      (let [doc-key (c/->id-buffer k)
+            seek-k (c/encode-doc-key-to (.get i/seek-buffer-tl) doc-key)]
+        (some->> (kv/get-value snapshot seek-k)
+                 (DirectBufferInputStream.)
+                 (DataInputStream.)
+                 (nippy/thaw-from-in!)))
+      (with-open [snapshot (kv/new-snapshot kv)]
+        (db/get-single-object this snapshot k))))
 
   (get-objects [this snapshot ks]
-    (->> (for [k ks
-               :let [seek-k (c/encode-doc-key-to (.get i/seek-buffer-tl) (c/->id-buffer k))
-                     doc (some-> (kv/get-value snapshot seek-k) <-nippy-buffer)]
-               :when doc]
-           [k doc])
-         (into {})))
+    (if (satisfies? kv/KvSnapshot snapshot)
+      (->> (for [k ks
+                 :let [seek-k (c/encode-doc-key-to (.get i/seek-buffer-tl) (c/->id-buffer k))
+                       doc (some-> (kv/get-value snapshot seek-k) <-nippy-buffer)]
+                 :when doc]
+             [k doc])
+           (into {}))
+      (with-open [snapshot (kv/new-snapshot kv)]
+        (db/get-objects this snapshot ks))))
 
   (put-objects [this kvs]
     (kv/store kv (for [[k v] kvs]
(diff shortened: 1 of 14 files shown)


hraberg:
It's worth noting that implementing a `Indexer`/`IndexStore` pair is currently quite hard, even if possible. Partly due to the fact that the current Indexer does too much, this should be split into to layers - with the upper layer being shared, dealing with the semantics of the operations, pre-conditions, ranges and transaction functions etc, while the lower layer being responsible for actually writing the data into the index.

Part of this can be facilitated by making `Snapshot+NewETXs` / `EntityHistory` protocol in `crux.tx` rely on the `IndexStore` instead of directly going to the kv store. This current PR does not yet deal with this, as there are other things necessary to split the `Indexer`. It is a stepping stone towards this split.

hraberg:
Practically, the content hashes could be found by scanning the right index instead of going to the kv store. The other parts of `EntityHistory` are already possible to move towards using the `IndexStore`.

There are also two steps at the end which writes to the `KvStore` - the first simply takes the `EntityTx`s and turns them into key-value pairs, this can easily be hidden in the lower protocol. Writing the meta data can be done via the indexer already, and finally, a method to write failed transactions could be added.

hraberg:
This PR now also includes the change to make the current `KvIndexer` rely more on the `IndexStore`, to simplify the future split of the actual indexer implementation and the common orchestration code.

hraberg:
This PR now also includes moving the boundary of the `Indexer` protocol so the consumer now contains most of the logic (though this still mainly sits in `crux.tx`, moving it around could be a later step). Changed some components which starts threads a bit taking the component itself in the consumption functions, this simplified some tests and give the argument an expected "type" (not just a dependencies map). Some parts of this isn't as clean as one would hope.

hraberg:
It's worth noting that there's a meta key used for the attributes, `:crux.kv/stats` that should in theory change name, but doing so would potentially require an index version bump, alternatively accepting losing earlier stats and queries changing their join orders to only take new documents into account. This key is never really exposed to the user as its hidden behind an API, so we can rename it as part of a later index bump.

crux-core/src/crux/object_store.clj
```diff
@@ -19,20 +19,26 @@
...
+        (db/get-single-object this snapshot k))))
 
   (get-objects [this snapshot ks]
-    (->> (for [k ks
-               :let [seek-k (c/encode-doc-key-to (.get i/seek-buffer-tl) (c/->id-buffer k))
-                     doc (some-> (kv/get-value snapshot seek-k) <-nippy-buffer)]
-               :when doc]
-           [k doc])
-         (into {})))
+    (if (satisfies? kv/KvSnapshot snapshot)
```
jonpither: satisfies is proved to be slow..

hraberg: As discussed, this is a known issue for now, if we run into noticeable performance issues we'll solve it. Eventually the need for this should go away if for example the object store goes and gets deprecated by the document store. Currently there's only one indexer - this check is there if you would use a new indexer together with the KvObjectStore.

PR #820, opened 2020-04-24
Add function to get entity-history (docs and diffs) and render

danmason:
Currently a draft, as I want to figure out how best to display this. You can request a full set of documents from an entity's history (in chronological order) using the `history=true` query parameter, which by default will display whole documents. To additionally request to show diffs between documents, you also pass `history-diffs=true`.


danmason:
How it currently appears (minus any styling):

---

**With documents:**

image


---

 **With diffs:**

image

danmason:
Closed for now - will be working based off of the new entity-history changes on a new PR.

PR #821, opened 2020-04-24
CAV index, optimising disk space

jarohen:
This PR refactors the indices in Crux to reduce the disk space usage (#465).

Currently, we have AVEC and AE covering indices, which result in a fair bit of duplication for documents that don't change significantly over time (e.g. the user frequently updates one or two keys from a larger document) - this PR reduces those to AVE and AE, and adds a CAV index for the bitemporal lookups.

Implementation notes (to be expanded)
* The object-store is now obsolete - objects can be reconstructed from the CAV index. That said, the standalone doc store uses the old object store implementation for its persistent record of the docs (outside of the Crux node's KV store), because we wanted the ability to blow the node's indices away and then reindex from the standalone doc store, and shredding these into CAV rather than just storing docs seemed like a waste of CPU time.
* Last time I profiled this branch (which, admittedly, was before #812), the ingest time was increased by 50-80% - I'm hoping that PR will have reduced that significantly. Will profile again before I even consider merging this


jarohen:
closing this in favour of some other investigation, see #465. will re-open if necessary

PR #823, opened 2020-04-28
tx-op validation without using spec

jarohen:
While I was profiling CAV, I found a significant proportion of the time to submit-tx was being taken by `clojure.spec.alpha/valid?` checking the tx-ops input by the user. Our tx-ops aren't particularly complicated, so I've hand-written the validation/checking logic.

Chances are the `submit-tx` thread isn't the critical path thread in any batch ingest - it's more likely to be the tx-consumer - but this change means that the `submit-tx` thread spends less time doing work and more time sleeping peacefully in `await-tx` (or, more likely, serving other requests).

On a micro-benchmark of the ts-devices ingest (50k docs):
- before the change: ~8.5s submit-tx, 3.5s await-tx
- after the change: ~2.5s submit-tx, 8s await-tx.

Related change: I've also made the remote API client rethrow `IllegalArgumentException`s, so that errors thrown locally and errors thrown from a remote client behave similarly.

### DIFF
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -17,58 +17,13 @@
 (s/def :crux.db.fn/body (s/cat :fn #{'fn}
                                :args (s/coll-of symbol? :kind vector? :min-count 1)
                                :body (s/* any?)))
-(s/def ::doc (s/and (s/map-of keyword? any?)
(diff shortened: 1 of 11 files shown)


hraberg (approved):
The case-statement in `->tx-event` could be replaced by making it also a multi method I think, more an observation than a necessary change.

jonpither (approved):

Issue #803, opened 2020-04-18
REST API /history/[:key] endpoint does not allow to pass a map as the :key


adam0z:
According to docs, this works:
```bash
curl -X POST $nodeURL/history/:ivan
```

Proposal:
Take data like POST /entity endpoint - in the request body:
```bash
curl -X POST \
     -H "Content-Type: application/edn" \
     -d '{:eid :ivan}' \
     $nodeURL/history
```

jarohen:
Thanks @adam0z for the report :) 

@luciodale, @danmason could you ensure that your updated `/_entity` endpoint accepts map keys as part of your work on #790? cheers!

PR #828, opened 2020-04-29
Crux Data Browser - Accept map keys

danmason:
The issue here has two parts to it - firstly, maps passed through URLs would be encoded, and there was no call to decode the URL before attempting to read as a Crux ID. Secondly, the `id-edn-reader` within `crux.codec` did not handle maps passed in as strings (there were `maybe-x-str` functions for the other types of IDs here, but none for maps). 

If there are issues to adding this to `crux.codec`, we could likely replace the `c/id-edn-reader` call in the handler with calls to `c/read-edn-string-with-readers` and `c/valid-id?`

### DIFF
--- a/crux-core/src/crux/codec.clj
+++ b/crux-core/src/crux/codec.clj
@@ -284,6 +284,12 @@
     (URL. s)
     (catch MalformedURLException _)))
 
+(defn- maybe-map-str [s]
+  (try
+    (let [edn (edn/read-string s)]
+      (when (map? edn) edn))
+    (catch Exception _)))
+
 (extend-protocol IdToBuffer
   (class (byte-array 0))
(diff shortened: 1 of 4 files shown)


jarohen (commented):
thanks @danmason - looks like a well-contained change. are there any tests that'd be suitable to adapt to show it working?

jarohen:
also, what'd be involved in making this work for #803?

danmason:
I think all that needs to be done for `/history` (and any others that get an **entity id** passed into the URL)  is decoding the URL before any call to converting the string into an EDN ID. As for tests that would be suitable to adapt for it - not sure, would need to check. Not sure how the remote API works, but I assume we could add a test to API test for a `history` call for an entity with a map key.

PR #829, opened 2020-04-29
Add `_status` endpoint to data browser handler

danmason:
Renders the HTML using the `entity->html` function, and makes a couple of minor changes to rendering.


jarohen (approved):
yep, no problem :)

does this still apply cleanly on master?

danmason:
Good question - will check after sorting out _query.

danmason:
(Would likely suggest merging this after query - would be easier to rebase this onto that than the other way around :) )

PR #834, opened 2020-04-30
Adding QueryEngine component

hraberg:
To scope config and resources, adding new bounde thread pool for the queries. Removing old CacheProvider interface, further removing direct dependencies on the kv store.


jarohen (approved):
as discussed, wouldn't mind us confirming the `IndexStore`/`Indexer`/`tx-consumer` names and responsibilities - the query engine having an 'indexer' when it doesn't write to it seems a little strange to me.

anyway, nothing blocking this one going in, IMO - let's get it merged and discuss naming later :smile: 

cheers @hraberg

jonpither (approved):

PR #836, opened 2020-05-01
Tidying up entity-history

jarohen:
Been tidying up the entity-history fns in index.clj in preparation for whatever changes we end up making to the entity-history API (no API changes in this PR).

* `idx/all-keys-in-prefix` now works in reverse order as well
* `idx/entity-history-seq-*` written in terms of `idx/all-keys-in-prefix` (`idx/entity-history-step` was pretty much doing this anyway)
* `idx/entity-history-seq-*` also now accept two bitemporal instants, and a `with-corrections?` flag
* removed some `idx/` fns that were only used in tests - these can also use `idx/entity-history-seq-*` now

RFC - regarding the range semantics: I've opted to have the minimum vt/tt as _exclusive_, maximum vt/tt as _inclusive_ - because of the common case of setting max-tt as the transaction time you've just got back from a transaction. This is intuitive for descending - start inclusive, end exclusive - but potentially counter-intuitive for ascending where it's start exclusive, end inclusive. I didn't want to go for either inclusive/inclusive, or having ascending not include the most recent transaction time. See the new `crux.tx-test/test-entity-history-seq-corner-cases` test for examples. Thoughts?

### DIFF
--- a/crux-core/src/crux/index.clj
+++ b/crux-core/src/crux/index.clj
@@ -564,13 +564,6 @@
            (not-empty)
            (map second)))))
 
-(defn all-entities [snapshot valid-time transact-time]
-  (with-open [i (kv/new-iterator snapshot)]
-    (let [eids (->> (all-keys-in-prefix i (c/encode-entity+vt+tt+tx-id-key-to (.get seek-buffer-tl)))
-                    (map (comp :eid c/decode-entity+vt+tt+tx-id-key-from))
-                    (distinct))]
-      (entities-at snapshot eids valid-time transact-time))))
-
 ;; TODO: Entity history would need to be able to use the Z order index
 ;; for us to be able to remove the old index. Outstanding questions
 ;; around how and if LITMAX would work together with seeks and prev
--- a/crux-core/src/crux/index.clj
+++ b/crux-core/src/crux/index.clj
@@ -569,34 +569,36 @@
   (-> (c/decode-entity+vt+tt+tx-id-key-from k)
       (enrich-entity-tx v)))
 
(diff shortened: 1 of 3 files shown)


crux-core/src/crux/index.clj
```diff
@@ -370,20 +370,21 @@
...
-       (let [k (f-cons)]
-         (when (and k (mem/buffers=? prefix k (mem/capacity prefix)))
-           (cons (if entries?
-                   [(mem/copy-to-unpooled-buffer k)
-                    (mem/copy-to-unpooled-buffer (kv/value i))]
-                   (mem/copy-to-unpooled-buffer k)) (step f-next f-next))))))
-    #(kv/seek i seek-k) #(kv/next i))))
+  ([i prefix] (all-keys-in-prefix i prefix prefix {}))
+  ([i prefix seek-k] (all-keys-in-prefix i prefix seek-k {}))
+
```
jonpither: blank line

crux-core/src/crux/index.clj
```diff
@@ -564,71 +565,48 @@
...
-     (->> (entity-history-seq i eid)
-          (take n)
-          (vec)))))
+(defn- ->entity-tx [[k v]]
+  (-> (c/decode-entity+vt+tt+tx-id-key-from k)
+      (enrich-entity-tx v)))
+
+(defn entity-history-seq-ascending
+  ([i eid] ([i eid] (entity-history-seq-ascending i eid {})))
+
```
jonpither: blank lines

jonpither (approved):

PR #839, opened 2020-05-04
Remove object store

hraberg:
Self explanatory. 

1. May have impact on ingest speed as it doesn't call straight into indexer anymore. 
2. KafkaDocumentStore has a known bug when fetching non-existing docs - this never happens during normal operation, but can occur due to the fact we expose the low-level document API, which is used in some tests. This could be removed.
3. Removes the sanity check of the doc store state from the tx consumer, as this is now not its business.
4. Lack of get-single-object leads to some verbose code in a few places, not refactored together.
5. As the document store doesn't use the snapshot concept, queries now lack snapshot visibility for documents. This is only an issue if someone evicts a document during a query, but I think this is OK.
6. Also, as there are no re-used snapshots, there might be some performance impact in places. My feeling is that this would be mitigated by moving the fast path value lookups onto the IndexStore protocol instead - this could also mitigate the snapshot visibility issue during evict.


hraberg:
Closing this, we'll revisit it in a slightly different manner together.

PR #841, opened 2020-05-04
Entity history API, #830

### LINKED ISSUE #830: Consistent Entity History API
Currently, we have four functions on the Crux API to get hold of history:

- `(crux/history node eid)`
  Eagerly returns all of the history of the given entity, with corrections, in index order.
- `(crux/history-range node eid vt-start tt-start vt-end tt-end)`
  Eagerly returns all of the history of the entity within the time co-ordinates, with corrections.
- `(crux/[open-,]history-seq-[a,de]scending db eid)`
  These four are on the DB, and also return the documents, but no corrections.

There's a couple of issues with this (beyond having a number of functions doing a similar but slightly different thing):
- history-seq-ascending on the DB starts from the valid-time/tx-time in the DB.
  This is the opposite behaviour of the other functions on the DB (for which the DB's vt/tt are upper bounds) - if I want the history since the start of time in ascending order, say, I have to get a DB at the start of time then call history-seq-ascending (and I can't then specify an upper bound without a `take-while`, or equivalent).
  More importantly, though, it also means that subsequent calls return more data, as the node indexes more transactions - the DB is not a value, in this case.
- (see #827) The `history-range` function on the node are vulnerable to the same issue as #546 - if I call it with a tt-end the node hasn't yet indexed, and then call it with the same parameters once the node's indexed the transaction, I'll get different results.

Broadly speaking, we're looking for a history API where users can:
- request data eagerly or lazily
- request corrections or no corrections
- request ascending or descending
- request docs or no docs

Ideally, this would also:
- follow the principles of 'db as a value' - i.e. if I call the same function with the same params on a db with the same params, I'm guaranteed to get the same result.
- return the same structure regardless of the flags the user supplies - although additive would also be fine (i.e. the difference between the user requesting docs and not requesting docs might be the addition of a `:crux/doc` key).

One such return value might be:
```clojure
[{:crux.db/valid-time ...
  :crux.tx/tx-time ...
  :crux.tx/tx-id ...
  :crux.db/content-hash ...
  :crux/doc {...}}
 ...]
```

Corrections, if requested, are included inline - the sequence is sorted first by valid-time, then tx-time.

Regarding _where_ this API lives - given we'd like to preserve both the DB's properties (repeatable queries, and ensuring that the node has indexed all of the transactions), we'd prefer to put this function on the DB. We'll accept `:start` (bitemp co-ordinate, inclusive) and `:end` (bitemp co-ordinate, exclusive) maps on the entity history call. This does have the tradeoff that the DB already contains one vt/tt - for the history API, we'll use this vt/tt as an upper bound to whatever the user provides (to the `:start`, in a descending history, and the `:end` in an ascending history). Separately, we'd also appreciate it if `(crux/db node ...)` would take a map of vt + tt rather than positional params (because it's valid to pass tt but not vt) - one for another day.

With this in mind - we'd like to keep the `entity-history` signature as close as possible to how DB _currently_ works (and then, if we change DB later, we can change this too). Proposal:

```clojure
(crux/entity-history db eid :asc
                     {:with-docs? true
                      :with-corrections? true
                      :start {:crux.db/valid-time ...
                              :crux.tx/tx-time ...}
                      :end {:crux.db/valid-time ...
                            :crux.tx/tx-time ...}})
```

We'd also add `open-entity-history`, with lazy semantics.

A nuanced difference here is that the behaviour of the order-ascending request changes to fix the above issue - this would return the history from the `:start` vt/tt provided up to the `:end` vt/tt, or the vt/tt of the DB, whichever's earlier.

Thoughts?

jarohen:
resolves #830 

This PR depends on #840 - to see the change on top, see the most recent two commits: https://github.com/juxt/crux/pull/841/files/91e517a66f5d22564eea3acf4a2e731d54bcfb7e..4fa9f0ed7e85bbf2eb7f681ce7eb27fcdfb624fc

I implemented it first on the node as suggested, and it worked fine, passed tests, etc. Trouble is - allowing the user to specify arbitrary maximum tx-times means that the result of calling the function with the same parameters at two different times is subject to what transactions the node has indexed in the meantime - and it's _this_ that I now think is the real value of the DB as a separate entity: not just a store of a bitemporal co-ordinate, but a bound on what transactions can possibly influence the result of any functions you call on it.

The way I've implemented now is to have the entity-history on the DB, but to allow the user to also specify min/max vt/tt to the function itself. Perfectly fine to elide max on the call and use the value from the DB; the semantics in that case are unambiguous. RFC?

This also demo's a pattern I'm considering to consolidate the Java APIs:
- The node object that the Java and Clojure API's use should both be the _same_ object - so that extensions can be written in either Java or Clojure. (with crux.api.alpha, we wrap the Clojure node, which means that extensions written by Java authors won't work in Clojure applications)
- The Java API uses strongly typed objects where possible (in this case, `HistoryOptions`), so that IDEs can auto-complete the various options. Particularly, they don't use any Clojure-specific data structures - either as parameters or return values.
- The Clojure node then implements the Java API directly (no change here), using the strong types - with the trade-off that this is quite an interop-heavy implementation.
- The Clojure protocol exposes Clojure data-types instead - the `extend-protocol` definition between the Clojure protocol and the Java interface is where we adapt both parameters and return values.
- This may mean that we do our computation in Clojure data types, convert to Java in the interface, and back to Clojure in the protocol, which may turn out to be an intolerable overhead.

TODO:
* [ ] docstrings/javadocs/etc

### DIFF
--- a/crux-bench/src/crux/bench/ts_devices.clj
+++ b/crux-bench/src/crux/bench/ts_devices.clj
@@ -223,7 +223,7 @@
                                         (reduce into []))
                        db (crux/db node #inst "1970")
                        histories (for [r reading-ids]
-                                   (crux/open-history-ascending db r))]
+                                   (crux/open-entity-history db r :asc))]
                    (try
                      (->> (for [history histories]
                             (for [entity-tx history]
--- a/crux-bench/src/crux/bench/ts_weather.clj
+++ b/crux-bench/src/crux/bench/ts_weather.clj
@@ -298,7 +298,7 @@
                                           (reduce into []))
                        db (api/db node #inst "1970")
                        histories (for [c condition-ids]
-                                   (api/open-history-ascending db c))]
+                                   (api/open-entity-history db c :asc))]
                    (try
                      (->> (for [history histories]
                             (for [entity-tx history]
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -5,7 +5,8 @@
             [crux.codec :as c]
             [clojure.tools.logging :as log])
   (:import [crux.api Crux ICruxAPI ICruxIngestAPI
-            ICruxAsyncIngestAPI ICruxDatasource ICursor]
+            ICruxAsyncIngestAPI ICruxDatasource ICursor
+            HistoryOptions HistoryOptions$SortOrder]
            java.io.Closeable
            java.util.Date
            java.time.Duration
@@ -163,6 +164,19 @@
 
   Returns an iterator of the TxLog"))
 
+(defn- ->HistoryOptions [sort-order
+                         {:keys [with-corrections? with-docs? start end]
+                          :or {with-corrections? false, with-docs false}}]
+  (HistoryOptions. (case sort-order
+                     :asc HistoryOptions$SortOrder/ASC
+                     :desc HistoryOptions$SortOrder/DESC)
+                   (boolean with-corrections?)
+                   (boolean with-docs?)
+                   (:crux.db/valid-time start)
+                   (:crux.tx/tx-time start)
+                   (:crux.db/valid-time end)
+                   (:crux.tx/tx-time end)))
+
 (extend-protocol PCruxNode
   ICruxAPI
   (db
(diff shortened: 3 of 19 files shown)


hraberg (approved):
LGTM, maybe except the missing Javadocs. Could also deprecate the other history APIs on ICruxAPI and ICruxDatasource?

PR #842, opened 2020-05-05
Replacing CachedIterator closed-state with an AtomicBoolean

hraberg:
About 10% faster than plain Clojure atom in `crux.kv-test/test-performance-off-heap`.
40% faster if only comparing opening and closing iterators. Should functionally be the same.

Note, the need for thread safety here is debatable, these iterators are only used from a thread at a time except if the system closes while used. This uses compare and swap as if the pool itself might be used from many threads, which it never is. The flag and lock would still be needed, due to the closing of the system. This PR doesn't deal with this, it just replaces the atom a with an AtomicBoolean.


jarohen (approved):
:ok_hand:

Issue #603, opened 2020-01-28
Crux shouldn't need to go to the object store for primitive-like attributes


jarohen:
Primitive values are already reversibly encoded in the triple indices, so we needn't go to the object store to decode them (especially if the object store moves off-node)

hraberg:
Copied from #843 :

It's worth noting that chars will be decoded to single char strings using this, which would change the behaviour in the -unlikely- case someone stored a document with a character in it. If we want to support this properly we need to introduce a special type id for char so we can detect it and turn it back into a char.

On a similar note, Crux encodes say byte, short, int and long all as longs, similarly float and double. This will decode long and doubles only, but this is unlikely to be an issue as this is semi-normal Clojure behaviour.

jarohen:
this is coming as part of #868, will close

PR #843, opened 2020-05-05
Decode primitive values

hraberg:
Running `curx.query-test` and comparing the time it takes to use this decoder for primitive values where possible instead of going to the document is about 20% for these cases, which are about 1/3 of all value accesses in the test. This does not show in the total runtime of the test as it does much more. A more targeted query test would likely show the improvement better.

This PR is mainly to implement and test the decoder, but also plugs it in into `crux.query` in `bound-results-for-var`. It does not yet try to move this logic into the `IndexStore`. (This isn't true anymore, the latter commits do this.)

### DIFF
--- a/crux-core/src/crux/tx.clj
+++ b/crux-core/src/crux/tx.clj
@@ -61,14 +61,14 @@
 (defn tx-event->doc-hashes [tx-event]
   (keep (conform-tx-event tx-event) [:content-hash :old-content-hash :new-content-hash :args-content-hash]))
 
-(defn tx-event->tx-op [[op id & args] index-store object-store]
+(defn tx-event->tx-op [[op id & args] index-store]
   (into [op]
         (concat (when (contains? #{:crux.tx/delete :crux.tx/evict :crux.tx/fn} op)
                   [(c/new-id id)])
 
                 (for [arg args]
                   (or (when (satisfies? c/IdToBuffer arg)
-                        (or (db/get-single-object object-store index-store arg)
+                        (or (db/get-document index-store arg)
                             {:crux.db/id (c/new-id id)
                              :crux.db/evicted? true}))
                       arg)))))
@@ -196,7 +196,7 @@
   {:pre-commit-fn (fn []
                     (let [content-hash (c/new-id v)]
                       (with-open [index-store (db/open-index-store indexer)]
-                        (assert (db/get-single-object object-store index-store content-hash)
+                        (assert (db/get-document index-store content-hash)
                                 (format "Put, incorrect doc state for: '%s', tx-id '%s'"
                                         content-hash (:crux.tx/tx-id tx))))
                       true))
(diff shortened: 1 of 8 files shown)


hraberg:
Note, this PR runs into the same `or` issue in `crux.lubm-test` that occasionally has a `nil` value buffer. It's mitigated by a check, doesn't try to solve the root cause (which was already there).

hraberg:
It's worth noting that chars will be decoded to single char strings using this, which would change the behaviour in the -unlikely- case someone stored a document with a character in it. If we want to support this properly we need to introduce a special type id for char so we can detect it and turn it back into a char.

On a similar note, Crux encodes say byte, short, int and long all as longs, similarly float and double. This will decode long and doubles only, but this is unlikely to be an issue as this is semi-normal Clojure behaviour.

hraberg:
This PR now contains a few valid cut-off points, building on each other:

- the basics.
- adding support for `nil`, `true` and `false`. Some additional refactoring and speed ups.
- extending `IndexStore` with `get-document` and `decode-value`, removing dependency on `ObjectStore` from the read side.
- using direct call to capacity on `DirectBuffer` instead of protocol, 4x faster. This is independent of the rest.

hraberg:
Related to #603.

jonpither (approved):

jarohen (approved):
lgtm, will merge :)

crux-core/src/crux/tx.clj
```diff
@@ -357,10 +365,32 @@
...
+            value-or-values (get doc a)]
+        (cond
+          (not (idx/multiple-values? value-or-values))
+          value-or-values
+
+          (nil? value-buffer)
+          (first (idx/vectorize-value value-or-values))
+
+          :else
+          (loop [[x & xs] (idx/vectorize-value value-or-values)]
```
jarohen, line 381: OOI - is this much quicker than `(first (filter ...))`?

crux-core/test/crux/codec_test.clj
```diff
@@ -58,3 +65,25 @@
...
+                    (t/is (if (and (double? v) (Double/isNaN v))
+                            (Double/isNaN (c/decode-value-buffer buffer))
+                            (= v (c/decode-value-buffer buffer)))
+                          (str (pr-str v) " " (class v)))
+                    (t/is (and (string? v)
+                               (> (count v) @#'c/max-string-index-length)
+                               (= @#'c/object-value-type-id
+                                  (.getByte (c/value-buffer-type-id buffer) 0))))))))
+
+(defn with-silent-test-check [f]
```
jarohen: will move this to `crux.fixtures` now that it's in crux-test/src

crux-core/src/crux/tx.clj
hraberg: Unsure, this code was already there, it was moved from `crux.query/bound-results-for-var`, but did wonder about its performance characteristics when moving it.

PR #847, opened 2020-05-05
rename Java `q` method to `query`, abbrevs aren't so common in Java

jarohen:


hraberg (approved):
I want to applaud the recursion of using an abbreviation in the title of this card :-)

hraberg:
Approved before seeing the RemoteDatasource issue in the Circle failure above, but assume that's just a missed rename and easy to fix.

PR #848, opened 2020-05-06
refactor common fixtures to one namespace `crux.fixtures` in crux-test/src

jarohen:
Refactor common fixtures to one namespace `crux.fixtures` in crux-test/src, so they can be reused easily.

I've put the common ones (crux.fixtures.api/crux.fixtures.standalone, and crux.fixtures.kv/with-kv-dir) all under the existing `crux.fixtures` namespace, so that we can just require that one in to individual modules and get started. 

Have marked the original ones as deprecated and have started migrating across usages when I see them. RFC?


hraberg (approved):

crux-test/test/crux/fixtures/standalone.clj
```diff
@@ -1,16 +1,11 @@
 (ns crux.fixtures.standalone
-  (:require [crux.fixtures.api :as fapi]
+  (:require [crux.fixtures :as fix]
```
jonpither: I would just kill this ns rather than deprecate.

crux-test/src/crux/fixtures.clj
```diff
@@ -22,43 +70,21 @@
...
-                        :sex       (rand-nth [:male :female])
-                        :age       (rand-int 100)
-                        :salary    (rand-int 100000)})
+(defn random-person []
+  {:crux.db/id (UUID/randomUUID)
+   :name (rand-nth ["Ivan" "Petr" "Sergei" "Oleg" "Yuri" "Dmitry" "Fedor" "Denis"])
+   :last-name (rand-nth ["Ivanov" "Petrov" "Sidorov" "Kovalev" "Kuznetsov" "Voronoi"])
+   :sex (rand-nth [:male :female])
+   :age (rand-int 100)
+   :salary (rand-int 100000)})
```
jonpither, line 87: At some stage we should look to see if `random-person` and `transact!` are stilling pulling their weight, if not, kill.

crux-test/test/crux/api_test.clj
```diff
@@ -25,15 +21,23 @@
...
+                                 fix/with-kv-dir
+                                 fh/with-http-server
+                                 fix/with-node
+                                 fh/with-http-client])
+       :h2 (t/join-fixtures [#(fj/with-jdbc-node :h2 %) fix/with-kv-dir fix/with-node])
+       :sqlite (t/join-fixtures [#(fj/with-jdbc-node :sqlite %) fix/with-kv-dir fix/with-node])
+       :local-kafka (-> (t/join-fixtures [fk/with-cluster-node-opts fix/with-kv-dir fix/with-node])
                         (with-meta {::embedded-kafka? true}))
-       :kafka+remote-doc-store (-> (t/join-fixtures [fk/with-cluster-node-opts fs/with-standalone-doc-store kvf/with-kv-dir fapi/with-node])
+       :kafka+remote-doc-store (-> (t/join-fixtures [fk/with-cluster-node-opts
```
jonpither, line 34: In the future, I'd like to consider the api-test being a fixture and to be used by the kafka module, therefore moving all kafka related stuff out of `crux-test` into `crux-kafka`.

jonpither (approved):
Approving, but I would personally delete the deprecated ns.

crux-test/test/crux/api_test.clj
jarohen: agreed :)

crux-test/src/crux/fixtures.clj
jarohen: agreed :)

crux-test/test/crux/fixtures/standalone.clj
jarohen: will do

PR #851, opened 2020-05-06
Adding new direct ByteBuffer based RocksDB kv store

### LINKED ISSUE #642: Migrate to RocksDB ByteBuffer APIs when Released
Direct ByteBuffer support is finally coming to RocksDB JNI:
https://github.com/facebook/rocksdb/pull/2283

Once this is released we want to migrate our implementation to use it. We should then also kill off the JNR version of Rocks we have under `crux.rocksdb.kv.jnr`.

And as it will be moving in that direction we can then also potentially consolidate the LMDB side to use ByteBuffers only. The default version does, but the lmdbjava version could also be made do so, depends on if we want to keep LWJGL dependency or not.

Later we can potentially get rid of Agrona and JNR completely, but that's another card. Hashing would still use it, but JNR can work with Pointers and direct ByteBuffers if we want to keep it. A potential option is to get rid of the JNR hashing completely for simplicity, but it's there for performance reasons (~10%).

hraberg:
Lives under crux.kv.rocksdb.nio, may replace the others.


hraberg:
crux.kv.rocksdb/kv
"Elapsed time: 18405.202506 msecs"
crux.kv.rocksdb.jnr/kv
"Elapsed time: 12024.220405 msecs"
crux.kv.rocksdb.nio/kv
"Elapsed time: 14550.30613 msecs"

hraberg:
Also removes the usage of pinned slices in JNR the Rocks backend.

hraberg:
Updated this PR to not introduce a new backend, but change the behaviour of the current one.

jonpither (approved):

PR #861, opened 2020-05-11
Encode values

hraberg:
Moving the responsibility of encoding values to the `IndexStore`, so it can chose the internal representation. The query engine now makes two assumptions of the buffers returned from `encode-value`:

1. The byte representations are sortable as unsigned big endian, so range searches depend on this.
2. The first byte represents the type id. This id is only used as a prefix for range searches and its values is never inspected.

Naturally, the existing `IndexStore` `decode-value` method takes a buffer respecting the above and returns the corresponding Java object representation.

The range searches will likely be pushed down into the `IndexStore` itself later, then the second assumption around type ids will be removed.

Ids in the index are mainly treated as buffers to be joined on, so they don't have to be stored as hashes, but there might be some places that still makes this assumption. Externally, for content hashes and entity ids on the TxLog, the hashes will still be used.

### DIFF
--- a/crux-core/src/crux/db.clj
+++ b/crux-core/src/crux/db.clj
@@ -34,7 +34,8 @@
   (entity-history-range [this eid valid-time-start transaction-time-start valid-time-end transaction-time-end])
   (entity-history ^crux.api.ICursor [this eid sort-order opts])
   (all-content-hashes [this eid])
-  (decode-value [this a content-hash value])
+  (decode-value [this a content-hash value-buffer])
+  (encode-value [this value])
   (get-document [this content-hash])
   (open-nested-index-store ^java.io.Closeable [this]))
 ;; end::IndexStore[]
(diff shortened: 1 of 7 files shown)


hraberg:
In theory, the `Indexer` could be responsible for encoding values as it's what writes them, but most code in the QueryEngine currently deals only with the `IndexStore`, so did put the method there. This method is also really for the read side when it needs create values compatible with what's stored in the index, and the `Indexer` won't likely use this method internally.

That said, the `Indexer` is available to the engine, as it's what's used to open the store in the first place.

jonpither (approved):

hraberg:
Further on the above, the `Indexer` may chose to represent its internal storage differently, the buffers returned from `encode-value` only needs to be compatible with the buffers returned from the `IndexStore`, which might not at all be what's happening behind the scenes.

hraberg:
There are places in `crux.query` that makes assumptions about the ids:

1. In the spec, when checking if an entity literal actually is a valid id, this could simply be dropped if we relax this.
2. In full results, when deciding when to look up a value as an entity. This check could also in theory be relaxed, but at the cost of trying to lookup every value as an entity. Potentially could be mitigated by caching all values, not just entities in a second level cache during the walk..

jarohen (approved):

PR #862, opened 2020-05-11
ability to open HTTP nodes in read-only mode, #854

### LINKED ISSUE #854: Crux HTTP nodes can be started in read-only mode
Supplied as a `:crux.http-server?` flag to the HTTP server module

jarohen:
resolves #854 

supply `:crux.http-server/read-only?` to the server-side topology to prevent submitting transactions via the HTTP node


danmason (approved):
Looks good to me!

Issue #867, opened 2020-05-13
Clarifying the index-store boundary


jarohen:
options to be documented here

hraberg:
We need to do something like this:

* User wants the full document for an entity at a given vt/tt
* Given an A, give me all the entities with that attribute (AEV-E)
* Given an AE, give me the current values (AEV-V)
* Given an A, give me all the Vs (AVE-V)
* Given an AV, give me all the current Es (AVE-E)
* Given an E, give me the ETx at a given vt/tt (entity-as-of)
* Given an ETx, give me the decoded V for a given A (bound-result-for-var)

Later we also want to decide if filters are per triple or per entity (document). And can bitemp be implemented as a normal filter or not?

In this design, where the IndexStore provides the services above, it would not return these as Indexes, but let the query execution engine deal with that and delegate to these functions.

In the IndexStore, they would need to provide seqs (or similar) starting at a prefix, the engine can then operate on these. We probably also want to push range filters down to the index store as well.

hraberg:
Related to #868.

PR #872, opened 2020-05-14
Remove EntityAsOfIndex and replace with entity-as-of method on IndexStore

hraberg:
Self-explanatory. Relies on caching of iterators and normally the resolved entity itself, as it creates and closes a new iterator internally every time instead of being a "resource" that's passed around.


hraberg:
Related to #867.

hraberg:
Branch also removes a few tests / micro benchmarks related to the entity cache, some that were living under `dev`.

jarohen (approved):

PR #879, opened 2020-05-19
Ensuring join cannot be smaller than the result index. Fixes #845.

### LINKED ISSUE #845: test-lubm-query 04 binds var at wrong level
There's a var, `t`, in `crux.lubm-test/test-lubm-query-04` that's accessed before its bound, this happens to work anyway as the entity `x` is already bound.

In `crux.query/bound-results-for-var` this happens here:
https://github.com/juxt/crux/blob/master/crux-core/src/crux/query.clj#L543

There's no known user facing bug caused by this, but it hints at a bug somewhere in the query engine. It might also be "OK" but some analysis would be required to judge this. The current code deals with this case with a `nil?` check, but it looks like a workaround to me.

I've encountered it twice last few days, so might be worth a look. Second time involved trying to replace `get` with a direct call to `.nth` which is faster as we know `join-keys` is a vector.

hraberg:


hraberg:
This fix actually messes with unification constraints a bit, but don't think they're really worth the complication. It works, but will have lost some of the perf benefit. Range constraints can be used instead. We should consider deprecating the unifcation stuff, `==` also clashes with the Clojure numeric equality operator.

hraberg:
Made fix smaller, avoiding to change the unification logic for now.

PR #880, opened 2020-05-19
Clarify index store boundary

### LINKED ISSUE #867: Clarifying the index-store boundary

hraberg:
This changes the IndexStore protocol to not return `db/Index` instances, but instead provide lower level methods returning seqs:

```
  (av [this a min-v entity-resolver-fn])
  (ave [this a v min-e entity-resolver-fn])
  (ae [this a min-e entity-resolver-fn])
  (aev [this a e min-v entity-resolver-fn])
```

The join engine has been refactored to deal with this. The above methods are taking in the entity-resolver-fn, but could instead take the bitemporal coordinates and let the IndexStore create a cached resolver internally for them.

The methods could be renamed and consolidated down to 2 or even 1, taking some form of options map.

### DIFF
--- a/crux-core/src/crux/db.clj
+++ b/crux-core/src/crux/db.clj
@@ -28,8 +28,10 @@
 
 ;; tag::IndexStore[]
 (defprotocol IndexStore
-  (new-attribute-value-entity-index-pair [this a entity-resolver])
-  (new-attribute-entity-value-index-pair [this a entity-resolver])
+  (av [this a min-v entity-resolver-fn])
+  (ave [this a v min-e entity-resolver-fn])
+  (ae [this a min-e entity-resolver-fn])
+  (aev [this a e min-v entity-resolver-fn])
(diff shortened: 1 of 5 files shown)


hraberg:
By splitting the logic here the upper levels are now much easier to reason about, those Index implementations share a lot of overlap and could be further simplified, but the important part of this PR is to create the boundary in the first place.

hraberg:
Current implementation is about 50% slower compared to master when running LUBM. A bit slower is OK, but not this much, haven't profiled yet. First cut has not been written with performance in mind.

hraberg:
Made the performance a bit better as the lazy-seq boundary was in the wrong place making it eagerly created the next value. Still slower than master.

hraberg:
This is now similar in performance running LUMB on master, after adding recycled nested index stores, which are really just a way of scoping resources (in this case, iterators).

jarohen (approved):

PR #881, opened 2020-05-19
Transaction functions replacing their arg docs, #866

### LINKED ISSUE #866: Transaction functions replace the arg doc with the result of the fn
Currently, transaction functions rely on an arg-doc containing the arguments for the function. This document, though, can't  be evicted when any of the documents contained in the args are evicted.

We can use the fact that transaction functions are pure to replace the args doc with the results of the transaction function, so that it looks (as far as possible) like any other transaction.

* When a node first sees the transaction fn arg doc it executes the function, giving it a list of operations
* The node then replaces the arg doc in the doc store with the list of operations, and adds the related documents to the doc store
* Any node that subsequently comes across this transaction sees an arg doc with just a list of operations - it plays these operations as it would any other transaction.
* There shouldn't be anything apart from transaction operations in the arg doc - so that this document never needs to be evicted (it has no PII)
* We can then expose a nicer UX to the user calling the transaction function - the more this can look like a normal Clojure call, the better
* Both `tx-log` and the transaction events on the public bus will need to be adapted (to make the view across doc-based operations seamless, or potentially/optionally indicate that they were generated by a transaction function)
* Failed transaction function operations (either through the functions returning a fail condition, or because of a sibling operation) should have their arguments automatically removed from the doc store

jarohen:
* This is now based on the doc-index-removal branch, #893, because they were conflicting - need to merge that one before this one.
* Have squashed the commits down - this tx-fn change is just the latest commit in this PR

Transaction functions now replace their arg docs when they've evaluated the function, so that the resultant documents can be evicted, and there's no PII in tx-fn args lingering in the doc-store.

Have also updated the syntax for calling transaction functions - previously you had to call them with an explicit argument document - now they're called with arguments passed as part of the tx-op:

```clojure
;; previously:
[:crux.tx/fn :the-fn {:crux.db/id :the-args, :crux.db.fn/args [arg1 arg2 arg3]}]

;; now
[:crux.tx/fn :the-fn arg1 arg2 arg3]
```

I briefly considered making a similar change to creating functions so that users don't need to know about `:crux.db.fn/body` - RFC?

How this works:
* The tx-consumer inlines the argument document in the tx-fn call tx-op, so that the arg docs don't need to be indexed
* The tx-fn `index-tx-event` implementation checks the arg doc to see whether it's a doc that's already been evaluated - if it hasn't, it'll see `:crux.db.fn/args`; if it has, it'll see `:crux.db.fn/tx-events` (or `:crux.db.fn/failed? true`)
* If it needs to evaluate the function, it does so, and responds with the docs to overwrite - this adapts the existing evict tombstone flow to both update the args doc and to submit any new documents in the resulting tx-ops. 
* If the transaction function fails, we still need to overwrite the args doc (this time with the `:failed? true`), a slight difference to the evict tombstones flow (which only updated documents if the transaction committed)
* (aside) we no longer need to store content-hashes in the Kafka tx-topic message metadata, because this is done by the tx-consumer for all tx-logs

TODO:
- [x] two failing tests on CI that I can't repro locally

### DIFF
--- a/crux-bench/test/crux/bench/rocksdb_microbench_test.clj
+++ b/crux-bench/test/crux/bench/rocksdb_microbench_test.clj
@@ -30,7 +30,4 @@
             (time
              (doseq [doc-batch (->> (take 10000 condition-docs)
                                     (partition-all 100))]
-               (db/index-docs (:indexer node) (->> doc-batch (into {} (map (juxt c/new-id identity)))))))
-
-            (with-open [snapshot (kv/new-snapshot (:kv-store node))]
-              (t/is (= first-doc (db/get-single-object (:object-store node) snapshot (c/new-id first-doc)))))))))))
+               (db/index-docs (:indexer node) (->> doc-batch (into {} (map (juxt c/new-id identity)))))))))))))
--- a/crux-core/src/crux/codec.clj
+++ b/crux-core/src/crux/codec.clj
@@ -26,7 +26,7 @@
 
 (def ^:const index-id-size Byte/BYTES)
 
-;; index for actual document store
+;; index for object store
 (def ^:const ^:private content-hash->doc-index-id 0)
 
 (def ^:const ave-index-id 1)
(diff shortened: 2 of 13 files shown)


hraberg (approved):
LGTM, but depends on #893 to be merged, where there are some outstanding questions.

PR #883, opened 2020-05-20
Moving kv-indexer impl into crux.kv-indexer

jarohen:


hraberg (approved):
There are a few more parts of crux.index which could be moved, for example the index version checks are really for the KvIndexer, not the KvStore in general (but get that this change might require further thinking). Similarly, read/store-meta could be directly in-lined into the KvIndexer. Any object store stuff (or doc store) could also be moved out. That is, at the end of the full move, we don't want crux.index to refer to crux.kv.

That said, this a step in the right direction, LGTM!

PR #885, opened 2020-05-20
Support falsey values in query :args

### LINKED ISSUE #882: Unused arg bound to false causes query to not return any results
(reported by Toyam Cox (@Vaelatern) on [Slack](https://clojurians.slack.com/archives/CG3AM2F7V/p1589946575361300))

* Including an unused argument with a value of `true` returns results (correctly?)
* Including an unused argument with a value of `false` no longer returns results (bug)

```clojure
dxpb-clj.db=> (crux/q (crux/db  @node) {:find ['e] :where '[[e :pkgname a]] :args [{'a "st"}]})
#{[:st:x86_64:x86_64:cross:false] [:st:x86_64:x86_64-musl:cross:false]}
dxpb-clj.db=> (crux/q (crux/db  @node) {:find ['e] :where '[[e :pkgname a]] :args [{'a "st" 'c false}]})
#{}
dxpb-clj.db=> (crux/q (crux/db  @node) {:find ['e] :where '[[e :pkgname a]] :args [{'a "st" 'c true}]})
#{[:st:x86_64:x86_64:cross:false] [:st:x86_64:x86_64-musl:cross:false]}
```

SevereOverfl0w:
This allows for things like :args [{'c false}], which currently don't
match anything.

This is related to #882 which is about 'c being unused (but _doesn't_ fix it).  This only
fixes what @Vaelatern actually intended, which is to match when `:cross`
was false.

### DIFF
--- a/crux-core/src/crux/query.clj
+++ b/crux-core/src/crux/query.clj
@@ -269,8 +269,9 @@
          (into {}))))
 
 (defn- arg-for-var [arg var]
-  (or (get arg (symbol (name var)))
-      (get arg (keyword (name var)))))
+  (second
+    (or (find arg (symbol (name var)))
+        (find arg (keyword (name var))))))
 
 (defn- update-binary-index! [index-store {:keys [entity-resolver-fn]} binary-idx vars-in-join-order var->range-constraints]
   (let [{:keys [clause names]} (meta binary-idx)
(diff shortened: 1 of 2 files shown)


jarohen:
hey @SevereOverfl0w, thanks for the PR :) 

Have added a test to confirm the fix for `false`-valued args, and also added a test for the separate issue raised in #882. Will continue tracking the latter issue.

James

jarohen:
cheers @SevereOverfl0w, merged

Issue #868, opened 2020-05-13
Clarifying the on-disk index structure


jarohen:
Context: we're currently debating the extent of the Indexer/IndexStore protocols (see #867)
Will keep this within the _current_ protocol.

The IndexStore exposes three index-related fns: AVE, AEV, and entity-as-of.

We need to support the following use cases efficiently:

1. User wants the full document for an entity at a given vt/tt
2. Given an A, give me all the entities with that attribute (AEV-E)
3. Given an AE, give me the current values (AEV-V)
4. Given an A, give me all the Vs (AVE-V)
5. Given an AV, give me all the current Es (AVE-E)
6. Given an E, give me the ETx at a given vt/tt (entity-as-of)
7. Given an ETx, give me the decoded V for a given A (bound-result-for-var)
8. Given a transaction, for each E, ensure all the documents have been indexed before I mark the transaction committed
9. Evict me an entity

## Current indices:

* EVT -> C
* AECV
* AVEC
* C -> Doc (object store)

With these indices, there's a lot of duplication in both the content indices.
For documents that have lots of incremental changes over time, we store all of the AV pairs every time.

### 'Removing C' (#754)

* EVT -> C
* AE
* AVE
* C -> Doc (object store)

Worked well at reducing disk space (~60-70%) and saw a corresponding decrease in ingest time (purely because there was less to ingest, presumably). Query time was bad, though, 80% increase in Watdiv - caused by needing to go to the object store much more frequently to check the current AVs for an entity.

### 'CAV' (#821)

* EVT -> C
* AE
* AVE
* CAV -> decoded V

Worked reasonably well at reducing the disk space (~40%) while keeping the query time down (Watdiv was comparable). Ingest time was 20% higher, though, and there was still plenty of duplication in the CAV index.

## Next iteration

(proposal)

Adding another level of indirection - a 'hash cache' - storing decoded values where we can't immediately decode them from the values in the index keys.

* `EVT -> C` (unchanged)
* `EV -> decoded V` (only for V not decodeable from the value)
* `AECV`
* `AVE`

Notes:

* We're ensuring that everything the query engine needs is local; all of the other APIs (entity, full-results, history docs, etc) can go to the doc-store
* No (on-disk) object store

Running through those reqmts again:

1. User wants the full document for an entity at a given vt/tt: EVT -> C, doc-store
2. Given an A, give me all the entities with that attribute (AEV-E): scan `AEV`
[...]

PR #886, opened 2020-05-21
Remove get-document from IndexStore

jarohen:
Initial PR to remove get-document from IndexStore. Most places now use the document-store where possible, but there are a few places that will eventually use the hash-cache - I've inlined the current object-store code into that with a view to now replace the object-store with the hash-cache.

first part of #868


jarohen:
I'll submit this as part of a wider doc-index removal PR

PR #887, opened 2020-05-21
Add wrapper to handle JWTs

### LINKED ISSUE #889: Users can authenticate to the HTTP API using JWT tokens
Currently, the HTTP API is open access, relying on the administrator to ensure it's not publicly accessible via networking - we'd like administrators to be able to restrict access based on the user's credentials.

* Administrators that want to limit access to their node's HTTP API can supply a JSON Web Key Set (JWKS) as part of the node config. Nodes with this config supplied should restrict access; nodes without it should remain open access. (tbc: how should a user provide this?)
* To authenticate, users can pass a bearer JWT token to the `authorization` header, in the form `Bearer <token>`
* Additionally, the node should look for a JWT in the AWS Cognito `x-amzn-oidc-data` header.
* If the JWT token is successfully validated against the provided JWKS, the request should proceed normally; otherwise, it should return a 401 Unauthorized response.
* Users of the remote API client should be able to provide an access token `Supplier` to the node. The remote API client should cache the access token so long as it remains valid, and call the `Supplier` again when the token expires.
* As a first pass, we'll say that any user who has successfully authenticated is fully authorized to perform any action.

danmason:
Resolves #889

### DIFF
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -6,11 +6,12 @@
             [clojure.tools.logging :as log])
   (:import [crux.api Crux ICruxAPI ICruxIngestAPI
             ICruxAsyncIngestAPI ICruxDatasource ICursor
-            HistoryOptions HistoryOptions$SortOrder]
+            HistoryOptions HistoryOptions$SortOrder
+            RemoteClientOptions]
            java.io.Closeable
            java.util.Date
            java.time.Duration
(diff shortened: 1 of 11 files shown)


danmason:
- Use HistoryOptions-esque pattern for passing options into the new-api-client call.
- Make clear you are supplying JWTs
- api-request-sync take a map with body as a key (rather than multi arity function)

PR #888, opened 2020-05-21
Fix falsy unbound args #882

### LINKED ISSUE #882: Unused arg bound to false causes query to not return any results

jarohen:
based on #885, will merge that first. resolves #882.

In this case, the unary index wasn't passing through tuples from the inner index where the value was falsy - by checking whether the _buffer_ is nil instead, we ensure false/nil values are passed through correctly.

### DIFF
--- a/crux-core/src/crux/index.clj
+++ b/crux-core/src/crux/index.clj
@@ -417,7 +417,7 @@
     (UnaryJoinIteratorState.
      idx
      (or value c/empty-buffer)
-     (when (and result-name results)
+     (when (and result-name value)
        {result-name results}))))
 
 (defrecord UnaryJoinVirtualIndex [indexes ^UnaryJoinIteratorsThunkFnState state]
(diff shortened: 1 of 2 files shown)


hraberg (approved):
Approving, assume this implies that there is a result, represented as a value buffer, but the value payload happens to be falsely? If so, the logic seems good to me!

PR #893, opened 2020-05-27
Removing the doc index, #868

### LINKED ISSUE #868: Clarifying the on-disk index structure

jarohen:
resolves #868, see that card for details of the new indices

* The kv-indexer no longer uses the object store. For most operations, this was relatively straightforward
  * For `decode-value` we go to the hash-cache
  * For eviction we go through all the attributes, find all the content hashes for all of the entities, and then evict using those. NB that we can't go to the doc-store during eviction because, in the case of multiple nodes, one node could have tombstoned the document before another node has had chance to read from it.
* We no longer include eviction in attribute stats
* We don't need `all-commit-hashes` any more, not going to miss it hugely
* I've moved some of codec (encoding/decoding content-index keys) to kv-indexer, given it's an implementation detail of the kv-indexer
* [ ] Do we want to refactor the standalone/Kafka doc store so that they don't explicitly depend on a kv-store?

### DIFF
--- a/crux-core/src/crux/kv_indexer.clj
+++ b/crux-core/src/crux/kv_indexer.clj
@@ -8,11 +8,11 @@
             [crux.memory :as mem]
             [crux.status :as status]
             [crux.morton :as morton])
-  (:import (crux.codec EntityTx EntityValueContentHash)
+  (:import (crux.codec Id EntityTx)
            java.io.Closeable
            java.util.function.Supplier
            (clojure.lang MapEntry)
-           (org.agrona DirectBuffer ExpandableDirectByteBuffer)))
+           (org.agrona DirectBuffer MutableDirectBuffer ExpandableDirectByteBuffer)))
 
 (set! *unchecked-math* :warn-on-boxed)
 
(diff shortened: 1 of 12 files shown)


crux-core/src/crux/kv_indexer.clj
```diff
@@ -109,6 +109,106 @@
...
+
+   (let [^MutableDirectBuffer b (or b (mem/allocate-buffer (+ c/index-id-size c/id-size (.capacity value))))]
+     (-> (doto b
+           (.putByte 0 c/hash-cache-index-id)
+           (.putBytes c/index-id-size entity 0 c/id-size)
+           (.putBytes (+ c/index-id-size c/id-size) value 0 (.capacity value)))
+         (mem/limit-buffer (+ c/index-id-size c/id-size (.capacity value)))))))
+
+;;;; Entity as-of
+
```
jonpither, line 308: At what point do we make a new `kv_index_store.clj` namespace?

jarohen: (discussed offline) this namespace is dealing with both the write and the read to the kv-store-backed indexer/index-store - the Indexer largely being the write side; the IndexStore being the read. 

Arguably it could be renamed to `crux.kv-index-store` but at that point I'd probably rename the `Indexer` protocol to `IndexStore`, and `IndexStore` to ... something else.

hraberg (changes requested):
You've reintroduced direct dependency on kv-store in KafkaDocumentStore, not necessary as the KvObjectStore can take nil as snapshot and then open one. But would prefer if this object store was just another "random access" doc store (defaulting to StandaloneDocumentStore).

I was wondering about was the new case in bound-results-for-var?

You can move more of the decode/encode fns into the kv-indexer, like most of them are implicitly tied to it.
Also, think you want to explicitly increase the index version :slightly_smiling_face:

hraberg:
Note, that once (and if) you remove the ObjectStore as a concept, you probably want to introduce a CachedDocumentStore (to mimic it) and wrap the document store with it in some places. As documents aren't used on the fast path anymore, this might not be strictly necessary in a first pass.

hraberg:
The crux.tx namespace should have no reference to crux.kv or the kv-store.

jarohen:
Cheers @hraberg :smile: 

> You've reintroduced direct dependency on kv-store in KafkaDocumentStore, not necessary as the KvObjectStore can take nil as snapshot and then open one. But would prefer if this object store was just another "random access" doc store (defaulting to StandaloneDocumentStore).

Yes - think this is the 'unresolved' issue we were talking about earlier. We could implement this as a 'kv-doc-store', I suppose? Will have a hack on this.

> I was wondering about was the new case in bound-results-for-var?

Hmm :thinking: I'd like to have just had the first case, where we pass the e-buffer in the binding directly through without needing to go via the ETx, but for some reason we didn't have that binding present in all cases.

That said, it's likely a premature optimisation, will remove.

> You can move more of the decode/encode fns into the kv-indexer, like most of them are implicitly tied to it.

Yeah - was considering that as I was moving the content ones, but didn't in the interests of a smaller change?

> Also, think you want to explicitly increase the index version

This is based on master-1.9 which already has a bump from your earlier branch, although no harm in incrementing it again I suppose? 

> Note, that once (and if) you remove the ObjectStore as a concept, you probably want to introduce a CachedDocumentStore (to mimic it) and wrap the document store with it in some places. As documents aren't used on the fast path anymore, this might not be strictly necessary in a first pass.

Agreed - ok to include this in a separate pass?

> The crux.tx namespace should have no reference to crux.kv or the kv-store.

This is a dead require, will clean the ns.

hraberg (approved):
LGTM, might be more encode/decode fns you could move over, like the meta stuff (which should be limited to the KvIndexer now). Also, as mentioned, you can avoid reintroducing the direct dependency on the kv store in KafkaDocumentStore as KvObjectStore will create a snapshot anyway if nil is provided. Though this will all go away later so not that important.

PR #897, opened 2020-05-31
Nested index store recycling

hraberg:
Contains a lot of internal tweaks and simplifications.


jarohen (approved):
Looking good :smile:

Still a little uneasy about the state/knowledge of 'levels' introduced into the KvIndexStore, as discussed - but let's not block this one (on the assumption we'll remember/have the time to come back to it later)

jonpither (approved):

PR #898, opened 2020-06-01
Transaction function contexts, #865

### LINKED ISSUE #865: Transaction function impls are passed a context
Currently, transaction functions are passed a database value at the current vt/tt and their args - we'd like to offer more:

* Ability to get a database at any vt/tt
* Details about the current transaction - transaction-time/transaction-id, other operations, etc - replacing the current dynamic var
* Anything else?

jarohen:
resolves #865.

See tx-test changes for what this looks like from a user's PoV

Anything else we should expose as part of this context?

Also:
* Documentation added for transaction functions
* If a transaction function returns `false` we abort the transaction
* Feature flag removed :smile:


hraberg:
Other things that could go in the context are the things currently sitting in the dynamic var, the submitted tx itself, that is, its timestamps and id (and maybe also its ops).

jarohen:
@hraberg - yep - I've put `:crux.tx/current-tx` on the context which has the tx-time and tx-id, could consider putting the ops on there too

hraberg (approved):

hraberg:
Reflecting a bit on this - would it be somewhat cleaner to provide access to the transaction via a method on the context instead of just grabbing a "magic" keyword?

jarohen:
@hraberg replaced magic keyword with `crux.api/indexing-tx` as discussed :smile:

PR #900, opened 2020-06-03
Optional z-curve index #896

### LINKED ISSUE #896: Users can choose not to use the z-curve index, to save disk space
The z-curve index is the same size as the 'normal' (EVtTtTid -> C) bitemp index, and is more performant when _both_ of the following hold:

* Users frequently correct entities back in valid-time (nb not just overriding an entity at current valid-time)
* Users frequently make queries back in transaction-time (ie without corrections - not just queries back in valid-time)

This may not be required in a decent proportion of Crux use cases, so we can save a not-insignificant amount of index space, ingest time and query speed (latter anecdotally, suspect due to a smaller index).

* The z-curve index should be turned off by default
* Users can turn on the index with configuration, on a clean index
* If the user starts indexing a node without the z-curve index and then turns it on, the node should throw an error at startup (ie the user should be forced to re-index to turn it on)

jarohen:
Adding `:crux.kv-indexer/z-curve-index?` configuration param, default false, to determine whether we keep a z-curve index.


crux-core/src/crux/kv_indexer.clj
```diff
@@ -761,7 +777,11 @@
...
-  {:start-fn (fn [{:crux.node/keys [kv-store]} args]
+  {:start-fn (fn [{:crux.node/keys [kv-store]} {::keys [z-curve-index?]}]
                (check-and-store-index-version kv-store)
-               (->KvIndexer kv-store))
-   :deps [:crux.node/kv-store]})
+               (check-z-curve-index-config kv-store z-curve-index?)
+               (->KvIndexer kv-store z-curve-index?))
+   :deps [:crux.node/kv-store]
+   :args {::z-curve-index? {:crux.config/type :boolean
+                            :default false
```
hraberg, line 786: I think the default should be true (that is, stay as is), as the cost can be quite high when it's not there for certain data sets.

hraberg (approved):
LGTM, but see line comment about that I think the config should default to true, turning it off should be an active space saving exercise I feel.

Also, but guess this sits in another commit later, will need docs, detailing the trade-off.

Another question - how big are these indexes (we had some measurements on this at some point I posted, but dataset wasn't that representative), like does it save much space, compared to the content indexes?

Ironically, if you need it - that is, you have a lot of history and corrections, it will be larger, while in cases where you can afford to be without it it will already be small.

jarohen:
As discussed, space saving not significant enough to justify the complexity - closing

PR #903, opened 2020-06-04
Improve next kv usage

hraberg:
Contains crux.indexer cleanup and simplification. Better namespace isolation.

Also adds ae and av indexes to avoid calls to seek, but this isn't directly proven to be faster. Adds ~5% size. Can back these changes out if we prefer to wait?


hraberg:
The added changes crux.kv-indexer are a mix, but the new indexes could back out. Another reason to keep the new indexes is that we can move AECV and ECAV.

jarohen (commented):
As discussed, could we move AECV -> ECAV?

hraberg:
Thanks, yes, will move to ECAV tomorrow or later today.

hraberg:
Now includes ECAV change, can be merged when build passes.

Issue #907, opened 2020-06-04
Use pred-fn to bind a leaf var


jarohen:
We currently use multiple `:where` clauses to project out multiple values from a document. Using multiple clauses means that we need to ensure that any document in the results must have all of the attributes, and therefore this entity takes part in the join algorithm (potentially many times, especially in SQL relations, see #514). 

(see #904 for an complementary approach to make these joins faster)

In practice, though, the user will often only need to explicitly filter based on one attribute, and then they'll know that the others are present - for example, if we know that any user document always has a name and an email, we only need to find the document itself, and then we can project out the other fields.

Let's add a `get` function to the query engine that will optionally bind a var, as follows:

```clojure
{:find [?u ?name ?email]
 :where [[?u :user/id]
         [(get ?u :user/name) ?name]
         [(get ?u :user/email) ?email]]}
```

(see also: pull syntax, #849)

jonpither:
Part of this task (which could moved to a separate issue) is to expose db and index-store to datalog fns. The `get` function will need this access.

jonpither:
We had also discussed naming this `optional` - final naming to be finalised.

jonpither:
I have spiked an approach for this that does yield some query-time performance. 

Using the TPCH dataset querying 300k line-items and 10 projected 'leaf var' fields, I reduced a `select *` query time down from ~55 secs to ~35 by introducing a hardcoded pred fn.

Old query:

```
  {:find [l_orderkey],
   :where [[e :l_orderkey l_orderkey]
           [e :l_partkey l_partkey]
           [e :l_suppkey l_suppkey]
           [e :l_linenumber l_linenumber]
           [e :l_quantity l_quantity]
           [e :l_extendedprice l_extendedprice]
           [e :l_discount l_discount]
           [e :l_tax l_tax]
           [e :l_returnflag l_returnflag]
           [e :l_linestatus l_linestatus]
           [e :l_shipdate l_shipdate]
           [e :l_commitdate l_commitdate]]
   :timeout 100000}
```

New query:

```
{:find [l_orderkey],
   :where [[e :l_orderkey l_orderkey]
           [(crux.query/optional e :l_partkey) l_partkey]
           [(crux.query/optional e :l_suppkey) l_suppkey]
           [(crux.query/optional e :l_linenumber) l_linenumber]
           [(crux.query/optional e :l_quantity) l_quantity40]
           [(crux.query/optional e :l_extendedprice) l_extendedprice]
[...]

PR #908, opened 2020-06-05
Function returns semantic changes

hraberg:
Proposed semantic changes: Support nil and false returns from functions. Treat sets and vectors as multiple results from functions.

Up for debate, but won't open as draft - either we merge and maybe tweak this a bit or simply decide close it, not much to wait for here.


hraberg:
Related to #907.

jarohen (approved):

jarohen:
some commented out tests in here still?

jonpither (approved):

Issue #906, opened 2020-06-04
Optimising equals/range constraints


jarohen:
We currently treat equals/inequalities as range constraints if one side's a literal; if both are logic-vars we treat it as a predicate (once both are bound). This means (for example) that queries using args as one side of a range/equals constraint don't make the best use of indexes - we have to fully realise the values before we can apply the constraint. We'd like to 

* Even if we do have two logic vars, we only need to wait until we have one side bound, and then apply a range constraint to the second index further down the join depth. 
* `=` would then be a special case of this - as soon as we bind one side, we can bind the other side. 
* We could take advantage of this in the query planning, too - if we know one side is a small relation, we bind that first, then immediately get the other side bound.
* We could consider deprecating/removing `==` (unify) once this change is in.

hraberg:
This is related to #348.

Issue #905, opened 2020-06-04
Optimising UnaryJoinVirtualIndex


jarohen:
We've identified `UnaryJoinVirtualIndex` as one area of the query engine that would particularly benefit from optimisation - any performance improvements to this index will likely result in a decent speed-up in the overall engine.

hraberg:
It doesn't seem like it's actually unary join itself that's the slow part, it's more the fact that it performs a join along several close to identical indexes it has to scan, having to read out the data n times.

hraberg:
A potential option here is to fold down the left-hand side of the E to a single index, more or less ignoring the clauses with the lower attribute count, and then point all right-hand side V indexes to the same left-hand peek state. This might require some cleanup in crux.index, and some changes to crux.query, but worth spiking.

An option is also to make indexes in general aware of the parent join keys, as that removes the need for peek state and coupling between them in this case.

hraberg:
Note that the smallest left hand side might still be a super set of a real intersection, but this would be pruned down later down in the join as the attribute values are bound.

hraberg:
Closing this, related work will be done as part of #907. The unary join step itself can be tweaked a bit, but isn't the bottleneck for most queries.

PR #909, opened 2020-06-06
refactor: make binary-index creation less stateful

jarohen:
We were creating a skeleton binary index for the query plan, and then statefully updating it with the concrete indexes at index creation - by passing query details through to idx-fn we can create the binary index in one go, removing `update-binary-join-order!` and friends

### DIFF
--- a/crux-core/src/crux/index.clj
+++ b/crux-core/src/crux/index.clj
@@ -276,17 +276,8 @@
   (max-depth [this]
     2))
 
-(defn new-binary-join-virtual-index
-  ([]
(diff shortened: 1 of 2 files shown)


hraberg (approved):
Like this. You could potentially change the names map to explicit e/v vars, but we might choose to create more advanced n-ary indexes here for the #905 case, so signature might change regardless.

Also think that this removes (most) of the mess where the wrong encode-value-fn from the compilation step is used. I think this might still be the case for range constraints, but that shall hopefully be tweaked and fixed as part of #906.

PR #910, opened 2020-06-08
Query planner refactoring

hraberg:
Moving relational virtual index to use TreeMaps under the hood. Cleaning out BinaryVirtualIndex.


hraberg:
Related to #905 and #906.

jonpither (approved):

crux-core/src/crux/index.clj
```diff
@@ -425,11 +352,23 @@
...
+
+(defrecord SingletonVirtualIndex [v]
+  db/Index
+  (seek-values [_ k]
+    (when-not (pos? (mem/compare-buffers (or k mem/empty-buffer) v))
+      v))
+
+  (next-values [_]))
+
+(defn new-singleton-virtual-index [v encode-value-fn]
```
jarohen, line 372: nit: this fn feels like its implementation details are in the wrong place to me - save the detail of nippy's freeze fallback, we're only passing the encoded value through to the record. better done in query.clj as a helper function?

PR #917, opened 2020-06-09
Rocksdb memory leak

hraberg:
Think we might want to duplicate the buffers returned from DirectBuffers.
Also, try using unpooled buffers when returning values out of Rocks (will be slower, but safer).


jarohen (changes requested):
This one failed the bench, OOM

jarohen:
Will close this one for the time being, we've reverted #851 in #924

PR #919, opened 2020-06-09
Leaf vars

### LINKED ISSUE #907: Use pred-fn to bind a leaf var

hraberg:
First stab at leaf vars, using a special built in pred fn.
Some other tweaks around var bindings and moving the nippy unfreezable to only deal with args.


jarohen (approved):

jonpither (approved):

crux-test/test/crux/query_test.clj
```diff
@@ -2916,13 +2978,123 @@
...
+                                     '{:find [?e ?age ?salary ?sex]
+                                       :where [[?e :age ?age]
+                                               [(get-attr ?e :salary) ?salary]
+                                               [(get-attr ?e :sex) ?sex]]}))))
+            (let [pred-fn-ns (- (System/nanoTime) pred-fn-ns-start)]
+              (prn :pred-fn-ns-------- pred-fn-ns)
+              (t/is (< (double (/ pred-fn-ns single-field-ns)) acceptable-slowdown-factor)
+                    (str (/ single-field-ns 1000000.0) " " (/ pred-fn-ns 1000000.0)))))
+
+
```
jonpither: extra blank line

crux-test/test/crux/query_test.clj
```diff
@@ -963,6 +963,68 @@
...
+    (t/is (= #{[:ivan 21]}
+             (api/q (api/db *api*) '{:find [e age]
+                                     :where [[e :name "Ivan"]
+                                             [(get-attr e :age) age]]})))
+
+    (t/is (empty? (api/q (api/db *api*) '{:find [e age]
+                                          :where [[e :name "Oleg"]
+                                                  [(get-attr e :age) age]]})))
+
+
```
jonpither: blank lines

PR #928, opened 2020-06-11
Fix CircleCI Cred Fail

### LINKED ISSUE #831: CI shouldn't fail if it doesn't have access to AWS credentials

danmason:
Fixes #831


jarohen (changes requested):
@danmason - looks like there's a fair few reflection warnings on this branch now that the clients are delays? the type hints on the client itself aren't correct either - it's no longer an S3Client, it's a delay containing an S3Client. happy to pair up if you're unsure?

PR #932, opened 2020-06-11
Optimising range constraints

### LINKED ISSUE #906: Optimising equals/range constraints

hraberg:


jarohen (commented):
hey @hraberg  - change looks fine in principle, wouldn't mind a quick walk through the changes to the planner in particular, when you've got a mo?

PR #933, opened 2020-06-11
Speculative transactions, #904

### LINKED ISSUE #904: Users can make queries based on speculative updates to the local database

jarohen:
resolves #904

This adds `with-tx` to database values, accepting a list of transaction operations.

See [crux.fork-test](https://github.com/jarohen/crux/blob/speculative-txs-904/crux-core/test/crux/fork_test.clj) for example usage.

Implementation notes:
* This is implemented by forking the indexer, yielding an indexer where any novelty is stored in an in-memory index - both the underlying indexer/index-store and the in-memory indexer/index-store are consulted for queries.
* This implementation is also used within the tx processing - we had a similar implementation here anyway (from #434) but this now uses the same forking code. 
* As a result, transaction functions can now easily see the in-transaction state of any entity they wish to.
* `IndexStore`'s `entity-as-of` now returns an `EntityTx` for deleted entities - previously it just returned `nil`. This is so that we can distinguish between the forked `IndexStore` knowing that an entity's explicitly been deleted since the fork or whether there's been nothing to affect that entity

To do:

* [x] Moar tests, particularly around the bitemporality implementation - history bounds, future transactions, etc
* [x] Squash down the commit history
* [x] Document up for opencrux.com

### DIFF
--- a/crux-core/test/crux/fork_test.clj
+++ b/crux-core/test/crux/fork_test.clj
@@ -1,0 +1,139 @@
+(ns crux.fork-test
+  (:require [crux.fork :as fork]
+            [clojure.test :as t]
+            [crux.fixtures :as fix :refer [*api*]]
+            [crux.api :as crux]
+            [crux.tx :as tx]
+            [crux.db :as db])
+  (:import [java.util Date]))
+
+(t/use-fixtures :each fix/with-standalone-topology fix/with-node)
+
+(t/deftest test-simple-fork
+  (fix/submit+await-tx [[:crux.tx/put {:crux.db/id :ivan, :name "Ivna"}]])
+
+  (let [all-names-query '{:find [?name]
+                          :where [[?e :name ?name]]}
+        db (crux/db *api*)
+        _ (Thread/sleep 10)    ; to ensure these two txs are at a different ms
+        db2 (crux/with-tx db
+              [[:crux.tx/put {:crux.db/id :ivan, :name "Ivan"}]])]
+
+    (t/is (= (crux/valid-time db) (crux/valid-time db2)))
+
+    (t/is (= #{["Ivna"]}
+             (crux/q db all-names-query)))
+    (t/is (= #{["Ivan"]}
+             (crux/q db2 all-names-query)))
+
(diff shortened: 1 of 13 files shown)


crux-core/test/crux/fork_test.clj
```diff
@@ -0,0 +1,139 @@
...
+  (let [ivan0 {:crux.db/id :ivan, :name "Ivan0"}
+        present-tx (fix/submit+await-tx [[:crux.tx/put ivan0]])
+        _ (Thread/sleep 10) ; to ensure these two txs are at a different ms
+
+        now+10m (Date. (+ (.getTime (Date.)) (* 10 60 1000)))
+        future-ivan {:crux.db/id :ivan, :name "Future Ivan"}
+        now+10m-tx (fix/submit+await-tx [[:crux.tx/put future-ivan now+10m]])
+        future-db (crux/db *api* now+10m)]
+
+    (let [now+5m (Date. (+ (.getTime (Date.)) (* 5 60 1000)))
```
jonpither: not need for an inner let blog?

hraberg (approved):

crux-core/test/crux/fork_test.clj
```diff
@@ -0,0 +1,139 @@
...
+
+(t/deftest test-speculative-from-point-in-past
+  (let [ivan0 {:crux.db/id :ivan, :name "Ivan0"}
+        tt0 (::tx/tx-time (fix/submit+await-tx [[:crux.tx/put ivan0]]))
+        _ (Thread/sleep 10)      ; to ensure these two txs are at a different ms
+        tt1 (::tx/tx-time (fix/submit+await-tx [[:crux.tx/put {:crux.db/id :ivan, :name "Ivan1"}]]))
+
+        db0 (crux/db *api* tt0 tt0)]
+
+
```
jonpither, line 66: excess blank line

jonpither (approved):

hraberg (approved):
Looks good, don't think LMDB will lock up, reads and writes to the kv store don't seem to be interleaved in the same thread.

PR #934, opened 2020-06-11
Add the tpch-stress-test to crux-bench

### LINKED ISSUE #927: Query Stress Test
Whilst tackling a recent memory leak (see https://github.com/juxt/crux/pull/924), I found this test setup was a good stress test to bring about an OOM: https://github.com/juxt/crux/blob/jp/calcite/dev/jon.clj.

My proposal in this issue, is to have a dedicated nightly stress test - beginning with using the above code,  running the bottom query ~25 times, against a single node setup using RocksDB. This will stress-test the Crux query engine. In the future we could have similar stress-tests for ingest and other areas if need be.

The existing bench setup in my mind, is to show relative timings to other databases and to track ingest/query speed across node variants. The point of this stress-test is to ensure we don't OOM again, testing a particular target area (querying against Rocks).

danmason:
WIP - Going to do a test run of this on AWS against the images deployed from this branch and against a branch with the memory leak.

### DIFF
--- a/crux-bench/README.md
+++ b/crux-bench/README.md
@@ -4,7 +4,7 @@ In order to run the benchmarks locally, you need the required datasets. Within t
 
 ## Running the benchmark locally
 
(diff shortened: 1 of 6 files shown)


jarohen:
Passed on first known bad - see [this comment](https://github.com/juxt/crux/issues/927#issuecomment-643105576)

PR #937, opened 2020-06-12
Run `deploy-commit` in parallel with `build`.

### LINKED ISSUE #936: We still deploy a bench commit even if tests fail
We have some speedup tests that don't always work when you're mid-way through a branch - we'd like to be able to bench these changes regardless.

Particularly, this happens when you write a failing speedup test at the start of an optimisation branch and then have to comment it out so that you get a build.

danmason:
Also removes 'deploy-master'.

Fixes #936

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,11 +77,7 @@ jobs:
       - run:
           command: |
             export LEIN_JVM_OPTS=-Xmx2G
-            lein sub install
-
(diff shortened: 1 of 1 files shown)


jarohen (changes requested):
Thanks Dan. As discussed, this builds everything twice - could you analyse the steps and dependencies of the build, figure out what actually depends on what? 

Cheers!

PR #938, opened 2020-06-12
Circleci  - Cache Dataset

### LINKED ISSUE #915: We should cache the bench dataset in CircleCI
No need to pull these down on every build, let's be good internet citizens :)

danmason:
Based on top of #937 

Fixes #915

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,6 +112,19 @@ jobs:
         - checkout
         - attach_workspace:
             at: ~/
+
+        - run: md5sum ~/crux/crux-bench/bin/download-dataset.sh > .circleci-bench-dataset-cache-key
+        - restore_cache:
+            keys:
+              - v1-dataset-{{ checksum ".circleci-bench-dataset-cache-key" }}
+              # fallback to using the latest cache if no exact match is found
+              - v1-dataset-
+
+        - run:
+            command: |
+              export LEIN_JVM_OPTS=-Xmx2G
+              lein sub install
+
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -102,40 +102,35 @@ jobs:
...
-            dockerfile: 'Dockerfile'
-            extra-build-args: '${BUILD_ARG}'
-            path: "crux-bench/"
-            repo: 'crux-bench'
-            tag: '${BUILD_TAGS}'
-
-  deploy-master:
-    docker:
-      - image: circleci/clojure:lein-2.9.1
+        - run: md5sum **/crux-bench/bin/download-dataset.sh > .circleci-dataset-cache-key
```
jarohen: don't think you want `**/crux-bench` here - this is saying 'in all the subdirectories of this directory, look for a `crux-bench` directory'. You should be able to run `md5sum **/crux-bench/bin/download-dataset.sh` in the top-level of your local Crux repo - what happens?

jarohen (changes requested):

.circleci/config.yml
jarohen: also, can this be 'bench-dataset', just to be unambiguous?

danmason: I mostly did it based off of the `project.clj` md5 also being ran using ** - can change. Although, when running it my own local crux repo not much different happens than just running it normally? It does say the md5 followed by `checkouts/crux-bench/bin/download-dataset.sh`

PR #945, opened 2020-06-16
LMDB Segfaults

hraberg:
Adding a stamped lock that has its write lock taken out when increasing the LMDB mapsize, and read lock while transactions are running, to avoid changing the underlying mmap while txs are running, causing seg faults.

### DIFF
--- a/crux-core/src/crux/kv_indexer.clj
+++ b/crux-core/src/crux/kv_indexer.clj
@@ -710,7 +710,7 @@
                                               av-i (kv/new-iterator snapshot)]
                                     (->> (for [eid eids
                                                ecav-key (all-keys-in-prefix ecav-i
-                                                                            (encode-ecav-key-to (.get seek-buffer-tl)
+                                                                            (encode-ecav-key-to nil
                                                                                                 (c/->id-buffer eid)))]
                                            ecav-key)
 
--- a/crux-core/src/crux/kv_indexer.clj
+++ b/crux-core/src/crux/kv_indexer.clj
@@ -687,8 +687,8 @@
 (defrecord KvIndexer [kv-store]
   db/Indexer
   (index-docs [this docs]
-    (with-open [snapshot (kv/new-snapshot kv-store)]
-      (let [docs (->> docs
+    (let [docs (with-open [snapshot (kv/new-snapshot kv-store)]
+                 (->> docs
                       (into {} (remove (let [crux-db-id (c/->id-buffer :crux.db/id)]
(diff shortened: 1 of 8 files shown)


hraberg:
See this ancient line:
```
;; TODO: Note, this has to be done when there are no open
;; transactions.
```
Added 2018-12-10

hraberg:
Same fix was applied for both LMDB backends in this PR.

hraberg:
While this fix might solve the underlying issue, it deadlocks in the consumer as it nests taking out snapshots and hold on to them while performing writes.

hraberg:
WatDiv tests still timeout, potentially due to some deadlock.

hraberg:
Local WatDiv results, doesn't seem to have a deadlock:

```
 *:watdiv-crux* (standalone-lmdb)
========
*ingest* (PT34M21.219S, *+0.00%*. 7D Min: PT0S, 7D Max: PT0S): `{:av-count 2823472, :bytes-indexed 3622119571, :doc-count 521585, :entity-count 521585}`
*docs-per-second*: 253
*avs-per-second*: 1369
*bytes-indexed-per-second*: 1757270
*compaction* (PT0.013S, *+0.00%*. 7D Min: PT0S, 7D Max: PT0S): `{:compacted-bytes-on-disk 5208109056, :bytes-on-disk 5208109056}`
*queries* (PT6M47.686S, *+0.00%*. 7D Min: PT0S, 7D Max: PT0S): `{}`
```

jarohen:
@hraberg have pushed this one to my fork too, and rebased on master - there was one conflict in `unindex-eids` with the 1.9.1 fix.

https://github.com/juxt/crux/compare/master...jarohen:lmdb-segfault

hraberg:
Rebased this on master again, to fix issue @jarohen mentions and has fixed in his fork.

jarohen:
@hraberg - have we got any idea why this one is taking 34m to ingest, even locally? it's a good 5-6x off Rocks.

Seen you have some changes afterwards, do they help?

hraberg:
My changes don't address performance. LMDB has always been much slower than Rocks to ingest into. There's a chance that the rw-lock has increased this a bit - easiest way would be to compare same ingest with or without the lock on the same machine.

This can be done by removing the references to the write lock when growing the mapsize. Don't think it works on master, due to the fact that master nests index-store with writes even in the same thread, so it can be tripped up and segfault by the mapsize growth call.

jonpither (approved):

PR #949, opened 2020-06-17
Pred constraint vars bug

hraberg:
Fixing issue with existing var calculation when the pred-fn is a built in method.

Related to issue reported from #735.

### DIFF
--- a/crux-test/test/crux/query_test.clj
+++ b/crux-test/test/crux/query_test.clj
@@ -1355,7 +1355,7 @@
         (t/is (< (double (/ args-ns literal-ns)) acceptable-slowdown-factor)
               (pr-str args-ns " " literal-ns))))))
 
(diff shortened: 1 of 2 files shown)


hraberg:
Related to #944 and should solve that as well in a better way.

crux-test/test/crux/query_test.clj
```diff
@@ -1355,6 +1355,17 @@
         (t/is (< (double (/ args-ns literal-ns)) acceptable-slowdown-factor)
               (pr-str args-ns " " literal-ns))))))
 
+(t/deftest test-or-range-vars-bug
```
jonpither: stick, the PR num in - 949, for some context and this conversaton on GitHub

jonpither (approved):

Issue #284, opened 2019-07-03
JSON API over HTTP - part 1


refset:
Users who aren't familiar with Java/Clojure will want to use the HTTP API directly (no client) and should not have to work with EDN. 

In Scope:

- [x] Users should be able to request JSON responses from all HTTP endpoints.
- [x] The endpoints that have streaming EDN responses should also have streaming JSON responses.
- [x] Where query params are currently submitted in EDN (e.g. `?eid=`, `?eid-edn=`), they should also accept `?eid-json=`
- [x] Query params submitted in camelCase should be accepted as if they were kebab-case (consider a middleware for this before it hits spec?)
- [x] JSON responses should have their keys camelCased, and any namespace qualification removed, for ease of use from JS. e.g. `"validTime"`, `"txTime"`, etc.  
- [x] Transaction function bodies (legacy and current) should be submitted as a string containing Clojure code, and read as EDN. 
- [x] Attributes in documents should be decoded to Clojure keywords and encoded to JSON strings, where appropriate (i.e. no string-keyed maps in Clojure). Kebab/camel casing in document attributes should be preserved.
- [x] tx operations (for submit-tx request and open-tx-log response) converted as above (`:crux.tx/put` -> `"put"` etc, dates parsed and encoded as necessary)
- [x] A separate set of tests specifically exercising the JSON API (assuming this won't get exercised by the standard api-test)

Out of Scope:

* Initially, the `/_crux/query` _request_ is considered out-of-scope (it should still be possible to receive query results in JSON format)
* Transaction function bodies will need to remain Clojure only. Transaction function arguments should be passed through as their JSON types (i.e. even if the transaction function expects keywords/symbols, JSON consumers will only be able to send them strings)

For reference & consideration:
https://github.com/tonsky/datascript/blob/master/test/js/tests.js
https://github.com/cognitect/transit-js

hraberg:
Here's an example of how to read and write Transit for Crux:

```clj
(require '[cognitect.transit :as transit])
(import [java.io ByteArrayInputStream ByteArrayOutputStream])

(with-open [out (ByteArrayOutputStream.)]
       (let [writer (transit/writer out :json
                                    {:handlers
                                     {crux.codec.EDNId
                                      (transit/write-handler
                                       "crux/id"
                                       c/edn-id->original-id)}})]
         (transit/write writer {:crux.db/id #crux/id "415c45c9-7cbe-4660-801b-dab9edc60c84", :value "baz"})
[...]

PR #952, opened 2020-06-18
Value ids

### LINKED ISSUE #244: Allow arbitrary strings as attributes
Hi! I've been playing around with Crux and I noticed that arbitrary strings aren't allowed as ids. It looks like the current implementation complains if it's not a stringy representation of a uuid/url/keyword or in hex format. Since we're able to serialize arbitrary URIs and keywords as ids, would loosening the requirements on strings be a possibility?

Lines of interest:

https://github.com/juxt/crux/blob/6ef4d8f7257f4a80824be7ad50403f9cefa732f9/src/crux/codec.clj#L255-L274

hraberg:
Implementation allows any valid primitive value that can be decoded directly to also be an id. Also separates the id space from the value space totally, so a value is never encoded to an id buffer and vice versa.

The temporal indexes (and the hash cache) uses ids still, so there's a small dance to move between values and ids in the index store. This could be pushed behind the cache, as the entity-as-of methods could deal with this if they get a buffer that's not an id.

Strings are hashed based on their bytes. This means keywords, urls and uuids can hash to the same value as a string with the same representation. Up for discussion.

This PR also contains along-standing tweak to ensure our ids from string values (like keywords etc.) are calculated using UTF-8 and not the default charset (which would likely be UTF-8). This means that there's an unlikely scenario where someone has run their JVMs using a different charset which would now maybe hash differently. If someone has this issue, we have to either offer a replay tool for their log, or back out this behaviour.

### DIFF
--- a/crux-test/test/crux/query_test.clj
+++ b/crux-test/test/crux/query_test.clj
@@ -1407,15 +1407,7 @@
                                              (or-join [e x]
                                                       [e :extra x]
                                                       (and [(identity :none) x]
-                                                           (not [e :extra])))]}))))
-
-  ;; NOTE: This is the actual root cause of the bug, when editing a
-  ;; message the id was transacted as a raw string.
-  (t/testing "entity ids cannot be raw strings"
-    (let [id (UUID/randomUUID)
-          id-str (str id)]
-      (t/is (thrown-with-msg? IllegalArgumentException #"invalid entity id"
-                              (fix/transact! *api* (fix/people [{:crux.db/id id-str :name "Ivan" :version 2}])))))))
+                                                           (not [e :extra])))]})))))
 
 (t/deftest test-arguments-bug-247
   (t/is (= #{} (api/q (api/db *api*)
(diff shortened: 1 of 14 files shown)


hraberg:
Somewhat related to #284 but we should consider adding an alias to `crux.db/id`, `_id` for JSON usage. At which level this should happen is to be discussed.

hraberg:
Due to the common hashing of strings and keywords, the following already works:
```
'{:find [e] :where [[e "name" "Ivan"]]}
```

hraberg:
The old limitation that strings longer than 128 characters automatically get hashed is still there. Now they will get hashed consistently with other strings though, as I've removed the object-id-type-id concept, and all hashes are now simply "ids".

hraberg:
And following on from that, supporting longer strings consistently during range searches (which didn't use to work before either) is another card IMO.

hraberg:
Actually, the long string hashing might be a problem here, as it would clash with an identical keyword (or url etc.) during decoding, which no way to seeing the difference.

jarohen:
> Strings are hashed based on their bytes. This means keywords, urls and uuids can hash to the same value as a string with the same representation. Up for discussion.

I'm not a fan of this, tbh - neither Java's nor Clojure's equals/hashCode works in this way, and I'm not convinced that we wouldn't introduce subtle inconsistencies in the process. Let's say you were to select a number of entities through a query, get the results back, put the documents into a Java/Clojure map, and then try to get the values by key - any lookup on that map would need to use the type in the DB rather than the type in the query, if they differed.

hraberg:
It's subtle, and partly for legacy reasons. Note that this applies to ids only, which is used for either content hashes (maps), the entity id on the tx log and the attributes.

One reason I went with this direction is that we already have this behaviour, for example `(c/new-id :http://example.com)` and `(c/new-id (URI. "http://example.com))` `(c/new-id "http://example.com")` all already return `#crux/id "89dce6a446a69d6b9bdc01ac75251e4c322bcdff"` This behaviour just extends this to any string.

Note that this doesn't apply to the indexed and returned values themselves, so you can have a document with `:crux.db/id` `:foo` and find it based on `"foo"` but if you access the value itself it will be `:foo` as that's what's in the document / index. Value and id encoding are fully separated on this branch.

Any string that can be coerced into a keyword, like `:foo` and `":foo"` already hashes to the same value, so not sure if this really buys us much. Same for UUID, URLs and EDN-string maps (!). The latter is I think a bit of a step to far, but it's there, I think it was added for the console at some point and I didn't really react to it.

hraberg:
Note that joins are performed on the values, so `:foo` and ":foo" wouldn't join against each other despite hashing to the same id. Arguably we kind of want to shut down this kind of stuff over time, and move to string / numeric ids as the preferred style, and document that you shouldn't mix and match id style in the same Crux instance and expect much joy.

hraberg:
Traditionally, the idea of coercing strings into ids was really meant as a helpful a parsing level coercing-thing, but it sits deeper than that in the code.

I don't think there's a super clean way to solve this TBH, except moving forward to something more sane.

crux-test/test/crux/query_test.clj
```diff
@@ -2421,6 +2406,44 @@
                     '{:find [x]
                       :where [[(hash-map) x]]})))))
 
+(t/deftest test-can-use-any-value-as-entity-id
+  (fix/transact! *api* [{:crux.db/id "ivan@example.com" :name "Ivan"}
```
jonpither, line 2410: nice.

hraberg:
As discussed, "normal" strings that cannot be coerced are hashed into their own space now, so `"foo"` and `:foo` will hash differently.

Issue #918, opened 2020-06-09
StandaloneDocumentStore fetch-docs always creates a snapshot


jonpither:
Regardless of a cache hit/miss on the oft wrapped `CachedObjectStore`.

hraberg:
The real solution to this is to get rid (or fold in) the object store and its cache, and add a CachedDocumentStore instead.

At a related note, you could recycle the snapshot inside the KvObjectStore, as you only need to update the snapshot when new docs are written as you don't care about other data in the store. In some stores, like LMDB, the snapshot is a read transaction, so not something one would want to keep around.

We could also re-add get-value on KvStore itself, if you don't care about snapshot versioning, which might be the case in the document store.

jarohen:
Done with #961

Issue #818, opened 2020-04-24
Standalone event-log-kv-store configuration is confusing


jarohen:
We've now had a number of support reqs related to the standalone event-log-kv configuration. If the user doesn't explicitly provide a value for `:crux.standalone/event-log-kv-store`, it has its own memdb default, which leads to some counter-intuitive behaviour - when the node's restarted, the indices are persisted but the tx-log isn't.

* We can't 'just' use the same KV store instance for both - because we'd like to be able to delete the indices' KV store (on an index version bump, for example) but not the tx-log's KV store instance.
* The topology system doesn't currently support being able to create multiple instances of the same component, so we can't just use the value under the `:crux.node/kv-store` key.

refset:
There has been additional confusion resulting from the likelihood of accidentally typing `:crux.standalone/event-log-kv-store` instead of `:crux.standalone/event-log-kv` - perhaps this can somehow be solved at the same time.

refset:
Regardless of how the mechanism works, I see two options:

1. Prevent use of kv persistence without a persisted event-log OR
2. fix:
2.a. event-log-dir not configured causes open-tx-log to throw as no protocol on ingest client AND
2.b. entity API not working on node restart when event-log-dir not configured

PR #961, opened 2020-06-22
Removing object store

hraberg:
Removing object store, replacing with document store in all places. Adding caches to all document stores for now, as this configuration isn't tied to a specific implementation.


jonpither (approved):
I like the saner way of doing things - having the CachedDocumentStore wrap the real DocStore, rather than the inverse way we have now.

jonpither:
This would fix https://github.com/juxt/crux/issues/918

hraberg:
I would prefer that to, but that would require us to introduce two document stores in the config, have other components use the cached one, while allowing the user to just override the other one. It's a valid option. The mess around Kafka's random access store and the standalone's internal event-log-kv one is a bit messier still to fix.

This PR doesn't strictly try to fix all that (there's a separate card for some of that, #818), just to bring it in line by:
1. removing the object store as a concept.
2. make all document stores cached.

jarohen (approved):

PR #964, opened 2020-06-22
await-tx no longer polls, #825

### LINKED ISSUE #825: Avoiding polling within await-tx?
We currently poll on a 100ms interval within await-tx, which means that, if a transaction hasn't been indexed as soon as the user calls await-tx (unlikely if they're trying to read their own writes), they have to wait a minimum of 100ms before they check again - this may be a significant chunk of an HTTP req/resp cycle.

We can obviously bring this polling interval down, but that means threads waking up more frequently, which may well have its own performance impact.

If this is indeed an issue, I wonder whether standard Java object wait/notifyAll or something else in j.u.concurrent be more suitable?

jarohen:
Resolves #825 - see that issue for rationale.
Based on #969, will merge that first (first four commits of this branch are from there). Latter three commits are all self-contained.

* Adds `await` to the internal bus so that we can block on the calling thread, waiting for a message of a certain shape.
* Adds `::tx/ingester-error` and `::node/node-closed` events to the bus s.t. `await-tx` can throw if either happen

### DIFF
--- a/crux-core/src/crux/bus.clj
+++ b/crux-core/src/crux/bus.clj
@@ -31,16 +31,24 @@
 
 (defrecord EventBus [!listeners]
   EventSource
-  (listen [this listen-ops f]
-    (let [{:crux/keys [event-type]} listen-ops
+  (listen [this listen-opts f]
+    (let [{:crux/keys [event-types]} listen-opts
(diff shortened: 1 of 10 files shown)


hraberg (approved):
LGTM, as discussed. My approval and review is really for the await parts, looks like this branch is based of the polling txs branch as well, which is independently approved, but haven't re-reviewed it.

PR #966, opened 2020-06-24
Various minor tweaks on some hot paths.

hraberg:


hraberg:
Also contains a fix to over-use of pooled buffers in Rocks, which hopefully should improve memory peak usage.

PR #969, opened 2020-06-25
Supporting non-polling transaction consumers

jarohen:
Currently, our tx-consumer polls using the tx-log's `open-tx-log` implementation - users can plug in a new tx-log implementation. We have one built-in tx-consumer component in the topology that starts up this polling loop. This has all the usual properties of a polling solution - particularly, that it defaults to polling every 100ms, which (if you're unlucky) provides a lower bound to the end-to-end latency of transactions.

This PR allows users to meaningfully plug in their own tx-consumer.

* We provide a polling implementation that's easy to integrate - see the new JDBC implementation in the PR for how to preserve the current behaviour
* This PR allows users to implement a non-polling tx-consumer, for tx-logs that support it. The new Standalone implementation is an example of this - as soon as the transaction is submitted it can be
* Kafka is somewhere in between - we still need to poll, but because it's now pluggable it means we can reuse the Kafka consumer (rather than having to open up a new Kafka consumer every time `open-tx-log`'s called, because we don't know what the caller's going to do with it).

Indeed, we did have something similar, previously, before we split out the Indexer/IndexStore/DocumentStore protocols - we didn't have any non-polling consumers at the time so it made sense to centralise the polling logic. This PR goes slightly back in the previous direction - keeping the polling logic centralised but allowing it to be overridden.

We do this by breaking up the transaction ingest process. Previously, `index-tx` could be thought of as three stages: begin, index-tx-events, and commit/abort - this PR exposes these three stages to allow the user to consume transactions however they want.

Notably, `with-tx` is now a doddle - we begin the tx, index the events, but don't commit the transaction.

### DIFF
--- a/crux-core/src/crux/db.clj
+++ b/crux-core/src/crux/db.clj
@@ -47,8 +47,14 @@
   (latest-submitted-tx [this]))
 ;; end::TxLog[]
 
-(defprotocol TxConsumer
-  (consumer-error [this]))
+(defprotocol TxIngester
+  (begin-tx [tx-ingester tx])
+  (ingester-error [tx-ingester]))
+
+(defprotocol InFlightTx
+  (index-tx-events [in-flight-tx tx-events])
+  (commit [in-flight-tx])
+  (abort [in-flight-tx]))
 
 (defprotocol DocumentStore
   (submit-docs [this id-and-docs])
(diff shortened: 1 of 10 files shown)


hraberg (approved):
I think I grok it now, and like the overall direction.

I don't find it very easy to follow in parts tbh, like there are various new states that are managed, components getting created directly with subset of the circular deps outside the module framework.

But that said, I think it's better to move forward here if you have momentum. My suggestion would be to spend some time to more simplify, reflect and refactor the pieces in the ingest pipeline now. Potentially some of this can happen as part of the next step of the module framework and configuration work.

I don't think we should expect external people to directly depend on any of these pieces, we can leverage them if needed, but be prepared to move them around.

Issue #967, opened 2020-06-24
Timeline as a value?


jarohen:
related to #921

jonpither:
Closing, we will use #968 to track this, which will encompass API changes.

PR #973, opened 2020-06-26
Initial simplistic cut of valid time range queries.

### LINKED ISSUE #968: Temporal range queries
The easiest way to implement this is to add a vt-index, which only contains the times vt has changed (by anything), and then scan this index from start-vt to end-vt and do a normal point-in-time query for each step. This can obviously get a bit slow for some cases, but might be usable enough for others. We can then present the result in a few ways:

1. a single massive set, without time visible.
2. a per vt-change set, preferably with diffing, pruning empty change-sets. `{#inst "2020" #{...} #inst "2021 #{... diff-to-previous}]`

There are other, better, but harder ways to implement this, but those require changes to the query engine itself, and ability to perform interval-joins, where the join order will affect the result set.

hraberg:


hraberg:
Not properly integrated with the actual API, just a proof of concept of the internals necessary to support it. Could be evolved to a proper PR.

hraberg:
Results are implemented as a map of valid-time->result-set, where the result set is always the full result, but the entire results are deduped so new valid time keys are only added when the actual result has changed. This valid-time will likely be a time from some entity in the actual tuple.

hraberg:
API side of things, this likely will be connected with the #967 work, we could roll it in as is, or wait. Technically the range queries doesn't belong to the db, as it takes out several dbs during the execution.

hraberg:
Closing but keeping around for reference.

PR #982, opened 2020-07-02
Pull syntax

### LINKED ISSUE #849: EQL-like syntax?
https://edn-query-language.org/eql/1.0.0/what-is-eql.html

jarohen:
#849 

Allows users to decouple specifying which entities they want from what data they'd like about those entities in their queries.

We add a `(pull logic-var pull-spec)` option to the `:find` part of queries, see the change to `query_test.clj` or `pull.adoc` in the PR for the syntax and usage.

Implementation notes:
* Like the rest of the query engine, we split the pull engine up into 'conform', 'plan' and 'execute'.
  * 'Conform' is handled by EQL (thanks!) 
  * 'Plan' builds up a tree of functions, each of which accepts the runtime value and the db instance (mainly for doc-store + index-store)
  * 'Execute' then runs those functions.
* The 'plan' part is cached in the same query cache as the rest of the query plan.
* `bound-result-for-var` no longer decodes values - for `pull` we need bytes.

TODO:
* [ ] RFC: we don't yet implement union/recursive queries - do we need to before release?
* [ ] Still makes one request at a time of the document-store - we could batch these for performance (at a cost of some complexity)
* [ ] RFC: do we want an independent `pull` function, taking `eid`

### DIFF
--- a/crux-core/project.clj
+++ b/crux-core/project.clj
@@ -12,7 +12,8 @@
                  [org.clojure/tools.reader "1.3.2"]
                  [com.taoensso/encore "2.114.0"]
                  [org.agrona/agrona "1.0.7"]
-                 [com.github.jnr/jnr-ffi "2.1.9" :scope "provided"]]
+                 [com.github.jnr/jnr-ffi "2.1.9" :scope "provided"]
+                 [edn-query-language/eql "1.0.0"]]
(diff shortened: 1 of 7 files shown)


jarohen:
I accidentally deleted the branch - see #992

PR #983, opened 2020-07-03
Combining query implementations #846

### LINKED ISSUE #846: Combining eager/lazy query implementations
Currently the eager and lazy query implementations are slightly different - ideally, the eager query would 'just' be a call to the lazy query that puts the result in an eager collection (see `open-entity-history`)

see: TODO in query.clj `DataSource`

jarohen:
fixes #846 

This PR combines the eager and the lazy query functions - the eager query now just takes the lazy query and pours it into a vector/set. Queries now run on the caller thread and are interrupted by a scheduled task, so timeouts apply to both eager and lazy queries now (they only applied to eager queries previously)

I've also re-ordered the query namespace a little with this PR, trying to group up conform/plan/execute a bit more - it's a separate commit so can take it out if that'd be preferable. Might be easier to review the PR a commit at a time (there's only two :smile:)

### DIFF
--- a/crux-core/src/crux/query.clj
+++ b/crux-core/src/crux/query.clj
@@ -1267,8 +1267,21 @@
 
   (query [this query]
     (with-open [res (.openQuery this query)]
-      (let [result-coll-fn (if (some (normalize-query query) [:order-by :limit :offset]) vec set)]
-        (result-coll-fn (iterator-seq res)))))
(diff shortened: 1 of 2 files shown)


jonpither (approved):

jonpither (changes requested):
As discussed, the time-out shouldn't apply to lazy queries.

jarohen:
Updated timeouts to only apply to eager queries, cheers :smile:

jonpither (approved):

PR #987, opened 2020-07-08
We also evict keys from bitemp indices on evict, fixes intermittent test failure

jarohen:
Previously, we only evicted keys from the content indices when we evict entities, not the bitemp indices - this was mostly safe:
* in the query engine, we look at the content indices first - evicted entities are removed from here, no problem.
* when returning a document (via entity, or history) we check to see whether the document's a tombstone

With Kafka, though, submitting a doc to the doc-store is asynchronous, so there was a slight delay between the transaction committing and the tombstone making it into the doc-store. This, combined with `entity-tx` still returning ETxs for the evicted entity due to the bitemp index (seems relatively harmless - not much you can do with just a content-hash), means that you could still get hold of the entity for a short while after the transaction's committed. 

Side effect of this PR is that `entity-tx` now returns nil for evicted eids where previously it would return an ETx.

I suspect adding document-store caches to this made it more visible - if the ordering is 'submit-doc', 'fetch-doc', 'index-doc', the cache is invalidated on submit, re-seeded in fetch, and not re-invalidated after the doc's indexed.

### DIFF
--- a/crux-core/src/crux/kv_indexer.clj
+++ b/crux-core/src/crux/kv_indexer.clj
@@ -678,6 +678,7 @@
 
   (unindex-eids [this eids]
     (let [{:keys [tombstones ks]} (with-open [snapshot (kv/new-snapshot kv-store)
+                                              bitemp-i (kv/new-iterator snapshot)
                                               ecav-i (kv/new-iterator snapshot)
                                               av-i (kv/new-iterator snapshot)]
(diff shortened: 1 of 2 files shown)


jonpither (approved):

jonpither:
Nice.

hraberg (approved):

PR #991, opened 2020-07-10
Various resolution of dependencies

danmason:


danmason:
Still a few more test level dependencies to do.

jarohen:
@danmason I've moved a load of crux-test dependencies out to the `:test` profile - seems like they were causing most of the conflicts, so a lot can now be removed. Will run the tests through and see what happens

PR #999, opened 2020-07-27
All value representations in content indices

jarohen:
This PR brings the representation of attributes in content indices to use 'value buffers' rather than 'id buffers'. @hraberg made the initial change in #952 to migrate entities and values over to value buffers - this PR migrates attributes so that all of the constituent parts of the content index keys (AV, AVE, AE, ECAV) are value-buffers.

This means that we have the flexibility to allow more types of attributes (eg strings) without a further index bump. I deliberately haven't relaxed the spec in this PR as it deserves more consideration - this merely lowers the barrier to doing so. Regardless of opinions on that particular debate, this simplifies the content indices as all the components now have the same encoding/decoding.

When we initially discussed this we were concerned about having more than one unknown length component in the key - in practice (as Håkan found out) in the majority of usages we still only have one unknown length component (in `av` we have `a`, `ave` we have `a` and `v`, `ae` we have `a`, `aev` we have `a` and `e`). The exception to the rule is evicting an entity, where we want to use the ECAV index but only have `E` in hand - `C` is fixed-length but `A` and `V` are both variable. I've fixed this by adding explicit lengths to the entity and attribute components (but not the values, to preserve the lexicographical ordering properties of values). 

Out of scope of this PR, but another related change to consider: migrating the eids in the bitemp indices to value buffers - this would mean that _everything_ in the kv-store would be value buffers.
* Value would be that we could extract all serde implementation from the KVIndexer - it could both accept and return bytes without understanding what they mean (so long as they're lexicographically comparable, for value types that we want to order)
* All Indexers will share the same binary representation, which will lower the barrier to users creating their own Indexer implementations.
* This will allow us to fix the assumption in speculative transactions that KVIndexer is the only implementation of Indexer - again, because of the shared binary representation.
* This is significantly more involved, though - in the tx-log we frequently have eid hashes which would need to be converted to value buffers. AFAIK we already store these values in the hash-cache, so it's theoretically possible - but clearly not as straightforward.

### DIFF
--- a/crux-core/src/crux/codec.clj
+++ b/crux-core/src/crux/codec.clj
@@ -21,7 +21,7 @@
 ;; Indexes
 
 ;; NOTE: Must be updated when existing indexes change structure.
-(def ^:const index-version 12)
+(def ^:const index-version 13)
 (def ^:const index-version-size Long/BYTES)
 
 (def ^:const index-id-size Byte/BYTES)
--- a/crux-core/src/crux/kv_indexer.clj
+++ b/crux-core/src/crux/kv_indexer.clj
@@ -100,31 +100,27 @@
    (encode-av-key-to b attr mem/empty-buffer))
   (^org.agrona.MutableDirectBuffer
    [^MutableDirectBuffer b ^DirectBuffer attr ^DirectBuffer v]
-   (assert (= c/id-size (.capacity attr)) (mem/buffer->hex attr))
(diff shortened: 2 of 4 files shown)


jarohen:
Discussed with @hraberg, closing this PR for the time being. 

The direction we're looking at taking here is to:
* have the Indexer implementation responsible for the encoding of the attribute - so that it can use whatever fields, keys etc as may be available
* extract the encoding of values/eids outside of the Indexer - so that indexes can be merge-sorted, and that individual indexers do not need to concern themselves with the range of value types in Crux (whatever that range may be)

PR #1027, opened 2020-08-12
Add "slowest queries" to the API.

### LINKED ISSUE #516: Crux exposes the slowest queries to have run on the node
Crux should report on the slowest <n> queries to have recently run on the node (n.b. not the query args, to avoid leaking data)

We'd like to be able to configure the recency, count and minimum time threshold for the queries that Crux stores (see 'slow query logging' for common databases)

- API call for 'recent slow queries' - last 100, say? same details as for #493 
- users can register a callback in crux-metrics to get all the slow queries (to log out, for example) - we can provide example sinks here to sink to a JSON-formatted stdout line (to be picked up by Docker log processing, for example), or one to log/info as plain text

danmason:
Resolves #516

Likely needs some iterating/cleanup but currently:
- List is implemented as a key, `:slowest` within the `!running-queries` atom
- Output is a list of QueryStates, as with `recent-queries` and `active-queries`.
- Adds an arg to `crux.node`, `slow-query-callback-fn` - user can pass a callback function that gets called when a 'slow-query' completes

### DIFF
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -129,7 +129,10 @@
     "Returns a list of currently running queries")
 
   (recent-queries [node]
-    "Returns a list of recently completed/failed queries"))
+    "Returns a list of recently completed/failed queries")
+
+  (slowest-queries [node]
+    "Returns a list of slowest completed/failed queries ran on the node"))
 
 (defprotocol PCruxIngestClient
   "Provides API access to Crux ingestion."
@@ -212,7 +215,10 @@
     (map qs/<-QueryState (.activeQueries this)))
 
   (recent-queries [this]
-    (map qs/<-QueryState (.recentQueries this))))
+    (map qs/<-QueryState (.recentQueries this)))
+
+  (slowest-queries [this]
+    (map qs/<-QueryState (.slowestQueries this))))
 
 (extend-protocol PCruxIngestClient
   ICruxIngestAPI
--- a/crux-metrics/src/crux/metrics/query.clj
+++ b/crux-metrics/src/crux/metrics/query.clj
@@ -21,17 +21,3 @@
      :current-query-count (dropwizard/gauge registry
                                             ["query" "currently-running"]
                                             (fn [] (count @!timer-store)))}))
-
-(defn add-slow-query-listeners
-  [{:crux.node/keys [bus]} {:crux.metrics/keys [slow-query-callback-fn] :as node-opts}]
-  (let [!in-progress-queries (atom {})]
-    (bus/listen bus {:crux/event-types #{::q/submitted-query}}
-                (fn [{::q/keys [query-id] :as query}]
-                  (swap! !in-progress-queries assoc query-id {:started-at (Date.)})))
-    (bus/listen bus {:crux/event-types #{::q/completed-query ::q/failed-query}}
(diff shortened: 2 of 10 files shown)


crux-metrics/src/crux/metrics/query.clj
```diff
@@ -1,20 +1,37 @@
...
-                (swap! !timer-store dissoc (:crux.query/query-id event))))
+                (fn [event]
+                  (dropwizard/stop (get @!timer-store (:crux.query/query-id event)))
+                  (swap! !timer-store dissoc (:crux.query/query-id event))))
     {:query-timer query-timer
      :current-query-count (dropwizard/gauge registry
                                             ["query" "currently-running"]
                                             (fn [] (count @!timer-store)))}))
+
+(defn add-slow-query-listeners
```
jarohen: could/should this be included in `assign-listeners`, above?

crux-test/test/crux/current_queries_test.clj
```diff
@@ -22,27 +25,68 @@
...
                     (map ::query-id))))
       (t/is (= [1 2]
-               (->> (clean-expired-queries queries
-                                           {::n/recent-queries-max-age (Duration/ofSeconds 8)
-                                            ::n/recent-queries-max-count 5})
+               (->> (clean-completed-queries queries
+                                             {::n/recent-queries-max-age (Duration/ofSeconds 8)
+                                              ::n/recent-queries-max-count 5})
+                    (map ::query-id))))))
+  (let [clean-slowest-queries @#'n/clean-slowest-queries
```
jarohen, line 52: given that these are now two code paths I'd split the tests into a different `t/deftest` - means that if one side breaks in the future you rerun fewer tests while you're fixing it

jarohen (changes requested):
Nice one Dan - a couple of moving-stuff-around suggestions but apart from that it's applied really cleanly, well done :smile:

crux-metrics/src/crux/metrics/query.clj
jarohen: that, or debating whether it has much to do with metrics at all, in hindsight. wdyt?

crux-core/src/crux/api.clj
```diff
@@ -212,7 +215,10 @@
     (map qs/<-QueryState (.activeQueries this)))
 
   (recent-queries [this]
-    (map qs/<-QueryState (.recentQueries this))))
+    (map qs/<-QueryState (.recentQueries this)))
+
+  (slowest-queries [this]
+    (map qs/<-QueryState (.slowestQueries this))))
```
jonpither, line 221: Question - do want to grow a list fns on the main API here, or have a single stats fn that returns a map of these?

crux-metrics/src/crux/metrics/query.clj
danmason: See latest commits - moved it out to the listeners within crux.node and did some refactoring. Unsure if it should, indeed, live there, but considering thats where the current configuration and listeners for it live it made sense.

crux-core/src/crux/api.clj
jarohen: Hey Jon - happy to go with this for now as we don't have any further additions in the pipeline - we've talked about splitting `ICruxAPI` in the past, maybe it's time to revisit that discussion?

PR #1028, opened 2020-08-12
Checkpointing and restoring query indices, #407

### LINKED ISSUE #407: Users can checkpoint/restore Crux nodes

jarohen:
resolves #407 

TODO: 
* [x] e2e test
* [x] upgrade Soak to 1.12.0
* [x] docs
* [ ] a blog

We split the job of checkpointing up into two protocols (names up for debate):
* a `CheckpointSource` (a data store that can yield checkpoints - all the KV stores now implement this protocol) 
* a `CheckpointStore` (a data store than can hold checkpoints - have written `crux.s3.checkpoint` but we could write many more)

Then, `crux.checkpoint/Checkpointer` wires up the two, and schedules the checkpointing job.

Basic configuration looks like this:

```clojure
{:crux/indexer {:kv-store {:crux/module `crux.kv.rocksdb/->kv-store
                           :db-dir ...
                           :checkpointer {:crux/module `cp/->checkpointer
                                          :store {:crux/module `s3.cp/->cp-store
                                                  :bucket "..."}
                                          :approx-frequency (Duration/ofHours 6)}}}}
```

* We use `approx-frequency` because Crux nodes form an un-coordinated cluster - we don't need _every_ node checkpointing every 6 hours. `approx-frequency` aims to have _one_ node submit a checkpoint about this often, and for the work to be shared between nodes (so that it's not the same node every time). This works by splitting time into periods of this length, and each node choosing a random time within the window. To prevent two checkpoints close together, a node won't take a checkpoint if it's been under half (tbd) the duration since the previous one - this means that the time between checkpoints will range between 0.5x and 2x freq.
* Checkpoints have a `cp-format`, which determines whether a stored checkpoint is suitable for restore. Examples of unsuitable checkpoints would be after an index version bump, or a different major version of the KV store.
* Checkpoint retention periods are assumed to be covered by the individual `CheckpointStore` - S3, for example, would use built-in lifecycle rules


iku000888:
This is def. a wanted feature!

iku000888 (approved):

crux-core/src/crux/checkpoint.clj
```diff
@@ -0,0 +1,111 @@
+(ns crux.checkpoint
```
jonpither, line 1: Feels like there's some unused ns requires/imports

examples/soak/src/crux/soak/main.clj
```diff
@@ -178,9 +180,9 @@
 
 (defmethod ig/init-key :soak/crux-node [_ node-opts]
   (let [node (api/start-node node-opts)]
-    (log/info "Loading Weather Data...")
-    (api/sync node)
-    (log/info "Weather Data Loaded!")
+    ;; (log/info "Loading Weather Data...")
+    ;; (api/sync node)
+    ;; (log/info "Weather Data Loaded!")
```
jonpither: seems immaterial to this commit

jonpither (approved):
Nice - I like iit. Like the concept of a checkpoint store and the testing looks good.

jonpither:
I couldn't tell, but I assume the old code we had has been purged?

jarohen:
> I couldn't tell, but I assume the old code we had has been purged?

@jonpither - thanks for the review :) yep - already purged on master as part of the module change, if memory serves

examples/soak/src/crux/soak/main.clj
jarohen: reverted :)

crux-core/src/crux/mem_kv.clj
```diff
@@ -85,31 +89,44 @@
...
-                                       :doc "Directory to (optionally) store K/V files"
-                                       :spec ::sys/path}
-                              :persist-on-close? {:required? true
-                                                  :default false
-                                                  :spec ::sys/boolean}}}
+    (cio/try-close cp-job)))
+
+(def ^:private cp-format
+  {:index-version c/index-version
+   ::version 1})
```
hraberg: Just wondering if this is correct, see the other KV stores have string versions?

crux-core/src/crux/io.clj
```diff
@@ -104,6 +104,9 @@
 
     f))
 
+(defn empty-folder? [^java.io.File f]
+  (<= 1 (count (take 2 (file-seq f)))))
```
hraberg: I would change this to something like `(empty? (.listFiles f))`.

hraberg (approved):

crux-core/src/crux/io.clj
jarohen: mm, me too. I wonder why I did it that way...

crux-core/src/crux/mem_kv.clj
jarohen: doesn't strictly matter so long as it's consistent for mem-kv - but yes, agreed, consistency's better :)

PR #1031, opened 2020-08-13
Data browser - Keep tab state

### LINKED ISSUE #1004: Don't lose tab state when switching between tabs

danmason:
(Currently based on top of #1017)

Resolves #1004 - stores previous query params of console-tab and loads them when tab is clicked, also preventing unnecessary re-fetching of data.

Some additional points/bug fixes:
- Small bug I noticed while testing this (unrelated to the PR itself): Attempting to delete elements of query history actually deletes elements at other positions - likely because the list is being reversed. Will push a fix on this branch (though could well bring in seperately)
- In entity, `not-found` and `missing eid` messages were not showing as errors - returning them from EDN within an error map serves as a fix. 
- Empty set of results from query was no longer showing 'no results found' - instead showing an empty table.

### DIFF
--- a/crux-http-server/src-cljs/crux/ui/events.cljs
+++ b/crux-http-server/src-cljs/crux/ui/events.cljs
@@ -43,8 +43,12 @@
 
 (rf/reg-event-fx
  ::console-tab-selected
- (fn [_ [_ tab]]
-   {:dispatch [:navigate tab {} {}]}))
+ (fn [{:keys [db]} [_ tab]]
+   (let [current-tab (get-in db [:current-route :data :name] :query)
+         current-params (get-in db [:current-route :query-params])]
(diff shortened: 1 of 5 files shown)


danmason:
Small bug I noticed while testing this (unrelated to the PR itself): Attempting to delete elements of query history actually deletes elements at other positions - likely because the list is being reversed. Will push a fix on this branch (though could well bring in seperately)

danmason:
Another quick couple:
- In entity, `not-found` and `missing eid` messages were not showing as errors - returning them from EDN within an error map serves as a fix. 
- Empty set of results from query was no longer showing 'no results found' - instead showing an empty table.

Will also add these to this PR - though happy to move it out.

PR #1044, opened 2020-08-19
Nippy 2.15.1 upgrade

keytiong:
A  [security advisory](https://github.com/ptaoussanis/nippy/issues/130) for users of nippy older than 2.15.0 to upgrade to 2.15.0.

The PR is to upgrade crux's nippy version to 2.15.0 plus associated dependency conflict resolutions due to the upgrade.


jarohen:
Hi @keytiong - many thanks for bringing this to our attention.

We'll unfortunately need to check whether we were inadvertently depending on the `Serializable` behaviour before we merge this one - will investigate.

Cheers!

James

jarohen:
Have included a `crux.nippy.allow-serializable-classes` JVM property so that users can specify an allow-list of classes that they're already using in their Crux documents. Pulling in @hraberg for another pair of eyes.

jarohen:
Let's see if we can put that JVM property in Nippy instead ^^

hraberg (approved):
IIRC there was a reason we're stuck on this version of Nippy and have an explicit version of encore we depend on. But cannot remember why atm.

That said, this exposure of general serialised Java objects is likely something we want to close down before it gets to Nippy, so that's another way of dealing with this, but for now, assuming everything still works, as I don't recall what it was that kept us back, let's go ahead with this.

jarohen:
Merged - thanks again @keytiong :smile:

PR #1048, opened 2020-08-21
Bring back the beautiful pictures!

mikepjb:
I think crux-venn-1.svg and crux-node-1.svg were moved from a main -> about folder and this broke their links


jarohen:
Cheers! Yep, we've been moving things around :)

PR #1054, opened 2020-08-24
Transaction function race condition #1049

### LINKED ISSUE #1049: Potential race condition using transaction functions
See #1045

jarohen:
resolves #1049, see #1045 for the original report.

When a transaction function runs successfully, it replaces the argument doc with the yielded tx-events. This is to ensure that, should one of the documents read by the transaction function later get evicted, the result of the transaction remains unaltered in subsequent replays of the tx-log - we wouldn't want a transaction to commit first time through and abort in a later run-through.

So, when a transaction function is evaluated, we either have two states - whether or not the arg doc has yet been replaced. If it hasn't, we eval the function and replace the document; if it has, we simply yield the tx-events calculated by an earlier node. It's not an issue for two nodes to both eval the function and both replace the document, because the transaction function is pure and deterministic (we can replace a call to such a function with its result) and the update idempotent.

This race condition was caused by two reads of the argument doc - the first picking up the un-replaced doc, the second the replaced doc. This was a problem because we only indexed documents referenced in the arg-doc on the first read, not on the second - so any documents yielded by the transaction function were not indexed in this case. We've fixed the issue by doing all of the document indexing based on the second read.

Separately (not in this PR, now) is handling document stores being eventually consistent, ensuring that all document stores behave the same way as Kafka currently does - i.e. waiting until all the documents are present before returning. We did see another issue whereby if the arg doc was replaced before the yielded docs were written, the same symptoms occurred.

### DIFF
--- a/crux-core/src/crux/tx.clj
+++ b/crux-core/src/crux/tx.clj
@@ -163,14 +163,15 @@
 
 (defmethod index-tx-event :crux.tx/fn [[op k args-doc :as tx-op]
                                        {:crux.tx/keys [tx-time tx-id] :as tx}
-                                       {:keys [query-engine index-store], :as tx-ingester}]
+                                       {:keys [query-engine document-store index-store], :as tx-ingester}]
   (let [fn-id (c/new-id k)
         {args-doc-id :crux.db/id, :crux.db.fn/keys [args tx-events failed?]} args-doc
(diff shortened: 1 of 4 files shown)


hraberg (approved):
I don't fully grok this change, but seems reasonable from skimming the problem description, and as it adds a test and everything else still passes, I assume it's OK.

PR #1055, opened 2020-08-24
Ensure that ints, shorts, bytes and floats as keys are translated to longs and doubles respectively

### LINKED ISSUE #1043: Unable to query documents with an `Integer` eid
from Slack

```
(def my-node
  (crux/start-node
    {:crux.node/topology ['crux.standalone/topology]}))

(crux/submit-tx my-node [[:crux.tx/put
                          {:crux.db/id (int 10)
                           :name "Foo Bar"}]])

(crux/q (crux/db my-node)
        '{:find [e]
          :where [[e :name _]]})
```

returns an empty set.

jarohen:
fixes #1043

If we were to be pedantic, this would require an index version bump given that documents with the aforementioned key types will have a different representation in the indices after this change. Discussion required!

Options:
* No bump - make it clear in the changelog
* Bump - make everyone re-index
* Some kind of flag that says 'trust me, I don't need one this time'?

### DIFF
--- a/crux-core/src/crux/codec.clj
+++ b/crux-core/src/crux/codec.clj
@@ -362,6 +362,22 @@
   (id->buffer [this to]
     (id-function to (nippy/fast-freeze this)))
 
+  Byte
+  (id->buffer [this to]
+    (id->buffer (long this) to))
+
+  Short
+  (id->buffer [this to]
+    (id->buffer (long this) to))
+
+  Integer
+  (id->buffer [this to]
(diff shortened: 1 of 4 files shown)


hraberg (approved):
Guess the overall "Number" could go, as we don't support BigDecimal etc, just things that can be coerced to long or double.

jarohen:
Thanks for this @hraberg, and for your other recent reviews too :)

What're your thoughts on the index version bump question?

PR #1068, opened 2020-08-28
Paginated history entry 1006

### LINKED ISSUE #1006: Entity history API/UI should be paginated if required
* [x] Ensure that the entity-history HTTP API streams its result, so that naively using `take` on the remote-client API doesn't realise the whole history.
* [x] Add pagination params to entity-history endpoint
* [x] Add 'link' HTTP headers for paginated requests so that consumers can easily get the URL of the next page

Thoughts:

* Let's not use a numeric 'offset' as we can't honour this efficiently (ie without calculating the elements we're skipping over) - we'd be looking for a way for consumers to skip straight to the start of the next page. (We order entity history by valid-time then transaction-time, in whatever direction the user's requested)
* Should we expose this through the local API too? We currently rely on the user calling `drop`.
* Re default limits - I'd propose we have a hard maximum limit for the HTML rendering (100, say) but not for other representations.

rhinocratic:
Added options `limit` (number of items per page) and `resume-from-tx-id` (transaction id of the first item in a page of history) to the options for requesting entity histories.

The header of the response to a history request contains a `Link` header with `rel=next` containing the URL of the next page.  Any history options contained in the original request are propagated to the link, as is the latest transaction time upon receipt of the first page request.

### DIFF
--- a/crux-http-server/src/crux/http_server/entity.clj
+++ b/crux-http-server/src/crux/http_server/entity.clj
@@ -164,6 +164,7 @@
     mfc/EncodeToOutputStream
     (encode-to-output-stream [_ {:keys [entity entity-history limit]} _]
       (fn [^OutputStream output-stream]
+        (println "Limit: " limit)
         (with-open [w (io/writer output-stream)]
           (if entity-history
             (if limit
@@ -230,7 +231,7 @@
                                 end-valid-time (assoc :end-valid-time end-valid-time)
                                 with-corrections? (assoc :with-corrections with-corrections?)
                                 with-docs? (assoc :with-docs with-docs?))]
-        (assoc {:entity-history (butlast page)} :continuation-opts continuation-opts)))))
+        (assoc {:entity-history (butlast page) :limit limit} :continuation-opts continuation-opts)))))
 
 (defn search-entity-history [{:keys [crux-node eid valid-time transaction-time sort-order history-opts limit resume-after-tx-id] :as params}]
   (try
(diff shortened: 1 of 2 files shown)


jarohen (changes requested):
Thanks @rhinocratic

Unfortunately the test (and hence, I'm fairly sure, the underlying code) doesn't cover the case we talked about - entity history returns the history of the entity sorted first by valid time and then by transaction time. In this case, the test's inserting data where valid-time is (almost) the same as transaction-time and there are no two entities with the same valid-time - so the second half of the sort isn't exercised.

We talked about an example where there was multiple revisions (throughout transaction-time) at each valid-time - say, ten transactions (let's call them tt 0 through 9) each submitting a single entity at ten different valid times (vt 0 through 9). If you were to make an entity history request, ascending, limit 15, you'd expect to see the first ten entries be vt 0, tt 0 through 9; the remaining 5 at vt 1, tt 0 through 4. At this point, the next page has start vt 1 (we update the start vt because it's the first entry in the sort order), start tt unbounded (because it's the second entry - we don't want vt 2 to have start tt 4), resume tx-id 4.

rhinocratic:
I've changed the tests to send batches of transactions, where the valid time for successive batches cycles through a predefined list.  There's a minor change to the URL building in entity.clj, but the method seems to return the expected results.

danmason (approved):
WIll need a separate push of work to fix it on the console (currently displays blank), but otherwise approved.

jarohen:
Have merged the streaming changes in separately as #1075, will park the pagination changes temporarily while we're making other changes to that API, and then look to re-incorporate this PR.

PR #1075, opened 2020-09-04
Streaming results in the new entity history HTTP API

jarohen:
Separating out the #1068 PR, this is a PR just containing the streaming API changes.

* [x] Check that the console UI works as before

### DIFF
--- a/crux-http-server/src/crux/http_server/entity.clj
+++ b/crux-http-server/src/crux/http_server/entity.clj
@@ -1,23 +1,22 @@
 (ns crux.http-server.entity
-  (:require [crux.http-server.util :as util]
-            [crux.http-server.entity-ref :as entity-ref]
-            [cognitect.transit :as transit]
-            [clojure.edn :as edn]
+  (:require [clojure.edn :as edn]
             [clojure.instant :as instant]
+            [clojure.java.io :as io]
             [clojure.string :as string]
-            [clojure.set :as set]
-            [crux.api :as api]
+            [cognitect.transit :as transit]
+            [crux.api :as crux]
             [crux.codec :as c]
+            [crux.http-server.entity-ref :as entity-ref]
+            [crux.http-server.util :as util]
+            [crux.io :as cio]
             [muuntaja.core :as m]
-            [muuntaja.format.core :as mfc]
-            [crux.api :as crux])
-  (:import [crux.api ICruxAPI NodeOutOfSyncException]
+            [muuntaja.format.core :as mfc])
+  (:import crux.http_server.entity_ref.EntityRef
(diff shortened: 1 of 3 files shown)


crux-http-server/src/crux/http_server/entity.clj
```diff
@@ -1,23 +1,22 @@
...
-            [cognitect.transit :as transit]
-            [clojure.edn :as edn]
+  (:require [clojure.edn :as edn]
             [clojure.instant :as instant]
+            [clojure.java.io :as io]
             [clojure.string :as string]
-            [clojure.set :as set]
-            [crux.api :as api]
+            [cognitect.transit :as transit]
+            [crux.api :as crux]
```
danmason, line 7: This is somewhat inconsistent with the naming elsewhere (see `http-server.clj` and the various other files in the `http-server` folder) - likely worth deciding if, within this section of the code, we should refer to crux calls as `crux/...` or `api/...`

crux-http-server/src/crux/http_server/entity_ref.cljc
```diff
@@ -26,11 +26,6 @@
       (.write w "#crux.http/entity-ref ")
       (print-method (:eid ref) w)))
 
-#? (:clj
```
danmason, line 29: Question: Is print-dup EntityRef now completely unused?

danmason (approved):
Happy with the change as a whole (so long as the Ui and tests still work fine then I can't see it causing any issues) - though I left a couple of nitpick/comments below.

crux-http-server/src/crux/http_server/entity.clj
jarohen: I'm trying to standardise on `crux.api :as crux`, because this is what I'd like it to be in any example code - our users won't be requiring `crux.api :as api`.

crux-http-server/src/crux/http_server/entity_ref.cljc
jarohen: I learned that we can't use `print-dup` for EDN, because it tries to preserve the concrete types (e.g.) of maps - `{:a 1, :b 2}` gets printed as `#=(clojure.lang.PersistentArrayMap/create ...)`, which isn't valid EDN.

Will double-check that we don't use it, though

jarohen:
Thanks @danmason, good feedback - will get this merged.

crux-http-server/src/crux/http_server/entity.clj
danmason: Sounds good - likely worth us going through and ensuring that's consistent everywhere else?

PR #1079, opened 2020-09-07
Or join bound vars

hraberg:
Previous behaviour looked like rules:
`(or-join [bound-var-x bound-var-y] free-var-z free-var-w)`

Correct behaviour in EDN Datalog is to add a level of nesting, like this:
`(or-join [[bound-var-x bound-var-y] free-var-z free-var-w])`

Normally an `or-join` looks like this, and only has free vars:
`(or-join [x y])`

This implies that previously `[x y]` were treated as bound vars, though Crux did demote them to free vars if necessary if they couldn't be bound in the parent query. New behaviour will treat `[x y]` free vars, but promote them if possible if they can be bound in the outer query.

Rules rely on bound vars, and compile down to or-join, but they directly generate an AST node, so they didn't run into this parser level issue.

### DIFF
--- a/crux-sql/test/crux/calcite_test.clj
+++ b/crux-sql/test/crux/calcite_test.clj
@@ -25,7 +25,7 @@
 (t/use-fixtures :each fix/with-standalone-topology cf/with-calcite-module fix/with-kv-dir fix/with-node with-each-connection-type with-sql-schema)
 
 (defn- inst->iso-str [^java.util.Date t]
-  (.format (ZonedDateTime/ofInstant (.toInstant t) (ZoneId/systemDefault)) DateTimeFormatter/ISO_INSTANT))
+  (.format (ZonedDateTime/ofInstant (.toInstant t) (ZoneId/of "UTC")) DateTimeFormatter/ISO_INSTANT))
 
 (t/deftest test-valid-time
   (let [id (java.util.UUID/randomUUID)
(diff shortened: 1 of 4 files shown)


crux-sql/test/crux/calcite_test.clj
```diff
@@ -25,7 +25,7 @@
 (t/use-fixtures :each fix/with-standalone-topology cf/with-calcite-module fix/with-kv-dir fix/with-node with-each-connection-type with-sql-schema)
 
 (defn- inst->iso-str [^java.util.Date t]
-  (.format (ZonedDateTime/ofInstant (.toInstant t) (ZoneId/systemDefault)) DateTimeFormatter/ISO_INSTANT))
+  (.format (ZonedDateTime/ofInstant (.toInstant t) (ZoneId/of "UTC")) DateTimeFormatter/ISO_INSTANT))
```
jonpither, line 28: Would put this in a different PR, or leave out entirely due to the related issue you raised.

crux-test/test/crux/replay_test.clj
```diff
@@ -46,7 +46,7 @@
                                            :crux.standalone/event-log-kv-store 'crux.kv.rocksdb/kv})]
           (t/is (= {:crux.tx/tx-id (dec n)}
                    (crux/latest-submitted-tx node)))
-          (t/is (crux/sync node (Duration/ofSeconds 5)))
+          (t/is (crux/sync node (Duration/ofSeconds 10)))
```
jonpither, line 49: Again, would probably do outside of this PR, but a nit pick.

jonpither (approved):

jarohen (approved):

PR #1082, opened 2020-09-07
Migrate HTTP api

### LINKED ISSUE #803: REST API /history/[:key] endpoint does not allow to pass a map as the :key

### LINKED ISSUE #1032: Crux HTTP API migrated to the new console API
We've recently been developing a parallel HTTP API for the console - let's make this the main Crux HTTP API.

* Port over the remaining HTTP endpoints (there are some in the current HTTP API that aren't yet reflected in the new API)
* Ensure we're happy to commit to the new HTTP API (ensure it's consistent, simple, etc)
* Make the remote client API use the new HTTP API
* Decomm the old API
* Use a 'proper' HTTP routing lib
* Parse parameters consistently between handlers (there'll be a modern Clojure pattern/middleware here, no doubt) and return decent error messages

danmason:
Resolves #1032, #803  - using reitit for routing and proper organized coercion of params against a `spec`.

TODO:
- [x] Clarify the use of crux/ids in the main API - currently need write handlers to pass them over the wire, should this be something we are exposing anyway?
- [x] Merge regular + console endpoints, and under same handler
- [x] Use reitit for routing
- [x] Ensure all routes use muuntaja for negotiation (?)
--> This currently supports EDN and transit+json as outputs, only currently accept EDN inputs
- [x] Ensure all non console routes use clojure API rather than Java API
- [x] Use reitit for handling spec validation + coercion of params
- [x] Migrate remote API to use new routes
- [x] Error messages for spec/coercion fail.
- [x] Cleanup remote API - no string encoding.
- [x] Cleanup the http-server namespace of unused content.
- [x] All endpoints prefixed with `_crux`.
- [x] Should empty `attribute-stats` call return an empty map or a nil?
- [x] Update REST API docs
- [x] Remove HTTP Special case from thrown (API test)
- [x] Remote API use post `query` - ensure ui route test for GET query still works.
- [x] Add transit decoder for body - muuntaja (and test it)
- [ ] Exception handler for muuntaja decode - likely want to think further on exception handlers
- [ ] May want some more specific tests for exceptions in `ui-routes-test`

Among these changes is a refactor and cleanup to the behaviour of `http-server.entity`.

Note: Errors are now (generally) thrown across the board, and error handling is done before muuntaja formatting (such that errors will now be formatted by muuntaja) - this also means that coercion errors within `query`/`entity` will still result in the user getting a page back when requesting HTML.

Likely some updates wanted to the REST API changes (minor tweaks, and perhaps a small explainer on types you're expected to pass?) - can see it here https://www.opencrux.com/http/reference/http.html

### DIFF
--- a/crux-http-client/src/crux/remote_api_client.clj
+++ b/crux-http-client/src/crux/remote_api_client.clj
@@ -83,9 +83,9 @@
        nil
 
        (= 400 status)
-       (let [{:crux.error/keys [error-key] :as error-data} (edn/read-string (cond-> body
-                                                                              (= :stream (:as http-opts)) slurp))]
-         (throw (ce/illegal-arg error-key error-data )))
+       (let [{::ce/keys [error-key] :as error-data} (edn/read-string (cond-> body
+                                                                       (= :stream (:as http-opts)) slurp))]
+         (throw (ce/illegal-arg error-key error-data)))
 
        (and (<= 200 status) (< status 400))
        (if (string? body)
--- a/crux-http-client/src/crux/remote_api_client.clj
+++ b/crux-http-client/src/crux/remote_api_client.clj
@@ -1,34 +1,31 @@
 (ns ^:no-doc crux.remote-api-client
   (:require [clojure.edn :as edn]
-            [clojure.string :as str]
-            [crux.io :as cio]
-            [crux.db :as db]
+            [clojure.string :as string]
             [crux.codec :as c]
+            [crux.error :as ce]
+            [crux.io :as cio]
             [crux.query-state :as qs]
-            [crux.query :as q]
-            [clojure.string :as string])
-  (:import (java.io Closeable InputStreamReader IOException PushbackReader)
-           (java.time Instant Duration)
-           java.util.Date
-           java.util.function.Supplier
-           com.nimbusds.jwt.SignedJWT
-           (crux.api Crux ICruxAPI ICruxDatasource NodeOutOfSyncException
-                     HistoryOptions HistoryOptions$SortOrder
-                     RemoteClientOptions)))
+            [crux.error :as ce])
+  (:import com.nimbusds.jwt.SignedJWT
+           [crux.api HistoryOptions HistoryOptions$SortOrder ICruxAPI ICruxDatasource NodeOutOfSyncException RemoteClientOptions]
+           [java.io Closeable InputStreamReader IOException PushbackReader]
+           java.time.Instant
+           java.util.function.Supplier))
 
 (defn- edn-list->lazy-seq [in]
-  (let [in (PushbackReader. (InputStreamReader. in))
-        open-paren \(]
-    (when-not (= (int open-paren) (.read in))
-      (throw (RuntimeException. "Expected delimiter: (")))
-    (->> (repeatedly #(try
-                        (edn/read {:readers {'crux/id c/id-edn-reader}
-                                   :eof ::eof} in)
-                        (catch RuntimeException e
-                          (if (= "Unmatched delimiter: )" (.getMessage e))
-                            ::eof
-                            (throw e)))))
-         (take-while #(not= ::eof %)))))
+  (let [in (PushbackReader. (InputStreamReader. in))]
+    (condp = (.read in)
+      -1 nil
+      (int \() (->> (repeatedly #(try
+                                   (edn/read {:readers {'crux/id c/id-edn-reader}
+                                              :eof ::eof} in)
+                                   (catch RuntimeException e
+                                     (if (= "Unmatched delimiter: )" (.getMessage e))
+                                       ::eof
+                                       (throw e)))))
+                    (take-while #(not= ::eof %)))
+
+      (throw (RuntimeException. "Expected delimiter: (")))))
 
 (def ^{:doc "Can be rebound using binding or alter-var-root to a
   function that takes a request map and returns a response
@@ -80,16 +77,15 @@
                                                                {"Authorization" (str "Bearer " (->jwt-token))}))
                                              :body (some-> body cio/pr-edn-str)
                                              :accept :edn}
-                                            http-opts))]
+                                            (update http-opts :query-params #(into {} (remove (comp nil? val) %)))))]
      (cond
        (= 404 status)
        nil
 
        (= 400 status)
-       (let [{:keys [^String cause data]} (edn/read-string (cond-> body
-                                                             (= :stream (:as http-opts)) slurp))]
-         (throw (IllegalArgumentException. cause (when data
-                                                   (ex-info cause data)))))
+       (let [{:crux.error/keys [error-key] :as error-data} (edn/read-string (cond-> body
+                                                                              (= :stream (:as http-opts)) slurp))]
+         (throw (ce/illegal-arg error-key error-data )))
 
        (and (<= 200 status) (< status 400))
        (if (string? body)
@@ -119,15 +115,19 @@
 
   ICruxDatasource
   (entity [this eid]
-    (api-request-sync (str url "/entity/" (str (c/new-id eid)))
-                      {:body (as-of-map this)
-                       :http-opts {:method :get}
-                       :->jwt-token ->jwt-token}))
+    (api-request-sync (str url "/_crux/entity")
+                      {:->jwt-token ->jwt-token
+                       :http-opts {:method :get
+                                   :query-params {:eid (pr-str eid)
+                                                  :valid-time (some-> valid-time (cio/format-rfc3339-date))
+                                                  :transact-time (some->  transact-time (cio/format-rfc3339-date))}}}))
 
   (entityTx [this eid]
-    (api-request-sync (str url "/entity-tx/" (str (c/new-id eid)))
-                      {:body (as-of-map this)
-                       :http-opts {:method :get}
+    (api-request-sync (str url "/_crux/entity-tx")
+                      {:http-opts {:method :get
+                                   :query-params {:eid (pr-str eid)
+                                                  :valid-time (some-> valid-time (cio/format-rfc3339-date))
+                                                  :transact-time (some->  transact-time (cio/format-rfc3339-date))}}
                        :->jwt-token ->jwt-token}))
 
   (query [this q args]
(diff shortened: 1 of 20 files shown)


danmason:
Current issue - query.csv negotiation isn't working properly post changes to muuntaja

crux-http-client/src/crux/remote_api_client.clj
```diff
@@ -113,21 +113,35 @@
     valid-time (assoc :valid-time valid-time)
     transact-time (assoc :transact-time transact-time)))
 
+(defn edn->query-params
```
danmason: No need for this - can always use `q` param to pass the query.

crux-http-server/src/crux/http_server.clj
```diff
@@ -1,269 +1,194 @@
...
-    (and (re-find path-pattern path)
-         (some #{method} valid-methods))))
-
-(defn- response
-  ([status headers body]
-   {:status status
-    :headers headers
-    :body body}))
 
 (defn- redirect-response [url]
```
danmason: Unnecessary - see ring util response

crux-http-client/src/crux/remote_api_client.clj
```diff
@@ -1,26 +1,23 @@
...
+           [java.io Closeable InputStreamReader IOException PushbackReader]
+           java.time.Instant
+           java.util.function.Supplier))
 
 (defn- edn-list->lazy-seq [in]
   (let [in (PushbackReader. (InputStreamReader. in))
         open-paren \(]
-    (when-not (= (int open-paren) (.read in))
-      (throw (RuntimeException. "Expected delimiter: (")))
+    (let [first-char (.read in)]
```
jarohen: reckon this'd be clearer as a `cond`, maybe even a `(condp = (.read in) ...)`

crux-http-client/src/crux/remote_api_client.clj
```diff
@@ -219,71 +221,75 @@
...
+                        :->jwt-token ->jwt-token})
+     (get :crux.tx/tx-time)))
 
   (awaitTxTime [_ tx-time timeout]
-    (api-request-sync (cond-> (str url "/await-tx-time?tx-time=" (cio/format-rfc3339-date tx-time))
-                        timeout (str "&timeout=" (cio/format-duration-millis timeout)))
-                      {:http-opts {:method :get}
-                       :->jwt-token ->jwt-token}))
+    (->
+     (api-request-sync (str url "/_crux/await-tx-time" )
```
jarohen: nit: trailing space, and can we have the first arg to `->` on the same line as the `->`?

crux-http-client/src/crux/remote_api_client.clj
```diff
@@ -219,71 +221,75 @@
...
+                                    :query-params {:tx-time (cio/format-rfc3339-date tx-time)
+                                                   :timeout (some-> timeout (cio/format-duration-millis))}}
+                        :->jwt-token ->jwt-token})
+     (get :crux.tx/tx-time)))
 
   (awaitTx [_ tx timeout]
-    (api-request-sync (cond-> (str url "/await-tx?tx-id=" (:crux.tx/tx-id tx))
-                        timeout (str "&timeout=" (cio/format-duration-millis timeout)))
-                      {:http-opts {:method :get}
+    (api-request-sync (str url "/_crux/await-tx" )
```
jarohen: nit: another cheeky trailing space

crux-http-server/src-cljs/crux/ui/events.cljs
```diff
@@ -20,6 +20,7 @@
                               (str "meta[title=" title "]"))
                              (.getAttribute "content"))
          edn-content (reader/read-string {:readers {'object pr-str
+                                                    'crux/id #(str %)
```
jarohen: nit: `#(str %)` === `str`

crux-http-server/src-cljs/crux/ui/http.cljs
```diff
@@ -12,7 +12,10 @@
 
 (def edn-response-with-readers
   (map->ResponseFormat {:read (fn [xhrio]
-                                (edn/read-string {:readers {'crux.http/entity-ref entity-ref/->EntityRef}} (-body xhrio)))
+                                (edn/read-string
+                                 {:readers {'crux.http/entity-ref entity-ref/->EntityRef
+                                            'crux/id #(str %)}}
```
jarohen: nit: `#(str %)` === `str`

crux-http-server/src/crux/http_server/entity.clj
```diff
@@ -128,61 +149,67 @@
 
 (defn ->entity-html-encoder [opts]
   (reify mfc/EncodeToBytes
-    (encode-to-bytes [_ {:keys [eid no-entity? not-found? error entity ^Closeable entity-history] :as res} charset]
+    (encode-to-bytes [_ {:keys [eid no-entity? not-found? cause entity ^Closeable entity-history] :as res} charset]
```
jarohen: why `cause` here rather than `error`?

crux-http-server/src/crux/http_server/entity.clj
```diff
@@ -128,61 +149,67 @@
...
-            (print-method entity w)))))))
+          (cond
+            entity-history (try
+                             (if (.hasNext entity-history)
+                               (print-method (iterator-seq entity-history) w)
+                               (.write w ^String (pr-str '())))
+                             (finally
+                               (cio/try-close entity-history)))
+             entity (print-method entity w)
+            :else (.write w ^String (pr-str res))))))))
```
jarohen, line 194: this seems like it'd return something like `{:entity nil}` for a not-found response? think we said to leave the response body empty in this case?

crux-http-server/src/crux/http_server/entity.clj
```diff
@@ -128,61 +149,67 @@
...
-          (if entity-history
-            (try
-              (print-method (iterator-seq entity-history) w)
-              (finally
-                (cio/try-close entity-history)))
-            (print-method entity w)))))))
+          (cond
+            entity-history (try
+                             (if (.hasNext entity-history)
+                               (print-method (iterator-seq entity-history) w)
```
jarohen: might be clearer to go with `(print-method (or (iterator-seq entity-history) '()) w) ` rather than this `if`?

crux-http-server/src/crux/http_server/entity.clj
```diff
@@ -128,61 +149,67 @@
...
-              (finally
-                (cio/try-close entity-history)))
-            (transit/write w entity)))))))
+        (let [w (transit/writer output-stream :json {:handlers {EntityRef entity-ref/ref-write-handler
+                                                                Id util/crux-id-write-handler}})]
+          (cond
+            entity (transit/write w entity)
+            entity-history (try
+                             (if (.hasNext entity-history)
+                               (transit/write w (iterator-seq entity-history))
```
jarohen: likewise

crux-http-server/src/crux/http_server/entity.clj
```diff
@@ -197,85 +224,65 @@
...
+                              :crux.tx/tx-time end-transaction-time}}
           entity-history (crux/open-entity-history db eid sort-order history-opts)]
-      (if-not (.hasNext entity-history)
-        {:not-found? true}
-        {:valid-time (crux/valid-time db)
-         :transaction-time (crux/transaction-time db)
-         :entity-history (cio/fmap-cursor (fn [entity-history]
-                                            (map #(update % :crux.db/content-hash str) entity-history))
-                                          entity-history)}))
+      {:entity-history (cio/fmap-cursor identity entity-history)})
```
jarohen: `(fmap-cursor identity entity-history)` === `entity-history`

crux-http-server/src/crux/http_server/query.clj
```diff
@@ -63,56 +102,44 @@
...
   (if (vector? param) param [param]))
 
-(defn- build-query [{:strs [find where args order-by limit offset full-results link-entities?]}]
-  (let [new-offset (if offset
-                     (Integer/parseInt offset)
-                     0)]
-    (cond-> {:find (c/read-edn-string-with-readers find)
-             :where (->> where vectorize-param (mapv c/read-edn-string-with-readers))
+
+;; TODO: Doesn't support all clauses a query can have - should do.
```
jarohen: what we doing about this TODO? :smile:

docs/reference/modules/ROOT/pages/http.adoc
```diff
@@ -1,6 +1,6 @@
 = HTTP
 
-Crux offers a REST API layer in the `crux-http-server` module that allows you to send transactions and run queries over HTTP.
+Crux offers a REST API layer in the `crux-http-server` module that allows you to send transactions and run queries over HTTP. (See the REST API reference xref:http-api.adoc[here])
```
jarohen:
```
Crux offers a xref:...[REST API] layer in the `crux-http-server` module that allows you to send transactions and run queries over HTTP.
```
?

Also, think we talked about putting the REST API at the bottom of this file and linking to it?

jarohen: Ah, we did, and you have - think this is a broken link now?

docs/reference/modules/ROOT/pages/http.adoc
```diff
@@ -73,183 +73,276 @@ If the node was started on `localhost:3000`, you can connect to it by doing the
...
-|uri|method|description
-|<<#entity, `/entity/[:key]`>>|GET|Returns an entity for a given ID and optional valid-time/transaction-time co-ordinates
-|<<#entity-tx, `/entity-tx/[:key]`>>|GET|Returns the transaction that most recently set a key
-|<<#entity, `/entity-history/[:key]`>>|GET|Returns the history of the given entity and optional valid-time/transaction-time co-ordinates
-|<<#query, `/query`>>|POST|Takes a datalog query and returns its results
-|<<#sync, `/sync`>>|GET| Wait until the Kafka consumer's lag is back to 0
-|<<#tx-log, `/tx-log`>>|GET| Returns a list of all transactions
-|<<#tx-log-post, `/tx-log`>>|POST|The "write" endpoint, to post transactions.
-|===
+All of the REST endpoints return both `application/edn` and `application/transit+json` (see transit https://github.com/cognitect/transit-clj[here]) - individual endpoints may return additional types - see the docs below. To set up your node to with the REST API, see the xref:http.adoc[HTTP Feature Module].
```
jarohen: can use '`application/transit+json`' as the link text, rather than a separate 'see ...'

jarohen: With adoc, I'm reliably informed (:smile:) that the idiom is one sentence per line

jarohen: I'd break the sentence before 'individual endpoints', too.

Also, 'see the HTTP Feature module' - you're already there :)

jarohen: in short:
```
All of the REST endpoints return both `application/edn` and xref:https://...[`application/transit+json`]
Individual endpoints may return additional types - see their docs below. 
```

docs/reference/modules/ROOT/pages/http.adoc
```diff
@@ -73,183 +73,276 @@ If the node was started on `localhost:3000`, you can connect to it by doing the
...
 ----
-{:crux.db/id :tommy, :name "Tommy", :last-name "Petrov"}
+curl -X GET \
+     -H "Content-Type: application/edn" \
+     $CRUX_URL/_crux/entity?eid=:tommy&history=true&sort-order=asc
 ----
 
+.*Required Parameters*
+* `eid` (Crux ID)
+* `sort-order` (_required_, either `asc` or `desc`)
```
jarohen: it's already in the 'required' section

docs/reference/modules/ROOT/pages/http.adoc
```diff
@@ -73,183 +73,276 @@ If the node was started on `localhost:3000`, you can connect to it by doing the
...
+* `eid` (Crux ID)
+
+.*Optional Parameters*
+* `valid-time` (date, defaulting to now)
+* `transaction-time` (date, defaulting to latest transaction time)
+
+[#query]
+=== GET `/_crux/query`
+
+Takes a datalog query and returns its results. Results are also available in `text/csv` and `text/tsv` formats (can force negotiation of these by using the `/_crux/query.csv` and `/_crux/query.tsv` endpoints respectively).
```
jarohen, line 167: I wonder whether we should extend this both to `.edn`, and to the other endpoints, as a general pattern - wdyt?

(Maybe not one for this PR)

docs/reference/modules/ROOT/pages/http.adoc
```diff
@@ -73,183 +73,276 @@ If the node was started on `localhost:3000`, you can connect to it by doing the
...
+     -H "Content-Type: application/edn" \
+     $CRUX_URL/_crux/query?q={%3Afind+[e]+%3Awhere+[[e+%3Acrux.db/id+_]]}
 ----
 
-[#entity-tx]
-=== GET `/entity-tx`
+==== Query Parameters
+
+.*Required Parameters*
+* `q` (URL Encoded datalog query)
```
jarohen: lowercase e

jarohen (changes requested):
Lots of comments below, but fundamentally this is a great PR, cheers Dan - much cleaner and more consistent now, and much less ad-hoc parameter encoding/decoding code.

re `/_crux/entity`'s `eid` param - I find it weird that this is EDN-encoded, it seems like it should be subject to a `Content-Type` header or something. Particularly, for future JSON users, we'd be saying 'everything on the API can send/receive JSON, except this query param, which has to be EDN'. 

Not one for this PR, but requires some thought :thinking:

docs/reference/modules/ROOT/pages/http.adoc
danmason: It is indeed - will sort!

docs/reference/modules/ROOT/pages/http.adoc
danmason: Perhaps - though it does wind up making the handler pretty messy and you wind up repeating a lot of things - perhaps if there were a cleaner way of doing it?

danmason: As discussed - not on this PR

jarohen:
This is at a good point to merge, IMO. Still got a few more thoughts to consider, which (if possible) I'd like to include in this release, as they'd also be breaking changes to the HTTP API:

* QPs in EDN format - I'm not 100% comfortable with these. It seems strange that QPs have an implicit content-type - as and when we add full JSON support to the API, we don't then want to say 'well, the exception to the rule is `?eid`, which has to be in EDN'. We could call this `?eid-edn`, and extend it with `?eid-json` later, but not sure whether this fits in with the spirit of URLs - specifically, what's the canonical URL for an entity at that point? edit: covered in #284 
* I'd like to keep the door open to a submit-then-query API, to go a decent way to solving #527. Will comment there with more details on that idea, but that option would involve us changing the `submit-tx` endpoint to take `{:tx-ops [tx-ops...]}` rather than juxt `[tx-ops...]`, so that we can extend it with a query. Maybe that's not a bad idea anyway.
* (non-breaking) I'm also considering newline-delimited EDN as a potential response format for query and entity-history (see newline-delimited JSON for more details). Neither clojure.edn nor Transit deal particularly well with converting a `(...)` response to a _lazy_ structure on the client side - they both want to read whole s-exps.

danmason:
All seem like fair points, wanted to address the following:

> QPs in EDN format - I'm not 100% comfortable with these. It seems strange that QPs have an implicit content-type - as and when we add full JSON support to the API, we don't then want to say 'well, the exception to the rule is ?eid, which has to be in EDN'. We could call this ?eid-edn, and extend it with ?eid-json later, but not sure whether this fits in with the spirit of URLs - specifically, what's the canonical URL for an entity at that point?

We could likely extend the `st/spec` for `query`/`eid` to attempt to decode to other types, if we wanted to keep a single query-parameters for them.

PR #1084, opened 2020-09-07
Tuple return vars

hraberg:
Allows to bind the elements of each returned tuple from a predicate, like this:

`[(identity #{[1 2] [3 4]}) [x y]]` 

In Crux, query returns are always assumed to be sets.
In normal EDN Datalog the above requires another level of nesting:

`[(identity #{[1 2] [3 4]}) [[x y]]]`

In Crux (currently), a set or vector always gets un-nested to one level, so `[(identity [1 2]) x]` binds x twice, and not to the vector itself. One reason for this behaviour is to be symmetric to how vector and set attributes are indexed. This PR doesn't change that behaviour, and only applies to the binding of each value.

### DIFF
--- a/crux-core/src/crux/query.clj
+++ b/crux-core/src/crux/query.clj
@@ -99,6 +99,13 @@
                                                   :body ::or-body)))
 (defmulti pred-args-spec first)
 
+(defmethod pred-args-spec 'q [_]
+  (s/cat :pred-fn #{'q} :args (s/spec (s/cat :query ::query
+                                             :args (s/* (s/cat :key (s/or :quoted-symbol (s/cat :quote #{'quote} :sym symbol?)
+                                                                          :keyword keyword?)
+                                                               :val some?))))
+         :return (s/? ::args-list)))
+
 (defmethod pred-args-spec 'get-attr [_]
   (s/cat :pred-fn  #{'get-attr} :args (s/spec (s/cat :e-var logic-var? :attr literal? :not-found (s/? literal?))) :return logic-var?))
 
@@ -209,7 +216,7 @@
 (set! *unchecked-math* :warn-on-boxed)
 
 (defmulti pred-constraint
-  (fn [{:keys [pred return] {:keys [pred-fn]} :pred :as clause} encode-value-fn idx-id arg-bindings return-vars-tuple-idxs-in-join-order]
+  (fn [{:keys [pred return] {:keys [pred-fn]} :pred :as clause} encode-value-fn idx-id arg-bindings return-vars-tuple-idxs-in-join-order rule-name->rules]
     pred-fn))
 
 (defn- blank-var? [v]
@@ -680,7 +687,7 @@
             (str "Clause refers to unknown variable: "
                  var " " (cio/pr-edn-str clause))))))
 
-(defmethod pred-constraint 'get-attr [_ encode-value-fn idx-id arg-bindings _]
+(defmethod pred-constraint 'get-attr [_ encode-value-fn idx-id arg-bindings _ _]
   (let [arg-bindings (rest arg-bindings)
         [e-var attr not-found] arg-bindings
         not-found? (= 3 (count arg-bindings))
@@ -696,6 +703,37 @@
               true)
           false)))))
 
+(defn- bind-pred-result [return encode-value-fn return-vars-tuple-idxs-in-join-order idx pred-result]
+  (if return
+    (if-let [values (not-empty (if (logic-var? return)
+                                 (c/vectorize-value pred-result)
+                                 pred-result))]
+      (do (if (logic-var? return)
+            (idx/update-relation-virtual-index! idx values encode-value-fn true)
+            (->> (for [tuple values
+                       :when (<= (count return-vars-tuple-idxs-in-join-order)
+                                 (count tuple))]
+                   (mapv tuple return-vars-tuple-idxs-in-join-order))
+                 (idx/update-relation-virtual-index! idx)))
+          true)
+      false)
+    pred-result))
+
+(defmethod pred-constraint 'q [{:keys [return] {:keys [pred-fn]} :pred :as clause} encode-value-fn idx-id arg-bindings return-vars-tuple-idxs-in-join-order rule-name->rules]
+  (let [query (normalize-query (second arg-bindings))
+        parent-rules (:rules (meta rule-name->rules))]
+    (fn pred-constraint [index-store db idx-id->idx join-keys]
+      (let [[_ _ & arg-kvs] (for [arg-binding arg-bindings]
+                              (if (instance? VarBinding arg-binding)
+                                (bound-result-for-var index-store arg-binding join-keys)
+                                arg-binding))
+            args [(apply hash-map arg-kvs)]
+            query (cond-> (assoc query :args args)
+                    (seq parent-rules) (update :rules (comp vec concat) parent-rules))]
+        (with-open [pred-result (.openQuery ^ICruxDatasource db query)]
+          (and (.hasNext pred-result)
+               (bind-pred-result return encode-value-fn return-vars-tuple-idxs-in-join-order (get idx-id->idx idx-id) (iterator-seq pred-result))))))))
+
 (defn- built-in-unification-pred [unifier-fn encode-value-fn arg-bindings]
   (let [arg-bindings (vec (for [arg-binding (rest arg-bindings)]
                             (if (instance? VarBinding arg-binding)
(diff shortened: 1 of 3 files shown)


hraberg:
Examples of using the sub-query predicate:
```
    (t/is (= #{[1 2 3]}
             (api/q (api/db *api*) '{:find [x y z]
                                     :where [[(q {:find [x y z]
                                                  :where [[(identity 2) y]
                                                          [(+ x y) z]]}
                                                 :x 1) [x y z]]]})))
```
Sub-queries without return bindings can be used as "exists" predicates:
```
    (t/is (empty? (api/q (api/db *api*) '{:find [x]
                                          :where [[(identity 2) x]
                                                  [(q {:find [y]
                                                       :where [[(odd? y)]]}
                                                      :y x)]]})))
```

hraberg:
Note that the sub query form can optionally be quoted, but doesn't need to be (built-in predicates are like macros). The arguments can also be given using symbols, like in the args map, but they need to be quoted: `'x 2` has the same effect as `:x 2`.

hraberg:
For brevity, sub queries inherit the rules from the parent query, but can also add their own rules which are merged in with the parent's.

crux-test/test/crux/query_test.clj
```diff
@@ -1065,6 +1065,139 @@
...
+                                   :where [[(q {:find [y]
+                                                :where [[(identity 2) x]
+                                                        [(+ x 2) y]]}) [x]]]})))
+
+  (t/is (empty?
+         (api/q (api/db *api*) '{:find [x]
+                                 :where [[(q {:find [y]
+                                              :where [[(identity 2) x]
+                                                      [(+ x 2) y]
+                                                      [(odd? y)]]}) [x]]]})))
```
jarohen: I find this hard to read when you have a floating parameter on a lower line - I have to visually match brackets to figure out what expr contains the `[x]`, rather than relying on indentation.

```clojure
;; good
(t/is (empty?
       (api/q (api/db *api*)
              '{:find [x]
                :where [[(q {:find [y]
                             :where [[(identity 2) x]
                                     [(+ x 2) y]
                                     [(odd? y)]]})
                         [x]]]})))

;; ok, if it's on the same line and there's not too much going on
(t/is (empty?
       (api/q (api/db *api*)
              '{:find [x]
                :where [[(q {:find [?u] :where [[?u :user/id]]}) [x]]]})))

;; bad
(t/is (empty?
       (api/q (api/db *api*)
              '{:find [x]
                :where [[(q {:find [y]
                             :where [[(identity 2) x]
                                     [(+ x 2) y]
                                     [(odd? y)]]}) [x]]]})))
```

crux-core/src/crux/query.clj
```diff
@@ -671,6 +706,37 @@
...
+            (->> (for [tuple values
+                       :when (<= (count return-vars-tuple-idxs-in-join-order)
+                                 (count tuple))]
+                   (mapv tuple return-vars-tuple-idxs-in-join-order))
+                 (idx/update-relation-virtual-index! idx)))
+          true)
+      false)
+    pred-result))
+
+(defmethod pred-constraint 'q [{:keys [return] {:keys [pred-fn]} :pred :as clause} encode-value-fn idx-id arg-bindings return-vars-tuple-idxs-in-join-order rule-name->rules]
```
jarohen: nit: `{:keys [pred-fn]} :pred` unused in the destructuring?

jarohen (commented):
Thanks @hraberg, a few points to discuss:

* Do we need to allow the user to submit a quoted query? The whole query is quoted regardless - I wouldn't expect other Clojure to work if I'd double quoted a form.
* We don't use kv-args anywhere else in Crux - could we have the args to the nested query in their own map - `(q <query> {:x x})`? maybe even `(q <query> {:args {:x x}})`?
* Am I right in thinking that if the sub-query returns zero results, the top-level query also doesn't return any results? Even if so, not sure if this is actually a problem or not :thinking: 
* Rules - I see why we'd want to specify rules only once within the query, but it feels a weird scoping rule. Also, if the user re-binds the rule inside the sub-query, does it overwrite the rule or add further clauses?

jonpither (approved):

hraberg:
The reason args are bound using key-value var-args is due to how Crux currently resolves the vars in function invocations, it simply assumes anything that's a logic var on top level, the logic vars in `(+ 1 x)` would be `x`, while it doesn't look into nested structures, which is why the query doesn't need to be quoted again.

So adding support for quoted structures is actually related - if we wanted to say nest and look for vars into a structure we need to either:
1. respect quoting.
2. have bespoke logic to deal with built-in predicates, like sub-queries for example.

To be honest, I don't really know what the best solution here is, and I think we can iterate on it. This PR currently implements an approach that strikes a balance of being simple to implement (in the current engine) and easy to use - but totally agree that it does ergonomically go against the Crux API.

I want to point out that we already have a way to do sub-queries, `or-join`, especially after the other PR gets merged, as that will clearly define in and out vars, `(or-join [[in-var] out-var] ...)`. Main reason to add full sub-queries is that there are certain things you cannot do like that, for example: `:limit` :order-by` and perform aggregates.

Regarding results, sub-queries returning 0 results will (and should) result in no results in the parent query.

Rules being inherited is again kind of a balance of what's simple to implement, leads to less verbose sub-queries, and matches  the `or-join` case you automatically get access to the parent rules, as it's essentially the same user-level query. So we could pass around a rule var (like `%`), but that's not how Crux currently works, and will be less ergonomic for the basic use-case. One could see parent rules like CTE (from the WITH-clause) being visible to all sub-queries in SQL. Like I'm not against adding some way of passing rules around, but it's a deeper change, which would require some thought.

So all the above issues kind of lead to one could do sub-queries more like an `or-join` construct, but with a better name (say `find`), but you would need some form of separating the find from the bound names, as we don't have an `:as` construct per find var/pull/aggregate. Again, performing the sub-query via a built-in function solves both these issues without adding new language constructs. To be further discussed.

hraberg:
Might be worth to add that another obvious way to do sub-queries is using rules, which is essentially an `or-join` template which can also recursively refer to itself.

Rules also have in and out vars clearly separated: `[(my-rule [in-var] out-var) ...]` - and they -also- have the ability to rename the vars, which means that in theory a construct like `[(my-rule [in-var] (count out-var)) ...]` would be possible, as the aggregate would be bound (and unified) with an other var in the main query. They still lack `:limit` and `:order-by`, and are also not anonymous (that would be the `or-join` variant) and inline, so you would need to separate them out from the main query and name them.

hraberg:
Note that aggregates in the presence of recursion is another discussion and one reason to not support the above example rule construct.

hraberg:
This PR now uses the nested destructure form, like normal EDN Datalog for function returns:

`a` - scalar
`[a b]` - tuple
`[a ...]` - collection
`[[a b]]` - relation

This is a (minor) breaking change, as previously the scalar form worked like the collection form if the returned value was a vector or a set. To preserve the behaviour of such queries one has to replace `a` with `[a ...]`.

hraberg:
Have added support for single argument tuples to be lifted up and bound (simply by converting them to the varargs version):
```
    '{:find [x y z]
      :where [[(identity 1) x]
              [(q {:find [z]
                   :args [{:x x}]
                   :where [[(+ x 2) z]]}) [[y]]]
              [(+ x y) z]]}
```
Now the sub-query just takes the query itself as an argument, and if there's a single args tuple, lifts it out, allow the var `x` inthis case in the outside query to be bound (as it happens, also to `x` in the sub query).

jarohen (approved):

crux-core/src/crux/query.clj
```diff
@@ -95,8 +101,16 @@
                                                   :body ::or-body)))
 (defmulti pred-args-spec first)
 
+(defmethod pred-args-spec 'q [_]
+  (s/cat :pred-fn #{'q} :args (s/spec (s/cat :query (s/or :quoted-query (s/cat :quote #{'quote} :query ::query)
+                                                          :query ::query)
+                                             :args (s/* (s/cat :key (s/or :quoted-symbol (s/cat :quote #{'quote} :sym symbol?)
+                                                                          :keyword keyword?)
+                                                               :val some?))))
```
jarohen: `some?` -> `any?`

crux-core/src/crux/query.clj
```diff
@@ -49,6 +50,8 @@
                                       :a (s/and c/valid-id? some?)
                                       :v (s/? (some-fn logic-var? literal?)))))
 
+(s/def ::args-list (s/coll-of logic-var? :kind vector? :min-count 1))
```
jarohen, line 53: rename?

crux-core/src/crux/query.clj
```diff
@@ -95,8 +101,16 @@
                                                   :body ::or-body)))
 (defmulti pred-args-spec first)
 
+(defmethod pred-args-spec 'q [_]
+  (s/cat :pred-fn #{'q} :args (s/spec (s/cat :query (s/or :quoted-query (s/cat :quote #{'quote} :query ::query)
```
jarohen: newline before `:args`

crux-core/src/crux/query.clj
```diff
@@ -669,7 +724,45 @@
...
+
+    :collection
+    (do (idx/update-relation-virtual-index! idx (mapv vector pred-result))
+        (not-empty pred-result))
+
+    (:tuple :relation)
+    (do (->> (for [tuple (if (= :relation return-type)
+                           pred-result
+                           [pred-result])
+                   :when (<= (count return-vars-tuple-idxs-in-join-order)
```
jarohen, line 741: removing `:when`, binding to `nil`

PR #1086, opened 2020-09-08
Aggregates

### LINKED ISSUE #809: Aggregations in the query engine

hraberg:


jonpither:
Addresses https://github.com/juxt/crux/issues/809

hraberg:
Support most expected aggregates. New aggregates can be added via a multi-method. This is a possible, but low-level extension point. There's also a corresponding spec for the multi-method. Aggregates can take literal arguments and a single, last, logic var.

Aggregates are computed lazily by consuming the stream and grouping tuple by tuple. That is, the entire result set isn't realised, but some aggregates will produce large intermediate results due to the way they work, like `(distinct x)` for example.

Grouping is done implicitly on all non-aggregate vars. As Crux doesn't produce the set until the end, and the aggregate is calculated as it goes, any valid path in the join tree will be visited, so duplicates are reported correctly (no need for the `:with` construct).

The aggregate multi-method itself is a factory that returns the aggregate function with 3 arities:
0. produce the accumulator.
1. finalise the aggregation.
2. reducing step taking the accumulator and one element.

The multi-method takes all the literal arguments, and can use them in the returned function, like in `(max 2 n)` for example.

Aggregates are computed using boxed maths, so they're not performance tuned in that sense. They are computed before pull does its projections, there's no way to aggregate a pull expression (which could in theory be possible). There's an assumption that a var binding and its pull expression will be a one to one mapping, so it's safe to do grouping first.

Order-by has been changed so one can refer to any find expression in it, this is just a syntactical reference to an index in the find, and does no computation itself. 

The aggregates do degrade to set-semantics over multiple valued attributes, which are always treated as sets in the index. My suggestion is that we don't try to solve this in this initial PR, as it complicates things for a corner use case, and instead clearly document this issue.

The real solution is to figure out the e-var for each aggregated v-var, and replace it with the corresponding pull expression and do the pull first, and allow aggregates over pulls respecting vector attributes. There are nuances to this, as this rewrite only works if the v-var itself is a leaf var. Another approach is to store count (or index position) in the indexes, which we looked at in ECAV a while back. But again, this complicates Crux and grows the index in general for a limited use case.

iku000888:
Thanks for this work! It def. has been an itch scratched that it is now available out of the box!

PR #1093, opened 2020-09-10
Making get-attr produce a collection, respecting return bindings.

hraberg:
Incremental tweak to get-attr so it respects bindings properly. When get-attr is produced by rewriting leaf vars, it expands to a collection binding, `[x ...]`. Binding to `x` will bind the entire collection. Normal usage would hence to be use the `[x ...]` binding.

Also changes to remove support for sets as the optional argument being treated as multiple values. That is, the optional argument will always just be a single value.


jarohen (approved):
:ok_hand:

jonpither (approved):

PR #1094, opened 2020-09-10
Lucene Module #977

### LINKED ISSUE #977: Users can find documents using full-text search
Plenty to consider here:
* node-local Lucene indices? (PRIORITY)
* opt-in on certain attributes?
* temporal/atemporal Lucene?
* integration with ElasticSearch for independent clustering?

## Elasticsearch

Possible modes for integration:

1. ES as remote "golden" doc-store (possible but not generally desirable)
2. ES embedded, like Lucene, but used only to augment indexes (per-node scaling would be 1:1 which has limited benefit)
3. ES as remote cluster, used only to augment indexes

Index-store augmentation should support regular bitemporal filtering.

Functionality could be exposed via `(crux.api/es-search db ...)` or within queries using a predicate like `[(crux.es/search ...) ?search-result]` (preferred)

Option 3:
image

It is unlikely we would decide to use a Clojure client library like https://github.com/clojurewerkz/elastisch or https://github.com/mpenet/spandex

jonpither:
Lucene module to provide full-text search #977 

## Todos:

 - [x] Consider cardinality (make test)
 - [x] Lucene Transactions
 - [x] Decorate indexer
 - [x] Pred fn Spec
 - [x] Test fuzzy search
 - [x] Eviction
 - [x] Docs
 - [x] Bind and order on Score
 - [x] Handling index bump (can defer?)
 - [x] Consider schema for docs indexed? Rather than just top level string fields
 - [x] Does score fit in the content of bitemp engine?
 - [x] Sanity check prior art contracts
 - [x] More sane eid string washing (hashing)
 - [x] Expose / use Lucene metrics?
 - [x] Microbench (volume + no of searches), compare indexing options
 - [x] Validate the module config (@jarohen, see lucene-test)
 - [x] Number of results? Sorting (by default I think top ten)
 - [x] Go through Lucene single field DSL in more detail (does it cater for range scans?)
 - [x] Test (using explain) (and score before/after) for document de-duplication / structural sharing
 - [x] Exception handling of bus fns
 - [x] change fn arity back to `A V`

## Potential improvements 

Options on indexing/searching Lucene docs. Currently we are on Option 3, with a backout plan back to Option 1.

See https://cwiki.apache.org/confluence/display/lucene/ImproveIndexingSpeed for tips on tuning Lucene.

### Option 1

* Documents contain A/V/E. 
* Having E on the document is useful for eviction purposes.

### Option 2

* As 1, and put the C into the doc to make temporal resolution / querying faster
* Docs don't benefit from structural sharing across Cs

### Option 3

* Docs are A/V only.
* Temporal resolution using AVE
* Has more structural sharing, and less susceptible to past-data affecting scores
* Fewer fields, speeding up ingest
* Requires sophisticated eviction, to use unindex-eids equivalent

### Option 4

* Match documents stored in Lucene to docs stored within Crux - no shredding
* Pros/Cons to be analysed - it *may* be easier for Lucene to have larger and fewer docs - benchmarking needed.

## References:

* https://github.com/juxt/crux-rnd/blob/master/src/crux/datalog/lucene.clj
* https://www.toptal.com/database/full-text-search-of-dialogues-with-apache-lucene
* https://lucene.apache.org/core/8_6_0/core/index.html

## Index Bumps

If we change the way we index lucene docs - which is forseeable as we iterate on performance etc - then we should bump the main index version. 

The other option is to maintain a separate lucene index version. For now I'm discounting this, as we would need a way to re-index all documents, outside the standard path of ingesting all docs when a blank node fires up. 

## Schema

We could have a schema document to do something more intelligent other than indexing all top-level string fields (and top-level collections containing strings for cardinality many) - but we will postpone this feature for the first iteration (considering also, `crux-schema` is still embryonic).

## Subsequent TODOs

- Expose Lucene metrics (i.e. docCount)
- Expose more Lucene config?
- Tweak scoring (potentially drop `idf`)

## Questions

 - Is there any opportunity to blend in Lucene's natural laziness?
 - Can the `lucene-store` participate in index-store snapshots (as we need similar life-cycle management of underlying components during query). 
 - Related to above, should the `lucene-store` be an `index-store`?


jonpither:
Scoring is currently an issue with this PR. Scores are computed by lucene, by factoring in the frequency/uniqueness of A/V matches across the whole index, which would take into account data-sets that would normally excluded by our bitemporal engine (i.e. past data).

The two broad options are: 1) getting Lucene to filter out past data as part of the query, or 2) adjusting our scoring algorithm.

## Filtering out the data via query.

One way is to add tt/vt into docs and use the Lucene query engine to filter out non relevant results. This is having Lucene do the work our query engine usually does.

The other option is explore any hooks we can use, to filter our candidate results pre-scoring. I haven't seen a hook yet, where you can filter out hits pre-scoring, and have this affect the score.

## Adapt the scoring algorithm

### Drop `idf`

We could for experiment with dropping `idf` (Inverse Document Frequency) from being part of the score, such that term matches would just be on the content of the A/V themselves, not taking into account A/V frequencies as part of the rest of the index.

### Supply `idf`

We could provide ourselves the values of `docFreq` and `docCount` used in the `idf` calculation.

`docFreq`, the number of documents in which the term t appears (matches)
`docCount`, num of docs with the field

This would likely a subquery for `docCount`, and setting `docFreq` to be the number of results returned by the query, after bitemporal filtering.

### Boost

We boost the scores back up again:

```
boost *=  (docFreq / number-of-bitemp-filtered results)
```

## References: 

https://opensourceconnections.com/blog/2015/10/16/bm25-the-next-generation-of-lucene-relevation/

jonpither:
I've added a disclaimer to the README for now. Other EDN Datalog DBs have a similar problem, with past data infringing on scoring, we can revisit this as needed, picking up up some of what I wrote in an earlier comment.

jonpither:
Would benefit from #1152

project.clj
```diff
@@ -103,6 +105,8 @@
                "crux-metrics/test"
                "crux-s3/test"
                "crux-azure-blobs/test"
+               "crux-sql/crux-sql"
```
jarohen: suspect these want to be `crux-sql/test`?

docs/reference/modules/ROOT/pages/lucene.adoc
```diff
@@ -0,0 +1,78 @@
+= Lucene
```
jarohen, line 1: The ADoc template's changed a little while this branch has been in-flight, comments below

jarohen: :toc: stuff at the top no longer required

docs/reference/modules/ROOT/pages/lucene.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+Lucene module for Crux making use of https://lucene.apache.org/[Apache Lucene].
+
+== Setup
+
+`crux-lucene` runs in-process as part of the Crux node as a Crux module.
+
+First, add the `crux-lucene` dependency to your project:
+
+.pom.xml
```
jarohen: see `[tabs]`

docs/reference/modules/ROOT/pages/lucene.adoc
```diff
@@ -0,0 +1,78 @@
...
+[source,xml]
+----
+<dependency>
+    <groupId>juxt</groupId>
+    <artifactId>crux-lucene</artifactId>
+    <version>20.09-1.11.0-alpha</version>
```
jarohen: there's an attribute for this now, `{crux_version}` - see another doc for setup

crux-lucene/src/crux/lucene.clj
```diff
@@ -0,0 +1,187 @@
...
+                        :required? true
+                        :spec ::sys/path}}}
+  [{:keys [^Path db-dir]}]
+  (let [directory (FSDirectory/open db-dir)
+        analyzer (StandardAnalyzer.)
+        node (LuceneNode. directory analyzer)]
+    (alter-var-root #'*node* (constantly node))))
+
+(defn ->index-store
+  {::sys/deps {:index-store :crux/index-store
```
jarohen: for the `:index-store` default, I don't think you want to refer to the `:crux/index-store` ref - this'll be circular if the user forgets to override it. At first glance, I'd copy the default out of the `:crux/node` component, but may need some further thought - a bigger but cleaner solution may be to explicitly add the concept of 'wrappers' to `crux.system`.

crux-lucene/src/crux/lucene.clj
```diff
@@ -0,0 +1,187 @@
...
+
+(defn ->node
+  {::sys/args {:db-dir {:doc "Lucene DB Dir"
+                        :required? true
+                        :spec ::sys/path}}}
+  [{:keys [^Path db-dir]}]
+  (let [directory (FSDirectory/open db-dir)
+        analyzer (StandardAnalyzer.)
+        node (LuceneNode. directory analyzer)]
+    (alter-var-root #'*node* (constantly node))))
```
jarohen: :/ this implies a miss in what the module system can do...

crux-lucene/src/crux/lucene.clj
```diff
@@ -0,0 +1,187 @@
...
+  (index-docs [this docs]
+    (db/index-docs index-store docs))
+  (unindex-eids [this eids]
+    (try
+      (evict! index-store lucene-node eids)
+      (catch Throwable t
+        (clojure.tools.logging/error t)
+        (throw t)))
+    (db/unindex-eids index-store eids))
+  (index-entity-txs [this tx entity-txs]
```
jarohen: why's this in `index-entity-txs`, OOI?

crux-lucene/src/crux/lucene.clj
```diff
@@ -0,0 +1,187 @@
...
+    (let [{:keys [^Directory directory ^Analyzer analyzer]} lucene-node
+          index-writer (IndexWriter. directory, (IndexWriterConfig. analyzer))]
+      (try
+        (let [content-hashes (entity-txes->content-hashes entity-txs)
+              docs (vals (db/fetch-docs document-store content-hashes))]
+          (write-docs! index-writer docs)
+          (db/index-entity-txs index-store tx entity-txs)
+          (.close index-writer))
+        (catch Throwable t
+          (clojure.tools.logging/error t)
```
jarohen: nit: require `:as log`

crux-lucene/resources/logback-test.xml
```diff
@@ -0,0 +1,11 @@
+<configuration>
```
jarohen: don't think we need another one of these files - we can refer to one of the common ones (see `crux-s3`'s `project.clj`)

crux-lucene/checkouts/crux-core
```diff
@@ -0,0 +1 @@
+../../crux-core
```
jarohen, line 1: do we need these checkouts? implies you're running your REPL from `crux-lucene` rather than the project root?

crux-lucene/README.adoc
```diff
@@ -0,0 +1,32 @@
...
+    <artifactId>crux-lucene</artifactId>
+    <version>20.09-1.11.0-alpha</version>
+</dependency>
+----
+
+For more details, see the https://opencrux.com/reference/lucene.html[Crux Lucene documentation]
+
+== Scoring
+
+This module uses the default BM25Similarity similarity to calculate
```
jarohen, line 27: nit: I'm reliably informed the idiom for adoc is one sentence per line rather than wrapping at 80 :)

crux-core/src/crux/kv/index_store.clj
```diff
@@ -698,6 +698,24 @@
       {:bytes-indexed (->> content-idx-kvs (transduce (comp (mapcat seq) (map mem/capacity)) +))
        :indexed-docs docs}))
 
+  (exclusive-avs [this eids]
+    (with-open [snapshot (kv/new-snapshot kv-store)
```
jarohen, line 772: can `unindex-eids` use this too?

crux-lucene/src/crux/lucene.clj
```diff
@@ -0,0 +1,187 @@
...
+           org.apache.lucene.queryparser.classic.QueryParser
+           [org.apache.lucene.search BooleanClause$Occur BooleanQuery$Builder DoubleValuesSource IndexSearcher Query ScoreDoc TermQuery]
+           [org.apache.lucene.store Directory FSDirectory]))
+
+(def ^:dynamic *node*)
+
+(defrecord LuceneNode [directory analyzer]
+  java.io.Closeable
+  (close [this]
+    (doseq [^Closeable c [directory]]
```
jarohen: `cio/try-close`

crux-lucene/src/crux/lucene.clj
```diff
@@ -0,0 +1,187 @@
...
+(defn- id->stored-bytes [eid]
+  (mem/->on-heap (cc/->value-buffer eid)))
+
+(defn- ^String eid->str [eid]
+  (mem/buffer->hex (cc/->id-buffer eid)))
+
+(defn- crux-doc->triples [crux-doc]
+  (->> (dissoc crux-doc :crux.db/id)
+       (mapcat (fn [[k v]]
+                 (for [v (if (coll? v) v [v])
```
jarohen: `coll?` returns true for maps as well - maybe `crux.codec/vectorize-value`?

crux-lucene/src/crux/lucene.clj
```diff
@@ -0,0 +1,187 @@
...
+  java.io.Closeable
+  (close [this]
+    (doseq [^Closeable c [directory]]
+      (.close c))))
+
+(defn- id->stored-bytes [eid]
+  (mem/->on-heap (cc/->value-buffer eid)))
+
+(defn- ^String eid->str [eid]
+  (mem/buffer->hex (cc/->id-buffer eid)))
```
jarohen: think `(str (cc/new-id eid))` does the same thing here without low-level buffer stuff

crux-lucene/src/crux/lucene.clj
```diff
@@ -0,0 +1,187 @@
...
+
+(defn write-docs! [^IndexWriter index-writer docs]
+  (doseq [d docs t (crux-doc->triples d)]
+    (.updateDocument index-writer (triple->term t) (triple->doc t))))
+
+(defn evict! [indexer, node, eids]
+  (let [{:keys [^Directory directory ^Analyzer analyzer]} node
+        attrs-id->attr (->> (db/read-index-meta indexer :crux/attribute-stats)
+                            keys
+                            (map #(vector (mem/buffer->hex (cc/->id-buffer %)) %))
```
jarohen: `(map (juxt eid->str identity))` ?
maybe `eid->str` should then be renamed to `->hex` or similar

crux-lucene/src/crux/lucene.clj
```diff
@@ -0,0 +1,187 @@
...
+        q (.build b)
+        q (FunctionScoreQuery. q (DoubleValuesSource/fromQuery q))
+        score-docs (.-scoreDocs (.search index-searcher q 1000))]
+
+    (when (seq score-docs)
+      (log/debug (.explain index-searcher q (.-doc ^ScoreDoc (first score-docs)))))
+
+    (cio/->cursor (fn []
+                    (.close directory-reader))
+                  (map (fn [^ScoreDoc d] (vector (.doc index-searcher (.-doc d))
```
jarohen: indentation here
```clojure
(map (fn [...]
       ...)
     score-docs)
```

otherwise it puts the onus on the reader to determine whether `score-docs` is part of the `fn` or not, by counting close parens

https://juxt.slack.com/archives/G018EA25K71/p1600275786012100

crux-lucene/src/crux/lucene.clj
```diff
@@ -0,0 +1,187 @@
...
+         (.content-hash entity-tx))))
+
+(defrecord LuceneDecoratingIndexStore [index-store document-store lucene-node]
+  db/IndexStore
+  (index-docs [this docs]
+    (db/index-docs index-store docs))
+  (unindex-eids [this eids]
+    (try
+      (evict! index-store lucene-node eids)
+      (catch Throwable t
```
jarohen: this should already be logged upstream?

crux-lucene/test/crux/lucene_test.clj
```diff
@@ -0,0 +1,200 @@
...
+                                   [?e :crux.db/id]]}))))
+
+      (with-open [db (c/open-db *api*)]
+        (t/is (= #{[:ivan "abar"]
+                   [:ivan "atar"]}
+                 (c/q db {:find '[?e ?v]
+                          :where '[[(text-search "a?ar" :foo) [[?e ?v]]]
+                                   [?e :crux.db/id]]})))))))
+
+(t/deftest test-can-search-string-across-attributes
```
jarohen, line 97: good that we _could_ do this, but worth mandating the attribute to keep it consistent with the rest of the query engine?
maybe allowing multiple search attributes if that's what this is required for?

crux-lucene/test/crux/lucene_test.clj
```diff
@@ -0,0 +1,200 @@
...
+               (c/q db {:find '[?e ?v ?a]
+                        :where '[[(text-search "Ivan") [[?e ?v ?a]]]
+                                 [?e :crux.db/id]]}))))
+
+    (t/testing "no match against a non-existant field"
+      (t/is (= #{}
+               (c/q db {:find '[?e ?v]
+                        :where '[[(text-search "Ivan" :non-field) [[?e ?v]]]
+                                 [?e :crux.db/id]]})))))
+
```
jarohen, line 112: double newline :wink: :joy:

docs/reference/modules/ROOT/pages/lucene.adoc
```diff
@@ -0,0 +1,78 @@
...
+[source,clojure]
+----
+{...
+ :crux.lucene/node {:db-dir "lucene-dir"
+                    :crux/indexer {:crux/module 'crux.lucene/->indexer
+                                   :indexer 'crux.kv.indexer/->kv-indexer}}
+----
+
+
+== Querying
```
jarohen, line 59: should we mention the four returned attributes in the tuple?

jarohen (changes requested):
hey @jonpither - comments below, hopefully the number doesn't detract from the overall impression that this'll be a great feature - nice one!

crux-lucene/checkouts/crux-core
jonpither: I do yeah :-) is that not OK? Repl runs faster, dog-fooding in that module.

crux-core/src/crux/kv/index_store.clj
jonpither: potentially - if we commit to this approach of eviction. Could also be done in a separate PR even then.

crux-lucene/README.adoc
jonpither: hmm - I'll park that one for a wider adoc discussion! cc @malcolmsparks

crux-lucene/src/crux/lucene.clj
jonpither: As oppose to index-docs? Will try there

crux-lucene/src/crux/lucene.clj
jonpither: Applied.

crux-lucene/src/crux/lucene.clj
jonpither: left as is, because it's not obvious hex is what's happening now

crux-lucene/test/crux/lucene_test.clj
jonpither: Fixed.

docs/reference/modules/ROOT/pages/lucene.adoc
jonpither: Cool. Fixed.

docs/reference/modules/ROOT/pages/lucene.adoc
jonpither: Done.

project.clj
jonpither: Fixed.

jonpither (commented):
Thanks for the review @jarohen. I've cleared out the low hanging fruit - some items will need a discusion.

hraberg (approved):
Not giving detailed notes, approved in principle (after fixing the most basic stuff, like lazy seqs escaping with-open), but comments were made and discussed on the call and captured there.

crux-lucene/test/crux/lucene_test.clj
jonpither: Moved out into a different pred-fn

crux-core/src/crux/kv/index_store.clj
jonpither: yes raise issue to capture unindex-eids using

PR #1098, opened 2020-09-14
Tpch datalog

hraberg:
Adds a bench mark running all 22 TPC-H queries as Datalog.
Uses scale factor 0.01 by default and compares the expected results from the `io.airlift.tpch` generator.

Generates the documents a bit differently than the tpch-stress generator:
1. turns date columns into actual `java,util.Date`, the generator returns numbers for the dates.
2. use single field primary key as the crux.db/id, prefixed with the table name, on the pattern of `customer_123`. Does the same for foreign keys.
3. dual field primary keys are still UUIDs, these are used form many-to-many tables and are never accessed directly themselves. Not the proper way for temporal, but this test doesn't (yet) deal with that.


jarohen:
notes from standup:

* @danmason - could you check this runs through within the bench harness?
* we're doubling up on TPC-H generation - might be worth seeing if we can coalesce with Jon's SQL work

jarohen (approved):
One for the future - we might want to consider splitting the bench tests up so that we get timings for each query, like we do for WatDiv, so that we can see if specific queries are slower than for our comparison databases.

PR #1099, opened 2020-09-14
In bindings

hraberg:
This is a tentative PR to support `:in` bindings in Crux:
```
(api/q (api/db *api*) '{:find [e]
                        :in [$ name]
                        :where [[e :name name]]} "Ivan")
```

The bindings build upon the recently added return bindings, so the expected scalar, collection, tuple and relation bindings are supported.

Changes sub-queries to only use this form when binding logic vars to sub-queries instead of rewriting the args-map.
Supports an (currently) ignored optional source var `$` at the front of the bindings.
Does not remove support for our current `:args` map.

Adds wrappers to the PCruxDatasource protocol to support varargs to provide the actual bindings for `q` and `open-q`.
The Java API uses varargs here. The HTTP API provides the varargs in the same map as the query, valid and tx-time.

This PR is a discussion point, I'm not sure we should add this or not?

### DIFF
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -309,10 +309,29 @@
   The tx-ops will only be visible in the value returned from this function - they're not submitted to the cluster, nor are they visible to any other database value in your application.
   If the transaction doesn't commit (eg because of a failed 'match'), this function returns nil."))
 
-(defn q [db q & args]
+(defn q
+  "q[uery] a Crux db.
+  query param is a datalog query in map, vector or string form.
+  This function will return a set of result tuples if you do not specify `:order-by`, `:limit` or `:offset`;
+  otherwise, it will return a vector of result tuples."
+  [db q & args]
   (q* db q args))
 
-(defn open-q ^crux.api.ICursor [db q & args]
+(defn open-q
+  "lazily q[uery] a Crux db.
+  query param is a datalog query in map, vector or string form.
+
+  This function returns a Cursor of result tuples - once you've consumed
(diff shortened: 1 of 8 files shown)


jarohen (commented):
As discussed - I'm concerned that given `:in` and `$` are present in other EDN Datalog databases it may be confusing to users if they don't behave in the same way - more so than if we used different names. If we differed sufficiently that users had to check the docs ('Crux doesn't have `:in`, oh, what _does_ Crux have?') this would be less of an issue than being _slightly_ different ('ah, Crux has `:in` - right, I know how `:in` works, I won't bother looking that up').

Things that differ:
* We have one data-source that's passed _before_ the query, with the args passed after - with an `:in` binding of `[$ foo bar]` these correspond to one parameter passed before and two after the query, rather than the parameters passed after the query.
* Lack of multiple data-sources
* Related - `:where` EAV triples don't/can't take `$` as an optional first param.
* `$` is hard-coded - inability to rename the variable.

I agree that we don't want to hold this functionality until(/if) we do opt for multiple data-sources, and that it's good to fix the cross-product issue with `:args`, and even that `:in` is a good pattern to base ourselves on, given familiarity elsewhere in the Clojure industry - but maybe we could consider an option whereby we explicitly state that we don't support `$` in `:in`, instead opting to have the `:in` bindings actually match up to what's passed after the query. If we considered it useful to be able to pass the database value to a predicate we could also, temporarily, specify that `$` would be a special var replaced by the database value - this would be forward compatible if we decided to iterate on it later.

PR #1104, opened 2020-09-16
Remove JS-less HTML views on query, entity and status

### LINKED ISSUE #1076: Remove JS-less HTML representation from console

danmason:
Based on top of #1082, see https://github.com/danmason/crux/compare/migrate-http-api...remote-js-less-html

Resolves #1076 

Currently, if JS is disabled, returns the `raw-html` template, with a `<noscript>` containing containing a `<pre>` with pretty printed results.


crux-http-server/src/crux/http_server/entity.clj
```diff
@@ -60,129 +50,55 @@
...
+      (let [^String resp (util/raw-html {:title "/_crux/entity"
+                                         :options opts
+                                         :results (cond
+                                                    no-entity? nil
+                                                    not-found? {:entity-results {"error" (str eid " entity not found")}}
+                                                    cause {:entity-results {"error" cause}}
+                                                    entity-history (try
+                                                                     {:entity-results (iterator-seq entity-history)}
+                                                                     (finally
+                                                                       (.close entity-history)))
```
jarohen: you can't close `entity-history` before the call to `raw-html` has returned, because `entity-history` is lazy - you'd be closing it before it's consumed

danmason: Ahhhh - think theres a similar problem within query, there, too! WIll fix :+1:

PR #1114, opened 2020-09-22
add crux.error to throw IAEs with data, s.t. they can be easily passed over the wire

jarohen:
Added a sub-class of Java's IAE so that Java consumers can continue to catch IAE, but also implemented IExceptionInfo so that these errors can be handled with `ex-data` in Clojure, used as an HTTP response nigh-on verbatim, and reconstructed on the client side (for #1082)

TODO:
* [x] I've migrated a few of the exceptions thrown in `crux.tx` but we'll need a wider migration to make this work for HTTP
* [x] Do we throw exceptions that aren't IAE, exceptions that should also be passed over the wire without just throwing our hands up and declaring 500?

### DIFF
--- a/crux-core/src/crux/IllegalArgumentException.java
+++ b/crux-core/src/crux/IllegalArgumentException.java
@@ -1,0 +1,20 @@
+package crux;
+
+import clojure.lang.IExceptionInfo;
+import clojure.lang.IPersistentMap;
+
+public class IllegalArgumentException extends java.lang.IllegalArgumentException implements IExceptionInfo {
+    private static final long serialVersionUID = 8569715935234823692L;
+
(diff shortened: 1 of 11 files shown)


danmason:
^ Gone through and migrated a number of IllegalArgumentExceptions inside of `codec`, `query`, and `conform` - theres a few more throughout the codebase - will continue porting them over (The ones I've currently ported over have made it such that, post changing how they handle errors to use this new format, http-server/remote-api pass api-tests)

Edit: I've migrated across all instances where we were throwing IllegalArgumentException - may want to rename some of associated keys and whatnot, and in a few cases they are thrown and caught quite close to each other.

PR #1122, opened 2020-09-28
Kafka Connect Tests

### LINKED ISSUE #1109: E2E tests for Kafka connector
- As a side piece, need to update the confluent install guide as the version on confluenthub is outdated

danmason:
Based on top of #1119 as the tx-log was causing issues with the connector otherwise

Resolves #1109 and fixes minor issues within the connector.

Some thoughts on the kafka connector:
- Sink only supports `put` and `delete` options (understandably, could perhaps support evict but I'm not sure how an evict would look in such a case) - and doesn't have a mechanism for entering either of these in at a `valid-time` - I wonder if there is something we could add there?

TODO:
* [ ] Notice about version on Confluent Hub being out of date, instruction on how to build and use manually

### DIFF
--- a/crux-test/test/crux/kafka_connect_test.clj
+++ b/crux-test/test/crux/kafka_connect_test.clj
@@ -1,0 +1,161 @@
+(ns crux.kafka-connect-test
+  (:require [crux.api :as api]
+            [crux.kafka :as k]
+            [crux.kafka.connect :as cfc]
+            [crux.codec :as c]
+            [clojure.test :as t]
+            [crux.fixtures.http-server :as fh :refer [*api-url*]]
+            [crux.fixtures :as fix :refer [*api*]]
+            [clojure.edn :as edn])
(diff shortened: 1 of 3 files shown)


crux-test/test/crux/kafka_connect_test.clj
```diff
@@ -0,0 +1,161 @@
...
+           org.apache.kafka.connect.storage.OffsetStorageReader))
+
+(t/use-fixtures :each fh/with-http-server fix/with-node)
+
+(defn new-sink-record [{:keys [topic partition key-schema key value-schema value kafka-offset]
+                        :or {partition 0
+                             kafka-offset 0}}]
+  (SinkRecord. topic partition key-schema key value-schema value kafka-offset))
+
+(defn get-tx-from-source-task [^CruxSourceTask source-task]
```
jarohen, line 22: I'd call this `source-task->tx`

PR #1129, opened 2020-09-29
ECAV Cache

hraberg:
Eagerly builds caches of ECAV, keyed by content hash. This incurs an extra cost at first access to a content hash (less than 10%), but TPC-H queries are about 50% faster once the caches are warm. As the cache is based on content hashes they live on the IndexStore itself and is shared across queries and time. Also adds chunked seqs to the index store.

### DIFF
--- a/crux-core/src/crux/kv/index_store.clj
+++ b/crux-core/src/crux/kv/index_store.clj
@@ -570,8 +570,7 @@
           eid-value-buffer (buffer-or-value-buffer e)
           eid-buffer (value-buffer->id-buffer this eid-value-buffer)]
       (when-let [content-hash-buffer (entity-resolver-fn eid-buffer)]
-        (let [prefix (encode-ecav-key-to nil eid-value-buffer content-hash-buffer attr-buffer)
-              a->vs (ecav-cache-lookup ecav-cache @ecav-iterator-delay eid-value-buffer content-hash-buffer)
+        (let [a->vs (ecav-cache-lookup ecav-cache @ecav-iterator-delay eid-value-buffer content-hash-buffer)
               vs ^NavigableSet (.get a->vs attr-buffer)]
           (seq (.tailSet vs (buffer-or-value-buffer min-v)))))))
 
--- a/crux-core/src/crux/kv/index_store.clj
+++ b/crux-core/src/crux/kv/index_store.clj
@@ -492,7 +492,7 @@
      (when k
        (if-let [k (k-fn k)]
          (cons k (lazy-seq (step (kv/next i))))
-         (lazy-seq (step (kv/next i))))))
+         (recur (kv/next i)))))
    (kv/seek i seek-k)))
 
 (defn- chunk-stepper [i k-fn seek-k]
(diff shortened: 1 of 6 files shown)


hraberg:
I've also spiked (not included here) pulling the cached values from the document store, removing the need for the ECAV index, but that could be a next iteration would we choose to do so. There are a couple issues with this (both solvable):

1. Eviction relies on the ECAV index, this could be solved in worst case by simply scanning AVE, or by using the evicted document (if this is at hand).
2. As this populates the index document by document, it would be slow if the document store caches are cold and the document store is remote without some form of batching. Batching is hard to add though due to the way the indexes are used.

crux-core/src/crux/kv/index_store.clj
```diff
@@ -452,22 +460,77 @@
...
+                         content-hash-buffer
+                         mem/copy-to-unpooled-buffer
+                         (fn [content-hash-buffer]
+                           (let [eid-size (mem/capacity eid-value-buffer)
+                                 a->vs (HashMap.)
+                                 prefix (encode-ecav-key-to (.get seek-buffer-tl)
+                                                              eid-value-buffer
+                                                              content-hash-buffer)
+                                 i (new-prefix-kv-iterator ecav-i prefix)]
+                             (loop [k (kv/seek i prefix)]
```
jarohen, line 474: think this is `all-keys-in-prefix`?

crux-core/src/crux/kv/index_store.clj
```diff
@@ -452,22 +460,77 @@
 (defn- value-buffer->id-buffer [index-snapshot ^DirectBuffer value-buffer]
   (c/->id-buffer (db/decode-value index-snapshot value-buffer)))
 
+(defn- ecav-cache-lookup ^java.util.Map [ecav-cache ecav-i ^DirectBuffer eid-value-buffer ^DirectBuffer content-hash-buffer]
```
jarohen: mind if we call this `cav-cache` given we don't store the `e`?

jarohen:
Is it specifically caching the whole entity/content-hash that makes this fast? Otherwise, we've also got A in hand - if we were to key the cache by CA instead, we'd reduce the risk of needlessly pulling in the whole of a large document when we only needed one or two attributes.

crux-core/src/crux/kv/index_store.clj
hraberg: It is, but this is slightly faster, as there's no immediate sequence.

crux-core/src/crux/kv/index_store.clj
hraberg: Good call, will do.

hraberg:
I did try segmenting the cache by A, at least for TPC-H that turned out to slower, part of the idea is exactly to pre-fetch attributes that might be touched later via next while in the area.

crux-core/src/crux/kv/index_store.clj
jarohen: Ah, right - in that case, can/should we apply these changes back to `all-keys-in-prefix`?

hraberg: `all-keys-in-prefix` is also used by the history API and the unindex stuff. History we likely want to be lazy, and unindex is off the fast path so isn't a big deal. So tempted to keep them separated for now.

PR #1140, opened 2020-09-30
Index caches

hraberg:
This PR goes further than the CAV cache and caches the top-level of the trees, AV and AE in memory for requested attributes.
This is an experimental PR. Both indexes each make TPC-H faster. Together it's about 2-3 times faster for some queries.

These caches are then kept up to date as more data arrives. An entire attribute column might be evicted from the cache, which is quite small (default size is 128) if it goes cold. The entire column as a whole is cached, and no attempt is made to identify hot ranges. This implies that a column can become very large while still needing to fit in memory. This is less of a problem for AE, as the AV space is usually much larger.

As queries run, the AV/AE data will be cross checked against, actual, temporal and snapshot-based data in AVE and AEV like previously so extra values at the top-level shouldn't affect the result.

Eviction might cause inconsistent reads as data might disappear from a cache during a query. One approach is to simply not remove data from these caches on entity eviction and accept that some data might sit in memory. An alternative approach is to have a global lock and postpone the cache eviction until all current queries are finished (or something similar).

The way the caches are kept up to date depends on usage of locks. When the cache is initiated the entire column is scanned and populated at once, which can cause non-trivial lock contention. As columns can get evicted, new columns might get initialised as the system is running. Other better ways to do this are likely possible.

### DIFF
--- a/crux-core/src/crux/kv/index_store.clj
+++ b/crux-core/src/crux/kv/index_store.clj
@@ -473,7 +473,7 @@
   (lru/compute-if-absent cav-cache
                          [content-hash-buffer attr-buffer]
                          #(mapv mem/copy-to-unpooled-buffer %)
-                         (fn [[content-hash-buffer attr-buffer]]
+                         (fn [_]
                            (let [eid-size (mem/capacity eid-value-buffer)
                                  vs (TreeSet. mem/buffer-comparator)
                                  prefix (encode-ecav-key-to nil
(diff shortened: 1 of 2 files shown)


hraberg:
A potentially simple idea is to keep a sentinel value in the cache for larger columns and fall back to the previous behaviour and use the kv store for these large columns. That is, one simply give up scanning after N elements on init (or growth) and replace the cached set with a sentinel value.

PR #1149, opened 2020-10-02
Bench fixes

hraberg:
Fixes to bench, the ts tests hadn't been updated to latest history API.
Improved logging of errors and reporting.
Fix to issue in merge sort when there's nothing to merge.

### DIFF
--- a/crux-bench/src/crux/bench/main.clj
+++ b/crux-bench/src/crux/bench/main.clj
@@ -97,6 +97,8 @@
                                            (System/exit 1))
                                        (update :selected-nodes disj "h2-rocksdb" "sqlite-rocksdb")))]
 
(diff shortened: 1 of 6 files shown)


danmason (approved):

crux-bench/src/crux/bench/main.clj
```diff
@@ -97,6 +97,8 @@
                                            (System/exit 1))
                                        (update :selected-nodes disj "h2-rocksdb" "sqlite-rocksdb")))]
 
+    (bench/post-to-slack (format "Finished Benchmark*, Crux Version: %s, Commit Hash: %s\n"
```
danmason: Needs an extra `*` before "Finished Benchmark*" (otherwise it won't bold the text) but asides from that, works a charm :)

Issue #1110, opened 2020-09-17
Swagger/OpenAPI generation for the HTTP API


jarohen:
* Figure out what's current, and let's deploy this to the docs site with every release

danmason:
Going to pause this until HTTP is finished - was originally writing by hand, may want to look into some aspects of generation for it.

PR #1150, opened 2020-10-02
HTTP JSON API

### LINKED ISSUE #284: JSON API over HTTP - part 1

danmason:
Resolves #284 

N.B.:
- We're choosing not to map `:crux.db/id` to `_id` nor `:crux.db/fn` to `_fn` within users' documents - these become exceptions to the 'everything is namespace-less and camelCased in the JSON API' rule.
- Lists within JSON responses are converted to strings iff their first element is `fn`/`fn*`. Functions are parsed to EDN for `:crux.db/fn` keys in maps.
- We considered parsing and emitting `#crux/id`s as `{"$oid": "<hash>"}` and byte-arrays as `{"$base64": "<base-64>"}`, although we reverted this in the interest of being conservative, as there are many other types to which this might apply. The trade-off is that JSON users cannot submit byte-arrays to Crux. (We do still have Transit writers for these types.)

Implementation:
- [x] Added the muuntaja default transit decoder to the default muuntaja instance (such that it can also be used for submit-tx)
- [x] Added JSON encoders for various endpoints.
- [x] Added a seperate muuntaja instance with a JSON decoder for submit-tx endpoint, (default muuntaja does not contain a default JSON decoder)
- [x] Keywords inside of documents/transactions are stringified explicitly before passing through to the JSON encoder, ensuring that these nested attributes aren't handled by JSON and CamelCased.
- [x] Add general tests for the API.
- [x] Adds docs for the JSON API
- [x] Added a custom encoder for `-queries` endpoints, returning queries as EDN strings (rather than returning invalid JSON encodings)
- [x] Transaction functions are read in/read out with the `"crux.db/fn"` key, values returned as strings in responses
- [x] Need to test eql-projects

Updated HTTP docs can be found [here](https://opencrux.com/json/reference/http.html).


jarohen:
> - Figure out what to do with entity-refs

'just' return the eids, as if link-entities is a no-op for JSON reqs

> - Investigate using spec-tools for conforming parsed JSON (doing `put` -> `:crux.tx/put`, handling dates etc - rather than doing these transformations in muuntaja)

tbc, will read through the PR. in the meantime, feel free to write up specs for the tx-ops if this makes life easier
could also put this into `conform.clj`, but bear in mind that it'll want to be in vector form to be passed to `crux/submit-tx`
_could_ update `crux/submit-tx` to accept map form too, but this is probably too much of a change to make in this PR

> - Need to document behaviour somewhere

tabs on the docs for now, we'll add Swagger/OpenAPI later (#1110)

crux-http-server/src/crux/http_server/util.clj
```diff
@@ -59,12 +79,58 @@
...
+                    (keyword? k) (name k)
+                    :else k))]
+    (walk/prewalk (fn [x]
+                     (cond
+                       (map? x) (into {} (map f x))
+                       (vector? x) (mapv g x)
+                       :else x))
+                   m)))
+
+(defn crux-object-mapper [{:keys [camel-case? mapper-options] :as options}]
```
jarohen, line 101: think this will indiscriminately convert keys to camelCase, even within users' documents?

jarohen (changes requested):
As discussed, on the docs:
* 'All of the REST endpoints return both EDN and Transit' - do they all return JSON now?
* Should the JSON request tabs pass their QPs in camelCase? In fact, do we need a generic note about kebab-case/camelCase...?
* entity-history 
  * missing `eid-json` in required params
  * 'under the `:crux.db/doc` key' - in fact, a few mentions of `crux.tx`, which doesn't apply to JSON
* entity-tx 
  * `eid-json` in required params
  * JSON tab shouldn't use `eid-edn` - just `eid`? (probably applies elsewhere as well)
* query
  * Shouldn't be two `#query` IDs - `#get-query` and `#post-query`?
  * `POST /_crux/query` doesn't accept a JSON body - worth an NB
  * `?query` QP (in hindsight) should be `?query-edn` - are there any more?
* submit-tx - this assumes the JSON reader is already familiar with the Clojure API - we're hoping they won't need to be.
* sync - 'timeout defaulting to 10s' - sure? thought it was 30.

jarohen:
@danmason - @hraberg's introducing a `#crux/base64` reader macro with #1168 - we'll want to encode byte arrays to print this reader macro when that PR's merged. Depends which one's merged first :)

danmason:
Should hopefully be as simple as adding a custom encoder to the object mapper!

crux-http-server/src/crux/http_server/query.clj
```diff
@@ -3,35 +3,31 @@
...
            java.util.Date))
 
-(s/def ::query
+(s/def ::query-edn
   (st/spec
    {:spec any? ; checked by crux.query
     :decode/string (fn [_ q] (util/try-decode-edn q))}))
 
-;; TODO: Need to ensure all query clasues are present + coerced properly
+;; TODO: Need to ensure all query clauses are present + coerced properly
```
danmason, line 28: Likely a redundant TODO - all clauses would be present + coerced when passed into `query-edn`

jarohen (approved):
Nice work @danmason, great to see :)

hraberg (approved):
Looks good, tried to grok if there's a JSON date encoder, or is this provided by default.?

Also, not 100% sure about the nested types in JSON (as opposed to just strings), but we can iterate this. The "oid" concept isn't used elsewhere in Crux.

PR #1156, opened 2020-10-06
Cache interface

### LINKED ISSUE #1153: New Cache Interface and add them to the Topology
Currently all caches are of the same type, created by the `crux.lru/new-cache` function. We have an existing protocol, but this would likely be better replaced by an interface, as to implement a cache you also need to implement some interfaces from clojure.lang (Counted, ILookup).

Some caches are now indirectly managed via other components, but the ones without clear lifespan should instead be their own components, implementing this interface. This would open up to add external, shared, caches (like memcached) for things where this makes sense (documents, values, cav), and making the caches modular like other parts of Crux. The caches which live per query have to be node-local and can be created directly like before, but via the new interface and factories.

We also want to remove the name LRU both from this interface and namespace, as this is an implementation detail.

hraberg:
Moves crux.lru down to crux.cache.lru and adds crux.cache.ICache interface.
Turned most caches into nested components, which are possible to override one by one.

The query engine's smaller caches (for compiled queries etc.) now work like this and the configuration is nested.
These caches are highly unlikely that someone would tune.

The index store's caches are nested onto the index store. The document store cache is added as a nested component to each cached" document store. A further branch, `cached-doc-store-as-config` turns this around, and make the cached document store explicit in the config and removes it from the code. This is cleaner, but requires users to upgrade their config to not lose caching behaviour.

Will close #1153.

### DIFF
--- a/crux-azure-blobs/src/crux/azure/blobs.clj
+++ b/crux-azure-blobs/src/crux/azure/blobs.clj
@@ -50,8 +50,7 @@
                                                       :doc "Azure Storage Account Name"}
                                     :container {:required? true,
                                                 :spec ::sys/string
-                                                :doc "Azure Blob Storage Container"}
-                                    :doc-cache-size ds/doc-cache-size-opt}}
+                                                :doc "Azure Blob Storage Container"}}}
   [{:keys [sas-token storage-account container document-cache] :as opts}]
   (ds/->cached-document-store
    (assoc opts
--- a/crux-azure-blobs/src/crux/azure/blobs.clj
+++ b/crux-azure-blobs/src/crux/azure/blobs.clj
@@ -3,8 +3,7 @@
             [taoensso.nippy :as nippy]
             [crux.db :as db]
             [crux.system :as sys]
-            [crux.document-store :as ds]
-            [crux.lru :as lru]))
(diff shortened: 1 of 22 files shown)


jonpither (approved):

hraberg:
We could consider pushing the cached document store to the config, which is clearer in many ways, but does make it a bit clunkier to swap stores:

```
                 {::j/connection-pool {:pool-opts pool-opts, :dialect dialect, :db-spec db-spec}
                  :crux/tx-log {:crux/module `j/->tx-log
                                :connection-pool ::j/connection-pool}
                  :crux/document-store {:crux/module 'crux.document-store/->cached-document-store
                                        :document-store {:crux/module `j/->document-store,
                                                         :connection-pool ::j/connection-pool}}}
```
JDBC example, but the other stores using caches would be configured in a similar fashion.

hraberg:
Have a second branch, `cached-doc-store-as-config`, where I've moved the cached document store into the node config itself, and removed the nested caches in the various stores. The main drawback of this is that it requires config changes for the user to not lose their document store, it has to be wrapped similarly to earlier comment.

hraberg:
Actually, the above isn't true, the store should keep working, but it would be without a cache if the config isn't updated.

PR #1158, opened 2020-10-07
Second chance + canonical buffer caches (and other performance tweaks)

hraberg:
* Using second chance cache as default cache, this is somewhat faster than our LRU cache, and less prone to overruns (where a single large scan fills and evicts the cache).
* Smaller index store caches by default, this is slower, but safer + have option to use soft values cold cache or adaptive cache.
* Ability to optionally adapt and grow cache sizes after available memory. The cache will be at its configured size at a configurable break even level, defaulting to 95% memory usage. Above this it will keep shrinking the cache to free up memory. Below the cache can grow beyond the configured limit to make use of available memory.
* The index store CAV and value caches use adaptive caching.
* Canonical buffer cache for the index store, used to de-duplicate buffers stored in caches.
* Larger entity cache during query execution to avoid redoing temporal lookup, related to the above.

(edit: pushed to master separately - @jarohen)
* Postpone temporal resolution on AE level until AEV. This was already the case for AVE. This is faster in the general case, but there might be queries where it turns out to be slower. It's a trade off of avoiding doing temporal lookup vs. having more candidates at this level in the join.
* Create attribute id buffers for the binary indexes once instead of per seek.
* Use `ByteBuffer/allocateDirect` in Rocks instead of converting it back and forth via `crux.memory/allocate-unpooled-buffer` which is slower.
* Also fixes a bug in ByteUtils/sha1 (which is not really used).

### DIFF
--- a/crux-bench/src/crux/bench/tpch_test.clj
+++ b/crux-bench/src/crux/bench/tpch_test.clj
@@ -583,4 +583,11 @@
   ;; SQL:
   (slurp (clojure.java.io/resource "io/airlift/tpch/queries/q1.sql"))
   ;; Results:
-  (slurp (clojure.java.io/resource "io/airlift/tpch/queries/q1.result")))
+  (slurp (clojure.java.io/resource "io/airlift/tpch/queries/q1.result"))
+
+  (let [node (user/crux-node)]
+    (time
+     (doseq [n (range 1 23)]
+       (time
+        (let [actual (crux.bench.tpch-test/run-tpch-query node n)]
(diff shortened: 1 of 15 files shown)


jarohen (approved):
Think these all look ok, nice one :) Wouldn't mind a walk-through of the memory characteristics of the soft caches and the related Java objects - one for after our planning catchup this afternoon?

Meta: I found this PR quite hard to review as the GitHub diff jumps around between the different changes, hard to see the whole diff for each change. Wonder if it'd help (so long as it's not too much effort) if the commits could be squashed down per-change, and then could review each change as a separate commit (even if it's all bundled in the same PR)?

hraberg:
Yeah, I agree it can be confusing.
There are a few different but related tweaks going on, let's simply go over it on the call.

PR #1160, opened 2020-10-07
Add training module to antora

luciodale:


jarohen:
suspect this needs a change to the submodule reference to bring in the changes in crux-site?

PR #1161, opened 2020-10-08
Bench variance - migrating nightly bench run to EC2

### LINKED ISSUE #1133: Investigate variance in crux-bench runs
**Tried so far:**

> Check whether we're close to the boundaries of the Fargate resources 
> Check whether doubling the size of the Fargate resources (without doubling the size of the JVM) fixes the issue

Both of the above somewhat linked - increasing memory and CPU provided to a single task didn't see, particularly, much in the way of a decrease in overall various - and we appear far from the fargate boundaries.

> Check whether anyone else is having these issues

After a cursory check, can't find much to suggest that anyone is having particular issues with consistency on Fargate - still, much more research to be done there.

**To check:**
- Investigate moving it to EC2 (hopefully with 'lights-out' policy)

danmason:
Still want to investigate the impact this has further, but a pass over what we currently have:
- `crux-bench-dev` tasks continue to run on Fargate
- `crux-bench` nightly run is on schedule, running on a `m5.xlarge` EC2 instance - this is started 5 minutes before the bench run starts, then stopped 3 hours after the bench run starts

We likely want to investigate multiple different EC2 instance types, compare general variance and performance


jarohen:
Thanks @danmason :)

Do you think it'd be worth re-running the comparison DBs on that setup too?

danmason:
Most definitely - in the name of it being a fair comparison. Can certainly do that (albeit one at a time :) )

PR #1164, opened 2020-10-09
typos in docs: replacing some missing closing parens

deobald:
These are in the EQL Projection stuff so probably no one is using it much yet anyway, but I noticed these a while ago while scanning through that section. No sense adding friction for anyone copy/pasting the examples. :)


jarohen:
Thanks!

PR #1168, opened 2020-10-12
Add nippy and byte-array value types

### LINKED ISSUE #864: Users can submit byte-arrays as values in documents
from @malcolmsparks 

Currently, explicitly prevented by `(UnsupportedOperationException. "Byte arrays as values is not supported.")`. 

It seems users could previously submit byte arrays as values but the ability to do so caused some confusion/bugs in the rest of the codebase (presumably because we use byte-arrays when moving values through Crux?)

@hraberg probably has more insight into the historical context

hraberg:
Adds Nippy and byte array value types. Similar to strings, allows values less than 128 bytes to be encoded inline
Covers things like keywords and uuids indirectly via Nippy, both which are common to use as eids.
Large values are hashed like before, but have their own value types. `object-value-type-id` for Nippy (as before), and `blob-value-type-id` for large byte arrays.

Adding byte arrays as values might have some unforeseen implication, as this was explicitly avoided before.

This PR further improves performance, as it removes most hashed values that need to be looked up indirectly.

[edit: pushed to master separately]
Removes the need for the value cache in `crux.kv.index-store`, as these values are usually either already cached in the temp hash cache (for value created during the query) or are uncommonly accessed larger values.

Requires an index bump.

### DIFF
--- a/crux-core/src/crux/codec.clj
+++ b/crux-core/src/crux/codec.clj
@@ -250,7 +250,8 @@
                                (binding [*sort-unordered-colls* true]
                                  (nippy/fast-freeze this))
                                (nippy/fast-freeze this))]
-      (if (< max-value-index-length (alength nippy-bytes))
+      (if (or (< max-value-index-length (alength nippy-bytes))
+              (not (nippy/freezable? this)))
         (doto (id-function to nippy-bytes)
           (.putByte 0 object-value-type-id))
         (mem/limit-buffer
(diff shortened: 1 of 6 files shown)


hraberg:
There are a few issues on this branch regarding termination of variable length values in the keys. Needs reflection and fixing before merging.

1. Strings used to have a termination byte, but this branch currently has removed that. Probably need to be added back.
2. Byte arrays on this branch have no termination character, so they are broken as prefixes. Needs length prefix which makes them not work for range constraints.
3. Nippy values have length prefixes by design, so they should work, no valid nippy byte array is a prefix of another. Easiest is to use Nippy to encode bytes as well.

hraberg:
Above comment resolved as discussed.

PR #1187, opened 2020-10-20
Fixing TPCH scale factor option on bench

hraberg:


crux-bench/src/crux/bench/tpch_test.clj
```diff
@@ -23,19 +23,20 @@
...
-          (tpch/load-docs! node scale-factor tpch/tpch-entity->pkey-doc)))
-
-      ;; TODO we may want to split this up, à la WatDiv, so that we can see if
-      ;; specific queries are slower than our comparison databases
-      (bench/run-bench :queries
-        {:success? (run-tpch-queries node opts)})
-
-      (bench/run-bench :queries-warm
-        {:success? (run-tpch-queries node opts)}))))
+  (let [scale-factor (or scale-factor 0.01)]
```
danmason, line 25: Think this might be a bit redundant as we are providing it with a default of 0.01 within `bench.clj`?

danmason (approved):
One very minor nitpick/comment but aside from that seems good :)

PR #1193, opened 2020-10-20
EQL unions, recursive queries and attribute renaming #1184

### LINKED ISSUE #1184: EQL projection union/recursive queries syntax
https://edn-query-language.org/eql/1.0.0/specification.html#_unions

https://edn-query-language.org/eql/1.0.0/specification.html#_recursive_queries

jarohen:
Adding unions, recursive queries and attribute renaming to our EQL support.

Notes:
* I've extracted the projection code out to its own namespace, `crux.project`, and simplified it significantly - would recommend reading the final state rather than the diff.
* Unions look like joins once EQL's parsed them, except that queries specified in a union are to apply to the _same_ document, at the same level - nested queries specified within a join apply to the child document.
* We add a `RecurseState`, a pair of the current recursion depth and the projection functions for the current level. When we recurse, we increment the recurse-depth (checking it against the limit if it's bounded) and use the same projection functions at the next level; for 'normal' nested queries, we reset the recurse-depth to 0 and use the next level's projection functions.
* Have also taken the liberty of adding an `:as` parameter to rename attributes in the result - see https://clojurians.slack.com/archives/CG3AM2F7V/p1603271755193300

### DIFF
--- a/crux-core/src/crux/eql_project.clj
+++ b/crux-core/src/crux/eql_project.clj
@@ -1,0 +1,154 @@
+(ns ^:no-doc crux.eql-project
+  (:require [crux.codec :as c]
+            [crux.db :as db]
+            [edn-query-language.core :as eql]
+            [clojure.string :as string])
+  (:import clojure.lang.MapEntry))
+
+(defn- recognise-union [child]
+  (when (and (= :join (:type child))
+             (= :union (get-in child [:children 0 :type])))
+    :union))
+
+(defn- replace-docs [v docs]
+  (if (not-empty (::hashes (meta v)))
+    (v docs)
+    v))
+
+(defn- lookup-docs [v {:keys [document-store]}]
+  (when-let [hashes (not-empty (::hashes (meta v)))]
+    (db/fetch-docs document-store hashes)))
+
+(defmacro let-docs [[binding hashes] & body]
+  `(-> (fn ~'let-docs [~binding]
+         ~@body)
+       (with-meta {::hashes ~hashes})))
+
+(defn- after-doc-lookup [f lookup]
+  (if-let [hashes (::hashes (meta lookup))]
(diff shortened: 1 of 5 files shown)


crux-core/src/crux/project.clj
```diff
@@ -0,0 +1,182 @@
...
+
+(defn- forward-joins-child-fn [{:keys [props special forward-joins unions]}]
+  (when-not (every? empty? [props special forward-joins unions])
+    (let [forward-join-child-fns (for [{:keys [dispatch-key] :as join} forward-joins]
+                                   (let [next-recurse-state (->next-recurse-state-fn join)
+                                         k (or (get-in join [:params :as] dispatch-key))]
+                                     (fn [doc db recurse-state]
+                                       (when-let [v (get doc dispatch-key)]
+                                         (when-let [recurse-state (next-recurse-state recurse-state)]
+                                           (->> (if (or (vector? v) (set? v))
```
hraberg: There's a crux.codec/multiple-values? for this check.

crux-core/src/crux/project.clj
```diff
@@ -0,0 +1,182 @@
+(ns ^:no-doc crux.project
```
hraberg: Not sure what a better name would be, but project.clj can be a bit confusing :-)

hraberg (approved):
LGTM.

crux-core/src/crux/project.clj
jonpither: `crux.eql.project` ?

jonpither (approved):
Nice - good factoring out of project concerns into project.clj

crux-core/src/crux/project.clj
jarohen: `crux.eql-project`

PR #1198, opened 2020-10-22
Removing meta from Nippy encodings when creating value-buffers, #1197

### LINKED ISSUE #1197: Hashes that include quoted lists are inconsistent
```clojure
(crux/submit-tx (crux-node) [[:crux.tx/put {:crux.db/id :with-list, :a-list '(1 2 3)}]])

(crux/q (crux/db (crux-node))
        '{:find [?e]
          :where [[?e :a-list '(1 2 3)]]})
;; => #{}
```

Caused by `->value-buffer` returning a different value for `'(...)` every time - nippy includes metadata when freezing lists, and quoted forms have line/column info in their metadata.

jarohen:
Not removing it when creating id buffers to preserve backwards compatibility of id hashing.

### DIFF
--- a/crux-core/src/crux/codec.clj
+++ b/crux-core/src/crux/codec.clj
@@ -3,18 +3,16 @@
   (:require [clojure.edn :as edn]
             [crux.error :as err]
             [crux.hash :as hash]
-            [crux.memory :as mem]
-            [crux.morton :as morton]
-            [taoensso.nippy :as nippy]
             [crux.io :as cio]
-            [clojure.walk :as walk])
-  (:import [clojure.lang IHashEq Keyword APersistentMap APersistentSet]
-           [java.io Closeable Writer]
+            [crux.memory :as mem]
+            [taoensso.nippy :as nippy])
+  (:import [clojure.lang APersistentMap APersistentSet IHashEq Keyword]
+           java.io.Writer
            [java.net MalformedURLException URI URL]
-           [java.nio ByteOrder ByteBuffer]
+           [java.nio ByteBuffer ByteOrder]
            java.nio.charset.StandardCharsets
-           [java.util Arrays Base64 Collection Date Map UUID Set]
-           [org.agrona DirectBuffer ExpandableDirectByteBuffer MutableDirectBuffer]
+           [java.util Base64 Date Map Set UUID]
+           [org.agrona DirectBuffer ExpandableArrayBuffer ExpandableDirectByteBuffer MutableDirectBuffer]
            org.agrona.concurrent.UnsafeBuffer))
 
 (set! *unchecked-math* :warn-on-boxed)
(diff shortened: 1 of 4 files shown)


crux-core/src/crux/memory.clj
```diff
@@ -227,7 +227,13 @@
              (.putByte idx (unchecked-byte (inc b))))))))))
 
 (defn <-nippy-buffer [buf]
-  (nippy/thaw-from-in! (DataInputStream. (DirectBufferInputStream. buf))))
+  (nippy/thaw-from-in! (-> (DirectBufferInputStream. buf)
+                           (DataInputStream.))))
 
 (defn ->nippy-buffer [v]
-  (->off-heap (nippy/fast-freeze v)))
+  (let [to (ExpandableDirectByteBuffer. 64)
```
hraberg, line 234: You probably want to support both on and off heap directly.
The above will allocate off heap, only to copy in back on heap. You can use ByteArrayOutputStream in a similar way and wrap it in a DataOutputStream.

Though often there's already a buffer at hand you can encode into though, value->buffer for example gets handed a `to` buffer. This would be a further optimisation though, as this isn't done today, there's always an extra on heap copy made in fast-freeze.

hraberg: There's also a `putBytes` overload on `MutableDirectBuffer` that can take another buffer (regardless if it's on or off heap), so you could also avoid to do the on-heap step that way.

jarohen: Good shout. I've added an arity to take an output buffer, defaulted to off-heap (because that's what it was previously). For the new usages I've added, I've explicitly specified an on-heap buffer - would that work?

jarohen: Ah, sorry, your second comment didn't show up on the UI until I replied. Will take a look

jarohen: as discussed, try not to copy buffers where possible in the Object value->buffer impl. particularly, id-function can accept a buffer - check the `on-heap` calls

jarohen: @hraberg have removed the extra `->on-heap` call. Looking for a way to use `slice-buffer` with an expandable ByteBuffer - we don't currently have a version of `slice-buffer` without a `limit` parameter (and, indeed, there isn't an `UnsafeBuffer` constructor that doesn't require a `limit` parameter) - I've left this one for now. Alternative might be to pass a buffer and an offset into `->nippy-buffer`, but given that's not a pattern we use elsewhere I've not opted for it for now?

crux-core/src/crux/codec.clj
```diff
@@ -231,24 +229,24 @@
 
   Class
   (value->buffer [this ^MutableDirectBuffer to]
-    (doto (id-function to (nippy/fast-freeze this))
+    (doto (id-function to (mem/->on-heap (mem/->nippy-buffer this)))
```
hraberg: Don't need to ->on-heap call here.

hraberg (approved):
Looks OK now, doesn't add any extra copying compared to how it was before, trying to avoid the extra copy at all and decode into the given buffer can be done later if deemed necessary.

PR #1199, opened 2020-10-22
Query capabilities - extend / update query documentation

### LINKED ISSUE #1088: Ensure that all the various things you can do in queries are documented up
Particularly, for the use cases where we're currently recommending `:args`, we should now recommend `:in`.

See #1199 for the addition of `:in`/sub-queries. Recommend that the implementer uses the query spec as the authority of what's possible in the query engine - we're still light on or-joins/not-joins etc.

danmason:
#1088 part 1

See changes here: https://opencrux.com/queries/reference/queries.html

[edit (James) still to come: or-joins/not-joins etc]

### DIFF
--- a/crux-test/test/crux/examples_test.clj
+++ b/crux-test/test/crux/examples_test.clj
@@ -28,6 +28,12 @@
       (ex/query-example-streaming node #(swap! !results conj %))
       (t/is (= [[:smith]] @!results)))))
 
+(t/deftest test-example-sub-queries
+  (with-open [node (crux/start-node {})]
+    (t/is (= #{[[[4]]]} (ex/query-example-subquery-1 node)))
+    (t/is (= #{[4]} (ex/query-example-subquery-2 node)))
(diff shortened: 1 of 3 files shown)


jarohen (changes requested):
Cheers @danmason :) 

One thing I'd change: previously, with only have one parameter to `crux/q`, it was fine to have just the query in the examples - now that we have multiple arguments, I'd be tempted to include the surrounding call to `crux/q` in the examples as well, to show that these are multiple arguments to the same function?

Wouldn't mind another pair of eyes (@hraberg, maybe?) over the query examples to make sure the brackets are all correct :) (probably easiest to review [here](https://opencrux.com/queries/reference/queries.html))

jarohen:
Thanks @danmason - looking good. We're still lacking in docs about or-joins/not-joins etc, and doubtless a few other things that are now possible in the query syntax - but this is a good change, so will merge it. Will keep the original card open, and add a note to the implementer to use the query spec as the authority on what's possible.

Issue #1155, opened 2020-10-05
Graphs for bench runs over time


jonpither:
* Line chart containing lines for Crux's Watdiv and TPC-H ingest and query performance over the last 3 months
* Posted via email/slack.

jarohen:
@jonpither - mind writing up what graphs you'd be looking for here? e.g. 'Crux performance over time', 'Crux vs comparison DBs, individual queries', etc? maybe what you'd want to do with the graphs, too, if it isn't sensitive?

jarohen:
CloudWatch insights spike link available, pinned to the #crux-bench channel

jarohen:
Discussed with @jonpither and @refset - let's merge this as an increment (after a code-review) but there's a wider discussion to be had around use-cases and context

jarohen:
Accidentally closed - @jonpither, as discussed, over to you :)

PR #1202, opened 2020-10-23
CloudWatch bench graphs, #1155

jarohen:
Pushing bench timings to CloudWatch metrics so that we can generate graphs and dashboards.

There's one large commit here standardising on `:success? true`, so that we can look out for it in result parsing - maybe best to filter to the middle commit.

TODO:
* [x] Add env var to docker image (but only overnight runs?)
* [ ] Pick graphs to be included on CloudWatch dashboard, and share link

### DIFF
--- a/crux-bench/cloudformation.yaml
+++ b/crux-bench/cloudformation.yaml
@@ -199,6 +199,9 @@ Resources:
             - Name: "SLACK_URL"
               ValueFrom:
                 "arn:aws:secretsmanager:eu-west-2:955308952094:secret:bench/slack-url-uumMHQ"
+          Environment:
+            - Name: 'CRUX_BENCH_CW_METRICS'
+              Value: 'true'
           LogConfiguration:
             LogDriver: awslogs
             Options:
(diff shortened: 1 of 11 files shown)


crux-bench/cloudformation.yaml
```diff
@@ -199,6 +199,9 @@ Resources:
             - Name: "SLACK_URL"
               ValueFrom:
                 "arn:aws:secretsmanager:eu-west-2:955308952094:secret:bench/slack-url-uumMHQ"
+          Environment:
```
danmason, line 202: Given that we're only adding this here (and not within `crux/run-bench`), am I correct in understanding that by default we're only ever collecting these metrics for the master `crux-bench` runs, and not for the `crux-bench-dev` runs?

jarohen: Yep :)

jarohen:
While CloudWatch graphs might be useful for development, their flexibility is some way off what we'd like for other contexts. Once @jonpither and @refset have confirmed contexts and use-cases on #1155, let's have another bash :)

hraberg (approved):

PR #1209, opened 2020-10-28
`nil` transact times

jarohen:
Fixing edge cases around the empty DB, and ensuring that we fix a `tx-time` in the HTTP API `db` call if the user doesn't provide one.

* For ease, we eagerly fetch the latest-completed tx when `db`'s called on the remote API client - we could consider including vt/tt as a response header, and fixing after the first response (which may well be a call to `transactionTime`)
* We accept (for now) the bug that if a user opens a `db` on an empty remote API client (no txs yet) that it may not act like a snapshot if the user makes subsequent calls on the same instance.


jonpither (approved):
Noticed the TODO, do we have an issue to link to?

hraberg (approved):

PR #1214, opened 2020-10-30
Using transaction-id as the transaction basis of `db` instances, #1205

### LINKED ISSUE #1205: Using tx-id as the transaction basis
context:
* currently, users can only specify a tx-time to `db` to use as a basis. this is an issue because there may (and have been, in our tests) be multiple transactions at the same tx-time.
* users may well want to submit a tx-id to `db`, given this is the actual identity of the transaction
* while we have a index version bump going through, it'd be a good time to consider changes

proposed solution:

* [x] Bitemp indices to have tx-id first - E VT TID TT C and E Z(VT, TID) TT C (option to remove tx-time if this is easier). 
* [x] Query engine to use `tx-id` throughout rather than `tx-time`
* [x] Existing `db` arities that accept `tx-time` to convert to `tx-id` early on using a (new) mapping index
* [x] `db` functions to accept another arity `(crux/db node vt {:keys [tx-time tx-id]})`
* [x] Entity history needs to additionally accept `tx-id` in `:start` and `:end`, and needs to convert start/end `tx-time` to `tx-id` early on

jarohen:
Index changes:

In the end, I opted for swapping the bitemp indices to E VT TID TT -> C and E Z(VT, TID) TT -> C, and added an addition TT TID index. 

TODO:
* [x] Docs
* [x] Calcite?
* [x] Re-do arities to `db`
* [x] Rename `transaction` to `as-of`
* [x] Check console

Notes/discussion points:

* Users can specify `tx-id`, `tx-time`, both, or neither to DB. If they just specify `tx-id`, we don't currently resolve the `tx-time` - we could, but the only way I can think of implementing this is an addition TID TT index? (If we decided to do this, I'd include it in this PR). Related - see the index-store test for how we resolve whatever the user passes us.
* I've added a `transaction` function to `db`, returning the transaction basis, but I'm not 100% convinced this conveys the essence of what it's returning? maybe `transaction-basis`, `transaction-upper-bound`, `snapshot-transaction` or something similar?
* The HTTP API should arguably use `tx-time` as the QP throughout, as that's what we (now) use in the rest of Crux - haven't made this change for the time being in the interest of not making this PR bigger than it has to be, although we'd be wise to do it before we release 1.13.0, given we're essentially re-writing the HTTP API in this release anyway.
* Type hinting can be especially onerous at times. You can't type hint a nil, hence needing to bind `nil` to a local variable and type hint the local. Also new to me was that I was still getting a reflection warning passing a literal map, and even after I'd type hinted the literal as `^Map` - so, again, I had to bind the map to a local and type hint the local, which appeased The Compiler.
* I'd like to discuss whether we should expose the `:crux.tx/*` and `:crux.db/*` namespaces in public APIs - it feels like the user shouldn't care that we have two different namespaces internally. Not one for here, though - have created a separate note.
* I've re-written `HistoryOptions` - previously it was a Java class, and we converted back-n-forth between that and a map - now it's an interface implemented by a record, so we don't need to convert. It's a breaking change to the Java API, technically.
* I've removed `next-monotonic-date` - now that we're using tx-id as the basis there's less of a need for it - in the end it was only being used in test fixtures.
* In a similar vein, there's now a slight discrepancy between the Java history options and the Clojure history options - the former has `startValidTime` and `startTx`, the latter a `:start` map with vt, tt and tid. I'm slightly leaning towards _not_ making these consistent until we've finalised the temporal semantics for DB - particularly, the 'tx-id as an upper bound' conversation.
* I'm still of the opinion that valid-time should also be in the map to `db`, given it's optional.


hraberg (approved):
It's quite chunky, but overall looks good when skimming it, no surprises!

crux-core/src/crux/api.clj
```diff
@@ -155,30 +161,23 @@
...
+       (.db this valid-time ^Date tx-or-tx-time)
+       (.db this valid-time ^Map tx-or-tx-time))))
 
   (open-db
     ([this] (.openDB this))
-    ([this ^Date valid-time] (.openDB this valid-time))
-    ([this ^Date valid-time ^Date transaction-time] (.openDB this valid-time transaction-time))))
+    ([this valid-time] (.openDB this valid-time))
+    ([this ^Date valid-time tx-or-tx-time]
+     (if (inst? tx-or-tx-time)
```
jarohen: instanceof Date

crux-core/src/crux/codec.clj
```diff
@@ -49,6 +49,8 @@
 (def ^:const av-index-id 10)
 (def ^:const ae-index-id 11)
 
+(def ^:const tx-time-mapping-id 12)
```
jarohen, line 52: bump

crux-test/test/crux/kv/index_store_test.clj
```diff
@@ -194,3 +180,40 @@
...
+        (t/is (thrown? IllegalArgumentException
+                       (db/resolve-tx index-snapshot #::tx{:tx-time #inst "2021", :tx-id 1})))
+
+        ;; This is an edge case - arguably it should throw IAE,
+        ;;   but instead returns an instant with tx-time "2022", but before tx-id 1 was ingested.
+        ;; The query engine would still behave consistently in this case (with tx-id = 0),
+        ;;   and it's unlikely that a user will hit it unless they explicitly ask for it,
+        ;;   so it's a philosophical question as to whether we should allow it or check for it :)
+        (t/is (= #::tx{:tx-time #inst "2022", :tx-id 0}
+                 (db/resolve-tx index-snapshot #::tx{:tx-time #inst "2022", :tx-id 0})))
```
jarohen: yes, fix this :)

jonpither (approved):

PR #1244, opened 2020-11-10
Add `--repeat` flag to bench.

danmason:
Allows you to rerun the entire bench run multiple times - in this instance, I could use it to run a bench on FARGATE of watdiv on standalone-rocksdb ten times by calling the following command: `./bin/run-bench --tests watdiv --nodes standalone-rocksdb --repeat 10`

### DIFF
--- a/crux-bench/src/crux/bench/main.clj
+++ b/crux-bench/src/crux/bench/main.clj
@@ -80,7 +80,12 @@
                          [nil "--tpch-scale-factor 0.01" "Scale factor for regular TPCH test"
                           :id :tpch-scale-factor
                           :default 0.01
-                          :parse-fn #(Double/parseDouble %)]])]
+                          :parse-fn #(Double/parseDouble %)]
+
+                         [nil "--repeat 10" "Number of times to repeat the current bench run"
+                          :id :repetitions
+                          :default 1
+                          :parse-fn #(Long/parseLong %)]])]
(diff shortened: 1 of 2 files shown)


crux-bench/src/crux/bench/main.clj
```diff
@@ -80,20 +80,28 @@
...
       options)))
 
-(defn run-benches [{:keys [selected-nodes selected-tests] :as opts}]
+(defn run-benches [{:keys [selected-nodes selected-tests repetitions] :as opts}]
   (let [nodes (select-keys bench/nodes selected-nodes)]
     (bench/with-embedded-kafka
-      (->> (for [test-fn (vals (select-keys bench-tests selected-tests))]
-             (test-fn nodes opts))
-           (into [] (mapcat identity))))))
+      (->>
```
jarohen: @danmason - doubt you need the double `for`/`into [] mapcat identity` here - one pair should do it?

jarohen:
```clojure
(->> (for [n (range repetitions)
           test-fn (vals (select-keys bench-tests selected-tests))]
       (test-fn nodes opts))
     (into [] (mapcat identity)))
```

danmason: Ahh, had wondered - thanks james, will give it a go :)

PR #1246, opened 2020-11-10
Build api docs on master

### LINKED ISSUE #1225: Javadoc/Swagger docs built on master
We should build this documentation on `master` CI runs - both to test it out on a more frequent basis than just releases, and also so that we can show `master` documentation on the site

danmason:
Resolves #1225 - adds links to master versions javadocs/redoc in the docs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,6 +200,12 @@ workflows:
       - test:
           requires:
             - build
+      - create-api-docs:
+          requires:
+            - build
+          filters:
+            branches:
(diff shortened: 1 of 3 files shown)


jarohen:
Thanks @danmason - any way we can use an Antora variable, say, so that we don't have to update the adoc on release?

danmason:
Well, I could probably use the *version* variable we already have in `antora.yml` - if I make circle deploy the master docs to `/master/<doc-name>`

PR #1247, opened 2020-11-11
Try to tweak MALLOC_ARENA_MAX

hraberg:
Try to tweak MALLOC_ARENA_MAX via mallopt if not already set.

### DIFF
--- a/crux-core/src/crux/cli.clj
+++ b/crux-core/src/crux/cli.clj
@@ -68,7 +68,7 @@
     shutdown?))
 
 (defn start-node-from-command-line [args]
-  (cio/try-set-malloc-arena-max 4)
+  (cio/try-set-malloc-arena-max 2)
   (cio/install-uncaught-exception-handler!)
 
   (let [{::keys [errors help node-opts]} (parse-args args)]
--- a/crux-core/src/crux/cli.clj
+++ b/crux-core/src/crux/cli.clj
@@ -68,6 +68,7 @@
     shutdown?))
 
 (defn start-node-from-command-line [args]
+  (cio/try-set-malloc-arena-max 4)
   (cio/install-uncaught-exception-handler!)
 
   (let [{::keys [errors help node-opts]} (parse-args args)]
--- a/crux-build/docker/Dockerfile
+++ b/crux-build/docker/Dockerfile
@@ -8,4 +8,5 @@ EXPOSE 3000
 ADD deps.edn deps.edn
 RUN clojure -Sforce -Spath >/dev/null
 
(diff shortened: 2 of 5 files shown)


crux-core/src/crux/cli.clj
```diff
@@ -68,6 +68,7 @@
     shutdown?))
 
 (defn start-node-from-command-line [args]
+  (cio/try-set-malloc-arena-max 4)
```
jarohen: Not all paths to Crux nodes come through this function - if the user's embedding Crux in their own JVM app, for example. One for `crux.node/start-node` maybe, or even a top-level side effect, given it applies to the JVM as a whole?

jarohen (changes requested):
Thanks @hraberg. Understand that we want to make this the default behaviour - should we allow the user an opt-out?

hraberg:
To opt out, one can set MALLOC_ARENA_MAX explicitly, like otherwise we end up in a situation where you have to put some other flag in, in which case we can just tell them to set this env var. I'm fine with that, in which case we don't strictly need this PR.

crux-core/src/crux/cli.clj
hraberg: The idea is that this is mainly interesting in the case where someone starts up the actual app via command line, but we want to avoid them having to set this environment variable, doing this in the node feels a bit too intrusive.

As mentioned in the other comment I'm not sure. Like this tries to nudge Crux to be a bit better behaved in the general case.

jarohen:
Agreed, that's preferable - is there something that they can set `MALLOC_ARENA_MAX` to that would behave as if neither they nor us had set it at all?

hraberg:
The implicit default is 8 * cores, so that would be similar. That said, an alternative would be to have a way to explicitly turn off this tweak, via another tweak..., in the unlikely case someone needs the default behaviour.

Again, the general idea here is only to have reasonable defaults without having to set the environment variable, most people won't notice or care about this (unless they run out of memory). I'm also fine with not merging this and only document the environment variable, as that's less magical, which was the original plan before we found this glibc hook (mallopt).

jonpither:
I'm good to merge this - leaves us in a better state - now with the warning in start-node. I agree we shouldn't set the flag when Crux is embedded within an application. The dockerfile and the start fn for the command line are the right places.

refset (approved):

jonpither (approved):

jonpither:
Upon consideration and conversation, I've removed the call in cli with a commit to master. With the docker config and the fn call available, with the warning, we have enough.

PR #1249, opened 2020-11-11
Nippy bump to 3.1.1

### LINKED ISSUE #1232: Bump to Nippy 3.1
* Check changelog
* Check tx-logs written under Nippy < 3.x are still readable with Nippy >= 3.x

danmason:
Resolves #1232

### DIFF
--- a/crux-core/src/crux/kv/index_store.clj
+++ b/crux-core/src/crux/kv/index_store.clj
@@ -659,7 +659,8 @@
       (or (.get temp-hash-cache value-buffer)
           (let [i @decode-value-iterator-delay]
             (when (advance-iterator-to-hash-cache-value i value-buffer)
-              (some-> (kv/value i) (mem/<-nippy-buffer)))))))
+              (cio/with-nippy-thaw-all
+                (some-> (kv/value i) (mem/<-nippy-buffer))))))))
 
   (encode-value [this value]
     (let [value-buffer (c/->value-buffer value)]
--- a/crux-azure-blobs/src/crux/azure/blobs.clj
+++ b/crux-azure-blobs/src/crux/azure/blobs.clj
@@ -2,6 +2,7 @@
   (:require [clj-http.client :as http]
             [taoensso.nippy :as nippy]
             [crux.db :as db]
+            [crux.io :as cio]
             [crux.system :as sys]
             [crux.document-store :as ds]))
 
@@ -34,12 +35,13 @@
          (run! deref)))
 
   (fetch-docs [_ docs]
-    (reduce
-     #(if-let [doc (get-blob sas-token storage-account container (str %2))]
-        (assoc %1 %2 (nippy/thaw doc))
-        %1)
-     {}
-     docs)))
+    (cio/with-nippy-thaw-all
+      (reduce
+       #(if-let [doc (get-blob sas-token storage-account container (str %2))]
+          (assoc %1 %2 (nippy/thaw doc))
+          %1)
+       {}
+       docs))))
 
 (defn ->document-store {::sys/deps {:document-cache 'crux.cache/->cache}
                         ::sys/args {:sas-token {:required? true
--- a/crux-core/project.clj
+++ b/crux-core/project.clj
@@ -8,7 +8,7 @@
                  [org.slf4j/slf4j-api "1.7.30"]
                  [org.clojure/spec.alpha "0.2.176"]
                  [com.stuartsierra/dependency "0.2.0"]
-                 [com.taoensso/nippy "2.15.1"]
+                 [com.taoensso/nippy "3.1.1"]
                  [org.clojure/tools.reader "1.3.2"]
                  [org.clojure/data.json "1.0.0"]
                  [org.clojure/tools.cli "1.0.194"]
--- a/crux-core/project.clj
+++ b/crux-core/project.clj
@@ -9,7 +9,7 @@
                  [org.clojure/spec.alpha "0.2.176"]
                  [com.stuartsierra/dependency "0.2.0"]
                  [com.taoensso/nippy "3.1.1"]
-                 [org.clojure/tools.reader "1.3.2"]
+                 [org.clojure/tools.reader "1.3.4"]
                  [org.clojure/data.json "1.0.0"]
                  [org.clojure/tools.cli "1.0.194"]
                  [org.agrona/agrona "1.7.2"]
--- a/crux-core/src/crux/document_store.clj
+++ b/crux-core/src/crux/document_store.clj
@@ -1,5 +1,6 @@
 (ns ^:no-doc crux.document-store
   (:require [clojure.java.io :as io]
+            [crux.io :as cio]
             [clojure.set :as set]
             [crux.codec :as c]
             [crux.db :as db]
@@ -14,19 +15,20 @@
 (defrecord FileDocumentStore [dir]
   db/DocumentStore
   (fetch-docs [this ids]
-    (persistent!
-     (reduce
-      (fn [acc id]
-        (let [doc-key (str (c/new-id id))
-              doc-file (io/file dir doc-key)]
-          (if-let [doc (when (.exists doc-file)
-                         (with-open [in (FileInputStream. doc-file)]
-                           (some->> in
-                                    (DataInputStream.)
-                                    (nippy/thaw-from-in!))))]
-            (assoc! acc id doc)
-            acc)))
-      (transient {}) ids)))
+    (cio/with-nippy-thaw-all
+      (persistent!
+       (reduce
+        (fn [acc id]
+          (let [doc-key (str (c/new-id id))
+                doc-file (io/file dir doc-key)]
+            (if-let [doc (when (.exists doc-file)
+                           (with-open [in (FileInputStream. doc-file)]
+                             (some->> in
+                                      (DataInputStream.)
+                                      (nippy/thaw-from-in!))))]
+              (assoc! acc id doc)
+              acc)))
+        (transient {}) ids))))
 
   (submit-docs [this id-and-docs]
     (doseq [[id doc] id-and-docs
--- a/crux-core/src/crux/io.clj
+++ b/crux-core/src/crux/io.clj
@@ -148,47 +148,55 @@
 
 (def ^:const default-external-sort-part-size (* 1024 1024))
 
+(defn with-nippy-thaw-all* [f]
+  (binding [nippy/*thaw-serializable-allowlist* #{"*"}]
+    (f)))
+
+(defmacro with-nippy-thaw-all [& body]
+  `(with-nippy-thaw-all* (fn [] ~@body)))
+
 (defn external-sort
   ([seq]
    (external-sort compare seq))
   ([comp seq]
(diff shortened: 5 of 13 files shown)


danmason:
Note: Fails dbpedia tests, I believe it's likely because of a breakage in the nippy serdes for kafka - 
EDIT: Been looking into this most of this morning, don't think it's actually a kafka specific problem, looks like dbpedia fails on all node types -  returns an error while thawing data: 
```
No reader provided for custom type id: -96
   {:type-id -96, :prefixed? nil}
```

danmason:
Attempted to comment out `dbpedia-test`, get a set of similar exceptions to the one about within `tx-test` - will likely investigate there. Seems, again, to be related to how we are loading ntriples within RDF.

danmason:
Eventually tracked the thing throwing the exception down to a particular set, and then removed elements from that set until it no longer threw an error. This allowed me a smaller repro, as described below.

The following is enough to get the `No reader provided for custom type id: -96` exception:
```
(crux/submit-tx
 (crux-node)
 [[:crux.tx/put
   {:crux.db/id :picasso
    :set #{:https://books.google.com/books?id=pbBIAQAAIAAJ&q=Becht-J%C3%B6rdens+Wehmeier&dq=Becht-J%C3%B6rdens+Wehmeier&hl=de&sa=X&ei=dbRVUvqzDc_GswaT-4CIAw&ved=0CDMQ6AEwAA
           :https://www.guggenheim.org/artwork/artist/pablo-picasso
           (keyword "http://collections.lacma.org/node/166980")}}]])
``` 

Removing any single one of those elements stops it from returning the error.

Though, removing the first one actually lets you get back something reasonable within the tx-ops of `open-tx-log`.

```
{:crux.db/id :picasso,
 :set
  #{:https://www.guggenheim.org/artwork/artist/pablo-picasso
     :http://collections.lacma.org/node/166980}}
``` 
Say if you remove the middle element, this is the document you get back:

```
{:crux.db/id :picasso, :set #{0 :}}
```

Also of note is that I did not directly run into this error upon using `submit-tx` - it wasn't until I called `open-tx-log` that I received the exception.

**EDIT**:

Some further findings. Taking out the set from the equation and working directly with these keywords, I found this as a further repro of the exception.

```
(crux/submit-tx (crux-node) [[:crux.tx/put {:crux.db/id :picasso
                                            :bookone :https://books.google.com/books?id=pbBIAQAAIAAJ&q=Becht-J%C3%B6rdens+Wehmeier&dq=Becht-J%C3%B6rdens+Wehmeier&hl=de&sa=X&ei=dbRVUvqzDc_GswaT-4CIAw&ved=0CDMQ6AEwAA
                                            :booktwo (keyword "http://collections.lacma.org/node/166980")}]])
```

Followed by submitting a seperate doc, and a call to `(iterator-seq (crux/open-tx-log (crux-node) nil true))` is enough to return the exception. Again, removing either one of these appears to prevent the exception being thrown. Also of note is that reversing the order of the two elements also prevents the error. With that in mind, I wrote this as another repro:

```
(crux/submit-tx (crux-node) [[:crux.tx/put {:crux.db/id :picasso
                                            :bookone :https://books.google.com/books?id=pbBIAQAAIAAJ&q=Becht-J%C3%B6rdens+Wehmeier&dq=Becht-J%C3%B6rdens+Wehmeier&hl=de&sa=X&ei=dbRVUvqzDc_GswaT-4CIAw&ved=0CDMQ6AEwAA
                                            :test "testing"}]])
```

It appears that the problem lies entirely with the content under `:bookone` - as this for some reason is treated as a custom type with no reader rather than a keyword.

danmason:
Issue appears to be related to the length of the keyword - the following also causes an exception.

```
(crux/submit-tx (crux-node) [[:crux.tx/put {:crux.db/id :picasso
                                            :bookone :aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                            :test "testing"}]])
```

It's important to note that within the 2.15 and 3.1 on nippy there were changes to `write-kw` - which are worth considering the impact of.

danmason:
Okay, had a bit of a look at the bytes outputted from calls to ->nippy-buffer, think I found the problem.

When writing a keyword with `write-kw`, 3 things are written to the dataoutput
- The type-id of the keyword (dependant on the length of keyword, but either 77 for medium sized keywords or 106 for small keywords)
- The length of the keyword string byte array.
- The bytes of the keyword string byte array.

For example, if I write `:aaaaaa` to the dataoutput, I get the following within the byte array:

```
...106 6 97 97 97 97 97 97...
```
Now, with the larger keywords, I get out the following (the length of the keyword used in this example was 175, a medium sized keyword):

```
...77 0 0 0 -81 97 97 97 97 97 97 97 (and so on)...
```

Now, attempting to call `thaw-from-in!` on this later returns the following exception:

```
No reader provided for custom type id: -81
```

Essentially, `thaw-from-in!` got a type-id from the last byte of the length, rather than the type-id of the medium sized keyword. Investigating `nippy.clj`, this is because the length is expected as a short (https://github.com/ptaoussanis/nippy/blob/b61bf594a1cae2c30bb77c1236d283d03db91f55/src/taoensso/nippy.clj#L1643 see `read-md-count`) - and is instead written as an integer (https://github.com/ptaoussanis/nippy/blob/b61bf594a1cae2c30bb77c1236d283d03db91f55/src/taoensso/nippy.clj#L652 see `write-lg-count`).

danmason:
Made a PR over on Nippy with the fix to `write-kw` - https://github.com/ptaoussanis/nippy/pull/138. Building this and using the JAR locally, it fixes the issues within our tests.

ptaoussanis:
Thanks for all the work tracking this down!

danmason:
My pleasure - thank you for the quick response!

crux-core/src/crux/codec.clj
```diff
@@ -94,6 +94,7 @@
   (mem/limit-buffer to id-size))
 
 (def ^:dynamic ^:private *sort-unordered-colls* false)
+(alter-var-root #'nippy/*freeze-serializable-allowlist* (fn [_] nippy/*thaw-serializable-allowlist*))
```
jonpither: Would this change Nippy for all uses of Nippy, even when Crux is embedded in a users application?

jonpither (commented):
Looks good @danmason. I left a Q that it would be good to touch on before I approve.

crux-core/src/crux/codec.clj
danmason: That's a very good question - in short, I'm not sure, but I think you may well be correct that it would have that behaviour. I'll check that out - I suppose alternatively we could make this a `binding` call and make a function to do it within `crux.io`, ensure that any calls we have to use freeze/thaw do the binding.

jonpither: yes, trying to avoid an alter-var-root with a binding would be preferable.

danmason:
As a note from the nippy docs:

> Allowlist values may be overridden with `binding`, `alter-var-root`, or:  
> - `taoensso.nippy.<freeze/thaw>-serializable-allowlist-base` JVM property
> - `taoensso.nippy.<freeze/thaw>-serializable-allowlist-add`  JVM property
>    - `TAOENSSO_NIPPY_<FREEZE/THAW>_SERIALIZABLE_ALLOWLIST_BASE` env var
>    - `TAOENSSO_NIPPY_<FREEZE/THAW>_SERIALIZABLE_ALLOWLIST_ADD`  env var

Seems like alter-var-root, the env var and the JVM property would all have the potential to impact a project outside of Crux - so it seems as if we should be using `binding` - though, not as low level as my previous commit, as that has a decent performance impact.

jarohen:
We appear to have gone down the route of making freeze more restrictive (no problem) but, in our case, it might be worth considering that we also have the option of making thaw more _lenient_, if this would reduce Crux's impact on the rest of the system. In Crux (unlike in the general case that Nippy's trying to protect against), I'm reasonably sure we can trust the serialiser to check everything it's serialising, so we may be able to assume that we can thaw anything that's been successfully frozen?

danmason:
Okay - went with that approach, made some higher level `binding` calls around various calls to thaw (rather than wrapping each call to thaw individually).

crux-core/src/crux/kv/tx_log.clj
```diff
@@ -81,25 +82,26 @@
...
-                                                (tx-log (kv/next iterator))))))]
-                               (let [after-tx-id (or (some-> after-tx-id (+ 1)) 0)]
-                                 (->> (tx-log (kv/seek iterator (encode-tx-event-key-to nil {::tx/tx-id after-tx-id})))
-                                      (take batch-size)
-                                      vec))))]
-                   (concat txs
-                           (when (= batch-size (count txs))
-                             (tx-log (::tx/tx-id (last txs))))))))]
-        (cio/->cursor (fn []) (tx-log after-tx-id)))))
+    (binding [nippy/*thaw-serializable-allowlist* #{"*"}]
```
jarohen: This returns a lazy-seq, so the binding won't be in effect when the call to `mem/<-nippy-buffer` actually happens - you'll have to wrap the eager part instead.

We may not need the binding in this case, though - I don't think there's any user-input objects being deserialised here?

crux-core/src/crux/kv/index_store.clj
```diff
@@ -654,12 +656,13 @@
 
   (decode-value [this value-buffer]
     (assert (some? value-buffer))
-    (if (c/can-decode-value-buffer? value-buffer)
-      (c/decode-value-buffer value-buffer)
-      (or (.get temp-hash-cache value-buffer)
-          (let [i @decode-value-iterator-delay]
-            (when (advance-iterator-to-hash-cache-value i value-buffer)
-              (some-> (kv/value i) (mem/<-nippy-buffer)))))))
+    (binding [nippy/*thaw-serializable-allowlist* #{"*"}]
```
jarohen: Most of the calls to this function won't need to do any nippy deserialisation (because a decent amount of the values being decoded now go via the `c/decode-value-buffer` branch) - in these cases, we'll be installing the binding without it being used. We're also not gaining anything from the binding being done at a higher level here because, even if it is used, it's only ever used once.

(It also happens that this is also a low-level function called within tight loops, so the binding at this level would have had quite a performance impact! :sweat_smile:)

jarohen (changes requested):
Thanks @danmason - yeah, I'm happier with applying the change on the thaw side, wdyt?

I'd be tempted to macro this, tbh - it's quite an implementation detail that you've had to put in each location. Something like `io/with-nippy-open-thaw`/`with-nippy-thaw-all`? (eurgh, naming)

crux-core/src/crux/kv/tx_log.clj
danmason: Right you are - only grabbing tx-events, saw `open-tx-log` and didn't realize.

Issue #664, opened 2020-02-19
Add bench run for kafka-lmdb + WatDiv


t-taylor:
Currently there is a configuration commented out. When run, lmdb segafaults (https://pastebin.com/XLwBuZxT). May be related to #280

hraberg:
I think there might be a chance that running LMDB/mmap on Fargate in the /tmp creates some issues. Might be worth trying to use the scratch volume space:
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/fargate-task-storage.html

Locally I cannot reproduce the slowness of running using LMDB on the bench.

jarohen:
It's now running on sorted maps, needs to work for the larger ingest tests

danmason:
Current state of this: LMDB now runs fine through all tests except Watdiv - where the ingest has taken up to **5 hours** on AWS. Going to investigate this further, and see how it runs locally (with more more memory available). 

The previous issues with Segfaults appear to be fixed, as does the bench hitting the MAX_READERS limit (thanks to #1261)

PR #1261, opened 2020-11-13
Increase LMDB max readers

hraberg:
Increase LMDB max readers to 1024 from default 126, and add a config for it.

This seems innocent, here's the comment around it from `mdb.c`:
```
        /**     Number of slots in the reader table.
         *      This value was chosen somewhat arbitrarily. 126 readers plus a
         *      couple mutexes fit exactly into 8KB on my development machine.
         *      Applications should set the table size using #mdb_env_set_maxreaders().
         */
#define DEFAULT_READERS 126
```

Related to #664

### DIFF
--- a/crux-lmdb/src/crux/lmdb.clj
+++ b/crux-lmdb/src/crux/lmdb.clj
@@ -91,6 +91,9 @@
                                flags
                                0664)))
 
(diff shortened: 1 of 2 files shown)


danmason (approved):
Would be interested to experiment with this over on bench 👍

jonpither (approved):

PR #1265, opened 2020-11-20
Returned `indexed-docs` on bus - use directly in lucene

danmason:
Saves roundtrip to get docs back from the document-store - open question on if there's any security concerns doing this on the bus?

### DIFF
--- a/crux-core/src/crux/tx.clj
+++ b/crux-core/src/crux/tx.clj
@@ -34,9 +34,10 @@
 (s/def ::bytes-indexed nat-int?)
 (s/def ::doc-ids (s/coll-of #(instance? crux.codec.Id %) :kind set?))
 (s/def ::eids (s/coll-of c/valid-id? :kind set?))
+(s/def ::docs (s/nilable (s/coll-of map?)))
 
 (defmethod bus/event-spec ::indexing-docs [_] (s/keys :req-un [::doc-ids]))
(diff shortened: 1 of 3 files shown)


jonpither:
Does the spec for ` ::indexed-docs` need to be updated? Otherwise LGTM, I'll let @jarohen comment on the security implications of putting docs on to the bus.

jonpither (approved):
I've approved, but also happy to wait for @jarohen to chip in.

jarohen:
I don't recall any _explicit_ blockers to putting full documents on the bus. Some thoughts, for now, in case this sparks others' ideas:

* security-wise, we don't protect access to the document store, so we're not really making the system any more secure by not putting the documents here - we have to assume that whoever has access to the bus also has access to the document store
* memory-wise, this means that documents may well stay in memory for longer, especially if a listener builds up a backlog. I'm reasonably sure we do but let's make sure each listener has a bounded queue so that memory use doesn't grow indefinitely if a listener locks up.
* there may be a halfway-house here if we were to not include docs in the events, but make sure that the document store has a decent LRU-like cache populated on ingestion, so that any listener that _does_ want access to the documents still requests them of the document store, but the document store likely has the documents available in a hot cache.

jarohen:
discussed in standup, let's ensure the documents are hot in the document store cache :)

PR #1266, opened 2020-11-20
Deref kafka producer futures, ensure exceptions are thrown.

### LINKED ISSUE #1267: Ensure exception in `submit-docs` in kafka are properly thrown

danmason:
Resolves #1267


jonpither (approved):

jarohen (changes requested):
Don't think this preserves the submit-tx-async behaviour? Not necessarily saying that we have to, but let's make it an explicit decision to remove it.

PR #1269, opened 2020-11-23
Ensure `await-tx` actually outputs original tx on timeout, and test timeout exceptions

### LINKED ISSUE #1268: Await-tx not properly which tx it is awaiting within timeout exceptions
When attempting to call `await-tx` on a particular transaction, it doesn't return the tx that is awaiting on.

For example, using the following on an empty:
```
(let [tx (crux/submit-tx node (for [n (range 100)] [:crux.tx/put {:crux.db/id (str "test-" n)}]))]
      (crux/await-tx node tx (Duration/ofNanos 1)))
```

Returns this exception:
```
java.util.concurrent.TimeoutException
   Timed out waiting for: nil, index has: nil
```

danmason:
Fixes #1268

### DIFF
--- a/crux-core/src/crux/node.clj
+++ b/crux-core/src/crux/node.clj
@@ -33,9 +33,8 @@
   (when @closed?
     (throw (IllegalStateException. "Crux node is closed"))))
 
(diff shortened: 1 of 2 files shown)


jonpither (approved):
Subtle.

PR #1282, opened 2020-11-27
Inlining calls to ByteUtils that have Clojure wrappers.

hraberg:
For stylistic reasons we have Clojure wrappers for some calls in `crux.ByteUtils`, like `compareBuffers` and `equalBuffers`
Adding Clojure inline metadata cuts down the call overhead with 10-15%.

Note that this isn't directly visible in end to end benchmarks, as there's a lot of other things going on, and many calls to `compareBuffers` go via `crux.memory/buffer-comparator`. The places where they will be inlined will mainly be in `crux.kv.index-store`.

There are potentially other places where utility functions (Java wrappers or small Clojure snippets) could benefit from simple inline metadata in a similar fashion.

### DIFF
--- a/crux-core/src/crux/memory.clj
+++ b/crux-core/src/crux/memory.clj
@@ -199,6 +199,11 @@
    (some-> b (ByteUtils/hexToBuffer to))))
 
 (defn compare-buffers
+  {:inline (fn [a b & [max-length]]
+             (if max-length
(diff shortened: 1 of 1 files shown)


hraberg:
As a note, I did try to use `memcmp` via JNR but the JNI overhead was too high, and it's about two times slower than the Java-based compare. This trade-off might be different for large buffers where SIMD and similar improvements in the C implementation can kick in.

Issue #1218, opened 2020-11-02
Allow pred-fns access to node components


jonpither:
For example #1094 needs to maintain a dynamic var, as pred-fns don't have access to the `lucene-store` component. 

This issue to make make more things available to the pred-fns, one extreme is the entire Crux node being available.

An alternative approach, to satisfy the lucene scenario, is that the Lucene store could itself be modeled as an index-store, and it may then benefit from being more exposed where it's needed.

PR #1283, opened 2020-11-27
Make Lucene query customisable

jonpither:
Refactoring that allows users to plug-in their own Lucene queries and predicate functions.

TODOs

- [x] Use `dep` instead of `fn` for configuration
- [x] Customize Analyzer https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-analyzers.html
- [x] Allow Vars into the lucene query string? parameterized prepared-statement idea for lucene queries?
- [x] How does query string work with ns keywords?
- [x] Resintate destructuring for `full-text` to not have `A`
- [x] Document multi-field
- [x] Document ALPHA and possible retiring of pred-fns
- [x] Code assert for multi-field that the indexing strategy is in place.

Separate issues / future work:
- [x] Consider having a single pred-fn for full-text and lucene-full-text
- [x] Compare av-docs vs full-docs on large volumes, looking at disk volume
- [x] As above, but think of looking at impact of history
- [ ] Would be nice if would sort #1218 
- [ ] limit (default limit is 1000, this is being applied)
- [ ] Is it possible to allow lucene to catch-up without the node needing to replay?

### DIFF
--- a/crux-core/src/crux/system.clj
+++ b/crux-core/src/crux/system.clj
@@ -276,5 +276,3 @@
                           (instance? TimeUnit t) t
                           (string? t) (TimeUnit/valueOf (str/upper-case t)))))
          #(instance? TimeUnit %)))
-
-(s/def ::fn fn?)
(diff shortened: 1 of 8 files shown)


jonpither:
This is NOT ready for merging until eviction is taken care of for the index-docs alternative.

jonpither:
This PR does not require a index bump. As made clear in the docs, if you add Lucene configuration to a node, you must always clear out the index directories prior to this.

PR #1284, opened 2020-11-30
crux.index refactor

hraberg:
Removes all the mutable value objects used by `crux.index` and replaces the records with deftype with mutable fields (and hence one less pointer reference). Also removes the thunk indirection in `UnaryJoinVirtualIndex` and removes the usage of `mod` from the calculations.

Also refactors `crux.index` a bit to remove `IndexStoreIndex` and replace it with `SeekFnIndex` and `DerefIndex`, the latter tracking the latest returned key, and also being used by `UnaryJoinVirtualIndex` instead of having a separate value type for this.

Fixes issue in `NAryJoinLayeredVirtualIndex` where nth on a seq was used on the indexes, replacing with array.

Replaces the persistent stack of join keys with a single array list + depth index for evaluating constraints.

Also removes the usage of off-heap when reading data out of Rocks, as it's faster to let the JNI layer allocate the right sized byte array directly. Buffers are still used to send data into Rocks.

About 10-15% faster than master.


jarohen:
Thanks @hraberg - particularly pleased to see the back of the thunk code, I never found it particularly easy to follow :)

PR #1285, opened 2020-11-30
Add note distinguishing between using symbols in EDN / Clojure

### LINKED ISSUE #1243: Doc note differentiating between Clojure and EDN
> The topology migration guide was super helpful, thanks for that!
> 
> One thing that confused me was, that the migration guide uses a backtick for module references.
> Using that in crux.edn gave me an error:
> Execution error at crux.system/read-opts (system.clj:26).
> Invalid leading character: `
> 
> I looked at the docs and they use a single quote. What worked was to use the symbol directly. I guess both docs are not targeted at users of the crux.edn file, but this took me a bit to figure out :sweat_smile:
> I think a simple but prominent one-liner will be helpful - I too ran into this very initially (mentally not distinguishing between the sample, in-code config and what goes into some config.edn), but soon figured out.

https://clojurians-log.clojureverse.org/crux/2020-10-30

danmason:
Resolves #1243 

Decided to add a little note about the usage of leading characters directly underneath where we state about being able to use configuration files to start nodes - can see it here: https://opencrux.com/doc-note/reference/configuration.html

### DIFF
--- a/docs/reference/modules/ROOT/pages/azure-blobs.adoc
+++ b/docs/reference/modules/ROOT/pages/azure-blobs.adoc
@@ -33,6 +33,24 @@ pom.xml::
 
 Replace the implementation of the document store with `+crux.azure.blobs/->document-store+`
 
+[tabs]
+====
+JSON::
++
+[source,json]
+----
+{
+  "crux/document-store": {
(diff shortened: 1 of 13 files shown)


jarohen:
splitting out Clojure/EDN tabs instead

PR #1281, opened 2020-11-26
BigDecimal and java.time classes as value types

jarohen:
Adding support for `BigDecimal`, `LocalDate`, `LocalTime`, `LocalDateTime`, `Instant` and `Duration` as value types, allowing them to be used in range filters.

`BigDecimal` implementation adapted from [Orderly](https://github.com/ndimiduk/orderly/blob/master/orderly-core/src/main/java/orderly/BigDecimalRowKey.java)

RFC: I haven't included `BigInteger` here, because in Clojure there's also `clojure.lang.BigInt` to consider (a separate type to `java.math.BigInteger`, neither being a sub-class/super-class of the other) - this is therefore more susceptible than the others to the known issue of Crux values not being comparable between types, but with (I'd wager) significantly less awareness, even within the Clojure community. Could also be argued that supporting `BigDecimal` but not `BigInteger` is a big gotcha, particularly for JVM devs who _really_ don't care about this Clojure nuance. Eurgh.

Overnight RFC: (definitely a separate PR, if at all) - for the time being, could/should we make range filters refuse to compare values from different types? i.e. `10` is neither `<`, `>` nor `=` to `10.0`? May be less counter-intuitive than having (say) `12.0 < 10` because the binary representation of doubles happen to be 'less' than longs.

PR #1288, opened 2020-11-30
Value types for BigInteger and BigInt

jarohen:
N.B. these are different value types - if some documents have BigInts and some BigIntegers, they won't compare properly.
this is consistent with the note in #1281 - we'll need a wider change to make these comparable.

### DIFF
--- a/crux-core/src/crux/codec.clj
+++ b/crux-core/src/crux/codec.clj
@@ -6,7 +6,7 @@
             [crux.io :as cio]
             [crux.memory :as mem]
             [taoensso.nippy :as nippy])
-  (:import [clojure.lang APersistentMap APersistentSet IHashEq Keyword]
+  (:import [clojure.lang APersistentMap APersistentSet BigInt IHashEq Keyword]
            crux.codec.MathCodec
            java.io.Writer
            java.math.BigDecimal
@@ -24,7 +24,7 @@
 ;; Indexes
 
 ;; NOTE: Must be updated when existing indexes change structure.
-(def index-version 15)
+(def index-version 16)
 (def ^:const index-version-size Long/BYTES)
 
 (def ^:const index-id-size Byte/BYTES)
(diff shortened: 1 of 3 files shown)


crux-core/src/crux/codec.clj
```diff
@@ -164,6 +166,26 @@
   (-freeze-without-meta! [this out]
     (freeze-map out this)))
 
+(defn- biginteger->buffer [^BigInteger x type-id ^MutableDirectBuffer to]
+  (let [signum (.signum x)]
+    (-> to
+        (doto (.putByte 0 type-id))
```
hraberg: Was considering commenting on this stylistic repeated `doto` I noticed - and should've spotted last PR - I prefer a single `doto` for all the puts (see most older encoders), the combination of `->` and multiple `doto` essentially does the same thing.

hraberg (changes requested):
LGTM! But taking this opportunity to request some stylistic changes to this namespace. See line comment about `doto`. This applies broader, not just to this latest change.

crux-core/src/crux/codec.clj
jarohen: Sure - in this case it works, happy to change - in the others, it's combined with `mem/limit-buffer`. I'll have a play around with it, it might be that something like the following is easier to read:

```clojure
(-> (doto to
      ...)
    (mem/limit-buffer ...))
```

Failing that, we usually know the size of the buffer before putting to it, so this might also work:

```clojure
(doto (mem/limit-buffer to)
  ...)
```

We could also separate the side effecting part into a separate form:

```clojure
(doto to
  ...)

(-> to (mem/limit-buffer ...))
```

Any preferences?

hraberg: A glance at the ns feels like most places can be replaced by the first form, my general preference is that `doto` itself shouldn't sit repeated within the threading, a single one will do (with the puts nested beneath it), as it's also a kind of threading macro, but not of the return value.

jarohen: done @hraberg - cheers :)

hraberg (approved):
Excellent, thanks.

PR #1291, opened 2020-12-02
Predicate allow list

### LINKED ISSUE #979: Users can configure an allow-list of predicate functions available during queries
Users should be able to control which core/built-in predicates and "custom" predicates (reachable via fully qualified names) are accessible during queries. This will help to secure the execution of queries that are constructed by untrusted clients. This could be achieved via a new topology option with a set of names for the allowed functions.

* If a user doesn't specify `:fn-allow-list` as an arg to the `:crux/query-engine` component, all functions are allowed
* `:fn-allow-list` should contain a set of symbols
  * unqualified symbols should be assumed to be namespaces, all symbols from those namespaces should be allowed
  * qualified symbols should be assumed to be functions
  * strings should be conformed to symbols, for Java folks
* should through a suitable `crux.error` error (probably IAE?) if a disallowed function is used in a query
* queries should be checked at query plan time (so that the check can be cached - we shouldn't need to re-check them at query execution)
* might be tempting, but `clojure.core` should _not_ be a blanket exception to the rule - there're functions like `spit` and `slurp` which access more than they potentially should. (We _could_ consider a built-in allow-list, or 'allow clojure.core but block some functions', if that's easier, but not a blanket allow)

danmason:
Resolves #979 

Added documentation, splitting out configuration into its' own section, see https://opencrux.com/query-config/reference/query-configuration.html


crux-test/test/crux/query_fn_allowlist_test.clj
```diff
@@ -0,0 +1,133 @@
...
+  (= 0 (mod n 3)))
+
+(defn multiple-of-seven? [n]
+  (= 0 (mod n 7)))
+
+(t/deftest test-default-allow-list
+  (let [with-fixtures (t/join-fixtures
+                       [(fix/with-opts {:crux/query-engine {:fn-allow-list '[]}})
+                        fix/with-node])]
+    (fn []
```
jarohen: missing a `with-fixtures` call here?

crux-core/src/crux/query.clj
```diff
@@ -1259,6 +1259,19 @@
            [sub-clause]))
        (reduce into [])))
 
+(defn- build-pred-fns [clauses {:keys [allowed-ns allowed-fns] :as fn-allow-list}]
+  (->> (for [[type clause :as sub-clause] clauses]
+         (let [pred-var (get-in clause [:pred :pred-fn])]
+           (if (and (= :pred type) (var? pred-var))
+             (let []
```
jarohen: `let []` -> `do`

crux-core/resources/query-allowlist.edn
```diff
@@ -0,0 +1,79 @@
+#{clojure.core/false? clojure.core/-' clojure.core/ratio?
```
jarohen: doubt we're going to read these often, but could they be one per line and sorted alphabetically?

(emacs will `M-x sort-lines`, remember to temporarily put the `#{` and `}` on their own lines, lest they be sorted too)

docs/reference/modules/ROOT/pages/query-configuration.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+* `entity-cache-size` (int, default 32768): query entity cache size.
+* `query-timeout` (int, default 30000): query timeout in milliseconds.
+* `batch-size` (int, default 100): batch size of results.
+* `fn-allow-list` (xref:#fn-allowlist[`Predicate Allowlist`], default nil): list of allowed namespaces/functions in predicate functions.
+
+[#fn-allowlist]
+== Predicate Allowlist
+
+The query engine can be provided with an allowlist to configure which predicate functions can be ran inside of queries. By default, any `clojure.core` or fully quantified function name is allowed within a query, though there are many reasons to make this restrictive.
```
jarohen: This doesn't read quite right to me, think it's mainly the use of the passive voice which makes it a little wordy. I'm aiming for the voice of the Reference section to read a lot more like a specification rather than a tutorial - authoritative, terse, dry, boring almost. Something like this?

```
By default, users can invoke any predicate function in their queries.
To restrict what functions a user can invoke, you can specify a `:fn-allow-list` of safe functions/namespaces.
We include a set of pure functions from `clojure.core`, for convenience.
```

I'm also assured that the adoc idiom is one sentence per line :)

docs/reference/modules/ROOT/pages/query-configuration.adoc
```diff
@@ -0,0 +1,86 @@
+= Query Engine Configuration
+
+There are a number of different defaults that can be overriden within the Crux query engine.
```
jarohen: *overridden

worth getting the `flyspell` package enabled in your adoc-mode :)

docs/reference/modules/ROOT/pages/query-configuration.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+== Overriding Defaults
```
jarohen: the `==` and the subsequent sentence aren't adding anything, IMO

docs/reference/modules/ROOT/pages/query-configuration.adoc
```diff
@@ -0,0 +1,86 @@
...
+The query engine can be provided with an allowlist to configure which predicate functions can be ran inside of queries. By default, any `clojure.core` or fully quantified function name is allowed within a query, though there are many reasons to make this restrictive.
+
+Allowlists can be a list of the following:
+
+* Unqualified symbols/strings, for example, `clojure.set`. These are assumed to be https://clojure.org/reference/namespaces[*namespaces*], and all functions within that namespace are allowed.
+* Fully qualified symbols/strings, for example, `clojure.string/capitalize`. These are assumed to be a permitted function.
+
+NOTE: When specifying an allowlist, we add in a set of safe `clojure.core` functions that will still be available to the query.
+
+=== Overriding
```
jarohen: Could this be included in the above example?

jarohen (changes requested):
Thanks @danmason - code change is good, few suggestions for the docs. Cheers!

PR #1294, opened 2020-12-07
Rocks block cache config

hraberg:
Making it possible to configure the size and share the block cache as its own dependency between Rocks instances.

See `dev.clj` for an example on how to use it. 

The default mimics the previous behaviour, which is to create a 8Mb block cache per each Rocks instance.
But this is really way to low, the recommended value is 1/3 of the memory budget:
https://github.com/facebook/rocksdb/wiki/Setup-Options-and-Basic-Tuning#block-cache-size

This hook also allows for creating a static, per JVM cache and do nothing on close, allowing to share it between nodes, but code for this is not included.

Somewhat related to this is that there's a old argument that is an Option instance, allowing users to fine tune their Rocks config. This should really be another dependency itself, but this PR doesn't try to address this, and only aims to fix the common case which is the need to tune the cache size properly.

### DIFF
--- a/crux-rocksdb/src/crux/rocksdb.clj
+++ b/crux-rocksdb/src/crux/rocksdb.clj
@@ -14,7 +14,7 @@
            java.nio.ByteBuffer
            (java.nio.file Files Path)
            java.nio.file.attribute.FileAttribute
-           (org.rocksdb Checkpoint CompressionType FlushOptions LRUCache
+           (org.rocksdb BlockBasedTableConfig Checkpoint CompressionType FlushOptions LRUCache
                         Options ReadOptions RocksDB RocksIterator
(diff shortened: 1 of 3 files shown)


hraberg:
One thing that has been discussed several times is to add support for the actual map and string versions of the options and populate them accordingly. The config is split into a db options and column family options, see `DBOptions#getDBOptionsFromProps` and `ColumnFamilyOptions#getColumnFamilyOptionsFromProps`

jarohen (changes requested):
Code all LGTM - can we get this included in the RocksDB documentation?

danmason:
Docs note here: https://opencrux.com/rocksdb-cache/reference/rocksdb.html#blocks-cache

PR #1308, opened 2020-12-09
Memory Allocators

hraberg:
Introduces an `Allocator` interface, and replaces the concept of unpooled buffers with a `DirectRootAllocator`. Each thread has a thread local `BumpAllocator` which works similar to how `allocate-buffer` used to work.

This PR initially aimed to introduce region allocators using a dynamic var, but this was backed out, see below, I've kept the discussion as it's still relevant to some extent going forward:

_Note: due to a thread being able to open many cursors the approach as implemented in this PR breaks down. Idea is to rework this, moving to smaller regions lower down, and more explicit memory management that can deal with both multiple cursors per thread, and laziness._

Introduces an `Allocator` interface and removes the concept of pooled buffers.

Instead there's `crux.memory/root-allocator` and a dynamic var `crux.memory/*allocator*`, which by default points to the former. All operations, like `crux.memory/allocate-buffer`, uses this dynamic var. `copy-to-unpooled-buffer` is replaced by `copy-buffer-to-root-allocator`.

The dynamic var is then rebound during queries to a scoped `BumpAllocator` wrapping a `DirectRegionAllocator`. The chunk size default to the page size. This also proactively closes buffers when the region is closed, to be used in a `with-open` call.

The following implementations exist:

1. `DirectRootAllocator`, wraps `ByteBuffer/allocateDirect`, without any tracking of buffers, an instance of this sits at `crux.memory/root-allocator`.
2. `DirectionRegionAllocator` wraps `ByteBuffer/allocateDirect`, and adds allocation book keeping. Frees all buffers on close.
3. `BumpAllocator` delegates to an underlying allocator and adds the chunking logic that Crux used to use in `crux.memory/allocate-buffer`.
4. `QuotaAllocator` just forwards all calls but also throws an exception if a memory quota is exceeded.

Allocators compose, but cannot be reused. They will attempt to close any wrapped allocator.

Currently about 5% faster on TPC-H that master. Exactly why is not obvious. One reason might be that `DirectRegionAllocator` actually proactively frees buffers by calling their cleaner (via `org.agrona.BufferUtil`). Another reason could be that more places now use the thread local buffers, as the scope is clearer and hence safer.

Queries and history have their own regions now. In theory other places, like caches for example, could have their own (long lived) allocators as well.

The `:query-memory-quota` argument to the `QueryEngine` component is set to 32Mb by default. This is the ceiling for in-use off heap buffers during query processing, the realised tuples might be much larger than this. If this is a good default or not is hard to tell. It does not include buffers sitting in the various caches that are shared between queries. These are currently always allocated by the root allocator. Buffers returned from the KV stores are owned by the underlying store, and not our allocators atm. I did try copy them into the current allocator, but that hit the quota limit quickly.

I also spiked allocators using Jemalloc and Netty's buffer pool on a different branch. And I've also spiked JNR calling the system's malloc and free. Pooled Netty buffers might be worth it if one redesigned the entire system around them, they're reference counted by design. Just plugging them in behind our current abstractions isn't a very good fit, tests shows the `BumpAllocator` is faster than Netty's pool, but is less flexible.

Similar to Netty's pool, a potential next step could likely to be to implement a more advanced malloc implementation in Clojure/Java and avoid JNI except when allocating large chunks from the system. The bump allocator we have is both easy and fast, but far from optimal. Netty does contain a port Jemalloc to Java: https://netty.io/wiki/new-and-noteworthy-in-4.0.html#pooled-buffers See also https://netty.io/wiki/using-as-a-generic-library.html This would also align towards Arrow, which also depends on `io.netty/netty-buffer`.

It's worth noting that JDK14 has a `MemorySegment` API to support deterministic deallocation of regions:
https://docs.oracle.com/en/java/javase/14/docs/api/jdk.incubator.foreign/jdk/incubator/foreign/package-summary.html

### DIFF
--- a/crux-bench/src/crux/bench/ts_devices.clj
+++ b/crux-bench/src/crux/bench/ts_devices.clj
@@ -240,7 +240,7 @@
                                        (first battery-levels)
                                        (last battery-levels)]))))
                        (finally
-                         (run! cio/try-close (reverse histories)))))
+                         (run! cio/try-close histories))))
 
             success? (= [[#inst "2016-11-15T12:00:00.000-00:00" 20.0 99.0]
                          [#inst "2016-11-15T13:00:00.000-00:00" 13.0 100.0]
--- a/crux-bench/src/crux/bench/ts_devices.clj
+++ b/crux-bench/src/crux/bench/ts_devices.clj
@@ -240,7 +240,7 @@
                                        (first battery-levels)
                                        (last battery-levels)]))))
                        (finally
-                         (run! cio/try-close histories))))
+                         (run! cio/try-close (reverse histories)))))
 
             success? (= [[#inst "2016-11-15T12:00:00.000-00:00" 20.0 99.0]
                          [#inst "2016-11-15T13:00:00.000-00:00" 13.0 100.0]
--- a/crux-bench/src/crux/bench/ts_weather.clj
+++ b/crux-bench/src/crux/bench/ts_weather.clj
@@ -316,7 +316,7 @@
                                        (trunc (first temperatures) 2)
                                        (trunc (last temperatures) 2)]))))
                        (finally
-                         (run! cio/try-close (reverse histories)))))
+                         (run! cio/try-close xhistories))))
 
             success? (= result
                         [[#inst "2016-11-15T12:00:00.000-00:00" 73.45 68.0 79.2]
--- a/crux-bench/src/crux/bench/ts_weather.clj
+++ b/crux-bench/src/crux/bench/ts_weather.clj
@@ -316,7 +316,7 @@
                                        (trunc (first temperatures) 2)
                                        (trunc (last temperatures) 2)]))))
                        (finally
-                         (run! cio/try-close xhistories))))
+                         (run! cio/try-close histories))))
 
             success? (= result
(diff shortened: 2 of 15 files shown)


hraberg:
The build was broken for a long time despite passing locally. This is why I added the `DirectRootAllocator` which works more similar to master, as it doesn't add any extra tracking layer, which was assumed to eventually get closed in a region.

hraberg:
Backed out regions for transaction processing for now as this failed an intermittent test, my guess is for timing reasons though.

hraberg:
Something is off when running bench, need to figure out why.

hraberg:
This PR has a fundamental flaw in that it assumes that one can tie cursors to threads exclusively, but nothing stops a thread to open more than one cursor. Accidentally this will share the thread bound region, but there's no guarantee about them being closed in the right order (as the first cursor would restore the parent allocator).

jarohen:
Cheers for the investigation @hraberg, sorry to hear that it didn't bear fruit - reckon it's worth closing this PR off for the time being?

hraberg:
Maybe not just yet, as mentioned, I've turned it into more of a refactoring of what's there, salvaging some changes.

PR #1309, opened 2020-12-09
Ingest performance

### LINKED ISSUE #1252: Avoid encoding KVs twice per document during ingest
Currently when ingesting documents we encode all the KV pairs at least twice, for both the Forked Kv Store and the Persistent Kv Store.

jarohen:
N.B. out of scope: moving the 'fork' functionality from the tx ingester to the index store - couple of gnarly evict bugs still to iron out, and haven't yet gotten to the bottom of the query (!) performance regression. Will bank this one, and then come back with a subsequent moving-the-fork PR.

Main changes:
* In-memory index-store uses a `j.u.TreeMap`, rather than `c.l.PersistentTreeMap`. The 'snapshots' yielded from this, strictly speaking, aren't snapshots - they're mutable, but they're only ever read from one thread, and that thread is _either_ reading or writing, never both at once - i.e. it's immutable through the sections that matter.
* We were calling `unindex-eids` even when there weren't any eids to unindex - this didn't have any `seq` checks, and opened/closed snapshots regardless. `(when (seq eids) ...)` cut about 10% of my microbench.
* I've replaced a few usages of `merge` with `into` - the former does `(reduce conj ...)`, which doesn't use transient data structures.
* I've not removed the check where we read the KV store to decide whether to index the doc, one for another PR.
* (not performance, but) have included more of the tx-ingestion in the try-catches that then report the error back to anyone awaiting - we should see fewer errors slipping through the net, and fewer awaits hanging indefinitely

TODO/RFC:
* [x] I've split out `IndexSnapshotFactory` as a separate protocol just containing `open-index-snapshot`, so that `IndexStoreTx` can also implement it. Should we migrate usages of `open-nested-index-snapshot` to this as well, or is the distinction helpful? I suspect the distinction's helpful, as the nested one doesn't open up a snapshot on the underlying KV store?

Microbench:
* Attach profiler
* Submit 25 transactions with 1k simple docs, `crux/sync`
* Restart node, time `crux/sync`

Timings (YourKit, 'tracing' - about 10-15x slower than just running at the REPL, but more interested in the relative speedup): before 25s, after 17s.

Bench (in progress, will update):
* TODO

### DIFF
--- a/crux-core/src/crux/db.clj
+++ b/crux-core/src/crux/db.clj
@@ -21,11 +21,17 @@
   (index-entity-txs [this tx entity-txs])
   (mark-tx-as-failed [this tx])
   (store-index-meta [this k v])
-  (read-index-meta [this k] [this k not-found])
   (latest-completed-tx [this])
   (tx-failed? [this tx-id])
   (open-index-snapshot ^java.io.Closeable [this]))
 ;; end::IndexStore[]
+
+(defprotocol IndexMeta
+  (-read-index-meta [this k not-found]))
+
+(defn read-index-meta
+  ([index-meta k] (-read-index-meta index-meta k nil))
+  ([index-meta k not-found] (-read-index-meta index-meta k not-found)))
 
 ;; tag::IndexSnapshot[]
 (defprotocol IndexSnapshot
--- a/crux-core/src/crux/fork.clj
+++ b/crux-core/src/crux/fork.clj
@@ -138,8 +138,7 @@
     (swap! !evicted-eids set/union (set eids))
     (db/unindex-eids transient-index-store eids)
 
-    (with-open [persistent-index-snapshot (db/open-index-snapshot persistent-index-store)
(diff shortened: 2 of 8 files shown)


jonpither (approved):

crux-core/src/crux/db.clj
```diff
@@ -59,5 +72,18 @@
   (abort [in-flight-tx]))
 
 (defprotocol DocumentStore
-  (submit-docs [this id-and-docs])
-  (fetch-docs [this ids]))
+  (submit-docs [document-store id-and-docs])
+  (fetch-docs [document-store ids]))
+
+(defn fetch-docs-sync [document-store ids]
```
jarohen: -> `crux.document-store`

hraberg (approved):

crux-core/src/crux/db.clj
jarohen: no longer in this PR

PR #1314, opened 2020-12-10
Transaction Info return in Java friendly format.

### LINKED ISSUE #1213: Java API shouldn't use `Map<Keyword, ?>` to represent transaction instants
These are pretty gnarly to use from anything that isn't Clojure - let's consider migrating to something similar to `HistoryOptions`, and using this throughout the Java API wherever we represent a transaction instant (tx-time/tx-id)

danmason:
(DRAFT/WIP)
Resolves #1213 

TODO:
- Sort out behaviour of `db` - potentially add artist for transaction instants?
- Rename `TransactionInstant`.
- Properly document, edits to examples.

### DIFF
--- a/crux-core/src/crux/api.clj
+++ b/crux-core/src/crux/api.clj
@@ -8,6 +8,7 @@
             [crux.ingest-client :as ic]
             [crux.io :as cio]
             [crux.query-state :as qs]
+            [crux.transaction-instant :as cti]
             [crux.system :as sys])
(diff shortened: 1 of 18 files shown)


jarohen:
Cheers @danmason - good PR, but unfortunately doesn't apply now that we've moved towards having the Clojure and Java APIs as two separate layers - will close.

Worth @AlistairONeill taking a look at this and porting it over, maybe even as a distinct first increment of the new Java strongly typed API.

PR #1319, opened 2020-12-11
Add lucene test, check query results not limited to 1000

### LINKED ISSUE #1336: Lucene predicate function limited to 1000 results
This is before bitemp filtering/resolution. This may lead to some unpredictable query behavour.

markthequark:
(edit, @jarohen)
Added fix. While the Lucene search lazily pages through, we later eagerly evaluate the results - we'll need to implement the standard query engine `Index` protocol to ensure this results are lazy end-to-end.


jarohen (changes requested):
This should probably use Lucene :wink:

jarohen (approved):

jonpither (approved):

Issue #1310, opened 2020-12-09
Atomic evicts from KV store


jarohen:
Currently, on commit, we have two writes to the KV store: one from `kv/store` and, if there're evicts, one from `kv/delete`.

* We should be able to call `kv/store` with both KVs to insert and delete, and remove `kv/delete`. Maybe a better name that encompasses both `store` and `delete`?
* Rocks's `WriteBatch`, at least, supports both `put` and `remove` in the same batch, atomically.
* If another KV store _doesn't_ support this, we can still fall back to the current behaviour
* Maybe we distinguish between them by having `nil` as the KV value? (if the KV doesn't have a value, we already pass `mem/empty-buffer`)

PR #1323, opened 2020-12-15
migrate kv/delete into kv/store

jarohen:
Rather than two separate calls, `kv/store` and `kv/delete`, this PR combines them s.t. previous calls to `kv/delete` can be replaced with calls to `kv/store` with nil values. This means that we can combine together additions and removals of KVs into one transaction on the underlying KV store - previously, we'd have one call to `store` and one to `delete`, which would open two transactions on the store.

Looks like we might need a wider refactor to get the value from this change - the `kv/delete` call in `crux.kv.index-store` is a fair way away from the `kv/store` call, will take the refactoring in my `move-the-fork` branch to actually get KV puts/deletes in the same write transaction.

#1310

### DIFF
--- a/crux-core/src/crux/mem_kv.clj
+++ b/crux-core/src/crux/mem_kv.clj
@@ -75,13 +75,14 @@
     (->MemKvSnapshot @!db))
 
   (store [_ kvs]
-    (swap! !db into (for [[k v] kvs]
-                     (MapEntry/create (mem/copy-to-unpooled-buffer (mem/as-buffer k))
-                                      (mem/copy-to-unpooled-buffer (mem/as-buffer v)))))
-    nil)
-
-  (delete [_ ks]
-    (swap! !db #(apply dissoc % (map mem/->off-heap ks)))
+    (swap! !db (fn [db]
+                 (reduce (fn [db [k v]]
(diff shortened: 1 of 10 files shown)


jonpither (approved):

crux-core/src/crux/mem_kv.clj
```diff
@@ -75,13 +75,14 @@
...
-    (swap! !db into (for [[k v] kvs]
-                     (MapEntry/create (mem/copy-to-unpooled-buffer (mem/as-buffer k))
-                                      (mem/copy-to-unpooled-buffer (mem/as-buffer v)))))
-    nil)
-
-  (delete [_ ks]
-    (swap! !db #(apply dissoc % (map mem/->off-heap ks)))
+    (swap! !db (fn [db]
+                 (reduce (fn [db [k v]]
+                           (let [k-buf (mem/copy-to-unpooled-buffer (mem/as-buffer k))]
```
hraberg: Mini thing, in theory you don't need to copy this buffer if you only use it for the delete case.

hraberg (approved):
LGTM. Nice one, solves long standing annoyance for atomic writes.

PR #1324, opened 2020-12-15
Removing the 'pre-commit-fn' abstraction - this is now executed directly

jarohen:
IIUC, we started off with this `pre-commit-fn` abstraction because we used to collect together all of the tx-events in a transaction - then, at the end of the transaction, we'd check all of the pre-commit-fns, before storing all the changed KVs. Now that we process transactions one tx-event at a time (partly so that later tx-events can see earlier tx-events, #529, and partly for transaction functions) we evaluate each pre-commit-fn shortly after it's created anyway, before the next tx-event.

I've refactored the code in this namespace s.t. the individual `index-tx-event` `defmethod`s evaluate what was previously their `pre-commit-fn`, and return `:abort? true` if their precondition fails.

### DIFF
--- a/crux-core/src/crux/tx.clj
+++ b/crux-core/src/crux/tx.clj
@@ -111,51 +111,51 @@
 (defmethod index-tx-event :crux.tx/match [[op k v at-valid-time :as match-op]
                                           {::keys [tx-time tx-id], :keys [crux.db/valid-time], :as tx}
                                           {:keys [index-snapshot] :as tx-ingester}]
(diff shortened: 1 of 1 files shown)


jarohen:
@hraberg, @jonpither have I understood the historical context correctly here?

jonpither (approved):

hraberg (approved):
The pre/post commit functions were a bit similar to the current InflightTx abstraction, they but as the ingest was a bit simpler and there were no forks one could roll back, they tried to split the work into safe chunks depending on what happened. Most likely this is redundant today.

PR #1325, opened 2020-12-16
fetch-docs waits for documents to be present before returning

jarohen:
related to #1178, paired with @refset 

Rather than polling waiting for eventually consistent doc stores to become consistent during transaction ingest, we now wait when the documents are actually required (for `entity`, `eql/project`, etc)

Existing test still applies here, it's at https://github.com/jarohen/crux/blob/fetch-docs-sync-read/crux-core/test/crux/eventually_consistent_doc_store_test.clj

NB - had to update the semantics of the JDBC doc store here. All the other doc stores return the tombstone of evicted docs, JDBC was filtering them out (returning nil). Worked fine before, but now we interpret the absence of a document as 'the doc store hasn't got the doc yet' (i.e. it's eventually consistent), and block waiting for it.

### DIFF
--- a/crux-core/src/crux/db.clj
+++ b/crux-core/src/crux/db.clj
@@ -1,4 +1,5 @@
-(ns ^:no-doc crux.db)
+(ns ^:no-doc crux.db
+  (:require [clojure.set :as set]))
 
 ;; tag::Index[]
 (defprotocol Index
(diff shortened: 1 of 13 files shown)


crux-core/src/crux/db.clj
```diff
@@ -66,4 +67,17 @@
 
 (defprotocol DocumentStore
   (submit-docs [this id-and-docs])
-  (fetch-docs [this ids]))
+  (-fetch-docs [this ids]))
+
+(defn fetch-docs [document-store ids]
+  (let [ids (set ids)]
+    (loop [indexed {}]
```
hraberg: Mini nitpick, the name `indexed` here doesn't feel clear, what is indexed here? Is this more simply `docs` or something similar? Assume this might been moved from a context where that name made more intuitive sense.

hraberg (approved):

crux-core/src/crux/db.clj
jarohen: Yep - it made sense in its original context but now doesn't, will change :)

PR #1326, opened 2020-12-16
Remove a second lein sub install

mattford63:
This doesn't look necessary - perhaps left in by accident?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,6 +76,7 @@ jobs:
 
       - run:
           command: |
+            rm -rf ~/.m2/juxt/crux-*
             export LEIN_JVM_OPTS=-Xmx2G
             lein sub install
 
@@ -97,7 +98,6 @@ jobs:
       - run: lein do sub test, check
       - store_test_results:
           path: "crux-test/target/test-results"
-      - run: rm -rf ~/.m2/juxt/crux-*
       - save_cache:
           paths:
             - ~/.m2
@@ -119,11 +119,6 @@ jobs:
               - v1-dataset-{{ checksum ".circleci-bench-dataset-cache-key" }}
               # fallback to using the latest cache if no exact match is found
               - v1-dataset-
-
-        - run:
-            command: |
-              export LEIN_JVM_OPTS=-Xmx2G
-              lein sub install
 
         - run: sh ./crux-bench/bin/download-dataset.sh
         - run: lein sub -s crux-bench uberjar
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,6 +86,12 @@ jobs:
           root: ~/
           paths: ./
 
+      - save_cache:
+          paths:
+            - ~/.m2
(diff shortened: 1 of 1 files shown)


danmason:
I suspect it likely is mistakenly there - may have added the `attach-workspace` at a later point (I think we may have originally ran a build+test job and `deploy-commit` in parallel). 

So long as this runs through and deploys properly, SGTM :)

mattford63:
It runs and deploys, but looking at the test step there seems to be a `rm -rf ~/.m2/juxt/crux-*`

Which is making me a bit nervous.  Also given the test and deploy-commit could run in parallel it seems there might be a bit of a race condition here.

It would perhaps make more sense if deploy-commit depended on test - after all why would we try and deploy a failed test?

So I'm thinking add the dependency and put these lines back.

jarohen:
> after all why would we try and deploy a failed test?

This is for bench testing - if we're spiking a proof-of-concept performance/memory change and want quick feedback on whether it's worth pursuing further, we might not fix all the tests for all the modules before wanting to run a benchmark.

Arguably `deploy-commit` could use a better name here, given this context - something that reflects that it's deploying the bench docker image, rather than official releasable Crux artifacts?

mattford63:
@jarohen ah okay, but there's the still the race condition.  Probably not hit as the test seems to take a lot longer than the deploy-commit.

mattford63:
I can't think of a quick fix around this at the moment.  Considered:
- commit message filtering
- PR commenting 
- CLI based submission
All a bit ugly, not officially supported and/or not quick.  Maybe we live with the race condition for now?

jarohen:
Right, yep. The reason for the `rm -rf` is so that we don't include any Crux artifacts as part of the build cache. Could this (and the cache save) run as a separate job at the end of the build, depending on both deploy-commit and test? (@danmason is more familiar with what's possible in CircleCI than I am :) )

mattford63:
Maybe, I'll have a look into it :-)  Might be a little awkward if the tests fail as we still want the cache clean up to work.

mattford63:
That doesn't seem an option either.

https://circleci.canny.io/cloud-feature-requests/p/option-to-allow-failures-in-fan-inout-workflow

Perhaps instead save the files to the cache but delete  artifacts prior to build....I'll see what that looks like.

jarohen:
as discussed, question about the impact of including Crux artifacts in the CircleCI cache, even if we then go on to remove them as the first step in the build process

mattford63:
My assumptions are that we save a cache to avoid fetching all of .m2 again.  As we provide the cache key the cache is independent of the contents.  The cache is dirtied with crux artifacts but as long as we always remove the crux artifacts before any install step we ensure the cache is cleaned and the build is fresh.  

Helping this is the other change, there's now only one place that an install step is run (the build job).  This stops the race and gives back a couple of minutes.

Alternately (still as part of the build job) we could temporarily move the crux jars out of .m2 before creating the cache and then move them back immediately after.  This ensures a clean cache and precludes the need for a removal.

PR #1331, opened 2020-12-17
Region as Macro

hraberg:
A different take on regions, using a macro and optional producer thread. Experimental. Related to #1308 but doesn't change `crux.memory` as much.

Introduces dynamic vars to control the existing bump allocator, and a `crux.memory/with-region` macro that starts with small chunks and grows them larger. The region frees all allocated chunks on exit.

This PR as a side effect introduces a few other changes to how `crux.index/layered-idx->seq` works, all buffers live within the region, and the tuples are realised within the scope of a region, and the lazy result never sees these buffers and only contains normal tuples.

Regions are used for both queries and transactions. The history API is left untouched.

The main questions here are if this is worth it, especially the extra complexity of the producer thread.
Want to compare memory and CPU usage with master. Initial impression is quite good.

I further plan to change the producer/consumer setup so more tuples are realised in a batch before they are passed over to the consumer. As one now has controlled scope for the regions thanks to the producer/consumer setup, one could more easily move this part to use Netty reference counted buffers and their Jemalloc Java implementation as a further step.

That said, this is an experimental PR, and might not be merged. Keeping it as DRAFT for now.

### DIFF
--- a/crux-core/src/crux/codec.clj
+++ b/crux-core/src/crux/codec.clj
@@ -366,7 +366,7 @@
        (ExpandableDirectByteBuffer. 32)))))
 
 (defn ->value-buffer ^org.agrona.DirectBuffer [x]
-  (mem/copy-buffer (value->buffer x (.get value-buffer-tl))))
+  (mem/copy-to-unpooled-buffer (value->buffer x (.get value-buffer-tl))))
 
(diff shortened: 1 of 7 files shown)


hraberg:
Closing this for now, there are some interesting ideas here, but it's not working good enough to warrant the extra complexity. Bench is also currently not finishing, but haven't figured out why - it doesn't end with out of memory, so might be some producer/consumer issue that gets stuck.

PR #1332, opened 2020-12-18
`QueryState` uses the same pattern as `HistoryOptions` #1327

### LINKED ISSUE #1327: `QueryState` uses the same pattern as `HistoryOptions`
* `HistoryOptions` is a Java interface backed by a Clojure record (in `crux.history-options`)
* Functions in the Clojure `PCruxNode` protocol accept a map that they then convert into a `HistoryOptions` record when they pass it to the Java API
* May be a slight adaptation needed for QueryState - HistoryOptions is only ever passed to Crux, QueryState is only ever returned from the API.

AlistairONeill:
Changed QueryState to be an interface and we now pass the relevant records through
active-queries, recent-queries and slowest-queries rather than converting back to a map


crux-core/src/crux/api/ICruxAPI.java
```diff
@@ -156,8 +156,8 @@
      */
     @SuppressWarnings("unchecked")
     public static final Map<Keyword, ?> TX_INDEXED_EVENT_OPTS = (Map<Keyword, Object>) PersistentArrayMap.EMPTY
-        .assoc(Keyword.intern("crux/event-type"), Keyword.intern("crux/tx-indexed"))
-        .assoc(Keyword.intern("with-tx-ops?"), true);
+            .assoc(Keyword.intern("crux/event-type"), Keyword.intern("crux/tx-indexed"))
```
jarohen: Could you exclude this indentation change from the PR if it's not related?

crux-core/src/crux/query_state.clj
```diff
@@ -1,28 +1,58 @@
...
+  (getErrorMessage [this] type))
+
+(defrecord QueryState [query-id started-at finished-at status query error]
+           IQueryState
+           (getQueryId [this] query-id)
+           (getStartedAt [this] started-at)
+           (getFinishedAt [this] finished-at)
+           (getStatus [this] (->query-status status))
+           (getQuery [this] query)
+           (getError [this] error))
```
jarohen: `defrecord` is two-space indented, see L11

```suggestion
(defrecord QueryState [query-id started-at finished-at status query error]
  IQueryState
  (getQueryId [this] query-id)
  (getStartedAt [this] started-at)
  (getFinishedAt [this] finished-at)
  (getStatus [this] (->query-status status))
  (getQuery [this] query)
  (getError [this] error))
```

crux-core/src/crux/query_state.clj
```diff
@@ -1,28 +1,58 @@
...
+(defn ->query-status [status]
+  (case status
+    :failed IQueryState$QueryStatus/FAILED
+    :completed IQueryState$QueryStatus/COMPLETED
+    :in-progress IQueryState$QueryStatus/IN_PROGRESS))
+
+(defrecord QueryState$QueryError [type message]
+  IQueryState$IQueryError
+  (getErrorClass [this] type)
+  (getErrorMessage [this] type))
```
jarohen: copy pasta? ;)
```suggestion
  (getErrorMessage [this] message))
```

crux-core/src/crux/query_state.clj
```diff
@@ -1,28 +1,58 @@
...
+  (:import (crux.api IQueryState IQueryState$IQueryError IQueryState$QueryStatus)
+           (java.io Writer)))
+
+(defn ->query-status [status]
+  (case status
+    :failed IQueryState$QueryStatus/FAILED
+    :completed IQueryState$QueryStatus/COMPLETED
+    :in-progress IQueryState$QueryStatus/IN_PROGRESS))
+
+(defrecord QueryState$QueryError [type message]
```
jarohen: I've not seen `$` in Clojure record names before, didn't even know it would be valid! I'd stick with just `QueryError`, the namespace makes it sufficiently unambiguous. (usages too)

```suggestion
(defrecord QueryError [type message]
```

crux-core/src/crux/query_state.clj
```diff
@@ -1,28 +1,58 @@
...
-             "COMPLETED" :completed
-             "IN_PROGRESS" :in-progress)
-   :query-id (.queryId query-state)
-   :query (.query query-state)
-   :started-at (.startedAt query-state)
-   :finished-at (.finishedAt query-state)
-   :error (when-let [error (.error query-state)]
-            {:type (.errorClass error)
-             :message (.errorMessage error)})})
+      {:status      (case (str (.getStatus query-state))
```
jarohen: indentation here:
* `defn`/`case`/`when-let` are two-space
* we don't line up KVs of a map in Crux because it adds noise to diffs (the wider Clojure community is split) - just one space between K and V.

Think there's settings in IntelliJ for these - can pair if it'd be helpful

crux-core/src/crux/query_state.clj
```diff
@@ -1,28 +1,58 @@
...
-               finished-at
-               (case status
-                 :failed QueryState$QueryStatus/FAILED
-                 :completed QueryState$QueryStatus/COMPLETED
-                 :in-progress QueryState$QueryStatus/IN_PROGRESS)
-               query
-               (when error
-                 (let [{:keys [type message]} error]
-                   (QueryState$QueryError. type message)))))
+      (QueryState. query-id
```
jarohen: two-space here, and for `when`/`let`

jarohen (changes requested):
Thanks @AlistairONeill - change broadly looks good, nice one :) 

Few nits to pick, comments below - mostly indentation changes that I suspect IntelliJ doesn't do automatically by default. Looks like the Kotlin docs change also made it into this PR - could that be separated out (if indeed it's ready to merge)?

Cheers!

crux-core/src/crux/api/ICruxAPI.java
AlistairONeill: Pants, I think intellij automatically redid the the indentation when I touched various things. 
I won't reply to all of the examples, but I'll sort out the indentation in all instances

crux-core/src/crux/query_state.clj
AlistairONeill: I was going based off how the inner interface was imported. I'll change it to just be QueryError

crux-core/src/crux/query_state.clj
AlistairONeill: Oops

AlistairONeill:
@jarohen 
I've made the changes in the formatting.

It seems that <-QueryState is unused and when I grep'd for "<-QueryState" I only found query_state.clj
When I tried to remove it, tests started failing though so I have left it in

PR #1335, opened 2020-12-21
On Heap as Default

hraberg:
Making on-heap the default for most buffers, but can still use off heap when needed.

A further experimental (but less so) PR in the wake of xtdb/xtdb#1331 and xtdb/xtdb#1308.

Buffers and Agrona were originally introduced in xtdb/xtdb#2009 (and later discussed further on xtdb/xtdb#2022), and off heap was made the default, even though it's not faster on its own. What actually is faster is when you can pass the underlying pointer from the KV store all the way to the query engine, or into it when you do seeks, avoiding allocations and copies. This still works in the main places where it's applicable. Usually one constructs a seek key using a thread local buffer, which is still kept off-heap. The Rocks backend has been tweaked so it uses the right form of seek method (on or off heap) instead of moving the incoming key around.

As part of that original PR, we also added our current bump allocator, which is needed to make off heap allocations fast. This PR removes that again, as the default is on heap buffers which are simply backed by new byte arrays. The bump allocator still works, but didn't seem to add any value.

The previous concept of "unpooled buffers" was related to this bump allocator, as holding on to a slice of the underlying chunks could cause memory leaks when placing it in a cache for example. This is now replaced by simply ensuring the buffers are on the heap, letting normal GC do its job to get rid of them.

This PR also removes the usage of JNR to hash buffers, and delegates to our own SHA1 implementation in `crux.ByteUtils/sha1` which works directly against `org.agrona.DirectBuffer` and doesn't care if it's on or off heap.

Contains a somewhat unrelated change to implement navigable sets backed by sorted array lists in some cases, seems consistently about 5% faster.

Adds `MethodHandle` for `Arrays/compareUnsigned` to be used when both buffers are backed by byte arrays. This is used to make it work on 1.8 as this method was added in JDK9. This method compiles to an intrinsic and is much faster when comparing arrays. This PR as a whole makes that often be the case.

Several tweaks to the main tuple projection to avoid allocating intermediate structures.

Adds use of `ThreadLocalRandom` in `crux.cache.second-chance` to pick the random element to evict.

Overall these changes seem to make Crux a 10% faster on TPC-H and in some places somewhat simpler.


jarohen:
hey @hraberg - what's the plan with this one?

hraberg:
Unsure, the idea was to merge it, but bench showed some oddities, might look at it later this week. Nothing strictly urgent here. Could potentially also spin-off a smaller PR with "safer" subsets of this.

hraberg:
I close this for now as the changes are not fundamentally needed as my focus will be elsewhere for a bit.

One standalone commit of this PR that is interesting and could be cherry picked into master is 7c1d908899d30c3e1d2cdb313253b718b899e46d which is very local and resulted in a small but noticeable speed up.

PR #1340, opened 2020-12-22
Java Transactions using concrete types

### LINKED ISSUE #1328: Java users can submit a strongly-typed transaction
Currently, to submit a transaction, Java users need to pass a Clojure data structure of transaction operations (tx-ops). We'd prefer them to use a stronger-typed API, so that their IDEs can help them out.

To consider:
* Before heading into a full implementation, let's discuss a proposal for what this could look like from a user's point-of-view.
* There are various places through `ICruxAPI` and `ICruxDatasource` that either accept or return transactions - `submitTx` and `openTxLog` being the two that spring to mind, but I dare say there's more.
* Let's make sure that whatever we implement here only needs a thin Kotlin layer to become an idiomatic Kotlin DSL.
* The concept of a 'document' is pretty pervasive through these APIs - this is arguably the most important thing to get right. At the moment it's just a map, but there's probably a few different routes we could take, ranging from more weakly to more strongly typed, from simple to complex. Let's list these out as part of the proposal, discuss pros/cons, etc.

Out of scope:
* Queries, and anything query related. Coming soon :)

AlistairONeill:
Main changes:
 - Deprecated NodeConfigurator and ModuleConfigurator in favour of NodeConfiguration, ModuleConfiguration and their respective builders. The builders have the same functionality but by building into a class, it gives immutable reusable configuration objects.
 
 - Transactions can now be made using Transaction.Builder and passed around.
 
 - Transaction objects can be recreated and reused from the TxLog
 
 - IDocument provides an interface for what a document should be
 
 - A few runtime checks to ensure nonsense cruxIds don't get sent to crux.

- JUnit 4 tests within CruxCore that run with the normal "lein test"


crux-core/src/crux/api/ModuleConfigurator.java
```diff
@@ -5,6 +5,7 @@
 import java.util.function.Consumer;
 
 @SuppressWarnings("unused") // entry points
+@Deprecated
```
jarohen, line 8: let's not deprecate these if we've removed their only usage - we may as well remove them

crux-core/project.clj
```diff
@@ -14,20 +14,23 @@
                  [org.clojure/tools.cli "1.0.194"]
                  [org.agrona/agrona "1.7.2"]
                  [com.github.jnr/jnr-ffi "2.1.9" :scope "provided"]
-                 [edn-query-language/eql "1.0.0"]]
+                 [edn-query-language/eql "1.0.0"]
+                 [junit/junit "4.12"]]
```
jarohen, line 18: including this here imposes a compile-scoped dependency on junit, which we don't want :) let's try popping it in the `:test` profile below, see whether `lein-junit` runs using the `:test` profile?

jarohen: it looks like it adds a hook into `lein test`, which does run using that profile - the question will be if running `lein junit` does. if not, `lein with-profile +test junit` should.

crux-core/src/crux/api/alpha/CruxNode.java
```diff
@@ -83,7 +79,7 @@ public Database db() {
      * @param validTime The valid time
      * @return Database instance at validTime
      * @see Database
-     */
+     *
```
jarohen, line 82: sure?

crux-core/src/crux/api/exception/CruxDocumentException.java
```diff
@@ -0,0 +1,8 @@
+package crux.api.exception;
```
jarohen, line 1: We have an existing `crux.IllegalArgumentException` class that we tend to throw in these cases - it has support within the Crux HTTP server/client too.

crux-core/src/crux/api/exception/CruxIdException.java
```diff
@@ -0,0 +1,9 @@
+package crux.api.exception;
```
jarohen, line 1: likewise :)

crux-core/src/crux/api/document/CruxId.java
```diff
@@ -0,0 +1,44 @@
...
+import clojure.lang.IPersistentMap;
+import clojure.lang.Keyword;
+import crux.api.exception.CruxIdException;
+
+import java.net.URI;
+import java.net.URL;
+import java.util.Objects;
+import java.util.UUID;
+
+public class CruxId {
```
jarohen, line 12: In hindsight, I'm not convinced this class is pulling its weight - we have runtime checks for this as soon as it gets into Clojure, I can't see a good way around the `#crux/id` hack, and the user can't do anything with a `CruxId` once they have it - shall we use `Object`?

crux-core/src/crux/api/configuration/NodeConfiguration.java
```diff
@@ -0,0 +1,61 @@
...
+        public Builder() {}
+
+        private final HashMap<String, Object> modules = new HashMap<>();
+
+        public Builder with(String module, ModuleConfiguration configuration) {
+            modules.put(module, configuration.getOpts());
+            return this;
+        }
+
+        public Builder with(String module, Consumer<ModuleConfiguration.Builder> f) {
```
jarohen, line 29: I wonder whether we'd be better off without the `NodeConfiguration.with(String, Consumer)` overload, if the `NodeConfiguration` factory can accept a consumer anyway? Without it, the use case from Java isn't much longer (`c.with("foo", ModuleConfiguration.build(c -> {...}))`, or maybe we rename `build` to `configureModule`/similar and advise a static import?) and it then allows a Kotlin extension function with the `ModuleConfiguration.Builder.() -> ()` signature.

crux-core/src/crux/api/ICruxIngestAPI.java
```diff
@@ -17,6 +21,14 @@
      */
     public Map<Keyword,?> submitTx(List<List<?>> txOps);
 
+    default Map<Keyword,?> submit(Consumer<Transaction.Builder> f) {
```
jarohen, line 24: This wants to be `submitTx` too, I suspect - I'm guessing you've held short because of the issues with overloading from Clojure that we talked about the other day? The Clojure implementation of `db` might be a good example, otherwise, fine to leave it and we'll pair up when I'm back.

jarohen (changes requested):
Cheers @AlistairONeill! Specific feedback inline.

More generally:
* I'd be tempted to migrate your tests into Clojure tests - we'll save ourselves both the additional test paths and the boilerplate involved. In terms of the aim of the tests, we don't really need to test the semantics of Crux on this occasion - it's an adapter layer, we're looking to see that it generates the correct transaction data.
* Meta: I'd break down this PR into 2, maybe 3 - a rule of thumb is that if a change can reasonably be merged independently, it's better as a separate PR (obviously within reason - if it's an uncontroversial two-line change I'll overlook it being bundled in with something else!). If it's separate, it's easier to review, gives us the opportunity to merge one but not the other, and doesn't block a viable, valuable change until the _whole_ PR's mergeable :). In this case, the Configuration change would be self-sufficient, and I wouldn't want to block the submit side of this PR on the ongoing discussions around `openTxLog`
* More meta: I often suggest a lot of things in PRs - doesn't mean it's not a good PR! I'm not _always_ right (!), I don't have as much context as the dev actually writing the PR, and a lot of suggestions about code tend to be subjective anyway. In short, if you don't agree with them, shout :)

For next steps, let's try to get the configuration change landed. If you have separate commits that you can raise in a new PR, great :) If not:
* This time - can `git reset master`, which will reset the commits but leave your changes. Then, make a new commit with the module stuff, push that to its own branch. You can then build commits for the changes to submitTx and openTxLog, and push them to this branch.
* One for next time - worth considering this in your commit history as you go. On your branch, your commits can be quite fluid - we frequently re-write/add to them as we go. Important thing with git commit history is to be helpful when you/others come back to read it. You'll want to know the overall change, its impact, as well as important context that you may forget/others may not notice - you won't care as much about the journey, or indeed a precise log of exactly what you did on each day - it can be quite noisy :) If it helps, I tend to make a load of temporary 'checkpoint' commits as I go, then squash those down once I've got a distinct change-set working, more checkpoints, squash, repeat.

crux-core/src/crux/api/transaction/TransactionWrapper.java
```diff
@@ -0,0 +1,43 @@
+package crux.api.transaction;
+
+import clojure.lang.Keyword;
+import clojure.lang.LazySeq;
+
+import java.util.Date;
+import java.util.Map;
+
+public class TransactionWrapper {
```
jarohen, line 9: This might be better named `TxLogEntry`?

Given it's never submitted by a use,r I suspect this might also be better applied using the QueryState pattern - an interface to specify the data available, and a Clojure record to implement it - will save us the wrapping layer?

jarohen: Ah, ok, so the issue there will be with the tx-ops. :thinking:

crux-core/src/crux/api/transaction/Transaction.java
```diff
@@ -0,0 +1,122 @@
...
+import crux.api.transaction.operation.*;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.Objects;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
+public class Transaction {
```
jarohen, line 15: Thinking on this - would it be easier to specify the Builder as an interface that we then implement in Clojure? The Clojure implementation can then build up the Clojure data structures directly. From a Java user's PoV it'd look similar, but we don't have to worry about building up an object tree in Java that we then translate back into Clojure?

jarohen: This might be a crazy idea, it's been a long night :baby:, but

Java side doesn't differ hugely from what you have:

```java
public interface ITransactionBuilder {
    ITransactionBuilder put(ICruxDocument doc);
    ITransactionBuilder put(ICruxDocument doc, Date fromValidTime);
    ITransactionBuilder put(ICruxDocument doc, Date startValidTime, Date endValidTime);
}

interface ICruxAPI {
    Map<Keyword, ?> submitTx(Consumer<ITransactionBuilder> f);
}
```

On the Clojure side:
```clojure
(defn <-i-crux-doc [^ICruxDocument doc]
  (assoc (.getData doc)
         :crux.db/id (.getId doc)))

(defrecord TransactionBuilder [!tx-ops]
  ITransactionBuilder
  (put [this doc]
    (swap! !tx-ops conj [:crux.tx/put (<-i-crux-doc doc)]))
  ...)

(defrecord CruxNode [...]
  ICruxAPI
  (submitTx [this ^Consumer c]
    (let [builder (->TransactionBuilder (atom []))]
      (.accept c builder)
      (.submitTx this @(:!tx-ops builder)))))
```

Tradeoff is that we lose the Transaction as Java data. The documents themselves still will be, the user supplies those, but we lose the tx-events as Java data.

wdyt? worth a spike?

crux-core/src/crux/api/alpha/CruxNode.java
AlistairONeill: Drat, I thought I caught all of these small changes and reverted them.

I commented out all the classes in alpha while I was working so I wouldn't accidentally clash with them

crux-core/src/crux/api/document/CruxId.java
AlistairONeill: The `CruxId` class doesn't allow allow instances for this reason as we discussed. 

Atm, it is basically a place for static type assertions to live.

I think it could be worth continuing to have these runtime checks when the transaction objects are created, as potentially this could be done far in advance of them actually being run on a node.

crux-core/src/crux/api/configuration/NodeConfiguration.java
AlistairONeill: I'm in two minds about this.
While I like automatically solving the future Kotlin signature problem, requiring a user to manually call `build` (or whatever it gets renamed to) introduces more syntax (and brackets) when there are nested modules.

e.g.

```
Crux.startNode( nc -> {
   nc.with("foo", ModuleConfiguration.build( foo -> {
       foo.with("bar", ModuleConfiguration.build( bar -> {
           
       }));
   }));
});
```

If we advise a static import of a renamed builder then we reduce the bloat on the `.with` lines, but we still have everything in an additional set of brackets.

Each end of the consumer is `}));` as opposed to just `});`

Overall, I think it is worth going down this route now that I think about it more. It also allows us to do things like changing the `ModuleConfiguration.build` to taking vararg Consumers without having to change the entry point in `Crux`. Plus, the Kotlin clash problem resolves itself!

crux-core/src/crux/api/transaction/TransactionWrapper.java
AlistairONeill: Atm, this is created from the path that required with-tx-ops to be true which is why the QueryState pattern might be a little fiddly.

I still think it will be doable, though. I'll have a try

crux-core/src/crux/api/exception/CruxDocumentException.java
AlistairONeill: I'm just entirely blind.
Yep, I'll use that instead

crux-core/project.clj
AlistairONeill: If "test" is not in `:java-source-paths`, then `lein install + lein test` doesn't pick up the Java tests

When the junit import is only in the `:test` profile then `lein install` falls over as the tests need to import from Junit

crux-core/src/crux/api/ModuleConfigurator.java
AlistairONeill: Sounds good to me.

AlistairONeill:
> I'd be tempted to migrate your tests into Clojure tests - we'll save ourselves both the additional test paths and the boilerplate involved.

I think having the tests in Java is the correct thing to do. If a potential user comes along and wants to see how to use common functionality, looking at tests are a great place to start (especially when something is new and there aren't third party open source implementations to look at). If there aren't tests in Java, it might put off someone who only knows and wants to use Java.

> In terms of the aim of the tests, we don't really need to test the semantics of Crux on this occasion - it's an adapter layer, we're looking to see that it generates the correct transaction data.

In the submitTx method of the test, I also test the recreation of the Transaction from the txLog and assert it is equal to the original Transaction data. This means that the Transaction has to be run on the node.

From my own perspective writing the code, as I was running the transactions anyway, it made sense to test the actual impact on the db.

> Meta: I'd break down this PR into 2, maybe 3 - a rule of thumb is that if a change can reasonably be merged independently, it's better as a separate PR (obviously within reason - if it's an uncontroversial two-line change I'll overlook it being bundled in with something else!). If it's separate, it's easier to review, gives us the opportunity to merge one but not the other, and doesn't block a viable, valuable change until the whole PR's mergeable :). In this case, the Configuration change would be self-sufficient, and I wouldn't want to block the submit side of this PR on the ongoing discussions around openTxLog

Yeah, I totally agree with this. I need to get the hang of making my Git usage better for other people. I'll try and get things split up :)

AlistairONeill:
I'm going to turn this PR into a draft so I still have visibility of the comments and submit split up PRs

PR #1345, opened 2020-12-24
Configuration changes for Java node startup

AlistairONeill:
Split out from the larger PR (#1340). 

In this part:

- Removed `NodeConfigurator` and `ModuleConfigurator` in favour of `NodeConfiguration`, `ModuleConfiguration` and their respective builders. The builders have the same functionality but by building into a class, it gives immutable reusable configuration objects.

- Changed the entry points in `Crux.startNode` and `Crux.newIngestClient` to take `NodeConfiguration` instead of `Consumer<NodeConfigurator`.

- Added JUnit4 tests which run when `lein test` is called.


AlistairONeill:
Ugh... I accidentally had my upstream set to origin rather than aon when I created this. Moving the branch to my fork

PR #1346, opened 2020-12-29
Added IDocument for Java interop

AlistairONeill:
Created the crux.document.Document record which implements IDocument and used it in the following places:

- ICruxDataSource.getEntity
- ICruxDataSource.entityHistory
- openTxLog.openTxLog (when withOps==true)


AlistairONeill:
KV store tests failed, most likely the OOM that kills my laptop

PR #1351, opened 2020-12-30
History Options doesn't use Case with Enums

AlistairONeill:
The clojure `case` statement doesn't play well with Java enums as `case` test constants must be compile time literals.

One way of using `case` is to use the ordinal of the enum, but this won't work nicely here as `<-sort-order` is intended to take either the enum or keyword so the most readable solution is to change to condp

### DIFF
--- a/crux-test/test/crux/java_api_test.clj
+++ b/crux-test/test/crux/java_api_test.clj
@@ -133,9 +133,7 @@
                                 (.db node)))))))
 
 (t/deftest test-history-options
-  (let [asc-keyword (his/->history-options :asc)
-        asc-enum (his/->history-options HistoryOptions$SortOrder/ASC)
-        desc-keyword (his/->history-options :desc)
-        desc-enum (his/->history-options HistoryOptions$SortOrder/DESC)]
-    (t/is (= asc-keyword asc-enum))
-    (t/is (= desc-keyword desc-enum))))
+  (t/is (= :asc (his/->history-options :asc)))
+  (t/is (= :asc (his/->history-options HistoryOptions$SortOrder/ASC)))
+  (t/is (= :desc (his/->history-options :desc)))
+  (t/is (= :desc (his/->history-options HistoryOptions$SortOrder/DESC))))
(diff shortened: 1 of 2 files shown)


crux-test/test/crux/java_api_test.clj
```diff
@@ -130,3 +131,11 @@
...
                                 #"Crux node is closed"
                                 (.db node)))))))
+
+(t/deftest test-history-options
+  (let [asc-keyword (his/->history-options :asc)
+        asc-enum (his/->history-options HistoryOptions$SortOrder/ASC)
+        desc-keyword (his/->history-options :desc)
+        desc-enum (his/->history-options HistoryOptions$SortOrder/DESC)]
+    (t/is (= asc-keyword asc-enum))
+    (t/is (= desc-keyword desc-enum))))
```
jarohen: idiom: we don't tend to explicitly let-bind a name if the definition is small enough and only used once. 

also, possible pedantry, but you're relying on `(his/->history-options :asc)` returning the correct thing :) 

```suggestion
(t/deftest test-history-options
  (t/is (= :asc (his/->history-options :asc)))
  (t/is (= :asc (his/->history-options HistoryOptions$SortOrder/ASC)))
  (t/is (= :desc (his/->history-options :desc)))
  (t/is (= :desc (his/->history-options HistoryOptions$SortOrder/DESC))))
```

crux-test/test/crux/java_api_test.clj
```diff
@@ -130,3 +131,9 @@
         (t/is (thrown-with-msg? IllegalStateException
                                 #"Crux node is closed"
                                 (.db node)))))))
+
+(t/deftest test-history-options
+  (t/is (= :asc (his/->history-options :asc)))
+  (t/is (= :asc (his/->history-options HistoryOptions$SortOrder/ASC)))
+  (t/is (= :desc (his/->history-options :desc)))
+  (t/is (= :desc (his/->history-options HistoryOptions$SortOrder/DESC))))
```
jarohen: Sorry @AlistairONeill - test fail here is my fault, serves me right for coding blind...

```suggestion
(t/deftest test-history-options
  (t/is (= :asc (his/<-sort-order :asc)))
  (t/is (= :asc (his/<-sort-order HistoryOptions$SortOrder/ASC)))
  (t/is (= :desc (his/<-sort-order :desc)))
  (t/is (= :desc (his/<-sort-order HistoryOptions$SortOrder/DESC))))
```

Issue #1353, opened 2020-12-30
Fix StackOverflowError in recursive projections


jarohen:
DataScript gets to (at least) 1500 recursion depth before throwing a StackOverflowError, we get to ~400.

I can't immediately think of a simple stackless implementation for this given we're creating a nested data structure, so this might just be a case of how many stack frames we have to add for each level of recursion (we add quite a lot :flushed: largely due to the batching)

refset:
I'm not advocating to bring `xodarap` in as a dependency (and `core.async` along with it), but it's been good to me before and might provide some inspiration: https://github.com/divs1210/xodarap (see also the Reddit thread https://www.reddit.com/r/Clojure/comments/9mxr83/xodarap_stackfree_recursion_in_clojure/)

jarohen:
That looks pretty impressive/magical, have opened up the source code and am still none the wiser how it achieves it, despite the main macro only being ~10LoC. One for a fresh brain!

PR #1356, opened 2020-12-30
More projection params - `:into`, `:default`, `:limit`

### LINKED ISSUE #1220: More projection params
* [x] `:into` param
* [x] `:limit` param
* [x] `:default` param

jarohen:
fixes #1220

### DIFF
--- a/crux-core/test/crux/datascript_pull_test.clj
+++ b/crux-core/test/crux/datascript_pull_test.clj
@@ -125,37 +125,40 @@
              (crux/q db '{:find [(eql/project ?e [* {:parent [:crux.db/id]}])]
                           :where [[?e :crux.db/id :david]]})))))
 
-;; TODO `:limit`
-#_
 (t/deftest test-pull-limit
-  (let [db (d/init-db
-            (concat
-             test-datoms
-             [(d/datom 4 :friend 5)
-              (d/datom 4 :friend 6)
-              (d/datom 4 :friend 7)
-              (d/datom 4 :friend 8)]
-             (for [idx (range 2000)]
-               (d/datom 8 :aka (str "aka-" idx))))
-            test-schema)]
+  (let [db (submit-test-docs (concat people-docs
+                                     [{:crux.db/id :elizabeth,
+                                       :name "Elizabeth"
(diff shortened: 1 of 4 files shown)


crux-core/test/crux/datascript_pull_test.clj
```diff
@@ -125,37 +78,40 @@
...
-             (for [idx (range 2000)]
-               (d/datom 8 :aka (str "aka-" idx))))
-            test-schema)]
+  (let [db (submit-test-docs (concat people-docs
+                                     [{:crux.db/id :elizabeth,
+                                       :name "Elizabeth"
+                                       :friend #{:matthew :eunan :kerri :rebecca}
+                                       :aka (into #{} (map #(str "liz-" %)) (range 2000))}]))]
 
+    ;; I'm tempted not to have a default - might be a gotcha? RFC?
```
jonpither: Do we want to resolve this RFC before merging? Or OK to follow it up after?

jonpither (approved):

crux-core/test/crux/datascript_pull_test.clj
jarohen: @jonpither yep, happy to resolve it now - wdyt?

refset: Assuming we did opt for a similar default limit mechanism, presumably it shouldn't be as high as 1000 based on the numbers in https://github.com/juxt/crux/issues/1353

I'm also inclined to think that a _silently_ default limit isn't a good idea as it could cause subtle errors. Short of implementing a more stack-friendly alternative approach, I think it would be better that a limit is always enforced as user-specified for all recursive usage.

jarohen: @refset this limit thankfully doesn't affect the stack depth. the one in #1353 is related to the nesting of the query (vertical depth of the tree, if you like) - this is elements in a collection at one level (horizontal/breadth).

jarohen: @jonpither, @refset - propose that, given we don't have a limit by default on the main query, we don't limit here?

PR #1357, opened 2020-12-31
Transaction and Queries Doc Changes

### LINKED ISSUE #1088: Ensure that all the various things you can do in queries are documented up

AlistairONeill:
I've restructured and reworded the Documentation sections for Transactions and Queries to be more useful as a reference as opposed to following a narrative.

I also extracted out the Event Subscription to its own section.
 - While we currently only have one Event that we expose (which is to do with transactions), it doesn't feel right to have the general subscription info within the Transaction section.


AlistairONeill:
As part of redoing the Query page, I have ensured that all of the code snippets used in the examples are fully tested.

The formatting of examples.clj looks weird in many places because I wanted the examples to not be indented when the docs are built.

jarohen:
> The formatting of examples.clj looks weird in many places because I wanted the examples to not be indented when the docs are built.

@AlistairONeill you can do this with an asciidoc flag, I think it's `[source,clojure,indent=0]` or similar - that should mean that you can indent the Clojure idiomatically and still not introduce excessive indentation in the docs.

Re the tests - the structure of these hasn't changed since I've joined, but maybe it's time. Particularly, in hindsight, there's no reason for the examples to live in `src` and the tests under `test` - we may as well put the lot under `test` and refer to them from there. Advantage being we shouldn't need a fixture - a standard `deftest` should do. Something like this?

```clojure
(t/deftest test-foo
  (let [node *api*]
    (fix/transact!
     node
     ;; tag::foo-docs[]
     [...]
     ;; end::foo-docs[]
     )
    (t/is (=
           ;; tag::foo-result[]
           #{[...]}
           ;; end::foo-result[]
           (crux/q (crux/db node)
                   ;; tag::foo-query[]
                   '{:find ...}
                   ;; end::foo-query[]
                   )))))
```

I also wouldn't be averse to a macro here, if this is a more general pattern - happy to pair. 

More widely, let's avoid single-char keywords ;)

docs/about/modules/ROOT/examples/test/crux/docs/examples/bitemporality_test.clj
```diff
@@ -0,0 +1,229 @@
+(ns crux.docs.examples.bitemporality-test
+    (:require [clojure.test :as t]
+              [crux.api :as crux]
+              [crux.fixtures :as fix :refer [*api*]]))
+(def ^:dynamic *storage-dir*)
```
jarohen, line 5: newline
```suggestion

(def ^:dynamic *storage-dir*)
```

jarohen (approved):
Approve the tests, thanks @AlistairONeill :) 

* Will run it through the emacs indenter and pop a commit in this PR
* `put-and-wait` - we have a fixture for this, `fix/submit+await-tx` - again, will pop a commit in

Cheers!

Will defer to @refset and @mattford63 for the documentation itself

docs/reference/modules/ROOT/pages/transactions.adoc
```diff
@@ -3,113 +3,74 @@
...
+<1> Document to add
+<2> (optional) <<#valid-time,`valid time`>>
+<3> (optional) <<#end-valid-time,`end valid time`>>
 
 [#delete]
 === Delete
 
-Deletes a document at a given `valid time`.
-Historical versions of the document will still be available.
+Deletes a document.
```
mattford63: Same again, maybe a note here to refer to the valid time section L197 for operational effect?

docs/reference/modules/ROOT/pages/transactions.adoc
```diff
@@ -3,113 +3,74 @@
...
-which is taken from the Kafka log.
-
-Crux `puts` into the past at a single point by default, so to overwrite several
-versions across a range in valid time, you can either submit a transaction
-containing several operations or supply a third argument to specify an `end
-valid time`. This period is inclusive-exclusive, such that the start of the
-validity period is included in the validity range, while the end is excluded.
+<1> Document to add
+<2> (optional) <<#valid-time,`valid time`>>
+<3> (optional) <<#end-valid-time,`end valid time`>>
```
mattford63, line 52: Maybe a note here to refer to the valid time section L197 for operational effect?

mattford63 (approved):

docs/reference/modules/ROOT/pages/transactions.adoc
AlistairONeill: Good point, I had links in the code annotation, but it is clearer to have them in the blurb. Added change

docs/reference/modules/ROOT/pages/transactions.adoc
AlistairONeill: Ditto my above

refset (approved):

PR #1368, opened 2021-01-05
Add 'pred-ctx' to query engine s.t. we don't have to rely on globals in Lucene, #1218

### LINKED ISSUE #1218: Allow pred-fns access to node components

jarohen:
Fixes #1218 - naming RFC.

### DIFF
--- a/crux-lucene/test/crux/lucene_test.clj
+++ b/crux-lucene/test/crux/lucene_test.clj
@@ -13,6 +13,9 @@
            org.apache.lucene.queryparser.classic.QueryParser
            [org.apache.lucene.search BooleanClause$Occur BooleanQuery$Builder Query]))
 
+;; tests in this namespace depend on the `(defmethod q/pred-constraint 'lucene-text-search ...)`
+(require 'crux.lucene.multi-field)
+
 (t/use-fixtures :each lf/with-lucene-module fix/with-node)
 
 (t/deftest test-can-search-string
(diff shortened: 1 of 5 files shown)


crux-lucene/test/crux/lucene_test.clj
```diff
@@ -11,8 +11,12 @@
   (:import org.apache.lucene.analysis.Analyzer
            org.apache.lucene.document.Document
            org.apache.lucene.queryparser.classic.QueryParser
+           [org.apache.lucene.index DirectoryReader]
+           [org.apache.lucene.store Directory]
            [org.apache.lucene.search BooleanClause$Occur BooleanQuery$Builder Query]))
 
+(require 'crux.lucene.multi-field)
```
jarohen, line 18: aside: this test depends on side effects in that namespace

crux-lucene/test/crux/lucene_test.clj
```diff
@@ -157,6 +162,10 @@
     (with-open [db (c/open-db *api*)]
       (t/is (= prior-score (c/q db q))))))
 
+(defn doc-count [{:keys [^Directory directory]}]
```
jonpither: May as well make pvt.

crux-lucene/test/crux/lucene_test.clj
```diff
@@ -168,7 +177,8 @@
     (submit+await-tx [[:crux.tx/put {:crux.db/id "ivan" :name "Ivan"}]])
     (submit+await-tx [[:crux.tx/put {:crux.db/id "ivan" :name "Ivan"}]])
 
-    (t/is (= 2 (l/doc-count)))
+    (t/is (= 2 (doc-count (-> @(get-in *api* [:query-engine :!pred-ctx])
```
jonpither: if we keep doing this access pattern, could establish an fn in the lucene fixtures ns

jonpither (approved):
A nice post-xmas present - thanks @jarohen

PR #1374, opened 2021-01-10
Extracting ICruxAPI and ICruxDatasource from the Clojure side of things

AlistairONeill:
(see #1375)

As discussed, removed the reliance on the Java interfaces to interact with PCruxNode, PCruxIngestClient, PCruxDatasource and PCruxAsyncIngestClient.

This leaves us free to change the Java interfaces to receive and return strong types without impacting pure Clojure crux users.


AlistairONeill:
`crux.api/new-api-client` was originally going through the static java `Crux.newApiClient` which calls to `crux.remote-api-client/new-api-client`
This hides the fact that you need to have included `crux-http-client` on the classpath. It feels wrong to have static entry methods that will just fall over conditional on other dependencies. It was outside of the scope of this change to fix this, so I have left the pattern as is for now.

crux-core/src/crux/api.clj
```diff
@@ -448,19 +248,170 @@
...
+  (alter-meta! #'entity-history assoc :arglists arglists)
+  (alter-meta! #'open-entity-history assoc :arglists arglists))
+
+(defrecord JQueryDatasource [^java.io.Closeable datasource]
+  ICruxDatasource
+  (entity [_ eid] (entity datasource eid))
+  (entityTx [_ eid] (entity-tx datasource eid))
+  (query [_ query args] (q* datasource query args))
+  (openQuery [_ query args] (open-q* datasource query args))
+  (entityHistory [_ eid opts] (entity-history datasource eid opts))
```
jarohen: as discussed, don't need the `HistoryOptions` Java object coming all the way through into Clojure now - can adapt it at this point

crux-core/src/crux/api/IUnwrappable.java
```diff
@@ -0,0 +1,5 @@
+package crux.api;
+
+public interface IUnwrappable {
```
jarohen: as discussed, let's remove

crux-core/src/crux/query.clj
```diff
@@ -1748,23 +1749,23 @@
...
-    (with-open [res (.openQuery this query args)]
+  (q* [this query args]
+    (with-open [res (api/open-q* this query args)]
       (let [result-coll-fn (if (some (normalize-query query) [:order-by :limit :offset]) vec set)
             !timed-out? (atom false)
             ^Future
             interrupt-job (when-let [timeout-ms (get query :timeout (:query-timeout this))]
                             (let [caller-thread (Thread/currentThread)]
-                              (.schedule interrupt-executor
+                              (.schedule ^ScheduledExecutorService interrupt-executor
```
jarohen: this type hint can go on the field on the `defrecord`, above

crux-core/src/crux/history_options.clj
```diff
@@ -1,36 +1,24 @@
-(ns crux.history-options
-  (:import crux.api.HistoryOptions$SortOrder))
+(ns crux.history-options)
```
jarohen: I don't think we need this in Clojure any more? Would it be worth us considering making `HistoryOptions` a concrete class?

crux-core/src/crux/api/alpha/CasOperation.java
```diff
@@ -5,6 +5,7 @@
```
jarohen, line 5: just in case I haven't mentioned it - intention is that `crux.api.alpha` gets deleted after your next block of work - if it's too much hassle to keep bringing it up to date, feel free to submit a PR deleting it

crux-core/src/crux/api/ICruxDatasource.java
```diff
@@ -117,7 +120,10 @@
      * @param eid The entity id to return history for
      * @return a cursor of changes to the given entity.
      */
-    public ICursor<Map<Keyword, ?>> openEntityHistory(Object eid, HistoryOptions options);
+    ICursor<Map<Keyword, ?>> openEntityHistory(Object eid, HistoryOptions.SortOrder sortOrder, HistoryOptions options);
```
jarohen: This is a _sneaky_ little breaking change :scream: :wink: 

I think you've managed to keep the whole of the rest of the change non-breaking (nice one!) so let's try avoid it here. I like the idea of the user not needing to create the full HistoryOptions object, but I think we can achieve that with:

```java
ICursor<Map<Keyword, ?>> openEntityHistory(Object eid, HistoryOptions options)
default ICursor<Map<Keyword, ?>> openEntityHistory(Object eid, HistoryOptions.SortOrder sortOrder)
```

jarohen (changes requested):
Cheers for the changes, @AlistairONeill :) 

As discussed, have squashed up the commits and re-indented. Couple of minor discussion points below, let's pair up on those if it'd be easier, doubt they'll take long.

I haven't looked through the JUnit tests in great detail, on the assumption that they're going to change significantly with the more strongly typed API?

crux-core/src/crux/api/alpha/CasOperation.java
AlistairONeill: I think they got hit when I was going through the code analysis auto-fixes. Was almost zero hassle :')
I think its best to leave in until the block of work is complete. Easier to refer to for any relevant inspiration

crux-core/src/crux/api/ICruxDatasource.java
AlistairONeill: I was trying to put it in line with the clojure open-entity-history which has the sort order distinct from the rest of the options, but you are right, it *is* breaking.

I'm happy to change it back, though, for consistency with the previous version.

AlistairONeill:
>I haven't looked through the JUnit tests in great detail, on the assumption that they're going to change significantly with the more strongly typed API?

Yeah. At the moment, they are pretty much "Does this method actually link up correctly with the clojure entity inside?".

As I go through each of them to make them typed, they will get proper tests of all the different options

crux-core/src/crux/api/ICruxDatasource.java
AlistairONeill: Changed it back. :) 

I left in the default versions which just take sort order as they don't break anything and can be convenient rather than having to manually create a `HistoryOption` instance.

Feels a little messy in the Java to Clojure conversion as we effectively take the sort order from the options to pass into `entity-history` as a separate argument only for the implementations to immediately `assoc` it back in ¯\\_(ツ)_/¯

PR #1382, opened 2021-01-18
TransactionInstant is a concrete class

### LINKED ISSUE #1213: Java API shouldn't use `Map<Keyword, ?>` to represent transaction instants

AlistairONeill:
Draft until the big PR is merged and this can be rebased properly (The tests for this use JUnit which is added in that PR)

Changed places that return or accept a transaction in the form `Map<Keyword, ?>` to use TransactionInstant.

Behaviour of factory methods:

- Fine to have one or both of `id` and `time`

- If supplied a `Map<Keyword, ?>` which contains neither, the factory will return `null`.
This is for compatibility with things like `latestCompletedTx` being called before any transactions are completed.


AlistairONeill:
Rebased on master now.

The test that failed is crux.query-test which this change doesn't touch. 
It is about a slow query.

I had a look and nothing in the test touches any of the Java side of things

jarohen:
This is great, thanks @AlistairONeill and @danmason :smile:

PR #1385, opened 2021-01-20
Transactions are now concrete types

### LINKED ISSUE #1328: Java users can submit a strongly-typed transaction

AlistairONeill:
Draft as will rebase after previous PR's are dealt with. Trying to make it independent of the TransactionInstant PR would be a pain as it modifies some of the same lines.

As part of this, created:
 - Transaction
 - TransactionOperation (Which is extended by Put, Match, Evict, Delete)
 - AbstractCruxDocument
 - CruxDocument
 
 Not dealt with:
 - Transaction Functions
 - TxLog


jarohen (commented):
Hey, nice one - looks like your previous PRs have made this quite a lot more straightforward! :smile: 

A few points for discussion:
* Completely subjective, but `factory` -> `create`? `CruxDocument.create` seems more familiar to me...
* When you create `Operations`, could you call the VT parameter `startValidTime` for Put and Delete (for the single-vt case too); Match is `atValidTime`.
* I know IDEs complete it, and I know Java's just generally verbose, but worth eliding the `Transaction` in `PutTransactionOperation` etc, just `PutOperation`, to cut down on noise?
* The more I see of Java manipulating Clojure keywords and data structures, the more I'm tempted to go back and implement the concrete types in Clojure - wdyt?
* In the absence of sealed classes, worth having a visitor for `TransactionOperation`s so that the user doesn't need to switch on `Type` and then cast?
* Wouldn't mind running through the rationale for `AbstractCruxDocument` - is the intention that users can create their own? 
* Even if you're not supporting users creating transaction functions (fair enough), supporting invoking them should still be possible - the operation takes a CruxId and a list of Object arguments

AlistairONeill:
> Completely subjective, but factory -> create? CruxDocument.create seems more familiar to me...

I've generally gone down the line of `factory` being things that (when we figure out the visibility issues) shouldn't be public. Mostly going from Clojure structures -> Java
The `create` methods are meant to be public facing

> When you create Operations, could you call the VT parameter startValidTime for Put and Delete (for the single-vt case too); 
Match is atValidTime.

Yup, will do

> I know IDEs complete it, and I know Java's just generally verbose, but worth eliding the Transaction in PutTransactionOperation etc, just PutOperation, to cut down on noise?

Yep, this is grand. Also TIL a new word :)

>The more I see of Java manipulating Clojure keywords and data structures, the more I'm tempted to go back and implement the concrete types in Clojure - wdyt?

I'm not sure what we would gain from doing this. As long as the visibility namespace issue is sorted, we can hide away almost all of the clojure data structures.
Also, for the Consumer/Receiver builder pattern, we need the builder to be mutable as we aren't chaining.

>In the absence of sealed classes, worth having a visitor for TransactionOperations so that the user doesn't need to switch on Type and then cast?

I don't see why not. I haven't had much experience with them before but I'll look into it :)

>Wouldn't mind running through the rationale for AbstractCruxDocument - is the intention that users can create their own?

Yeah, the idea is that it is easy to define a class and just use it as a document (e.g. PUT it directly). It had to be an abstract class rather than an interface so that I could override equals and allow the comparison between the CruxDocument you receive back from `.entity` or similar.

E.g. 
````
class PersonDocument extends AbstractCruxDocument {
        private static final Keyword NAME = Keyword.intern("name");
        private static final Keyword LAST_NAME = Keyword.intern("lastName");
        private static final Keyword HEIGHT = Keyword.intern("height");

        private final String id;
        private final String name;
        private final String lastName;
        private final int height;

        public PersonDocument(String id, String name, String lastName, int height) {
            this.id = id;
            this.name = name;
            this.lastName = lastName;
            this.height = height;
        }


        @Override
        public Object getId() { // <1>
            return id;
        }

        @Override
        public Map<Keyword, Object> getData() { // <2>
            HashMap<Keyword, Object> ret = new HashMap<>();
            ret.put(NAME, name);
            ret.put(LAST_NAME, lastName);
            ret.put(HEIGHT, height);
            return ret;
        }
    }
````

In the future I would quite like an AutoCruxDocument that can be extended where it would look something like:

````
class Person extends AutoCruxDocument {
        @CruxId private final String id;
        @CruxField private final String name;
        @CruxField private final String lastName;
        @CruxField private final int height;

        public PersonDocument(String id, String name, String lastName, int height) {
            this.id = id;
            this.name = name;
            this.lastName = lastName;
            this.height = height;
        }
    }
````

And it reflectively generates the bits needed. But this was a *little* outside the scope of the change ;)

> Even if you're not supporting users creating transaction functions (fair enough), supporting invoking them should still be possible - the operation takes a CruxId and a list of Object arguments

That's fair, I'll add the invokation. Going to make the docs a little awkward.
"Here's how you call the thing" "How do you make it?" "¯\\_(ツ)_/¯"

**Comment by AlistairONeill** [comment]
Made the changes discussed above :)

**Comment by AlistairONeill** [comment]
Using the Visitor managed to eliminate any reference to IPersistentVector from all the TransactionOperations which is great.

Removed the AbstractCruxDocument concept as we discussed. So now the class that someone puts in is what they get out.

Allowed the creation of documents which are TransactionFunctions. As discussed, they take a String which is parsed by `Clojure.read`. I don't add any additional checks on it being valid.

File: crux-core/src/crux/api.clj [hunk 1]
```diff
@@ -413,7 +413,13 @@
   (^ICruxDatasource openDB [_ ^Map basis] (->JCruxDatasource (open-db node basis)))
   (status [_] (status node))
   (attributeStats [_] (attribute-stats node))
-  (submitTx [_ tx-ops] (TransactionInstant/factory ^Map (submit-tx node tx-ops)))
+
+  (submitTx [_ tx]
+    (->> tx
```
jarohen: I get that you _can_ thread here, but I probably wouldn't - I try to keep usage of threading when you're passing a data structure through a side-effect free pipeline. Here, the `submit-tx` isn't really part of a transformation pipeline, so I wouldn't thread.

crux-core/src/crux/api/CruxDocument.java
```diff
@@ -0,0 +1,155 @@
...
+    public Object getId() {
+        return id;
+    }
+
+    public IPersistentMap toMap() {
+        return data.assoc(DB_ID, id);
+    }
+
+    private static void assertNotReserved(Keyword key) {
+        if (DB_ID.equals(key)) throw new RuntimeException(":crux.db/id is a reserved key");
```
jarohen: `IllegalArgumentException`?

crux-core/src/crux/api/CruxDocument.java
```diff
@@ -0,0 +1,155 @@
...
+
+    public static CruxDocument createFunction(Object id, String rawFunction) {
+        return new CruxDocument(id, PersistentArrayMap.EMPTY.assoc(FN_ID, Clojure.read(rawFunction)));
+    }
+
+    public static CruxDocument create(Object id, Map<String, Object> data) {
+        return create(id).putAll(data);
+    }
+
+    public CruxDocument put(String key, Object value) {
```
jarohen: `put`/`remove` on a Java Object are _usually_ (IME?) mutative - maybe borrowing `plus`/`minus` from Kotlin?

jarohen: I might be tempted to open up a Builder here based on the current Crux document, delegating through to its `put` method, and then building a new `CruxDocument` - save yourself duplicating the business logic?

(same for the other methods below)

crux-core/src/crux/api/tx/FunctionOperation.java
```diff
@@ -0,0 +1,52 @@
+package crux.api.tx;
+
+import java.util.Arrays;
+import java.util.Objects;
+
+public final class FunctionOperation extends TransactionOperation {
```
jarohen: naming is hard... :thinking: 

Think this needs some reference to `Invoke`? (aware the Clojure one doesn't)

crux-core/src/crux/api/tx/Transaction.java
```diff
@@ -0,0 +1,214 @@
...
+        public final Builder add(TransactionOperation operation) {
+            operations.add(operation);
+            return this;
+        }
+
+        public final Builder put(CruxDocument document) {
+            return add(PutOperation.create(document));
+        }
+
+        public final Builder put(CruxDocument document, Date validTime) {
```
jarohen: more `startValidTime`s here

crux-core/src/crux/api/tx/Transaction.java
```diff
@@ -0,0 +1,214 @@
...
+
+        public final Builder matchNotExists(Object id) {
+            return add(MatchOperation.create(id));
+        }
+
+        public final Builder match(CruxDocument document) {
+            return add(MatchOperation.create(document));
+        }
+
+        public final Builder matchNotExists(Object id, Date validTime) {
```
jarohen: `atValidTime`

crux-core/src/crux/api/tx/TransactionOperation.java
```diff
@@ -0,0 +1,34 @@
...
+        Type(String idRaw) {
+            keyword = Keyword.intern(idRaw);
+        }
+
+        final Keyword getKeyword() {
+            return keyword;
+        }
+    }
+
+    public interface Visitor {
```
jarohen: this is usually `Visitor<R>`, with each of the `visit` methods returning `R` instead - means that you can write less side-effecting visitors

crux-core/src/crux/api/tx/Transaction.java
```diff
@@ -0,0 +1,214 @@
...
+        }
+    }
+
+    public final IPersistentVector toVector() {
+        EdnVisitor visitor = new EdnVisitor();
+        accept(visitor);
+        return visitor.vector;
+    }
+
+    public void accept(TransactionOperation.Visitor visitor) {
```
jarohen: (see `Visitor<R>` comment below first)

I see what you're trying to do by adding the `accept` method on `Transaction` too - agree that it's a common enough case to warrant such a method. I'd be tempted to make this `public <R> List<R> accept(Visitor<R> visitor)`, which would be less mutable/side-effecty?

The individual `visit` methods would then become `return toAdd`, and `toVector` becomes `return accept(new EdnVisitor())`.

crux-test/test/crux/api/TransactionInstantTest.java
```diff
@@ -91,7 +93,20 @@ public void mapIsFineWithExtraneousTest() {
...
         TransactionInstant tx = TransactionInstant.factory(map);
 
-        Assert.assertEquals(now, tx.getTime());
-        Assert.assertEquals(1L, (long) tx.getId());
+        assertEquals(now, tx.getTime());
+        assertEquals(1L, (long) tx.getId());
+    }
+
+    @Test
+    public void foo() {
```
jarohen: guessing this wasn't meant to be included? :wink:

crux-test/test/crux/api/TransactionOperationVisitorTest.java
```diff
@@ -0,0 +1,162 @@
...
+        private void throwRuntime() {
+            throw new RuntimeException("We called into a method we shouldn't have!");
+        }
+
+        Object get() {
+            return object;
+        }
+    }
+
+    private static class ObjectWrapper {
```
jarohen: `Visitor<R>` would save you this mutable object too

crux-core/src/crux/api/tx/MatchOperation.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public static MatchOperation create(CruxDocument document, Date validTime) {
+        return new MatchOperation(document.getId(), document, validTime);
+    }
+
+    public Object getId() {
+        return id;
+    }
+
+    public CruxDocument getCompare() {
```
jarohen: `compare` isn't a word we use in this context elsewhere in Crux, so not sure I'd introduce it here. `document`?

crux-test/test/crux/api/JCruxNodeTest.java
```diff
@@ -107,66 +96,66 @@ public void statusTest() {
 
     @Test(expected = NodeOutOfSyncException.class)
     public void hasTxCommittedThrowsTest() {
-        TransactionInstant tx = put();
+        TransactionInstant tx = p();
```
jarohen: not a fan of single-letter methods - could this go back to `put`?

jarohen (changes requested):
Hey @AlistairONeill, thanks! another round of feedback included below - let me know if it'd be useful to pair up :)

crux-core/src/crux/api.clj
AlistairONeill: That's fair. I thought it was just for reducing bracket craziness

crux-test/test/crux/api/TransactionInstantTest.java
AlistairONeill: ... I'mma go with no

crux-core/src/crux/api/tx/MatchOperation.java
AlistairONeill: Yup, happy to change this

crux-test/test/crux/api/JCruxNodeTest.java
AlistairONeill: Yep. This was a holdover from when it was conflicting with the static import of TestUtils. This test now uses `TestUtils.tx` instead so I've changed it to the full word.

crux-core/src/crux/api/tx/Transaction.java
AlistairONeill: Yep

crux-core/src/crux/api/tx/Transaction.java
AlistairONeill: Yep

crux-core/src/crux/api/tx/FunctionOperation.java
AlistairONeill: Sure thing

crux-core/src/crux/api/tx/Transaction.java
AlistairONeill: Made the change :)

crux-core/src/crux/api/CruxDocument.java
AlistairONeill: I've changed the names to `plus`/`minus`. Also, the fact that `remove` usually returns the value associated with the removed key (if present) whereas we are returning the new `CruxDocument`

I've changed the `plus`/`minus` to use the builder

AlistairONeill: Although, just had a thought that this might cause issues when `plus`/`minus` on a function `CruxDocument`.
Will write a test and see

crux-core/src/crux/api/CruxDocument.java
AlistairONeill: Changed :)

crux-core/src/crux/api/CruxDocument.java
AlistairONeill: I was overreacting, it is fine :)

PR #1397, opened 2021-01-25
Added failing tests for Java Map values in documents

### LINKED ISSUE #1398: Should be able to use java.util.Map as a value in documents
We can't currently use a `java.util.Map` as a value in a document.

This is because we call into `nippy/write-map` in [codec.clj](https://github.com/juxt/crux/blob/master/crux-core/src/crux/codec.clj#L151) which requires the map supplied to implement `reduce-kv` which a java.util.Map doesn't.

AlistairONeill:
Using `java.util.Map` as a value in a document is broken. This PR adds tests to check that they work as intended when the fix is implemented.

### DIFF
--- a/crux-test/test/crux/tx_test.clj
+++ b/crux-test/test/crux/tx_test.clj
@@ -880,37 +880,37 @@
   (t/testing "URL"
     (test-putting-retrieving-querying-id-and-value (URL. "https://github.com/juxt/crux")))
   (t/testing "IPersistentMap"
-    (test-putting-retrieving-querying-id-and-value (.assoc (PersistentArrayMap/EMPTY) "foo" "bar")))
+    (test-putting-retrieving-querying-id-and-value (-> (PersistentArrayMap/EMPTY)
+                                                       (.assoc "foo" "bar"))))
   (t/testing "Set"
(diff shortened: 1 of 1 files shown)


jarohen:
thanks for these :smile: mind including them with the other related tests, `tx_test.clj` `map-ordering-362` and `incomparable-colls-1001`?

(might well only be the former)

jarohen (changes requested):
thanks for these :smile: mind including them with the other related tests, `tx_test.clj` `map-ordering-362` and `incomparable-colls-1001`?

(might well only be the former)

AlistairONeill:
Moved to `tx-test`

AlistairONeill:
Failed Test: https://app.circleci.com/pipelines/github/juxt/crux/1838/workflows/b2348fa6-7fe7-473a-9631-1615148aa2e5/jobs/7104

It looks like this one fails intermittently as I haven't touched any of the code only added extra tests so can't have slowed anything down crux wise

jarohen:
Merged in 8d8e6a54e567aeade8372e99f0409ae1de352fec, thanks @AlistairONeill :pray:

PR #1400, opened 2021-01-26
Made ModuleConfiguration and NodeConfiguration classes in their own right

AlistairONeill:
Changed from starting nodes with `Consumer<NodeConfigurator>` to an instance of `NodeConfiguration`.

This allows for reusing configurations more easily. The Configuration objects themselves are immutable once built.

Potentially, we could leave in a default implementation for startNode and newIngestClient that takes `Consumer<NodeConfiguration.Builder>`. This would make the change non-breaking as the `Builder` classes have the same methods on them as the original `Configurator`s.


crux-core/src/crux/api/Crux.java
```diff
@@ -85,9 +85,9 @@ public static ICruxAPI startNode(URL url) throws IndexVersionOutOfSyncException
      * Starts a Crux node using the provided configuration.
      * <p>
      * <pre>
-     * ICruxAPI cruxNode = Crux.startNode(n -> {
+     * ICruxAPI cruxNode = Crux.startNode(buildNode(n -> {
```
jarohen: This looks like it's on the Javadoc for the overload expecting a Builder, so a separate call to `buildNode` shouldn't be necessary?

crux-core/src/crux/api/ModuleConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+        public Builder set(Map<String, Object> options) {
+            opts.putAll(options);
+            return this;
+        }
+
+        public Builder with(String module) {
+            return set(module, ModuleConfiguration.EMPTY);
+        }
+
+        public Builder with(String module, ModuleConfiguration configuration) {
```
jarohen, line 51: this could also have a Consumer arity, like the NodeConfiguration builder?

jarohen (changes requested):
Hey, cheers! Couple of minor nits and this is good to go. 

Any doc updates required? If so, could you tack them on to this PR?

crux-core/src/crux/api/Crux.java
AlistairONeill: Yep. I forgot to change the Javadoc back after adding the `Consumer` signature back in

AlistairONeill:
No docs changes needed I believe as the docs use the `Consumer<Foo>` approach but don't specify the type as they are just usage examples

PR #1403, opened 2021-01-28
Adds a note to the JDBC documentation regarding the PostgreSQL driver.

coyotesqrl:


jarohen:
As discussed, have updated the deps in the `project.clj` to avoid the confusion - thanks for bringing it to our attention! :pray:

PR #1404, opened 2021-01-29
Added Java to Transaction Docs

AlistairONeill:
Split out Transactions into:

Transactions <- General concepts about valid times and awaiting
  - Clojure <- Didn't touch much apart from extracting snippets for tests and removing the valid time bit to the top level
  - Java


AlistairONeill:
As discussed, will merge the Java and Clojure pages to use tabbed views

jarohen:
Thanks @AlistairONeill, have deployed these to https://opencrux.com/_jms/reference/transactions.html.

Have pushed an indentation change through for the Clojure examples, will see if that passes tests or whether there's something else up.

AlistairONeill:
Not sure what the issue with the test was. I couldn't replicate on my machine (without the indentation changes)

PR #1405, opened 2021-01-29
Made DBBasis a class

AlistairONeill:
Another purge of `Map<Keyword, ?>`


jarohen:
Grand, thanks! :pray:

PR #1408, opened 2021-02-03
NIO FileSystem doc-store + checkpoint-store

jarohen:
We're trialling using an NIO-based ObjectStore in other current spikes, have ported that back to Crux, used it in the existing local file-system implementation and have GCS support in #1458 .

Java's NIO adds a `FileSystem` abstraction which can be implemented by anything that looks roughly like a file system. By default, it uses the local file-system, but there're also officially supported implementations for [GCS](https://github.com/googleapis/java-storage-nio) and [Azure](https://github.com/Azure/azure-sdk-for-java/tree/master/sdk/storage/azure-storage-blob-nio), and a few third-party ones for AWS S3.

We may find that this abstraction isn't ideal - it's not specifically meant for remote file systems, so a few of the operations may not be quite as efficient as if they were hand-rolled (hence why I haven't ported the hand-rolled Azure implementation over) - but it might provide a quick route to getting something working.


jarohen:
@danmason to road-test this on upcoming project work before it gets merged.

jonpither (approved):

crux-s3/test/crux/s3_test.clj
```diff
@@ -10,7 +10,8 @@
 
 (def test-s3-bucket
   (or (System/getProperty "crux.s3.test-bucket")
-      (System/getenv "CRUX_S3_TEST_BUCKET")))
+      (System/getenv "CRUX_S3_TEST_BUCKET")
+      #_"jms-crux-test"))
```
hraberg: Seems to been left behind after local testing..

hraberg (approved):

PR #1419, opened 2021-02-10
Seperate out function for `->crux-routes`.

### LINKED ISSUE #1412: Make the crux routes public to allow embeding in existing applications
It would be nice if the routes below were public and not wrapped in rr/router, this would allow the routes to be embedded into an exiting handler, seems the data itself could be split into its own def which is public then ->crux-router can still apply rr/router to the data.

https://github.com/juxt/crux/blob/951a2584fee4888318e8ca1d20955dc1de2eb093/crux-http-server/src/crux/http_server.clj#L317

This makes trying crux in existing projects easier, in my case I have a ring reitit api and want to just add the routes to my existing setup to experiment with crux this avoids me creating a separate project and deploy setup to use crux.

I spoke to @refset about this and mentioned it had been discussed previously and I have created this issue off the back of that discussion.

danmason:
Makes the crux-routes public, resolves #1412

### DIFF
--- a/crux-http-server/src/crux/http_server.clj
+++ b/crux-http-server/src/crux/http_server.clj
@@ -314,10 +314,9 @@
                                                    (m/slurp)
                                                    (json/read-value))}}))))
 
-(defn- ->crux-router [{{:keys [^String jwks, read-only?]} :http-options
-                       :keys [crux-node], :as opts}]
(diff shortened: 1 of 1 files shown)


jarohen:
As discussed on #1412, should be sufficient to embed at the handler level, which is routing-library independent.

Cheers @danmason

PR #1420, opened 2021-02-12
WIP: Add `model` section.

danmason:
Draft - still more to add here. Adds `model` section to the docs.


jarohen:
Let's re-open this one if we get back to it.

PR #1431, opened 2021-02-21
Add Egeria crux-lucene configuration example

refset:


cmgrote:
Sorry to make this even more complicated, but after some additional thought on the patterns we want to support in Egeria, I think it would make sense if we could push-down _both_ case-sensitive and case-insensitive queries to Lucene.  This would avoid the need to do a secondary check of matches against a regex in the Crux query logic, which creates potentially significant extra work as the same value coming back from Lucene may be confirmed / rejected multiple times with this approach.

By pushing down the query (whether case-sensitive or case-insensitive) to Lucene, we ensure that whatever we get back from Lucene is already confirmed to be the value we want to use to directly reverse-lookup against the Crux index store.  (No additional, potentially duplicative, regex-based confirmation of the values needed.)

However, this would mean that we need the same field indexed by Lucene as both case-insensitive and case-sensitive.

I've done some very basic testing of an approach to this which would avoid the need to create two separate indexes (taking advantage of our reliance on the simplicity of `KeywordAnalyzer` itself) -- more details here: https://github.com/odpi/egeria-connector-crux/issues/45

Would this be something to consider adding some further nuances / complexity to the example here?

cmgrote:
Here is an undoubtedly _very_ hacky approach to implementing the ideas above, but it seems to work (presumably it could be simplified so there's not so much code duplication):

https://github.com/odpi/egeria-connector-crux/blob/main/src/main/clojure/crux/lucene.clj

refset:
The immediate need for this no longer exists after @jarohen pointed out that the Egeria project requiring a dedicated Clojure build was not necessary, since custom Clojure can be compiled dynamically at run-time.

My recommendation on the Zulip thread:

> you just need to have your custom lucene-egeria.clj file on the classpath (as well as any deps it needs), such as alongside your Java in src/main/resources (or you could create src/main/clojure as a resource directory) and Crux will take care of loading/compiling everything. This should work by virtue of Crux having Clojure embedded as a dependency and making it available at runtime

...which was validated and implemented successfully here:
https://github.com/odpi/egeria-connector-crux/commit/c12f66b053a015fb433649c616e19bd5390aa413

PR #1437, opened 2021-02-25
A start at fixing crux-site infuriations

jarohen:
see juxt/crux-site#35 and juxt/crux-site#36

* CF moved to crux-site
* Main build/deploy scripts moved to crux-site. Convenience build script left here for testing changes locally - can still build the docs without access to crux-site.
* package.json created so that dep installation is just `npm install`


danmason:
Both commits look good to me and is a welcome change :+1:

PR #1440, opened 2021-02-26
Renaming project to pull

### LINKED ISSUE #1355: Rename 'projection' to 'pull'
After these changes we'll be close enough to make it confusing that we're _not_ calling it 'pull'

jarohen:
see #1355


crux-core/src/crux/api/ICruxDatasource.java
```diff
@@ -55,38 +55,38 @@
...
      * @return the requested projection starting at the given entity
      */
-    Map<Keyword, ?> project(Object projection, Object eid);
+    Map<Keyword, ?> pull(Object projection, Object eid);
 
     /**
      * Returns the requested data for the given entity IDs, based on the projection spec
      *
-     * e.g. `db.projectMany("[:film/name :film/year]", Arrays.asList("spectre", "skyfall"))`
+     * e.g. `db.pull("[:film/name :film/year]", Arrays.asList("spectre", "skyfall"))`
```
deobald: Should this not be `pullMany`?

Issue #1316, opened 2020-12-10
Segfault on ingestion when closing node?


jarohen:
Not sure the KV tx-log respects the node's closing latch?

PR #1448, opened 2021-03-05
Not shutting down IndexStore til all the snapshots are closed

jarohen:
related to #1316, #1445 (merge after #1445, diff of this PR: https://github.com/jarohen/crux/compare/ingest-segfault-1316...jarohen:index-store-thread-cleanup)

We keep track of the open snapshots in the IndexStore. When the IndexStore is then closed, it interrupts any threads that haven't closed their snapshots, so that they can get on and close them. This then means that the KV store doesn't shut down til all snapshots are shut, which prevents segfaults in Rocks caused by use-after-close bugs.

Also fixed a snapshot leak in node/db, unearthed by the test's index-store not closing :confused:

### DIFF
--- a/crux-core/src/crux/kv/index_store.clj
+++ b/crux-core/src/crux/kv/index_store.clj
@@ -12,7 +12,8 @@
             [crux.memory :as mem]
             [crux.morton :as morton]
             [crux.status :as status]
-            [crux.system :as sys])
+            [crux.system :as sys]
+            [clojure.tools.logging :as log])
   (:import clojure.lang.MapEntry
            crux.api.IndexVersionOutOfSyncException
            [crux.codec EntityTx Id]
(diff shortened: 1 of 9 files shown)


crux-core/src/crux/kv/index_store.clj
```diff
@@ -535,6 +536,35 @@
...
+  (snapshot-closed [_ snapshot]))
+
+(defrecord ThreadManager [^Map snapshot-threads]
+  PThreadManager
+  (snapshot-opened [this snapshot]
+    (locking this
+      (.put snapshot-threads snapshot (Thread/currentThread))))
+
+  (snapshot-closed [this snapshot]
+    (locking this
```
hraberg, line 555: I recommend using a ConcurrentHashMap instead of explicit locks.

crux-core/src/crux/kv/index_store.clj
```diff
@@ -535,6 +536,35 @@
...
+      (.remove snapshot-threads snapshot)
+      (.notifyAll this)))
```
hraberg: I would also recommend not using `notifyAll` / `wait` directly, but some higher level `java.util.concurrent.`. construct depending on what you trying to do, would also remove the need for explicit `locking` most likely. In this case you might not need anything else, but just use a simple polling loop in close.

hraberg (changes requested):

crux-core/src/crux/kv/index_store.clj
jarohen: migrated to use a Semaphore

crux-core/src/crux/kv/index_store.clj
jarohen: We still need locks here because of the atomic access required to both the map and the semaphore, so I've left it as a map?

hraberg (approved):

PR #1450, opened 2021-03-05
Adds support for single find arg, #1449

### LINKED ISSUE #1449: Adding support for a single unwrapped find-arg
(from @refset)

https://github.com/juxt/crux/blob/master/crux-core/src/crux/query.clj#L125

See `.` and `...` find specs: https://github.com/tonsky/datascript/blob/master/test/datascript/test/query_find_specs.cljc#L18-L28

jarohen:
NOTE: we cannot support single find-args in Java, because the return type is specified to be `Collection<List<?>>`. Given we're planning to re-visit the Java query API shortly, I've left it as unsupported for now.


hraberg:
As discussed, I prefer to add the real find specs from EDN Datalog and not add our own extensions to the map form of the queries, especially not when there's overlap.

jarohen:
Will close this for the time being, it's turning out to not be as quick a win as I'd hoped for a quick Friday hack

PR #1468, opened 2021-03-19
Use Lucene SearcherManager instead of a new IndexWriter per tx

refset:
This somewhat improves ingestion performance with `crux-lucene` for small transactions although batching operations into large transactions is still a far better strategy.

In relation to https://github.com/juxt/crux/pull/1468/files#diff-97fb172b1e6d0b14e8f78e6bdcb364d8fef607cf1ad68e42d97e6404744b9e9fR246-R247

- It's _far_ less than ideal to .commit here per transaction (this was essentially happening before this PR also, as the commits happened as part of the `.close` of the short-lived IndexWriters)
- It's approximately the same tradeoff as the `:sync? true` arg for crux.rocksdb/->kv-store (when used for the index-store)
- Removal of the `.commit` here requires alignment/resolution with KV store fsync behaviour to avoid mismatch errors following a crash
- The commit would also have to happen before checkpointing (once that's implemented for crux-lucene)
- It's possible that ControlledRealTimeReopenThread can improve performance further when removing .commit, see https://stackoverflow.com/questions/17993960/lucene-4-4-0-new-controlledrealtimereopenthread-sample-usage (EDIT: actually it looks like it only helps with concurrency/consistency challenges we don't have due to tx-time, ~no performance benefit)
- The .maybeRefreshBlocking here will still be required regardless

### DIFF
--- a/crux-lucene/src/crux/lucene.clj
+++ b/crux-lucene/src/crux/lucene.clj
@@ -16,12 +16,13 @@
            [org.apache.lucene.index DirectoryReader IndexWriter IndexWriterConfig Term]
            org.apache.lucene.queries.function.FunctionScoreQuery
            org.apache.lucene.queryparser.classic.QueryParser
-           [org.apache.lucene.search BooleanClause$Occur BooleanQuery$Builder DoubleValuesSource IndexSearcher Query ScoreDoc TermQuery TopDocs]
+           [org.apache.lucene.search BooleanClause$Occur BooleanQuery$Builder DoubleValuesSource IndexSearcher Query ScoreDoc SearcherManager TermQuery TopDocs]
            [org.apache.lucene.store Directory FSDirectory]))
 
-(defrecord LuceneNode [directory analyzer indexer]
+(defrecord LuceneNode [directory analyzer index-writer searcher-manager indexer]
   Closeable
   (close [this]
+    (.close ^IndexWriter index-writer)
     (cio/try-close directory)))
 
 (defn- ^String ->hash-str [eid]
@@ -36,9 +37,8 @@
 (def ^:const ^:private field-crux-val "_crux_val")
 (def ^:const ^:private field-crux-attr "_crux_attr")
 
-(defn- ^IndexWriter index-writer [lucene-store]
-  (let [{:keys [^Directory directory ^Analyzer analyzer]} lucene-store]
-    (IndexWriter. directory, (IndexWriterConfig. analyzer))))
+(defn- ^IndexWriter ->index-writer [^Directory directory ^Analyzer analyzer]
+  (IndexWriter. directory, (IndexWriterConfig. analyzer)))
 
 (defn- index-tx! [^IndexWriter index-writer tx]
   (let [t (Term. "meta" "latest-completed-tx")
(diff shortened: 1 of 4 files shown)


crux-lucene/test/crux/lucene_microbench_test.clj
```diff
@@ -58,3 +58,33 @@
...
+  ((t/join-fixtures [lf/with-lucene-module fix/with-node])
+   (fn []
+     (time
+      (let [last-tx (->> (customers 1500)
+                         (partition-all 1)
+                         (reduce (fn [last-tx chunk]
+                                   (c/submit-tx *api* (vec (for [doc chunk]
+                                                             [:crux.tx/put doc]))))
+                                 nil))]
+        (c/await-tx *api* last-tx)))))
```
jonpither: no parens on their own line!

crux-lucene/src/crux/lucene.clj
```diff
@@ -225,20 +227,28 @@
...
-                    (index! indexer index-writer (db/fetch-docs document-store (:doc-ids ev)))
-                    (when-let [evicting-eids (not-empty (:evicting-eids ev))]
-                      (evict! indexer index-writer evicting-eids))
-                    (index-tx! index-writer (:submitted-tx ev)))))
+                  (index! indexer index-writer (db/fetch-docs document-store (:doc-ids ev)))
+                  (when-let [evicting-eids (not-empty (:evicting-eids ev))]
+                    (evict! indexer index-writer evicting-eids))
+                  (index-tx! index-writer (:submitted-tx ev))
+                  (.commit index-writer)
+                  ;; it's _far_ less than ideal to .commit here per transaction
```
jonpither: I would remove these comments and place them in this GH issue. We don't know there's a valid alternative to calling `commit` on each tx. and if users partition the tx-ops into txes, then it's not potentially an issue at all (subject to your microbench results)

jonpither (changes requested):

crux-lucene/test/crux/lucene_microbench_test.clj
refset: Noted!

crux-lucene/src/crux/lucene.clj
refset: I think the need to avoid a `commit` per tx might be more apparent at much larger scales, and particularly when `:sync false` is set on the KV index-store. The solution shouldn't be too hard - I think we just need some kind of background service thread that calls `commit` followed by `fsync` on the KV protocol.

crux-lucene/src/crux/lucene.clj
```diff
@@ -47,17 +47,21 @@
...
+
 (defn latest-submitted-tx [lucene-store]
-  (let [{:keys [^Directory directory]} lucene-store]
-    (when (DirectoryReader/indexExists directory)
-      (with-open [directory-reader (DirectoryReader/open directory)]
-        (let [index-searcher (IndexSearcher. directory-reader)
-              q (TermQuery. (Term. "meta" "latest-completed-tx"))
-              d ^ScoreDoc (first (.-scoreDocs (.search index-searcher q 1)))]
-          (when d
+  (let [[^IndexSearcher index-searcher index-searcher-release-fn] (->index-searcher lucene-store)
```
jarohen: suspect you want some kind of try-finally here so that the release-fn is called whatever happens in the rest of this function?

jarohen: maybe you want the return from `->index-searcher` to implement `Closeable` so that you can use `with-open`?

jonpither (approved):

crux-lucene/src/crux/lucene.clj
refset: I agree returning something `Closeable` would be nicer but I wasn't certain if it was worth the effort to figure out. I couldn't see a way to wrap this other usage instance inside `cio/->cursor` in a `with-open`: https://github.com/juxt/crux/pull/1468/files#diff-97fb172b1e6d0b14e8f78e6bdcb364d8fef607cf1ad68e42d97e6404744b9e9fR93-R94

Would you call `(.close index-searcher)` there explicitly instead? Or is there some better way to tie the two `Closeable` things together?

> suspect you want some kind of try-finally here so that the release-fn is called whatever happens in the rest of this function?

I'm not sure if anything in the rest of the function is ever going to throw (at least without having much more severe consequences) - are you thinking to avoid possibilities of memory leaks?

jarohen: > I'm not sure if anything in the rest of the function is ever going to throw

Assume it might :smile: We should always take responsibility for closing resources that we own.

> are you thinking to avoid possibilities of memory leaks?

Without knowing the specific implementation details of `IndexSearcher`, I don't _know_ what the impact of not releasing it might be. In though, for example, not closing a snapshot has the effect of preventing Rocks from compacting any transactions older than that snapshot - i.e. the impact of _not_ closing something can be large.

jarohen: In this case, because you're handing off a cursor to your caller, you don't actually want to close the resource in a `finally` (and hence `with-open` is out too), because you're not done with it when the function returns.

```clojure
(defn search [lucene-store, ^Query q]
  (assert lucene-store)
  (let [[^IndexSearcher index-searcher index-searcher-release-fn] (->index-searcher lucene-store)]
    (try
      (let [q (FunctionScoreQuery. q (DoubleValuesSource/fromQuery q))
            score-docs ...]

        (when (seq score-docs)
          (log/debug (.explain index-searcher q (.-doc ^ScoreDoc (first score-docs)))))

        (cio/->cursor (fn []
                        (index-searcher-release-fn))
                      (->> score-docs
                           (map (fn [^ScoreDoc d]
                                  (vector (.doc index-searcher (.-doc d))
                                          (.-score d)))))))
      (catch Throwable t
        (index-searcher-release-fn)
        (throw t)))))
```

jarohen (changes requested):
(can't delete this comment :roll_eyes:)

crux-lucene/src/crux/lucene.clj
refset: I've now added the try+catch, thanks for the feedback!

For posterity: we also discussed the merits of creating a `IndexSearcherWrapper` record that implements `closeable`...but it's probably overkill :)

jarohen (approved):

jarohen (approved):

PR #1469, opened 2021-03-23
Second Chance Cache Downsizing

hraberg:
Shrink down cache again if it gets small to avoid having to scan a large empty map when finding random entries to evict.
ConcurrentHashMap doesn't itself seem to shrink the array when elements are removed from the map.

### DIFF
--- a/crux-core/src/crux/cache/second_chance.clj
+++ b/crux-core/src/crux/cache/second_chance.clj
@@ -33,10 +33,8 @@
   (^java.util.Map getHot [])
   (^void maybeResizeCache []))
 
-(def ^:private ^:const ^{:tag 'double} resize-load-factor 0.01)
-
 (deftype SecondChanceCache [^:unsynchronized-mutable ^ConcurrentHashMap hot ^Queue cooling ^double cooling-factor ^ICache cold
-                            ^long size adaptive-sizing? ^double adaptive-break-even-level
+                            ^long size adaptive-sizing? ^double adaptive-break-even-level ^double downsize-load-factor
                             ^Semaphore resize-semaphore]
   Object
   (toString [_]
@@ -49,7 +47,7 @@
   (maybeResizeCache [this]
     (when-not (.isEmpty hot)
       (when-let [table (ConcurrentHashMapTableAccess/getConcurrentHashMapTable hot)]
-        (when (< (double (/ (.size hot) (alength table))) resize-load-factor)
+        (when (< (double (/ (.size hot) (alength table))) downsize-load-factor)
           (set! (.hot this) (doto (ConcurrentHashMap. 0)
                               (.putAll hot)))))))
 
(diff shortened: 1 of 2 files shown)


jarohen (approved):

jarohen:
Assume @refset wants to check this against his recent benchmarks before we merge?

hraberg:
Yes, it would be great if @refset could see if it works for his test, this was honestly written in the blind a bit.

refset:
Unfortunately I am still finding this change to be ~6x slower than when switching the default in cache.clj to LRU. This is my test code (not very minimal, sorry!):
```clojure
  ;; five-clique-q

  (defn create-docs [id-key haystack? offset n]
    (apply concat (for [i (map #(- (- 0 %) offset 1) (range n))]
                    (let [e (* i 5)
                          d (inc e)
                          c (inc d)
                          b (inc c)
                          a (inc b)]
                      [{id-key a
                        :a true
                        :ref [b c d e]}
                       {id-key b
                        :ref [c d e]}
                       {id-key c
                        :ref [d e]}
                       {id-key d
                        :ref (if haystack? [c] [e])}
                       {id-key e
                        :e true}]))))

  (defn haystack [id-key needles size]
    (concat (create-docs id-key false 0 needles)
            (create-docs id-key true (* needles 3) size)))

  (def shape-q '{:find [?a ?b ?c ?d ?e]
                 :timeout 90000
                 :where [[?a :a true]
                         [?a :ref ?b]
                         [?a :ref ?c]
                         [?a :ref ?d]
                         [?a :ref ?e]
                         [?b :ref ?c]
                         [?b :ref ?d]
                         [?b :ref ?e]
                         [?c :ref ?d]
                         [?c :ref ?e]
                         [?d :ref ?e]]})

  (def my-node (let [n (crux/start-node {})]
                 (->> (last (doall (for [t (partition-all 1000 (for [doc (haystack :crux.db/id 4 20000)]
                                                                 [:crux.tx/put doc]))]
                                     ( crux/submit-tx n t))))
                      (crux/await-tx n))
                 n))

  (with-open [db (crux/open-db my-node)]
    (time (crux/q db shape-q))
    (time (crux/q db shape-q))
    (time (crux/q db shape-q)))
```

hraberg:
Not sure if it means anything, but the cache that this happens for in the provided query is the per-query entity resolver cache, which is short-lived. So one workaround is to replace this one with a LRU cache, but obviously, the underlying issue can still happen in the general case.

refset:
## Looking once again at behaviour on `master`

In my efforts to construct a microbench namespace for these cyclic queries I've discovered that a very subtle change to the above example code reliably _doesn't_ trigger the same `computeIfAbsent` thrashing(?) effect with the second-chance cache, specifically switching to use string IDs instead of the negative integers (they were negative to help interop with DataScript's tempids):
```clojure
(defn create-docs [id-key haystack? offset n]
    (apply concat (for [i (map #(- (- 0 %) offset 1) (range n))]
                    (let [e (* i 5)
                          d (inc e)
                          c (inc d)
                          b (inc c)
                          a (inc b)
                          e (str e)
                          d (str d)
                          c (str c)
                          b (str b)
                          a (str a)]
                      [{id-key a
                        :a true
                        :ref [b c d e]}
                       {id-key b
                        :ref [c d e]}
                       {id-key c
                        :ref [d e]}
                       {id-key d
                        :ref (if haystack? [c] [e])}
                       {id-key e
                        :e true}]))))
```

Could there be something problematic with the sorting/scanning order of integer entity IDs in the cache? I tried doing `(* id -1)` to invert the negative integers to be positive, but I don't think that made any appreciable difference.

The contrast of the effect is especially clear when reducing the default cache size to be very small, e.g. 100 times smaller.

## Changes from https://github.com/juxt/crux/pull/1469/commits/32a59116a8c607810487c04a76ef596af21c4246

The PR change definitely helps the integer example code, which is good, however if I again reduce the default cache size by 100x then the `cav_cache_lookup` (which is still using second-chance) locks up with this same thrashing behaviour. If I then change the query to the string version then all is happy again.

I've also noticed there seems to be a fairly dramatic threshold for triggering the effect. Like it won't be noticeable at all with haystack@15k, but the query could be 5x slower at @16k due to the thrashing `computeIfAbsent` second-chance calls.

In summary, I believe the PR will help reduce the surface area and likelihood for triggering the behaviour, by limiting usage of second-chance, but the issue can still be reproduced if I reduce the second-chance cache size, or presumably (I haven't measured it enough), if I proportionally increase the size of the data set.

I would be happy to see the PR merged regardless, but perhaps my analysis offers some further clues.

## Query timeout interrupt

As a tangential note, this `(.interrupt caller-thread)` query timeout: https://github.com/juxt/crux/blob/6d602bb5b6caed199f10fd8c3711cb034d49248a/crux-core/src/crux/query.clj#L1798 ...doesn't seem to have any effect during this thrashing second-chance behaviour. Could there be an easy fix for that?

refset (approved):

Issue #1105, opened 2020-09-16
Handling document stores being eventually consistent


jarohen:
We currently index documents once when the transaction arrives on the tx-log. This isn't usually a problem, because documents are submitted first, before the transaction - but, theoretically, if the document store is eventually consistent, the document may not be available in the document store by the time the transaction is received.

We should investigate making the `fetch-docs` implementation of other document stores behave like Kafka, which waits for all documents to be present before returning.

See discussion in #1054

jarohen:
As discussed, let's block on ingest, essentially moving the Kafka logic higher up for ingest only

PR #1470, opened 2021-03-23
Fix doc-store hanging on missing doc, #1387

### LINKED ISSUE #1387: If, for whatever reason, the doc store cannot find a doc, or the doc is corrupted, Crux ingestion is blocked, and queries time out
Introduced after changes to support eventually consistent doc stores. Need to better consider what options we can give the user in this case.

See https://clojurians-log.clojureverse.org/crux/2021-01-13/1610497911.043400 from @nivekuil, and #1378

jarohen:
Having the tx-ingester wait for documents to be returned from `fetch-docs` caused the tx-ingester to hang indefinitely if (for whatever reason) the requested document wasn't present in the doc-store. This was introduced with eventually-consistent doc-store support in #1105.

* Instead, we change the contract of `db/submit-docs` s.t. once it returns successfully, docs are guaranteed to be fetchable through `db/fetch-docs`. S3 is now [strongly consistent](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html#ConsistencyModel), as is [Azure Blob Storage](https://docs.microsoft.com/en-us/azure/storage/blobs/concurrency-manage?tabs=dotnet). 
* ~TODO: confirm behaviour of KV doc-store - do we need to mandate fsyncing here?~ see #1471 
* The tx-ingester now halts on missing docs. I don't think there's a _good_ solution here - now that we've changed the above consistency contract, the only reason I can see that we'll get into this situation is a breakdown in the _durability_ contract of the doc-store.
* The Kafka doc-store is still a nuisance here - docs aren't fetchable until they've been indexed locally, so the Kafka doc-store `fetch-docs` takes on the previously general polling logic. This means that the Kafka doc-store _can_ hang indefinitely in this case.

See https://github.com/orgs/juxt/teams/crux-core/discussions/25


crux-kafka/src/crux/kafka.clj
```diff
@@ -245,8 +248,23 @@
   (submit-docs [this id-and-docs]
     (submit-docs id-and-docs this))
 
-  (-fetch-docs [this ids]
-    (db/-fetch-docs local-document-store ids)))
+  (fetch-docs [_ ids]
+    ;; TODO will hang forever if the requested ids aren't ever submitted
```
jonpither: make a subsequent issue and link to?

jonpither (approved):

crux-kafka/src/crux/kafka.clj
jarohen: good shout - #1479

hraberg (approved):

PR #1471, opened 2021-03-23
fsyncing KV tx-log and document-store by default

jarohen:


hraberg (approved):
You can do this by configuring the kv-store itself to be fsync, but get that it's harder to get it to default it to that.

PR #1472, opened 2021-03-23
Adding attribute distinct cardinalities using HyperLogLog statistics, #1296

### LINKED ISSUE #1296: Improve the join order with index stats
- keeping stats on the number of distinct values for an attribute

jarohen:
resolves #1296 

We now keep [HyperLogLog](https://en.wikipedia.org/wiki/HyperLogLog) buffers for each attribute on ingestion so that we know, for each attribute, how many distinct values and distinct entity ids we have. This is predominantly to fix the query join order in the case where two attributes have similar doc-counts but very different selectivities. Previously, when a user is filtering both by `:type` (say only a handful of different types) and `:name` (assuming mostly unique between entities), Crux wouldn't necessarily choose to filter first by `:name`. See `picks-more-selective-join-order` test. 

See https://github.com/orgs/juxt/teams/crux-core/discussions/22

Benchmarks-wise, this currently adds 4-5% to ingest (makes sense, we're doing more work, although we may be able to shave off some time), and no appreciable difference in Watdiv nor TPC-H queries.
* NOTE - we could consider caching the stats KVs between transactions so that we don't have to go look them up. This may not be that much of a performance improvement given Rocks et al will likely cache recently changed/accessed KVs in memory. In practice, it seems that the hashing of the values for the HLL dominates this time.
* NOTE - the cache key for `compile-sub-query` doesn't take the relative attribute cardinalities into account. if these change significantly, the cache won't get invalidated, and we may have sub-optimal queries.

### DIFF
--- a/crux-core/src/crux/fork.clj
+++ b/crux-core/src/crux/fork.clj
@@ -77,11 +77,9 @@
                            capped-valid-time
                            capped-tx-id))
 
-  (attribute-cardinalities [_]
-    (db/attribute-cardinalities index-snapshot))
-
-  (attribute-cardinality [_ attr]
-    (db/attribute-cardinality index-snapshot attr))
+  (attribute-stats [_] (db/attribute-stats index-snapshot))
+  (attribute-doc-count [_ attr] (db/attribute-doc-count index-snapshot attr))
+  (attribute-cardinality [_ attr] (db/attribute-cardinality index-snapshot attr))
 
   db/IndexMeta
   (-read-index-meta [_ k not-found]
@@ -156,13 +154,17 @@
                            (db/open-nested-index-snapshot transient-index-snapshot)
                            evicted-eids))
 
-  (attribute-cardinalities [_]
-    (merge (db/attribute-cardinalities persistent-index-snapshot)
-           (db/attribute-cardinalities transient-index-snapshot)))
+  (attribute-stats [_]
+    (merge (db/attribute-stats persistent-index-snapshot)
+           (db/attribute-stats transient-index-snapshot)))
+
+  (attribute-doc-count [_ attr]
+    (or (db/attribute-doc-count transient-index-snapshot attr)
+        (db/attribute-doc-count persistent-index-snapshot attr)))
 
   (attribute-cardinality [_ attr]
-    (or (db/attribute-cardinality persistent-index-snapshot attr)
-        (db/attribute-cardinality transient-index-snapshot attr)))
+    (or (db/attribute-cardinality transient-index-snapshot attr)
+        (db/attribute-cardinality persistent-index-snapshot attr)))
 
   db/IndexMeta
   (-read-index-meta [_ k not-found]
--- a/crux-core/src/crux/fork.clj
+++ b/crux-core/src/crux/fork.clj
@@ -77,6 +77,12 @@
                            capped-valid-time
                            capped-tx-id))
 
+  (attribute-cardinalities [_]
+    (db/attribute-cardinalities index-snapshot))
+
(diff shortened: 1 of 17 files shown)


crux-core/src/crux/fork.clj
```diff
@@ -150,6 +155,22 @@
                            (db/open-nested-index-snapshot transient-index-snapshot)
                            evicted-eids))
 
+  (all-attrs [_]
+    (set/union (db/all-attrs persistent-index-snapshot)
+               (db/all-attrs transient-index-snapshot)))
+
+  (doc-count [_ attr]
+    (or (db/doc-count transient-index-snapshot attr)
```
hraberg, line 165: Think this does want to be addition?

crux-core/src/crux/fork.clj
```diff
@@ -150,6 +155,22 @@
...
+        (db/doc-count persistent-index-snapshot attr)))
+
+  (value-cardinality [_ attr]
+    (or (db/value-cardinality transient-index-snapshot attr)
```
hraberg, line 169: This should in theory be `hll/estimate-union` but as that is likely hidden at this level, it can maybe be `max`?

crux-core/src/crux/query.clj
```diff
@@ -1388,7 +1387,7 @@
                                                                        var->joins
                                                                        type->clauses
                                                                        in-vars
-                                                                       stats)
+                                                                       db)
```
hraberg: Bit unsure here, so likely fine, but the entire idea of the encode-value-fn is to clearly separate compile time from runtime of the query, and not use the index-snapshot during the former.

hraberg (commented):
LGTM on a glance, I would probably test caching of the HLL to see what, if any, benefit it has during ingest.

Also, see note about the compile time  / runtime separation, not clear to me if there's any real issue here the way its used.

An idea I had earlier was to cache the relative cardinalities used in the query as part of the cache key, so if they change, the query plan would get regenerated. But that's not how it works now, so not part of this PR.

crux-core/src/crux/fork.clj
jarohen: Don't _think_ so - the attribute HLL gets copied into the transient store, so it completely replaces the one in the persistent store?

crux-core/src/crux/fork.clj
jarohen: likewise?

crux-core/src/crux/query.clj
jarohen: Ah, ok, thanks - seems a quick chat will be best to resolve this one?

crux-core/src/crux/fork.clj
hraberg: Ah, ignore me then, didn't follow it through, just assumed that the transient one was an empty, local one.

crux-core/src/crux/query.clj
jarohen: TODO: hide this behind a wrapper for now so that it's obvious that the HLL provider shouldn't be passed through to query runtime.

jarohen: done :smile:

jarohen:
@hraberg - attacked this with a profiler this afternoon, seems like the looking up of the HLL buffers pales into insignificance compared to constructing the hashes, and (due to the two index snapshots) caching this is non-trivial. In profiling, however, I did  find that I was constructing the attr keys for every doc - adding a `(distinct)` call here per doc batch halved the overhead of HLL on 
a ts-devices microbench :smile:

PR #1473, opened 2021-03-23
MemKv records to `deftype`s

jarohen:
Making MemKv records `deftype`s so that their hashcodes are based on identity rather than content, fixes performance issue raised by @refset with IndexStore's thread manager storing these in a HashMap.


jarohen:
merged, c1c936d9

PR #1475, opened 2021-03-24
Less fsyncing in Rocks/LMDB

jarohen:
We don't need to fsync in Rocks if sync is already set or the WAL's enabled - this change brings the Rocks ingest benchmarks back into similar territory than they were before #1471. 

Likewise, in LMDB, we don't need to manually sync if `:sync? true` is already set.

This arguably makes `fsync` mis-named, we ideally want to convey something along the lines of 'ensure everything written thus far is made durable', but that seemed a bit of a mouthful.

### DIFF
--- a/crux-lmdb/src/crux/lmdb.clj
+++ b/crux-lmdb/src/crux/lmdb.clj
@@ -214,7 +214,7 @@
 (def ^:dynamic ^{:tag 'long} *mapsize-increase-factor* 1)
 (def ^:const max-mapsize-increase-factor 32)
 
-(defrecord LMDBKv [db-dir env env-flags dbi ^StampedLock mapsize-lock]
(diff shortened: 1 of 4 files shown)


refset:
`commit` is probably a good candidate, to align with both Kafka:
https://www.confluent.io/blog/transactions-apache-kafka/
> Marking an offset as consumed is called committing an offset. In Kafka, we record offset commits by writing to an internal Kafka topic called the offsets topic. A message is considered consumed only when its offset is committed to the offsets topic

And Lucene:
http://blog.mikemccandless.com/2012/03/transactional-lucene.html
> Durability: once commit returns, all changes have been written to durable storage (assuming your I/O system correctly implements fsync). If the computer or OS crashes, or the JVM crashes or is killed, or power is lost to the computer, all changes will still be present in the index.

hraberg (approved):

PR #1476, opened 2021-03-24
Asynchronousification of TxLog and DocStore

jarohen:
Based on `less-fsyncing` and `doc-store-hanging-1387`, see https://github.com/jarohen/crux/compare/doc-store-hanging-1387...async-things for real diff

Changing the three functions on doc-store and tx-log to be `-async`, using patterns established in other ongoing spikes.  
- `submit-tx` was already async, this is just a rename to `submit-tx-async` and returning a `CompletableFuture`.
- updated ts-devices and ts-weather benchmarks to use asynchronous ingestion to test performance, bench ongoing (TODO)
- third-party implementations of these two will be required to migrate - at a minimum, wrapping their existing implementations in `(CompletableFuture/completedFuture ...)`

TODO
- [ ] post bench results
- [x] need to rebase this on #1470, suspect a few conflicts
- [ ] in the KV implementations, suspect this may even reduce performance unless we queue the actual writes into a single thread - both Rocks and LMDB benefit from single-threaded batch ingest.
- [ ] we may well not want to use the default asynchronous executor throughout - the components may benefit from having their own pools

### DIFF
--- a/crux-azure-blobs/src/crux/azure/blobs.clj
+++ b/crux-azure-blobs/src/crux/azure/blobs.clj
@@ -1,10 +1,13 @@
 (ns crux.azure.blobs
   (:require [clj-http.client :as http]
-            [taoensso.nippy :as nippy]
             [crux.db :as db]
+            [crux.document-store :as ds]
             [crux.io :as cio]
             [crux.system :as sys]
-            [crux.document-store :as ds]))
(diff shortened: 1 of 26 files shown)


jarohen:
out of date now, closing - will have another go another day

PR #1477, opened 2021-03-25
Cursors can also implement IReduceInit so that people can transduce over them

jarohen:
Quick hack from @danmason and I this morning. Spent 5 minutes figuring out whether we could, haven't yet fully considered whether we _should_ :smile:


jarohen:
We'd want the `reduce` function itself to open the query, would require a separate API. Closing for now, can always re-open later.

PR #1483, opened 2021-03-30
Fix unbundled architecture image in README

jonassvalin:


jarohen:
Cheers @jonassvalin! :pray:

PR #1495, opened 2021-04-06
remove `full-results?`

### LINKED ISSUE #1317: Remove `full-results?`
Depends on `pull`. 

Given we're renaming eql/project (see #1220), let's release that in its 'final' form, and then kill this in the subsequent release.

jarohen:
resolves #1317.

I've left a defensive check in so that users' queries throw an error rather than silently returning different results - RFC?


jarohen:
cc @johantonelli - I've migrated the examples to use `(pull e [*])` rather than `:full-results? true` but suspect there may also need to be updates made elsewhere?

PR #1499, opened 2021-04-14
Add cyclic queries microbench

refset:
I'm looking for feedback on whether/how to integrate this ns further, e.g. add it to the nightly runs

Cyclic queries highlight the differences between Crux's WCOJ multi-way join execution and clause-by-clause hash-join execution (i.e. total query time and peak memory usage).

### DIFF
--- a/crux-bench/src/crux/bench/cyclic_microbench.clj
+++ b/crux-bench/src/crux/bench/cyclic_microbench.clj
@@ -1,0 +1,168 @@
+(ns crux.bench.cyclic-microbench
+  (:require [crux.api :as crux]
+            [crux.bench :as bench]
+            [datomic.api :as d])
+  (:import java.time.Duration))
+
+(def ^:dynamic *needle-count* 4)
+(def ^:dynamic *haystack-size* 10000)
+
+(defn haystack [create-docs-fn id-key needles size]
+  (concat (create-docs-fn id-key false 0 needles)
+          (create-docs-fn id-key true needles size)))
+
+(defn create-triangle-docs [id-key haystack? offset n]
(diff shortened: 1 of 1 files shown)


jarohen:
@refset: if it's something that might realistically get broken by a future change, let's add a test. if it's something that may become less performant in a future change, we can add it to the bench, if a bench run would be accurate enough to spot a regression? if not, no harm in keeping it as a micro-bench, as-is, for use in local development on such changes.

refset:
> if it's something that might realistically get broken by a future change

In the spirit of not adding new "slowdown" tests I don't think we'd want to apply it as a regular test, and it doesn't exercise any user-facing query features not already covered extensively in query_test.clj

> if it's something that may become less performant in a future change, we can add it to the bench

This is reasonably likely and I think it would be useful for spotting a regression by itself. However, I'd be surprised if it picked up a regression that didn't show up clearly in WatDiv also, so the value in promoting it would probably be limited. Admittedly, I haven't got my head around the effort to extend the nightly bench (dashboards, emails etc.), so if it only takes 30m then its probably a no-brainer.

> if not, no harm in keeping it as a micro-bench, as-is, for use in local development on such changes.

I expect it will help us most when reasoning about any changes involving multi-way join execution specifically. For example, if we were to move to a different relational execution model that treated multi-way joins as an option rather than the universal default (or even removes WCOJ multi-way joins altogether). I don't believe WatDiv, at the scale we typically run it, would help us in quite the same way, or at least the impact would be obscured (and I'm not certain that any of the WatDiv queries are actually cyclic in the first place).

jarohen:
Ok - in that case, let's at least merge this as a microbench for now :smile:

Issue #1426, opened 2021-02-17
Asychronous indexing for crux-lucene


refset:
The behaviour of crux-lucene currently relies on synchronous indexing alongside the main KV index store, where both stores are updated at the ~same time within the ingestion thread. This is a pragmatic design and the strong consistency behaviour is often desirable, but it limits the maximum ingestion throughput.

Asynchronous indexing could enable:
- (re)building of search indexes without also requiring rebuilding of the main KV index store
- delayed local I/O to avoid contention with the indexing thread
- use of networked/distributed search indexes

New API options would be necessary to allow the user to specify whether a query should:
1. tolerate stale Lucene indexes if there is a lag (which is likely), or
2. block until the index is sufficiently caught up (subject to a timeout), or
3. abort entirely if there is any lag (perhaps within some bound?)

The pattern for this could inform other kinds of secondary indexes in future, and might even be relevant when thinking about the possibilities for multi-instance queries.

nivekuil:
+1, one use case that would be unlocked here is adding additional fields to the Lucene doc based off of out-of-band information (e.g. popularity). And as an example of one of those other kinds of secondary indexes, one might want to keep track of aggregations like association counts, stored as some sketch data structure (https://datasketches.apache.org/) while making use of other crux facilities like checkpoints.

Issue #1221, opened 2020-11-03
Lucene-store is compatible with checkpointing


jonpither:
Relates to #1094 

The checkpoints for Lucene and the KV need not be synchronous/atomic as Lucene indexing is (in principle) idempotent, and so for a given checkpoint pair, the Lucene checkpoint just needs to be more recent (but with as small a delay as possible).

PR #1501, opened 2021-04-16
Remove unnecessary forced commit call to index writer

### LINKED ISSUE #1500: Lucene indexing process abysmally slow due to unnecessary commit call

kevinmershon:
fixes https://github.com/juxt/crux/issues/1500

before:
image

after:
image

### DIFF
--- a/crux-lucene/src/crux/lucene.clj
+++ b/crux-lucene/src/crux/lucene.clj
@@ -248,6 +248,5 @@
                     (when-let [evicting-eids (not-empty (:evicting-eids ev))]
                       (evict! indexer index-writer evicting-eids)))
                   (index-tx! index-writer (:submitted-tx ev))
-                  (.commit index-writer)
                   (.maybeRefreshBlocking searcher-manager)))
     lucene-store))


kevinmershon:
It looks like there may need to be additional work using [throttler](https://github.com/brunoV/throttler) or a similar approach so that ~5-10 seconds after the last call to the bus/listen fn, a commit write happens.

jarohen:
Hey @kevinmershon, thanks for the PR :pray: 

As mentioned on Slack, I'm afraid I don't think we'll be able to make that exact fix though, due to a current constraint that the main query indices and Lucene stay in sync, even if the power cord's pulled out - if not, Crux unfortunately fails on next startup.
We're looking at ways to relax this constraint in future releases - best that I defer to @jonpither and @refset who are a little more in context :)

refset:
Hi :) to add context to what @jarohen just wrote, you can see my fairly recent investigation and write-up about this on the `SearcherManager` PR: https://github.com/juxt/crux/pull/1468#issue-596668106

`crux-lucene` doesn't currently support the online checkpointing feature of the main index-store (see https://github.com/juxt/crux/issues/1221), which means it is especially important that a node can routinely restart after a crash, unless:
1) you roll your own offline index backup & restore mechanism, or
2) you can cope with the delay of replaying the tx-log from scratch

...only in those circumstances does removing the `commit` feel like a practical option.

Admittedly 1) is somewhat unavoidable in serious deployments as things stand today - if you want to use `crux-lucene` _and_ be able to quickly spin up new nodes from index backups, then you have to `.close` a given node before copying the kv + Lucene directories, etc.

In any case, leaving the `commit` where it is seems to be the best compromise for the moment, until either https://github.com/juxt/crux/issues/1221 or https://github.com/juxt/crux/issues/1426 is resolved. Solving both of those is a high priority for us though, so you shouldn't have to wait ~long.

An interim fix / improvement over the current situation could be to skip the `commit` calls only whilst a node is undergoing it's initial `sync`, but we don't have anything else that works like that so it would require some thinking and design work. I believe this is more or less what you were thinking too when you wrote the following on Slack:
> At minimum on the crux side it should "batch" transactions with lucene and flush only once caught up. When realtime you can reasonably ask it to flush to disk every tx and ask for the callers to use bulkier transactions where possible

You can of course also fork `crux-lucene` for the time being, which I know isn't ideal, but might be the simplest thing for your prototyping timelines.

If you would like to discuss timelines or chat these options through on a call, please do get in touch - my email is jdt@juxt.pro (or DM me on Slack / Zulip)

Thank you for your report, it is very useful to know that this is affecting users in the wild, and I'm sorry you hit this snag!

kevinmershon:
>You can of course also fork crux-lucene for the time being, which I know isn't ideal, but might be the simplest thing for your prototyping timelines.

Exactly what I have done for the time being. :)

refset:
That's great, I will close this PR for now then and leave the issue open. Thanks in advance for your patience!

PR #1507, opened 2021-04-29
Re-enable :in binding args for crux-http-server

### LINKED ISSUE #1506: crux-http-server fails to handle :in parameters
From: https://clojurians.slack.com/archives/CG3AM2F7V/p1619629553259200

I can reproduce Tomas Brejla's test case below. It does appear to fail on 21.04-1.16.0-beta.

```clojure
(ns tut
  (:require [crux.api :as crux]))

(def crux-via-api-client (crux/new-api-client "http://localhost:3000"))

(def crux-inmemory-node
  (crux/start-node {}))

;; store sample account in in-memory node
(crux/submit-tx
 crux-inmemory-node
 [[:crux.tx/put
   {:crux.db/id :account/joe
    :account/id :account/joe
    :balance 100000M
    :person :person/joe}]])
;; => #:crux.tx{:tx-id 1, :tx-time #inst "2021-04-28T16:45:58.339-00:00"}

;; store the same account in standalone node behind HTTP client
(crux/submit-tx
 crux-via-api-client
 [[:crux.tx/put
   {:crux.db/id :account/joe
    :account/id :account/joe
    :balance 100000M
    :person :person/joe}]])
;; => #:crux.tx{:tx-id 18, :tx-time #inst "2021-04-28T16:46:23.794-00:00"}


;; retrieve the account using in-memory node
(crux/q
 (crux/db crux-inmemory-node)
 '{:find [(pull e [*])]
   :where [[e :crux.db/id :account/joe]]})
;; => #{[{:crux.db/id :account/joe, :account/id :account/joe, :balance 100000M, :person :person/joe}]}

;; retrieve the account using standalone node behind HTTP client
(crux/q
 (crux/db crux-via-api-client)
 '{:find [(pull e [*])]
   :where [[e :crux.db/id :account/joe]]})
;; => #{[{:crux.db/id :account/joe, :account/id :account/joe, :balance 100000M, :person :person/joe}]}

;; hmm... I can grab the data from both in-memory and via http client.. so far so good...


;; retrieve the account using :in works just fine when using in-memory node
(crux/q
 (crux/db crux-inmemory-node)
 '{:find [(pull e [*])]
   :where [[e :crux.db/id ?account-id]]
   :in [?account-id]}
 :account/joe)
;; => #{[{:crux.db/id :account/joe, :account/id :account/joe, :balance 100000M, :person :person/joe}]}

;; the same thing (using :in) yields no results when using http client node
(crux/q
 (crux/db crux-via-api-client)
 '{:find [(pull e [*])]
   :where [[e :crux.db/id ?account-id]]
   :in [?account-id]}
 :account/joe)
;; => #{}

;; Why am I not able to get the result when using `:in`? The same stuff works for inmemory node. And the data is actually there on the node behind remote http api.

```

refset:
This issue was reported via #1506 and here is the background:
1. Support for `:in` bindings, including use via `crux-http-server`, was added correctly in PR https://github.com/juxt/crux/pull/1099, specifically https://github.com/juxt/crux/pull/1099/commits/da33146d22262d426cb906ab9547fc54ded43354#diff-34510033ce6b20631fcf93dce84fa0a0140be6828350f2ad0bdc72c8d47ad872L125-R127
2. Unfortunately however, this PR only added `:in` test coverage to `query_test.clj`, and not `api_test.clj` (which is run against multiple API implementations by default, including `:remote`)
3. The subsequent commit to `http_server.clj` was part of a very large PR and accidentally broke the use of `:in` args for `crux-http-server`, presumably whilst consolidating `q`/`open-q` calls, e.g. there should probably have been references to `args` somewhere around here https://github.com/juxt/crux/commit/e7e3b1c5f83c89804151dbd40370c050bc18f7b0#diff-e80d14b2675646aa754c45c247fbf90a0a4652097ea0e03cec17e586c64d6c7fR137-R146

The fix as implemented in this PR isn't ideal. There is potentially some much cleaner Reitit/Ring and/or `spec-tools` way to handle coercion of body params, e.g. something like this:
```clojure
;; unfortunately neither the inline spec conformer or st :decode/string are sufficient, therefore try-decode-edn coercion also takes place data-browser-query
(s/def ::args
  (st/spec
   {:spec
    (s/and
     (s/conformer (fn [x] (util/try-decode-edn x)))
     (s/coll-of any? :kind vector?))
    :swagger/example (pr-str '["foo" 123])
    :description "EDN formatted :in binding arguments"
    :decode/string (fn [_ a] (util/try-decode-edn a))}))
```

In general it seems best to use these tools for such coercions, but the obvious configurations haven't helped in this case. We may well want to change strategy in this area altogether before worrying about finding a cleaner resolution here.

Whilst investigating and fixing this I noted some outstanding questions about crux-http-server:
1. Should the JSON-over-HTTP `/_crux/query` calls be non-lazy by default? The bag semantics are undocumented and likely to be surprising, and the deduplication should probably happen on the server. Replacing https://github.com/refset/crux/commit/2bfab4838856411c8b0af31ab1225d7278c5e3d4#diff-e80d14b2675646aa754c45c247fbf90a0a4652097ea0e03cec17e586c64d6c7fR76 with `(cio/->cursor (fn []) (apply crux/q db query in-args))` will demonstrate this alternative behaviour
2. Should `query` be coerced/normalised before hitting `crux/q`? Currently it is being passed as a string
3. Could it be sensible to differentiate between `args` and `args-edn`/`args-json`? i.e. similarly to `query-edn`. Are there some tests we could add to explore this?

Additionally, could/should we run `query_test.clj` against all API implementations also?

(Thank you again @danmason for the pairing and input!)

### DIFF
--- a/crux-http-client/src/crux/remote_api_client.clj
+++ b/crux-http-client/src/crux/remote_api_client.clj
@@ -91,7 +91,8 @@
                                            (= :stream (:as http-opts)) slurp))]
          (throw (case (::err/error-type error-data)
                   :illegal-argument (err/illegal-arg (::err/error-key error-data) error-data)
-                  :node-out-of-sync (err/node-out-of-sync error-data))))
+                  :node-out-of-sync (err/node-out-of-sync error-data)
+                  (ex-info "Generic remote client error" error-data))))
 
        (and (<= 200 status) (< status 400))
        (if (string? body)
@@ -130,20 +131,20 @@
                                                         {:eid-edn (pr-str eid)})}
                        :->jwt-token ->jwt-token}))
 
-  (q* [this query args]
-    (with-open [res (api/open-q* this query args)]
+  (q* [this query in-args]
+    (with-open [res (api/open-q* this query in-args)]
       (if (:order-by query)
         (vec (iterator-seq res))
         (set (iterator-seq res)))))
 
-  (open-q* [this query args]
+  (open-q* [this query in-args]
     (let [in (api-request-sync (str url "/_crux/query")
                                {:->jwt-token ->jwt-token
                                 :http-opts {:as :stream
                                             :method :post
                                             :query-params (temporal-qps this)}
                                 :body {:query (pr-str query)
-                                       :args (vec args)}})]
+                                       :in-args (vec in-args)}})]
       (cio/->cursor #(.close ^Closeable in) (edn-list->lazy-seq in))))
 
   (pull [this projection eid]
--- a/crux-http-server/src/crux/http_server/query.clj
+++ b/crux-http-server/src/crux/http_server/query.clj
@@ -33,15 +33,34 @@
     :description "EDN formatted Datalog query"
     :decode/string (fn [_ q] (util/try-decode-edn q))}))
 
+(s/def ::in-args
+  (st/spec
+   {:spec (s/coll-of any? :kind vector?)
+    :swagger/example ["foo" 123]
+    :description ":in binding arguments"}))
+
+(s/def ::in-args-edn
+  (st/spec
+   {:spec ::in-args
+    :swagger/example (pr-str '["foo" 123])
+    :description "EDN formatted :in binding arguments"
+    :decode/string (fn [_ a] (util/try-decode-edn a))}))
+
+(s/def ::in-args-json
+  (st/spec
+   {:spec ::in-args
+    :swagger/example (http-json/write-str '["foo" 123])
+    :description "JSON formatted :in binding arguments"
+    :decode/string (fn [_ a] (http-json/try-decode-json a))}))
+
 ;; TODO: Need to ensure all query clauses are present + coerced properly
 (s/def ::query-params
-  (s/keys :opt-un [::util/valid-time ::util/tx-time ::util/link-entities? ::query-edn]))
-
-(s/def ::args (s/coll-of any? :kind vector?))
+  (s/keys :opt-un [::util/valid-time ::util/tx-time ::util/link-entities?
+                   ::query-edn ::in-args-edn ::in-args-json]))
 
 (s/def ::body-params
   (s/keys :req-un [::query]
-          :opt-un [::args]))
+          :opt-un [::in-args]))
 
 (defn with-entity-refs
   [results db]
@@ -55,16 +74,16 @@
                              (cond-> el
                                (get entity-links el) (entity-ref/->EntityRef))))))))))
 
-(defn run-query [{:keys [link-entities? query]} {:keys [crux-node valid-time tx-time]}]
+(defn run-query [{:keys [link-entities? query]} in-args {:keys [crux-node valid-time tx-time]}]
   (let [db (util/db-for-request crux-node {:valid-time valid-time
                                            :tx-time tx-time})]
     {:query query
      :valid-time (crux/valid-time db)
      :tx-time (crux/transaction-time db)
      :results (if link-entities?
-                (let [results (crux/q db query)]
+                (let [results (apply crux/q db query in-args)]
                   (cio/->cursor (fn []) (with-entity-refs results db)))
-                (crux/open-q db query))}))
+                (apply crux/open-q db query in-args))}))
 
 (defn- ->*sv-encoder [{:keys [sep]}]
   (reify mfc/EncodeToOutputStream
@@ -174,11 +193,13 @@
 (defn data-browser-query [options]
   (fn [req]
     (let [{query-params :query body-params :body} (get-in req [:parameters])
-          {:keys [valid-time tx-time query-edn]} query-params
-          query (or query-edn (get body-params :query))]
+          {:keys [valid-time tx-time query-edn in-args-edn in-args-json]} query-params
+          query (or query-edn (get body-params :query))
+          in-args (or in-args-edn in-args-json (get body-params :in-args))]
       (-> (if (nil? query)
             (assoc options :no-query? true)
             (run-query (transform-req query req)
+                       in-args
                        (assoc options
                               :valid-time valid-time
                               :tx-time tx-time)))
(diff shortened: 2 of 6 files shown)


crux-http-server/src/crux/http_server/query.clj
```diff
@@ -33,11 +33,20 @@
     :description "EDN formatted Datalog query"
     :decode/string (fn [_ q] (util/try-decode-edn q))}))
 
+;; unfortunately neither the inline spec conformer or st :decode/string are sufficient, therefore try-decode-edn coercion also takes place data-browser-query
+(s/def ::args
```
danmason: Not sure I'd personally include the `:args` spec if isn't really doing anything - could likely leave it as an `any?` spec for now, and leave the comment in case we go back to fix it later.

danmason (approved):
👍 All looks good to me - left a small comment re the `args` spec but I think it's good to get this fixed.

crux-http-server/src/crux/http_server/query.clj
refset: Yep I agree it's not really necessary to merge this bit, but I felt I should include it during this review stage for completeness :)

refset: Is there a better process for discussing such snippets? e.g. just create another branch on my clone and reference it in the description?

danmason: Hmmm - I'd probably just create a comment on the line itself and copy the snippet into that.

refset:
Update: managed to get the `:decode/string (fn [_ a] (util/try-decode-edn a))` approach working in the end - maybe there was some conflict happening before when Dan & I were pairing.

crux-http-client/src/crux/remote_api_client.clj
```diff
@@ -91,7 +91,8 @@
                                            (= :stream (:as http-opts)) slurp))]
          (throw (case (::err/error-type error-data)
                   :illegal-argument (err/illegal-arg (::err/error-key error-data) error-data)
-                  :node-out-of-sync (err/node-out-of-sync error-data))))
+                  :node-out-of-sync (err/node-out-of-sync error-data)
+                  (err/illegal-arg "Generic remote client error" error-data))))
```
jarohen: Fair enough to add a default case, although it's not an illegal argument at that point - I'd go with `(ex-info "..." error-data)`.

crux-http-server/src/crux/http_server/query.clj
jarohen: assuming this is now used?

refset: yep :) since my latest force-push @ https://github.com/juxt/crux/pull/1507/commits/60c6fc387aef3bb05fe5322fa9bc0c6a8e12b8b4

jarohen (changes requested):
Cheers @refset & @danmason :pray: 

One suggestion - I'd call this `:params` rather than `:args` in the HTTP API given `:args` already has a usage in the query spec, with a different format - RFC?

To your questions:

* Should the JSON-over-HTTP `/_crux/query` calls be non-lazy by default? 
  Possibly, but definitely not in this PR. Given how it works currently, I'd be tempted to have this resolved on the client-side so that we don't break the API.
* Should query be coerced/normalised before hitting `crux/q`? 
  Again, out of scope for this PR. What're the pros/cons here?
* Could it be sensible to differentiate between args and args-edn/args-json? 
  I would, yes. Query parameters do also allow positional params out of the box, so it might be worth using these for the top-level params and then EDN/JSON for anything nested - or, you could go the whole way and use QP's in-built structure throughout (although I inevitably have to look up how it works). Tbh, for any substantial in-params, I'd use the request body anyway, even if it's not 'textbook REST'.

danmason:
:+1: for naming it `:params`.

jarohen:
as discussed:
- `::args` -> `::in-args`, not `::params`
- addition of `in-args-edn` and `in-args-json` query parameters
- test all of the above in JSON test
- let's add a pull test to api-test if there isn't already one present
- add new parameters to the `opencrux` docs

crux-http-server/test/crux/http_server/json_test.clj
```diff
@@ -90,6 +90,21 @@
...
+              {:url "/_crux/query"
+               :qps {"queryEdn" (pr-str '{:find [(pull e [*])]
+                                          :in [n c k s [i ...] m]
+                                          ;; :limit 1 ;; can use this to mitigate streaming open-q bag effect
+                                          :where [[e :firstName n]
+                                                  [(= c 123)]
+                                                  [(= k :firstName)]
+                                                  [(contains? s i)]
+                                                  [(= m {"a[b][c]" :d "a[b][e]" "f"})]]})
+                     "args" [["Sally" 123 :firstName [0 1 2] [0 1 2] {:a {:b {:c :d :e "f"}}}]]}})))
```
jarohen: add pr-str here for `in-args-edn`

jarohen: and add to HTTP reference docs

crux-http-server/test/crux/http_server/json_test.clj
```diff
@@ -90,6 +90,36 @@
...
+                                          :in [n c k s [i ...] m]
+                                          ;; :limit 1 ;; can use this to mitigate streaming open-q bag effect
+                                          :where [[e :firstName n]
+                                                  [(= c 123)]
+                                                  [(= k :firstName)]
+                                                  [(contains? s i)]
+                                                  [(= m {:a {:b {:c :d :e "f"}}})]]})
+                     "inArgsEdn" (pr-str ["Sally" 123 :firstName [0 1 2] [0 1 2] {:a {:b {:c :d :e "f"}}}])}})))
+
+    (t/is (= [[{"crux.db/id" "sal", "firstName" "Sally", "lastName" "Example"}]
```
danmason, line 108: JSON decode - see differences to above (since we cannot pass through keywords)

jarohen:
> > Should the JSON-over-HTTP /_crux/query calls be non-lazy by default?

> Possibly, but definitely not in this PR. Given how it works currently, I'd be tempted to have this resolved on the client-side so that we don't break the API.

@refset looks like it already works this way :smile: 
https://github.com/juxt/crux/pull/1507/files#diff-85e2a423fc7062d253791fc2a5b438618e5430ea20df30adcc2077c2bdd2f66dR136

jarohen (approved):

Issue #1341, opened 2020-12-23
NPE in crux.metrics.index-store on startup


nivekuil:
Tested on `dev-SNAPSHOT`. Seems like this one actually causes a fatal error that prevents the application from starting, unlike https://github.com/juxt/crux/issues/1320 which was apparently mostly benign. 
```
{:clojure.main/message
 "Execution error (NullPointerException) at crux.metrics.index-store/assign-tx-id-lag$fn (index_store.clj:12).\nCannot invoke \"Object.getClass()\" because \"x\" is null\n",
 :clojure.main/triage
 {:clojure.error/class java.lang.NullPointerException,
  :clojure.error/line 12,
  :clojure.error/cause
  "Cannot invoke \"Object.getClass()\" because \"x\" is null",
  :clojure.error/symbol crux.metrics.index-store/assign-tx-id-lag$fn,
  :clojure.error/source "index_store.clj",
  :clojure.error/phase :execution},
 :clojure.main/trace
 {:via
  [{:type clojure.lang.ExceptionInfo,
    :message "Error on key :app.db/crux when building system",
    :data
    {:reason :integrant.core/build-threw-exception,
     :system
     {:app.crux/s3
      #object[app.crux$eval4155$fn__4157$fn__4161 0x6505a530 "app.crux$eval4155$fn__4157$fn__4161@6505a530"]},
     :function
     #object[clojure.lang.MultiFn 0x598aa68e "clojure.lang.MultiFn@598aa68e"],
     :key :app.db/crux,
     :value
     {:crux.http-server/server {:port 3000},
      :crux.metrics.prometheus/http-exporter {:port 9000},
      :crux/index-store
      {:kv-store
       {:crux/module crux.rocksdb/->kv-store,
        :db-dir "/tmp/crux-rocksdb-index",
        :checkpointer
        {:crux/module crux.checkpoint/->checkpointer,
         :store
         {:crux/module crux.s3.checkpoint/->cp-store,
          :bucket "crux-checkpoint",
          :configurator
          #object[app.crux$eval4155$fn__4157$fn__4161 0x6505a530 "app.crux$eval4155$fn__4157$fn__4161@6505a530"]},
         :approx-frequency "PT6H"}}},
      :crux.jdbc/connection-pool
      {:dialect {:crux/module crux.jdbc.psql/->dialect},
       :pool-opts {:maximumPoolSize 3, :maxLifeTime 120000},
       :db-spec
       {:jdbcUrl
[...]

PR #1508, opened 2021-05-01
Fix crux-metrics npe on startup

nivekuil:
Ran into https://github.com/juxt/crux/issues/1341 again, extra when-let seems to fix it.


jarohen:
Cheers @nivekuil :pray: 

That's strange - if there's a `latest-completed-tx` there should always be a `latest-submitted-tx` - the latter happens before the former... :thinking:

Do you have a repro for this?

nivekuil:
I do not, only tried on my own project, but it seems possible to me as they read from different kv stores (completed from the index-store kv, submitted from the tx-log kv) and go through different code paths

PR #1512, opened 2021-05-14
Reconstructing docs in the query indices, #1511

### LINKED ISSUE #1511: Reconstructing docs from query indices
Currently, when required to reconstruct documents (e.g. for `entity` or `pull [*]`) we have to go to the document store, because the query indices don't contain enough information. Specifically, they cannot reconstruct values that are sets or vectors - e.g. if we find a single value in the ECAV index we don't know whether it's actually a single value or a single element set/vector, and we don't store the ordering/duplicate values of vectors.

For remote/pay-per-req doc-stores, we do recommend as large a local doc-store cache as is practical, although this is mostly a duplication of data that's already stored in local query indices.

Proposal:
- the values in the ECAV index are unused - we could store information here about whether the value is a scalar, a set, or its indices within a vector.
- the attribute hashes in the ECAV index currently aren't reversible - we'll have to either make these reversible, or keep a cached mapping from the `all-attrs` function

jarohen:
We use the value in the ECAV index to reconstruct entities using the query indices alone. Resolves #1511. 

Pros:
- no need to go to doc-store for `entity` or `pull`. we could also remove the recommendation to keep an as-full-as-possible local doc-store cache, which will reduce per-node disk space reqmts.
- `pull` is vastly simpler - can get rid of the monadic doc-store lookup batching

Cons:
- entity reconstruction from the query indices is more involved than simply deserialising it out of the doc-store

Notes:
- Java API now returns longs in documents if ints are passed in, because the docs are reconstructed from the indices which transform numbers to longs/doubles.
- I've added the distinction between hashing which needs to preserve consistent ids and hashing which needs to preserve consistent values - the former is so that hashing behaviour is preserved for hashes that go on the log/in the doc-store; the latter for values hashed in the indices. The distinction means that we can fix some hashing behaviours in the indices on an index version bump, rather than holding off because the hashes are on the tx-log. RFC.

### DIFF
--- a/crux-core/src/crux/pull.clj
+++ b/crux-core/src/crux/pull.clj
@@ -2,46 +2,14 @@
   (:require [crux.codec :as c]
             [crux.db :as db]
             [edn-query-language.core :as eql]
-            [clojure.string :as string])
+            [clojure.string :as string]
+            [crux.api :as api])
   (:import clojure.lang.MapEntry))
 
 (defn- recognise-union [child]
   (when (and (= :join (:type child))
              (= :union (get-in child [:children 0 :type])))
     :union))
-
-(defn- replace-docs [v docs]
-  (if (not-empty (::hashes (meta v)))
-    (v docs)
-    v))
-
-(defn- lookup-docs [v {:keys [document-store]}]
-  (when-let [hashes (not-empty (::hashes (meta v)))]
-    (db/fetch-docs document-store hashes)))
-
-(defmacro let-docs {:style/indent 1} [[binding hashes] & body]
-  `(-> (fn ~'let-docs [~binding]
-         ~@body)
-       (with-meta {::hashes ~hashes})))
-
-(defn- after-doc-lookup [f lookup]
-  (if-let [hashes (::hashes (meta lookup))]
-    (let-docs [docs hashes]
-      (let [res (replace-docs lookup docs)]
-        (if (::hashes (meta res))
-          (after-doc-lookup f res)
-          (f res))))
-    (f lookup)))
-
-(defn- raise-doc-lookup-out-of-coll
-  "turns a vector/set where each of the values could be doc lookups into a single doc lookup returning a vector/set"
-  [coll]
-  (if-let [hashes (not-empty (into #{} (mapcat (comp ::hashes meta)) coll))]
-    (let-docs [docs hashes]
-      (->> coll
-           (into (empty coll) (map #(replace-docs % docs)))
-           raise-doc-lookup-out-of-coll))
-    coll))
 
 (defrecord RecurseState [seen-vs child-fns])
 
@@ -55,9 +23,7 @@
                 (->> (mapv (fn [f]
                              (f v db recurse-state))
                            (.child-fns recurse-state))
-                     (raise-doc-lookup-out-of-coll)
-                     (after-doc-lookup (fn [res]
-                                         (into {} (mapcat identity) res)))))))]
+                     (into {} (mapcat identity))))))]
     (cond
       (= '... query) pull-child
       (int? query) (fn [v db ^RecurseState recurse-state]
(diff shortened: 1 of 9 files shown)


crux-core/src/crux/kv/index_store.clj
```diff
@@ -701,6 +712,37 @@
...
+                                            (c/->id-buffer content-hash))]
+
+      (when-let [ecav-kvs (seq (all-keys-in-prefix @level-1-iterator-delay
+                                                   ecav-k-prefix (.capacity ecav-k-prefix)
+                                                   {:entries? true}))]
+        (->> (for [[k idxs] ecav-kvs
+                   :let [^Quad quad (decode-ecav-key-from k eid-size)
+                         a (db/decode-value this (.buffer ^Id (.attr quad)))
+                         v (db/decode-value this (.value quad))]
+                   idx (or (decode-ecav-value-from idxs) [nil])]
```
hraberg: This will result in a boxed iteration over the int array, might normally not be an issue unless you decode many large docs with long vectors, just mentioning it.

crux-core/src/crux/kv/index_store.clj
```diff
@@ -193,6 +193,17 @@
...
+(defn- decode-ecav-value-from ^ints [^DirectBuffer v]
+  (let [v-size (.capacity v)]
+    (when-not (zero? v-size)
+      (let [len (/ v-size Integer/BYTES)
+            out (int-array len)]
+        (dotimes [idx len]
+          (aset out idx (.getInt v (* idx Integer/BYTES))))
+        out))))
+
+(defrecord ECAVEntry [a v idx])
```
hraberg: Might want to be a deftype with an actual `int` idx, see next comment.

crux-core/src/crux/kv/index_store.clj
```diff
@@ -867,25 +909,58 @@
...
 
+(defn- val-idxs [v]
+  (let [val->idxs (HashMap.)]
+    (cond
+      (vector? v) (dorun
+                   (map-indexed (fn [idx el]
+                                  (.compute val->idxs el
+                                            (reify BiFunction
+                                              (apply [_ _ acc]
+                                                (let [^List acc (or acc (ArrayList.))]
```
hraberg: Similar thing here, the actual elements here are ints, so there will be boxing. Worth testing indexing and pull of documents with large vectors inside to sanity check if it matters or not, but that applies more broadly to the entire approach as well.

crux-test/test/crux/kv/index_store_test.clj
```diff
@@ -274,3 +274,17 @@
       (t/is (= {:crux.db/id {:doc-count 3675, :values 3554, :eids 3554}
                 :sub-idx {:doc-count 3675, :values 50, :eids 3554}}
                (->stats *index-store*))))))
+
+(t/deftest test-entity
```
hraberg: This namespace might be a good place for a slowdown factor-style test, sanity checking the indexing and reconstruction cost of the new approach?

hraberg (approved):
Approving, but recommend looking at and sanity checking with a microbench indexing and reconstructing documents with large vectors.

We've discussed the data model issues elsewhere, that now it won't be possible to reconstruct the document verbatim the way it was stored.

This is in theory a feature (as we want a controlled data model), but might caught someone off guard if they rely on the equality of stored documents against documents from outside Crux where ints etc. still are there. My guess is that this is minor issue in practice, but might be worth a note in the release notes.

crux-core/src/crux/kv/index_store.clj
jarohen: Did give this a go, but can't just make this an `int` because it's nullable in the case where the value isn't a collection. Could use `-2` instead, say, if it'd make a reasonable difference, but I'm not sure there's much we can do about passing it to `sort-by` fairly soon afterwards?

jarohen:
Performance for both indexing and small docs was reasonable, but turns out this wasn't adequately performant for reconstructing entities containing large vectors - I'll close this for now.

Part one of this (the changes to the indices) was merged in f5ee85bf15abc4843f93b88bae7e33d0ab146ecd - this means that the change to enable it shouldn't require an index version bump (or even, indeed, a breaking API change).

Issue #1463, opened 2021-03-17
Sync extremly slow when jdbc used


beetleman:
I using crux in this configuration:
```clojure
{:crux.jdbc/connection-pool {:dialect   {:crux/module
                                         'crux.jdbc.psql/->dialect}
                             :pool-opts {:maximumPoolSize 100}
                             :db-spec
                             {:jdbcUrl
                              "jdbc:postgresql://localhost:6432/postgres"
                              :user     "postgres"
                              :password "postgres"}}
 :crux/index-store          {:kv-store
                             {:crux/module 'crux.rocksdb/->kv-store
                              :db-dir      (io/file
                                            "crux/index")}}
 :crux/tx-log               {:crux/module     'crux.jdbc/->tx-log
                             :connection-pool :crux.jdbc/connection-pool}
 :crux/document-store       {:crux/module     'crux.jdbc/->document-store
                             :connection-pool :crux.jdbc/connection-pool}}

```
versions:
```clojure
juxt/crux-core{:mvn/version "21.02-1.15.0-beta"}
juxt/crux-rocksdb{:mvn/version "21.02-1.15.0-beta"}
juxt/crux-jdbc{:mvn/version "21.02-1.15.0-beta"}

;; JDBC
org.postgresql/postgresql{:mvn/version "42.2.19"}
```

and  `crux/sync` for new node is slower than it should be. I checked execution plan for doc fetching query and it do not use index, crux have index https://github.com/juxt/crux/blob/master/crux-jdbc/src/crux/jdbc/psql.clj#L31 but its for 2 fields. I added index for `event_key` and query now for ~100 docs improved from ~6s to ~10ms. I do not know how it transfer to other dialects but for postgres its huge difference.

jarohen:
Hey @beetleman, thanks for the report :pray: I'm not sure why the index is ordered in that way, will check on the history and let you know.

beetleman:
After quick look on code i think we should have both indexies:
```clojure
"CREATE INDEX IF NOT EXISTS tx_events_compacted_event_key_idx ON tx_events(compacted, event_key)"
"CREATE INDEX IF NOT EXISTS tx_events_event_key_idx ON tx_events(event_key)"
```
but I'm not SQL expert so somebody with more experience should look on this:D

kevinmershon:
Even with the additional index, syncing/indexing new data with a postgresql backend is slow to the point of unusable, if `crux.lucene/lucene-store` is used

Follow-up edit:
[...]

PR #1521, opened 2021-05-28
remove compacted column in the migrated tx_events index #1463

keytiong:
Index migration code still contains the compacted column


jarohen:
Good spot, thanks! :pray:

PR #1526, opened 2021-06-08
fixed sql syntax error on mysql dialect to specify table

yueda27:
current mysql/->dialect function will throw a SQL syntax error on this line when executed.

Current dialect only works if tx_events_event_key_idx is currently not in table. if it is, it will fail on startup


jarohen:
Thanks @yueda27 :pray:

PR #1527, opened 2021-06-08
Cast ByteBuffer to Buffer before calling .limit, fixes #1524

### LINKED ISSUE #1524: Error locating module crux.kv.index-store/->kv-index-store
After updating from 1.16.0 to 1.17.0 I get a new error from my unit tests.

Previously I've initialized a "mock" node by just passing an empty configuration map to `crux.api/start-node`. Which worked fine.

With 1.17.0 I get the following error:

```
Error locating module
   {:module crux.kv.index-store/->kv-index-store}
                system.clj:  113  crux.system.ModuleRef/prepare_dep
                system.clj:  131  crux.system/opts-reducer/f
     PersistentVector.java:  343  clojure.lang.PersistentVector/reduce
                  core.clj: 6852  clojure.core/reduce
                  core.clj: 6835  clojure.core/reduce
                system.clj:  156  crux.system/prep-system
                system.clj:  141  crux.system/prep-system
                system.clj:  142  crux.system/prep-system
                system.clj:  141  crux.system/prep-system
                   api.clj:  231  crux.api/start-node
                   api.clj:  218  crux.api/start-node

```

refset:
I've not been able to reproduce the error so far, but reading around the issue (e.g. https://github.com/eclipse/jetty.project/issues/3244) I _think_ this should fix it.

### DIFF
--- a/crux-core/src/crux/memory.clj
+++ b/crux-core/src/crux/memory.clj
@@ -5,10 +5,10 @@
             [crux.io :as cio]
             [taoensso.nippy :as nippy])
   (:import [java.io DataInputStream DataOutputStream]
+           java.nio.Buffer
            java.nio.ByteBuffer
-           java.util.Comparator
            java.util.function.Supplier
-           [org.agrona BufferUtil DirectBuffer ExpandableDirectByteBuffer MutableDirectBuffer]
+           [org.agrona DirectBuffer ExpandableDirectByteBuffer MutableDirectBuffer]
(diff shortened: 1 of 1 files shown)


jarohen:
Well TIL!

Looks like this method was then removed in Java 10, although I can't find anything in the changelog. Strange. Think it's also going to depend on the versions of Java that our users are using to build/run Crux, because we don't ship AoT compiled classes. This fix looks fine though, thanks! :pray:

PR #1528, opened 2021-06-09
Add optional commit debounce to crux-lucene, fixes #1500

### LINKED ISSUE #1500: Lucene indexing process abysmally slow due to unnecessary commit call

refset:
This is not an ideal long-term solution to #1500, but it will unblock our users (who are prepared to handle unrecoverable mismatch errors by restoring nodes from offline backups, not checkpoints) and some kind of debounce is necessary anyway. My comments here still stand: https://github.com/juxt/crux/pull/1501#issuecomment-821294963

The other context around this change was here: https://github.com/juxt/crux/pull/1468#issue-596668106

## Background reading on Lucene behaviours

Background info on Lucene commit / flush:
> Both commit and flush write indexing data that is currently in memory to disk. Commit, however, does something extra. It also updates the index, indicating that the data on disk is ready to be used for searching.
> So if you always flush but never commit, your index cant be searched. If you always commit, but never flush, that is fine since a commit implicitly flushes. Flushing normally occurs automatically when you index a large amount of data that wouldn't be feasible to keep in memory. You only commit when you have reached a state that you want to really persist.
from https://stackoverflow.com/questions/48557670/what-is-the-difference-between-commit-and-flush-for-indexwriter-in-lucene

Further details on flush config:
https://lucene.apache.org/core/4_5_0/core/org/apache/lucene/index/IndexWriter.html

And on maybeRefreshBlocking:
https://lucene.apache.org/core/7_4_0/core/org/apache/lucene/search/SearcherManager.html
https://blog.mikemccandless.com/2011/11/near-real-time-readers-with-lucenes.html

## Debounce

The "debounce" implementation matches the requirements described here: https://engineering.vena.io/2020/05/19/how-to-java-debounce/
> When we make a request:
> If it hasn't been made within the past delayMillis milliseconds (and there isn't a call queued), then the request gets made immediately
> If there IS a recent call, then we check to see if one is queued
> - If there is a call queued, we drop the current call
> - If there is not, we queue up the current call to be called after delayMillis milliseconds pass

I'd be happy to move it out of the Lucene namespace if there's a good home somewhere else.

### DIFF
--- a/crux-lucene/src/crux/lucene.clj
+++ b/crux-lucene/src/crux/lucene.clj
@@ -7,9 +7,13 @@
             [crux.io :as cio]
             [crux.query :as q]
             [crux.system :as sys])
-  (:import crux.query.VarBinding
+  (:import clojure.lang.IFn
+           crux.query.VarBinding
            java.io.Closeable
+           java.io.IOException
            java.nio.file.Path
+           java.time.Duration
+           java.util.Date
            org.apache.lucene.analysis.Analyzer
            org.apache.lucene.analysis.standard.StandardAnalyzer
            [org.apache.lucene.document Document Field$Store StoredField StringField TextField]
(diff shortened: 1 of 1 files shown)


refset:
I'll tweak that slowdown test again, separately :sweat_smile:

crux-lucene/src/crux/lucene.clj
```diff
@@ -10,19 +10,54 @@
...
+                              (reset! !is-scheduled?
+                                      (.schedule executor
+                                                 ^Runnable
+                                                 (fn []
+                                                   (reset! !is-scheduled? false)
+                                                   (reset! !last-run-time (.getTime (Date.)))
+                                                   (f))
+                                                 ^long (- next-run now)
+                                                 TimeUnit/MILLISECONDS))))))]
+    [(if (= wait-ms 0) f schedule-run!)
```
jarohen: I'd be tempted to combine these return values - you can also implement `clojure.lang.IFn` as part of the `reify` form.

```clojure
(reify
  IFn
  (invoke [_] ...)

  Closeable
  (close [_] ...))

Comment (jarohen) [reply crux-lucene/src/crux/lucene.clj]: Also worth checking `wait-ms` at the top, and not creating the Executor if it's 0?

File: crux-lucene/src/crux/lucene.clj [hunk 2]
```diff
@@ -10,19 +10,54 @@
...
+     (reify
+       Closeable
+       (close [_]
+         (when-not (false? @!is-scheduled?)
```
jarohen: if there is an fsync scheduled when you're being asked to close, would you want to call it immediately?

crux-lucene/src/crux/lucene.clj
```diff
@@ -10,19 +10,54 @@
...
+           [org.apache.lucene.store Directory FSDirectory]
+           [java.util.concurrent Future Executors TimeUnit]))
 
-(defrecord LuceneNode [directory analyzer index-writer searcher-manager indexer]
+(defn debounce [name f wait-ms]
+  (let [!last-run-time (atom (.getTime (Date.)))
+        !is-scheduled? (atom false)
+        executor (Executors/newSingleThreadScheduledExecutor (cio/thread-factory name))
+        schedule-run! (fn run []
+                        (when (false? @!is-scheduled?)
```
jarohen: * There's a couple of potential race conditions in here - (_very_ generally speaking) any time you're reading state twice, or reading it and writing it in two distinct operations, you may have a race. `swap!` might be out given it retries, and you've also got side effects to think about - it might be that an approach based on `compare-and-set!` is your best option.
* 'reading state' here includes getting the current time :wink: time is fun.
* Worth considering what you want to happen if the fsync fails, for whatever reason
* `!is-scheduled?` is a bit of a misnomer given it might contain the job - maybe `!fut` is sufficient (i.e. anything without the `?`). fine to have an atom that's either a Future or nil.
* Looking at lodash, I think this may be 'throttling' rather than 'debouncing'?

jarohen (changes requested):

crux-lucene/src/crux/lucene.clj
refset: I've now tried to consolidate all the atom usage and current time lookups. However I already foresee that I may need to rename my three-valued `fut` :)

refset:
Thanks for all the feedback so far. It feels much closer now!

In case it helps anyone, this is how I've been testing `throttle`:

```clj
(comment
  (def f #(do ( prn "hi" (rand-int 100))))
  (def g (throttle "hi" f 500))
  (do
    (g)
    (g)
    (g)
    (g)
    (g)
    (Thread/sleep 440)
    (g)
    (g)
    (Thread/sleep 640)
    (g)
    (g))
  (.close ^Closeable g))
```

jarohen:
As discussed - well implemented :clap: but going for a simpler schedule now that we have #1535

PR #1529, opened 2021-06-10
Allow crux-lucene to search for keywords and symbols as strings

refset:
WIP - assuming this strategy looks okay we would also need to adapt the wildcard and multi-field versions similarly to finish this PR.

Outstanding questions:

1. In this PR's current implementation, keywords and symbols are tokenized in the same way as all the other strings - is this actually useful? What is an example?
    1. If "Yes" - should we also index a non-tokenized version?
    2. If "No" - what are the use-cases where having these values indexed by Lucene enables something that is not possible (or highly inefficient) when retrieving the values via Datalog instead


jarohen:
Will close this one for now - it's preferable that users write their own indexing strategy in this case, rather than making this the general case.

PR #1535, opened 2021-06-16
Lucene indexes asynchronously, #1426

### LINKED ISSUE #1426: Asychronous indexing for crux-lucene

jarohen:
Previously, Lucene had to be indexed in lockstep with the main Crux indices, using bus events. Now it can be indexed independently - it's fine to (e.g.) re-index either the Crux or the Lucene indexes without affecting the other.

In order to read their own Lucene writes, users must call `l/await-tx` in addition to the usual Crux `await-tx`.

Implementation notes:
- Most of this PR is decoupling the tx-log, tx-indexer and tx-ingester. The latter two have been split out to decouple the `begin-tx`/`index-tx-events`/`commit` API from any notion of ingestion processes - there are use cases for the former (e.g. speculative txs, or as a target for more bespoke ingestion processes) and the latter (e.g. Lucene doesn't have this same ability to stop and take a break mid-transaction).
- Lucene will always be behind the main Crux indexer, because it needs to know whether the transaction committed or not - Lucene has no knowledge (nor should it) of whether a `match` succeeded, or how to run tx-fns.
- Now that this is split out, TxLogs don't need to worry about indexing/submit-only nodes - this is done by including (or not!) the tx-ingester in the node. Calling `crux/start-node` includes it, starting an ingest-client (for a submit-only node - rather confusingly named now!) doesn't.
- TxLogs must now implement `subscribe-async`, which takes a reduce-like f + init, returning `CompletableFuture`. This is trivially implemented for tx-logs - implementers can call `tx-sub/handle-polling-subscription` or, if the tx-log supports some kind of 'notify', it can use `tx-sub/->notify-subscriber-handler`.
- Lucene now uses AVE docs rather than AV to ease eviction - it can no longer rely on the corresponding Crux indices being able to answer its `exclusive-avs` questions. This will involve a Lucene index version bump (which doesn't yet exist out of the box - something we should do before the next release, IMO - #1538).
- I've removed `committing-tx` and `aborting-tx` bus events, on the assumption they're only used by Lucene. Likewise `db/exclusive-avs` function.

Still to do:
- [x] Ensure `crux-lucene` works with (optionally nested) transaction functions - it'll need to flatten out the resulting tx-ops to ensure all documents are indexed. 
- [x] Implement a 'notifying transaction subscriber' for the KV tx-log to restore its previous behaviour.
- [x] Update Lucene docs
- [ ] Worth running past a few folk that we know are using `crux-lucene`
- [ ] Worth running past the folk that we know have created their own TxLogs.


crux-core/src/crux/kv/tx_log.clj
```diff
@@ -30,42 +30,52 @@
...
+          (kv/fsync kv-store))
 
-      (when fsync?
-        (kv/fsync kv-store))
+        (tx-sub/notify-tx! subscriber-handler next-tx)
 
-      next-tx)))
+        next-tx)
+      (catch Exception e
+        (.printStackTrace e)
```
jonpither: Presumably here for debugging - can take out?

jonpither (changes requested):
A lot of code changes here - feel we need an person session to go over it briefly.

crux-core/src/crux/kv/tx_log.clj
jarohen: cheers @jonpither :pray:

jarohen:
> A lot of code changes here - feel we need an person session to go over it briefly.

@jonpither yep, let's :smile:

crux-core/src/crux/tx.clj
```diff
@@ -338,148 +335,149 @@
 
       (catch Throwable e
         (reset! !tx-state :abort-only)
-        (raise-ingester-error! this e))))
-
-  (commit [this]
-    (try
-      (when-not (compare-and-set! !tx-state :open :committed)
```
jarohen, line 345: why was this removed...?

crux-core/src/crux/tx.clj
```diff
@@ -338,148 +335,149 @@
...
+(defn ->tx-ingester {::sys/deps {:tx-indexer :crux/tx-indexer
+                                 :index-store :crux/index-store
+                                 :tx-log :crux/tx-log
+                                 :bus :crux/bus}}
+  [{:keys [tx-log tx-indexer bus index-store]}]
+  (log/info "Started tx-ingester")
+  (let [!error (atom nil)
+        fut (db/subscribe-async tx-log
+                                (::tx-id (db/latest-completed-tx index-store))
+                                (fn [_acc tx]
```
jarohen: do we need the reduce-like functionality here?

crux-core/src/crux/node.clj
```diff
@@ -286,30 +254,30 @@
...
-                                                  :default (Duration/ofHours 24)
-                                                  :spec ::sys/duration}
-                           :slow-queries-max-count {:doc "Max number of finished queries to retain on the slow query queue"
-                                                    :default 100
-                                                    :spec ::sys/nat-int}
-                           :slow-queries-min-threshold {:doc "Minimum threshold for a query to be considered slow."
-                                                        :default (Duration/ofMinutes 1)
-                                                        :spec ::sys/duration}}}
+(defn ->node {::sys/deps {:index-store :crux/index-store
+                          :tx-ingester :crux/tx-ingester
```
jarohen, line 258: does this need tx-indexer too?

crux-lucene/src/crux/lucene.clj
```diff
@@ -170,83 +176,128 @@
...
+
+(defrecord LuceneAveIndexer []
+  LuceneTxIndexer
 
-(defrecord LuceneAvIndexer [index-store]
-  LuceneIndexer
+  (index-tx! [_ index-writer evicted-eids docs]
+    (let [qs (for [eid evicted-eids]
+               (TermQuery. (Term. field-crux-eid (->hash-str eid))))]
+      (.deleteDocuments ^IndexWriter index-writer ^"[Lorg.apache.lucene.search.Query;" (into-array Query qs)))
```
jarohen, line 188: can we tidy up `exclusive-avs`?

crux-lucene/src/crux/lucene.clj
```diff
@@ -170,83 +176,128 @@
...
 
+(defn- transform-tx-events [document-store tx-events]
+  (let [conformed-tx-events (map txc/<-tx-event tx-events)
+        docs (db/fetch-docs document-store
+                            (txc/conformed-tx-events->doc-hashes conformed-tx-events))]
+    (reduce (fn [acc {:keys [op] :as tx-event}]
+              (case op
+                :crux.tx/evict (-> acc (update :evicted-eids conj (:eid tx-event)))
+                :crux.tx/fn (let [{:keys [args-content-hash]} tx-event
+                                  ;; doc replaced by this point
```
jarohen, line 223: sure? Kafka race condition...

crux-lucene/src/crux/lucene.clj
```diff
@@ -170,83 +176,128 @@
...
-                    (index! indexer index-writer (db/fetch-docs document-store (:doc-ids ev)))
-                    (when-let [evicting-eids (not-empty (:evicting-eids ev))]
-                      (evict! indexer index-writer evicting-eids)))
-                  (index-tx! index-writer (:submitted-tx ev))
-                  (.commit index-writer)
-                  (.maybeRefreshBlocking searcher-manager)))
-    lucene-store))
+
+    (-> lucene-store
+        (assoc :ingester-fut
```
jarohen: -> job

crux-lucene/src/crux/lucene.clj
```diff
@@ -170,83 +176,128 @@
...
+                                         (let [committed? (not (db/tx-failed? index-store tx-id))]
+                                           (when committed?
+                                             (let [{:keys [docs evicted-eids]} (transform-tx-events document-store tx-events)]
+                                               (index-tx! indexer index-writer evicted-eids docs)))
+
+                                           (complete-tx! index-writer tx)
+                                           (.commit index-writer) ; TODO aware of #1500
+                                           (.maybeRefreshBlocking searcher-manager)
+
+                                           (bus/send bus {:crux/event-type ::indexed-tx
```
jarohen, line 280: fq these event-types?

crux-core/src/crux/tx/subscribe.clj
```diff
@@ -0,0 +1,118 @@
...
+                   (recur last-tx-id v))))))))
+
+(defprotocol PNotifyingSubscriberHandler
+  (notify-tx! [_ tx])
+  (handle-notifying-subscriber [_ tx-log after-tx-id f init]))
+
+(defrecord NotifyingSubscriberHandler [!latest-submitted-tx-id !semaphores]
+  PNotifyingSubscriberHandler
+  (notify-tx! [_ tx]
+    (let [semaphores (dosync
```
jarohen: -> atoms?

crux-lucene/src/crux/lucene.clj
```diff
@@ -170,83 +176,128 @@
...
+                                         (tx/await-tx await-deps tx)
+
+                                         (let [committed? (not (db/tx-failed? index-store tx-id))]
+                                           (when committed?
+                                             (let [{:keys [docs evicted-eids]} (transform-tx-events document-store tx-events)]
+                                               (index-tx! indexer index-writer evicted-eids docs)))
+
+                                           (complete-tx! index-writer tx)
+                                           (.commit index-writer) ; TODO aware of #1500
+                                           (.maybeRefreshBlocking searcher-manager)
```
refset, line 278: On reflection, it is now perhaps best to do this refresh work in the new `await-tx` - WDYT?

jarohen: it's an option - although someone (you, I think?) mentioned in the call earlier not to put all this work on one unfortunate reader?

(assuming you mean `.maybeRefreshBlocking`? given I'm pretty sure `complete-tx!` has to stay here (and it's quick enough anyway), and `.commit`'s not required for readers, can do that on a schedule)

crux-core/src/crux/node.clj
jarohen: nope - covered in `crux.api/start-node`

crux-lucene/src/crux/lucene.clj
refset: (Yep, just meant the `.maybeRefreshBlocking`, sorry :))

> you, I think?

I did yep, quoting from https://lucene.apache.org/core/7_0_0/core/org/apache/lucene/search/SearcherManager.html
> In addition you should periodically call ReferenceManager.maybeRefresh(). While it's possible to call this just before running each query, this is discouraged since it penalizes the unlucky queries that need to refresh. It's better to use a separate background thread, that periodically calls ReferenceManager.maybeRefresh().

The benefit of putting it on await-tx is that it will make the ingest/catch-up even snappier (which might well be _very_ significant, it's hard to judge at this stage) and also then the user has complete control to schedule their own periodic (not immediately before a query) `l/await-tx` _if_ they want to avoid penalizing random queries. Granted it's a little bit of extra complexity for the user though, so perhaps we could just have another background thread anyway.

jarohen: fair. will leave it as-is for this PR, #1500 seems better placed to handle it :smile:

crux-lucene/src/crux/lucene.clj
jarohen: fixing in #1541

jarohen:
Closed in favour of #1555

Issue #600, opened 2020-01-28
Get cljr-clean-ns working for kafka-test


jonpither:

jonpither:
Found another bug - do it on crux.moberg

It takes out `[crux.moberg.types]` from the require, but then you can't compile the ns, because of stuff in imports. Looping in @benedekfazekas

benedekfazekas:
https://github.com/clojure-emacs/clj-refactor.el/blob/master/clj-refactor.el#L238 `cljr-libspec-whitelist` might provide a good workaround

jarohen:
`crux.moberg` was removed long ago - `cljr-clean-ns` looks good now given we no longer have any requires-for-side-effects

PR #1545, opened 2021-06-25
Minor cleanup

vemv:
## Context

I'm trying to add Crux (as a git submodule) to the [Eastwood](https://github.com/jonase/eastwood) build, since it's a very interesting project (as it's large, uses `lein-sub`, has many dependencies, type hints, etc).

This PR adds the things for the project to be readable to it; it doesn't try fixing the linting faults it found (some of which are pretty interesting!).

## Commit summary

* `crux.calcite-microbench-test`: don't require `user` ns which isn't necessarily present
* Ensure all keywords and symbols can be read by tools.reader
  * tools.reader is slightly more strict than the Clojure compiler, which doesn't always enforce its constraints due to backwards compatibility concerns ([ref](https://clojure.org/guides/faq#keyword_number)).
  * By making the codebase play better with tools.reader, one can consume Crux while making full use of various tooling.

### DIFF
--- a/crux-test/test/crux/lubm_test.clj
+++ b/crux-test/test/crux/lubm_test.clj
@@ -40,15 +40,15 @@
 ;; This query bears large input and high selectivity. It queries about just one class and
 ;; one property and does not assume any hierarchy information or inference.
 (t/deftest test-lubm-query-01
-  (t/is (= #{[:http://www.Department0.University0.edu/GraduateStudent101]
-             [:http://www.Department0.University0.edu/GraduateStudent124]
-             [:http://www.Department0.University0.edu/GraduateStudent142]
-             [:http://www.Department0.University0.edu/GraduateStudent44]}
+  (t/is (= #{[(keyword "http://www.Department0.University0.edu/GraduateStudent101")]
+             [(keyword "http://www.Department0.University0.edu/GraduateStudent124")]
+             [(keyword "http://www.Department0.University0.edu/GraduateStudent142")]
+             [(keyword "http://www.Department0.University0.edu/GraduateStudent44")]}
            (api/q (api/db *api*)
                   (rdf/with-prefix {:ub "http://swat.cse.lehigh.edu/onto/univ-bench.owl#"}
                     '{:find [x]
                       :where [[x :rdf/type :ub/GraduateStudent]
-                              [x :ub/takesCourse :http://www.Department0.University0.edu/GraduateCourse0]]})))))
+                              [x :ub/takesCourse #=(keyword "http://www.Department0.University0.edu/GraduateCourse0")]]})))))
 
 ;; TODO: subOrganizationOf is transitive, should use rules.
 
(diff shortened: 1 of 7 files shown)


crux-test/test/crux/lubm_test.clj
```diff
@@ -40,15 +40,15 @@
...
+  (t/is (= #{[(keyword "http://www.Department0.University0.edu/GraduateStudent101")]
+             [(keyword "http://www.Department0.University0.edu/GraduateStudent124")]
+             [(keyword "http://www.Department0.University0.edu/GraduateStudent142")]
+             [(keyword "http://www.Department0.University0.edu/GraduateStudent44")]}
            (api/q (api/db *api*)
                   (rdf/with-prefix {:ub "http://swat.cse.lehigh.edu/onto/univ-bench.owl#"}
                     '{:find [x]
                       :where [[x :rdf/type :ub/GraduateStudent]
-                              [x :ub/takesCourse :http://www.Department0.University0.edu/GraduateCourse0]]})))))
+                              [x :ub/takesCourse #=(keyword "http://www.Department0.University0.edu/GraduateCourse0")]]})))))
```
vemv, line 51: Read-eval isn't necessarily a clean solution. It trades one dubious thing for another 😄

Maybe one alternative would be changing `http:` to `http`, making the keywords/symbols valid. I tried `#=` first as it doesn't change semantics.

jonpither:
I expect this could fix this issue: https://github.com/juxt/crux/issues/600

jarohen:
Cheers @vemv :pray: I realise you were looking to make an change that preserved existing behaviour (which is much appreciated!) - it may be more straightforward to turn those `:http://...` keywords into strings now that Crux supports string eids. I realise that's quite a boring change given you've _already_ been through that namespace - happy to apply it if you can check the 'maintainers can make changes' box on the PR :)

Thanks again,

James

jarohen:
@jonpither thanks for reminding me of that one - that was fixed a while back by virtue of removing the code in question :smile: closed

vemv:
Thanks for the quick reply! It sounds good to me. I already hinted beforehand that `#=` didn't seem ideal so I'm fine with the alternative.

I'll close the PR since removing the second commit would leave a trivial first commit - not worth the CLA. I'd appreciate if you can replicate both.

jarohen:
Hi again @vemv - sorry to mess you about, I've realised in making that change that the keywords are part of the `crux-rdf` public API, and I would have hence broken people's code if I'd continued changing it to strings. With that in mind, I'd like to re-open your PR and merge it as-is - would you mind sending us through the CLA?

Cheers,

James

vemv:
Alright! Can do 👍

I assume you realise that all keyword changes are limited to `test/` dirs? So they wouldn't directly break consumers. OTOH it might be good to have explicit test coverage for keywords as an intended API.

jarohen:
> I assume you realise that all keyword changes are limited to test/ dirs? So they wouldn't directly break consumers. 

True - although my changing them to strings _would_ have required changes to `crux-rdf` itself - it returns keywords by default.

> OTOH it might be good to have explicit test coverage for keywords as an intended API.

That's what these tests do - again, before I proposed changing them! :flushed:

vemv:
👍! CLA sent

jarohen:
Thanks - all merged :smile:

PR #1558, opened 2021-07-01
Lucene infrequent fsyncing, #1500

### LINKED ISSUE #1500: Lucene indexing process abysmally slow due to unnecessary commit call

jarohen:
Based on #1555, see most recent commit on this branch for real diff.

Only 'committing' (fsyncing) Lucene's IndexWriter every so often (5m, by default) now that the Lucene index doesn't need to be kept strictly in sync with the main Crux indices.

Resolves #1500.


jarohen:
got other Lucene work queued up, will bundle up into a bigger PR

PR #1559, opened 2021-07-01
Lucene bash

### LINKED ISSUE #1221: Lucene-store is compatible with checkpointing

### LINKED ISSUE #1500: Lucene indexing process abysmally slow due to unnecessary commit call

### LINKED ISSUE #1538: Lucene should maintain and validate its own lucene-index-version
As a follow up to https://github.com/juxt/crux/pull/1535

### LINKED ISSUE #1540: In-memory Lucene
Currently, when running tests that use `crux-lucene`, a user will have to specify a directory and handle the cleanup afterwards. It would be preferable to align with the in-memory nature of Crux, where no directories need to be specified for basic usage.

Potential solutions:

1. If a directory is not specified in the config, then `crux-lucene` could write to an in-memory index via https://lucene.apache.org/core/7_5_0/core/org/apache/lucene/store/RAMDirectory.html - though this may be deprecated in future
2. If a directory is not specified in the config, then `crux-lucene` could write to a `/tmp` directory that the user doesn't have to worry about cleaning up afterwards.

Note this is unlikely to help solve/overlap with `crux-lucene`'s lack of support for speculative transactions (i.e. `with-tx`).

jarohen:
Based on #1555, see https://github.com/jarohen/crux/compare/secondary-indices-1426...lucene-bash for a better diff

* Scheduled fsyncing, fixes #1500 
* fixes #1221 - TODO doc
* fixes #1538
* fixes #1540

### DIFF
--- a/crux-lucene/src/crux/lucene.clj
+++ b/crux-lucene/src/crux/lucene.clj
@@ -40,15 +40,26 @@
 (defn ^String keyword->k [k]
   (subs (str k) 1))
 
+(def ^:const ^:private index-version 1)
 (def ^:const ^:private field-crux-id "_crux_id")
 (def ^:const ^:private field-crux-val "_crux_val")
 (def ^:const ^:private field-crux-attr "_crux_attr")
 (def ^:const ^:private field-crux-eid "_crux_eid")
 
+(defn- validate-index-version [^IndexWriter writer]
+  (let [found-index-version (some-> (into {} (.getLiveCommitData writer))
+                                    (get "crux.lucene/index-version")
+                                    (Long/parseLong))]
+    (when-not (or (zero? (.numDocs (.getDocStats writer)))
+                  (= index-version found-index-version))
+      (throw (IllegalStateException. (format "Lucene index structure out of date - please remove the Lucene dir and re-index. (expected: %d, actual: %s)"
+                                             index-version found-index-version))))))
+
 (defn- ^IndexWriter ->index-writer [{:keys [directory analyzer index-deletion-policy]}]
-  (IndexWriter. directory,
-                (cond-> (IndexWriterConfig. analyzer)
-                  index-deletion-policy (doto (.setIndexDeletionPolicy index-deletion-policy)))))
+  (doto (IndexWriter. directory,
+                      (cond-> (IndexWriterConfig. analyzer)
+                        index-deletion-policy (doto (.setIndexDeletionPolicy index-deletion-policy))))
+    (validate-index-version)))
 
 (defn- user-data->tx-id [user-data]
   (some-> user-data
@@ -294,7 +305,8 @@
                                 (evict! indexer index-writer evicting-eids))
                               (index! indexer index-writer docs)))
 
-                          (.setLiveCommitData index-writer {"crux.tx/tx-id" (str tx-id)})
+                          (.setLiveCommitData index-writer {"crux.tx/tx-id" (str tx-id)
+                                                            "crux.lucene/index-version" (str index-version)})
                           (.maybeRefreshBlocking searcher-manager)))
 
     lucene-store))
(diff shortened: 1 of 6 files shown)


crux-lucene/src/crux/lucene.clj
```diff
@@ -174,79 +172,142 @@
...
+                acc))
+            {:docs docs
+             :evicted-eids #{}}
+            conformed-tx-events)))
+
+(defn- checkpoint-src [^IndexWriter index-writer]
+  (let [^SnapshotDeletionPolicy snapshotter (.getIndexDeletionPolicy (.getConfig index-writer))]
+    (reify cp/CheckpointSource
+      (save-checkpoint [_ dir]
+        (.commit index-writer) ; RFC: do we need this?
```
refset, line 252: I think so! Based on "When you make a snapshot, it grabs the current commit point and holds a reference to it" via https://freecontent.manning.com/hot-backups-with-lucene/

jarohen: It does, but we also have the option of using the last commit point from however long ago the schedule ran.

I quite like having this `commit` here as it means the checkpoint is as up-to-date as it can be - the question is whether it's superfluous, or even hinders if the schedule runs at the same time.

crux-lucene/src/crux/lucene.clj
```diff
@@ -174,79 +172,142 @@
...
+                        (fn [{:keys [::tx/tx-id ::txe/tx-events committing?]}]
+                          (when committing?
+                            (let [{:keys [docs evicted-eids]} (transform-tx-events document-store tx-events)]
+                              (when-let [evicting-eids (not-empty evicted-eids)]
+                                (evict! indexer index-writer evicting-eids))
+                              (index! indexer index-writer docs)))
+
+                          (.setLiveCommitData index-writer {"crux.tx/tx-id" (str tx-id)
+                                                            "crux.lucene/index-version" (str index-version)})
+                          (.maybeRefreshBlocking searcher-manager)))
```
refset, line 330: In the other PR I debated whether to include this as part of the `l/await-tx` call https://github.com/juxt/crux/pull/1535#discussion_r653487129

...but now we're back to ~sync indexing I think it's fine/necessary to stay living here :+1:

crux-lucene/src/crux/lucene.clj
refset: Agreed, the commit frequency and the snapshot frequency feel like they should stay unrelated, and therefore it seems the least surprising option for the user.

jarohen:
> might Lucene have an option to fsync regularly anyway?

It _flushes_ periodically, depending on memory usage - this pushes in-memory segments down to persistent storage (to avoid memory pressure, it seems). This isn't a _commit_ though. The `IndexWriter` (by default) commits on normal close but, if it closes abnormally, we have to assume it won't commit, and the flushed-but-not-committed docs will be rolled back on next start?

jarohen:
> consider what situations would mean that we should abort the fsync loop. potentially IOException is the _only_ exception that we should try to recover from - should everything else kill the job?

On second thoughts, even `IOException` can represent both recoverable and unrecoverable errors. Happy to go with this for now - if we start to see errors that should abort the Crux ingestion, let's re-assess.

jarohen:
> confirm that `.commit` is thread-safe

https://lucene.apache.org/core/7_4_0/core/org/apache/lucene/index/IndexWriter.html

NOTE: IndexWriter instances are completely thread safe, meaning multiple threads can call any of its methods, concurrently.

Issue #1424, opened 2021-02-17
crux-lucene shouldn't need to fully evaluate all search results if only the first page are required


refset:
Limit behaviour was relaxed in https://github.com/juxt/crux/pull/1319 (where previously the result sets were arbitrarily limited to 1000) however now the full set of results must be held in memory and sorted before being made available for further processing in Datalog. For large result sets this behaviour could be problematic.

For real-time text-search based queries that require joins across the results of multiple Lucene predicates (e.g. powering a search form with multiple user-provided search input fields) it would be helpful to limit the number of results per predicate either by time (with a non-error timeout parameter) or space (with a limit parameter) to ensure that queries return within an acceptable time-frame for the end-user, even if the results are partial / incomplete from the database's perspective.

- offset by score? would need to sort by eids before returning

PR #1572, opened 2021-07-16
Enable Lucene predicates to apply limits for raw and returned results

refset:
A candidate resolution for #1424 

Whilst contemplating implementing equivalent `limit` functionality on top of the new `l/search` API (and also doing the temporal filtering in userspace), I thought it would be best to have something to that compare to.


jarohen:
Closing in favour of #1578, as per PR message over there

PR #1576, opened 2021-07-23
Create new Lucene analyzers per query

refset:
On the basis of https://stackoverflow.com/a/34498217 ...it seems Lucene's analyzers are not thread-safe, so queries may currently conflict with each other and the indexer. It should be fine to still use the long-lived analyzer created at startup (as a module dep) for index writing, but each query should generate its own new one instead.

I also included a few small tidy ups of the ns.

### DIFF
--- a/crux-lucene/src/crux/lucene.clj
+++ b/crux-lucene/src/crux/lucene.clj
@@ -154,7 +154,7 @@
           search-results))
 
 (defmethod q/pred-args-spec 'text-search [_]
-  (s/cat :pred-fn  #{'text-search} :args (s/spec (s/cat :attr keyword? :v (some-fn string? symbol?))) :return (s/? :crux.query/binding)))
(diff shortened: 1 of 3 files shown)


jarohen:
As discussed:

* merged the three tidy-up changes into master directly - cheers!
* we think the Lucene Analyzers _are_ thread-safe, because the [implementation](https://github.com/apache/lucene/blob/0a1dd10d5262153f4188dfa14a08ba28ec4ccb60/lucene/core/src/java/org/apache/lucene/analysis/Analyzer.java#L77) seems to use a ThreadLocal, which would mean they're safe to use from multiple threads.
* in the use-case that required different analyzers per field, it seems idiomatic to use `PerFieldAnalyzerWrapper` to specify the analyzer for each field - this would then propagate to the query as well

Cheers @refset!

refset:
I was hoping to run this ThreadLocal conclusion past the users on the SO thread, but I can't :) (perhaps someone else can post on our behalf?)

image

PR #1577, opened 2021-07-23
Support multiple Lucene stores via pred opts map with :lucene-store-k

refset:
Enabled by the recent secondary indexes changes in https://github.com/juxt/crux/pull/1555

### DIFF
--- a/crux-lucene/src/crux/lucene.clj
+++ b/crux-lucene/src/crux/lucene.clj
@@ -119,13 +119,18 @@
      (search* lucene-store
               (parse-query lucene-store query opts)))))
 
-(defn pred-constraint [query-builder results-resolver {:keys [arg-bindings idx-id return-type tuple-idxs-in-join-order ::lucene-store]}]
+(defn pred-constraint [query-builder results-resolver {:keys [arg-bindings idx-id return-type tuple-idxs-in-join-order] :as pred-ctx}]
   (fn pred-lucene-constraint [index-snapshot db idx-id->idx join-keys]
     (let [arg-bindings (map (fn [a]
                               (if (instance? VarBinding a)
                                 (q/bound-result-for-var index-snapshot a join-keys)
                                 a))
                             (rest arg-bindings))
+          last-arg (last arg-bindings)
+          [arg-bindings opts] (if (map? last-arg)
+                                [(butlast arg-bindings) last-arg]
+                                [arg-bindings nil])
+          lucene-store (get pred-ctx (or (:lucene-store-k opts) ::lucene-store))
           query (query-builder (:analyzer lucene-store) arg-bindings)
           tuples (with-open [search-results ^crux.api.ICursor (search* lucene-store query)]
                    (->> search-results
@@ -154,7 +159,7 @@
           search-results))
 
 (defmethod q/pred-args-spec 'text-search [_]
-  (s/cat :pred-fn  #{'text-search} :args (s/spec (s/cat :attr keyword? :v (some-fn string? q/logic-var?))) :return (s/? :crux.query/binding)))
+  (s/cat :pred-fn  #{'text-search} :args (s/spec (s/cat :attr keyword? :v (some-fn string? q/logic-var?) :opts (s/? (some-fn map? q/logic-var?)))) :return (s/? :crux.query/binding)))
 
 (defmethod q/pred-constraint 'text-search [_ pred-ctx]
   (let [resolver (partial resolve-search-results-a-v (second (:arg-bindings pred-ctx)))]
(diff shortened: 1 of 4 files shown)


crux-lucene/src/crux/lucene.clj
```diff
@@ -209,6 +217,7 @@
                         (.add (TextField. (keyword->k a), v, Field$Store/YES))
                         ;; Used for wildcard searches
                         (.add (TextField. field-crux-val, v, Field$Store/YES))
+                        ;; Used for eviction
```
refset, line 217: Added for completeness - but correct me if I'm wrong!

crux-lucene/src/crux/lucene/multi_field.clj
```diff
@@ -54,11 +54,9 @@
         search-results))
 
 (defmethod q/pred-args-spec 'lucene-text-search [_]
-  (s/cat :pred-fn #{'lucene-text-search} :args (s/spec (s/cat :query (some-fn string? symbol?) :bindings (s/* :crux.query/binding))) :return (s/? :crux.query/binding)))
```
refset, line 57: I'm fairly sure `:crux.query/binding` wasn't right here, since it covers more than just logic-vars.

crux-lucene/src/crux/lucene/multi_field.clj
```diff
@@ -54,11 +54,9 @@
...
+  (s/cat :pred-fn #{'lucene-text-search} :args (s/spec (s/cat :query (some-fn string? q/logic-var?) :bindings (s/* (some-fn string? q/logic-var?)) :opts (s/? (some-fn map? q/logic-var?)))) :return (s/? :crux.query/binding)))
```
refset, line 57: I added `string?` since it's free and useful.

crux-lucene/test/crux/lucene/extension_test.clj
```diff
@@ -0,0 +1,26 @@
+(ns crux.lucene.extension-test
+  (:require  [clojure.test :as t]
```
refset, line 2: ah, I left more requires than strictly needed here, but I will make use of them in the next PR

crux-lucene/test/crux/lucene/extension_test.clj
```diff
@@ -0,0 +1,26 @@
...
+(require 'crux.lucene.multi-field :reload) ; for dev, due to changing protocols
+
+(t/deftest test-multiple-lucene-stores
+  (let [n (c/start-node {:ave {:crux/module 'crux.lucene/->lucene-store
+                               :analyzer 'crux.lucene/->analyzer
+                               :indexer 'crux.lucene/->indexer}
+                         :multi {:crux/module 'crux.lucene/->lucene-store
+                                 :analyzer 'crux.lucene/->analyzer
+                                 :indexer 'crux.lucene.multi-field/->indexer}})]
+    (c/await-tx n (c/submit-tx n [[:crux.tx/put {:crux.db/id :ivan
```
jarohen: nit: submit+await-tx 🙂

crux-lucene/src/crux/lucene.clj
```diff
@@ -119,13 +119,21 @@
...
                                 a))
                             (rest arg-bindings))
+          {:keys [lucene-store-k] :as opts} (let [m (last arg-bindings)]
+                                              (when (map? m)
+                                                m))
+          arg-bindings (if opts
+                         (butlast arg-bindings)
+                         arg-bindings)
+          lucene-store (or (get pred-ctx lucene-store-k)
+                           lucene-store)
```
jarohen: A pattern I've seen for this kind of situation (i.e. where you're picking off optional args one at a time) is using tuple destructuring in a multiple-return sort of way:

```clojure
(let [last-arg (last arg-bindings)
      [arg-bindings {:keys [lucene-store-k] :as opts}] (if (map? last-arg)
                                                         [(butlast arg-bindings) last-arg]
                                                         [arg-bindings nil])
      lucene-store (or (get pred-ctx lucene-store-k)
                       lucene-store)]
  )
```

jarohen: Also, maybe slightly cleaner if you don't destructure `::lucene-store` out at the top?

```clojure
(let [last-arg (last arg-bindings)
      [arg-bindings opts] (if (map? last-arg)
                            [(butlast arg-bindings) last-arg]
                            [arg-bindings nil])
      lucene-store (get pred-ctx (or (:lucene-store-k opts) ::lucene-store))]
  )
```

jarohen (changes requested):
Nice one @refset - couple of small suggestions but otherwise good to go :smile:

crux-lucene/src/crux/lucene.clj
refset: nice suggestions, thanks!

PR #1578, opened 2021-07-28
Lucene extension examples

### LINKED ISSUE #1318: Bench/compare Lucene indexing strategies
Since we're introducing another alpha indexing/query strategy in https://github.com/juxt/crux/pull/1283 - single docs housing multiple fields, it would be good to bench this against the current A/V approach side by side. The outcome might be - if the multi-field doc route benches out OK - to retire the current A/V approach.

Note that the current A/V approach benefits from structural sharing, whereas the new approach has a Lucene doc per Crux document-version. An outcome could be that we offer both approaches in light of contrasting trade-offs.

### LINKED ISSUE #1424: crux-lucene shouldn't need to fully evaluate all search results if only the first page are required

### LINKED ISSUE #1425: Support a user-defined Lucene indexing transformer to handle specific document shapes
Users might wish to index particular fields with certain values (e.g. Date values) on certain types of documents using a Lucene/text-search friendly string encoding (e.g. ISO 8601 date-time).

This transformation configuration could only be safely (consistently!) altered prior to indexing, so any changes would require a full re-indexing.

### LINKED ISSUE #1460: Refactor crux-lucene to support slightly more advanced configurations
Decomposing the existing module into additional namespaces might help facilitate new requirements, in addition to the standard functionality, without ~forking the module. Motivated by https://github.com/juxt/crux/pull/1431

E.g. to support parallel usage of the AV index and the multi-field index (requiring twice the storage space, along with their respective predicates)

idea from the community:  "include the tx containing the tx-events that docs came from as well to include the info there, although I suppose crux in general doesn't use that info anywhere yet (vt index when?)"

use-case: be able to specify an :allowed-attributes or :ignored-attributes as a function (or, by default, set's IFn impl) of options to the lucene indexer to avoid indexing every string into lucene. E.g. indexing a million or so documents that can have 50-80 strings per document

### LINKED ISSUE #1532: verify that crux-lucene respects Crux's query timeout
particularly, wonder whether we'll need to convert Lucene's `ThreadInterruptedException` to a normal `InterruptedException` so that the query engine knows that the exception was caused by a timeout

refset:
1. Upgraded Lucene dependencies
2. Made a few things in lucene.clj public
3. Multiple Lucene stores test case (exercising the new `:lucene-store-k` opt)
4. Limits test case
5. Egeria test case (new take on #1431)
6. How-to test case (a guide will be written about this)
7. Updated documentation
8. l/search doc string

Propose to close #1318, #1572, #1532, #1460, #1424, #1425

### DIFF
--- a/crux-lucene/project.clj
+++ b/crux-lucene/project.clj
@@ -14,8 +14,9 @@
   :dependencies [[org.clojure/clojure "1.10.3"]
                  [pro.juxt.crux/crux-core]
 
-                 [org.apache.lucene/lucene-core "8.6.1"]
-                 [org.apache.lucene/lucene-queryparser "8.6.1"]]
+                 [org.apache.lucene/lucene-core "8.9.0"]
+                 [org.apache.lucene/lucene-queryparser "8.9.0"]
+                 [org.apache.lucene/lucene-analyzers-common "8.9.0" :scope "test"]]
 
   :profiles {:dev {:dependencies [[ch.qos.logback/logback-classic "1.2.3"]
                                   [pro.juxt.crux/crux-test]
--- a/crux-lucene/test/crux/lucene/extension_test.clj
+++ b/crux-lucene/test/crux/lucene/extension_test.clj
@@ -1,17 +1,27 @@
 (ns crux.lucene.extension-test
   (:require  [clojure.test :as t]
-             [clojure.spec.alpha :as s]
              [crux.api :as c]
-             [crux.checkpoint :as cp]
-             [crux.db :as db]
+             [crux.codec :as cc]
+             [crux.system :as sys]
              [crux.fixtures :as fix :refer [*api* submit+await-tx]]
-             [crux.fixtures.lucene :as lf]
-             [crux.lucene :as l]))
+             [crux.lucene :as l])
+  (:import org.apache.lucene.analysis.Analyzer
+           org.apache.lucene.queries.function.FunctionScoreQuery
+           [org.apache.lucene.analysis.core KeywordAnalyzer KeywordTokenizerFactory LowerCaseFilterFactory]
+           [org.apache.lucene.analysis.standard StandardTokenizerFactory]
+           org.apache.lucene.analysis.custom.CustomAnalyzer
+           org.apache.lucene.analysis.miscellaneous.PerFieldAnalyzerWrapper
+           [org.apache.lucene.document Document Field$Store StringField TextField]
+           [org.apache.lucene.index IndexWriter Term]
+           org.apache.lucene.queryparser.classic.QueryParser
+           [org.apache.lucene.search BooleanClause$Occur BooleanQuery$Builder TermQuery Query DoubleValuesSource]))
 
 (require 'crux.lucene.multi-field :reload) ; for dev, due to changing protocols
 
+(t/use-fixtures :each (fix/with-opts {::l/lucene-store {}}) fix/with-node)
+
 (t/deftest test-multiple-lucene-stores
-  (with-open [node (c/start-node {:ave {:crux/module 'crux.lucene/->lucene-store
+  (with-open [node (c/start-node {:eav {:crux/module 'crux.lucene/->lucene-store
                                         :analyzer 'crux.lucene/->analyzer
                                         :indexer 'crux.lucene/->indexer}
                                   :multi {:crux/module 'crux.lucene/->lucene-store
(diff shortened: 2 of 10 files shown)


crux-lucene/project.clj
```diff
@@ -14,8 +14,9 @@
   :dependencies [[org.clojure/clojure "1.10.3"]
                  [pro.juxt.crux/crux-core]
 
-                 [org.apache.lucene/lucene-core "8.6.1"]
-                 [org.apache.lucene/lucene-queryparser "8.6.1"]]
+                 [org.apache.lucene/lucene-core "8.9.0"]
+                 [org.apache.lucene/lucene-queryparser "8.9.0"]
+                 [org.apache.lucene/lucene-analyzers-common "8.9.0"]]
```
jarohen: Unless we need this for `crux-lucene` code, could we add this as a test dependency?

crux-lucene/test/crux/lucene/extension_test.clj
```diff
@@ -22,5 +32,253 @@
...
+                                                             :indexer 'crux.lucene.extension-test/->custom-eav-indexer}})]
+    (submit+await-tx node [[:crux.tx/put {:crux.db/id 0, :description "Some Entity"}]
+                           [:crux.tx/put {:crux.db/id 1, :description "Another entity"}]])
+
+    (let [db (c/db node)]
+      ;; text-search + leading-wildcard
+      (t/is (= [[0 "Some Entity" 2.0]
+                [1 "Another entity" 1.0]]
+               (let [a :description]
+                 (custom-text-search node db a (build-custom-query (->per-field-analyzer nil) (l/keyword->k a) "so* *En*")))))
```
jarohen, line 181: That the user has to pass the analyzer here is a bit of a smell, I wonder if there's a way to get hold of the analyzer from the node?

Doesn't have to be one for now :)

crux-lucene/test/crux/lucene/extension_test.clj
```diff
@@ -22,5 +32,253 @@
...
+                [1 :description "Another entity" 1.0]]
+               (custom-wildcard-text-search node db (build-custom-query (->per-field-analyzer nil) l/field-crux-val "*So* *En*"))))
+
+      ;; case-sensitive wildcard-text-search + leading wildcard
+      (t/is (= [[0 :description "Some Entity" 1.0]]
+               (custom-wildcard-text-search node db (build-custom-query (->per-field-analyzer nil) field-crux-val-exact "*En*")))))))
+
+;;; how-to example
+
+(defrecord exampleEavIndexer []
```
jarohen: nit: `ExampleEavIndexer`

crux-lucene/test/crux/lucene/extension_test.clj
```diff
@@ -22,5 +32,253 @@
...
+                       (distinct) ;; rely on the later resolve step to find all the entities sharing a given AV
+                       (mapcat (fn resolve-atemporal-av [[a v s]]
+                                 (c/q db {:find '[e a v s]
+                                          :in '[a v s]
+                                          :where [['e (keyword a) 'v]]}
+                                      (keyword a) v s))))
+                  (cond->> result-limit (take result-limit))
+                  (->> (into [])))))]
+    (with-open [db (c/open-db *api*)]
+      (t/is (= 5 (count (user-text-search-fn db "Entity*" 5)))))))
```
jarohen: subjective style: I'm not sure the `letfn` is making it clearer here?

```suggestion
(t/deftest test-userspace-wildcard-text-search-limit
  (submit+await-tx (for [n (range 10)] [:crux.tx/put {:crux.db/id n, :description (str "Entity " n) :foo (str "Entitybar" n)}]))
  (submit+await-tx (for [n (range 4)] [:crux.tx/put {:crux.db/id n, :description (str "Entity v2 " n) :foo (str "Entitybaz" n)}]))
  (submit+await-tx (for [n (range 4)] [:crux.tx/put {:crux.db/id (+ n 4), :description (str "Entity v2 " n) :foo (str "Entityqaz" n)}]))

  (with-open [db (c/open-db *api*)]
    (with-open [s (l/search *api* v {})]
      (t/is (= 5 (count (-> (iterator-seq s)
                            (->> (map (fn doc->rel [[^Document doc score]]
                                        [(.get ^Document doc l/field-crux-attr) (.get ^Document doc l/field-crux-val) score]))
                                 (distinct) ;; rely on the later resolve step to find all the entities sharing a given AV
                                 (mapcat (fn resolve-atemporal-av [[a v s]]
                                           (c/q db {:find '[e a v s]
                                                    :in '[a v s]
                                                    :where [['e (keyword a) 'v]]}
                                                (keyword a) v s))))
                            (cond->> result-limit (take result-limit)))))))))

```

docs/reference/modules/ROOT/pages/lucene.adoc
```diff
@@ -184,6 +181,45 @@ You can escape your input strings when constructing Lucene query strings by call
...
+
+[source,clojure]
+----
+(with-open [search-results (crux.lucene/search node "Ivan")]
+  (into [] (iterator-seq search-results)))
+----
+
+Each item returned will be a vector of `org.apache.lucene.document.Document` and a Double representing the matched score.
+
+See the https://github.com/juxt/crux/blob/master/crux-lucene/test/crux/lucene/extension_test.clj[extension tests] for examples of decoding the contents of the result document and performing userspace temporal filtering.
```
jarohen, line 203: I'm not generally a fan of pointing people at our tests within the docs, feels like it's putting the work onto the user to understand what's going on in our tests?

jarohen: Maybe something akin to the Corda reference docs would work better here?

jarohen (changes requested):
Cheers @refset - a few minor nits here. Worth a discussion offline regarding whether we should be pointing users to our tests as a means of documentation?

docs/reference/modules/ROOT/pages/lucene.adoc
jarohen: As discussed, one for a separate how-to

crux-lucene/test/crux/lucene/extension_test.clj
refset: Yeah...that's exactly why `custom-text-search` is just example code for the time being :)

crux-lucene/test/crux/lucene/extension_test.clj
refset: As discussed, I was thinking of users that might want to quickly copy-paste the function, but I'm happy either way :)

PR #1580, opened 2021-07-28
Make tx-committed? throw on nil tx-id, fixes #1579

### LINKED ISSUE #1579: tx-committed? strange results
`(crux/tx-committed? crux {:what :ever})` returns true but probably shouldn't. I ran into this after copy-pasting the resulting transaction map in the REPL.

refset:
Cases:

1. If tx-id is nil due to either the user providing it as `nil` in the passed map, or by leaving out the key from whatever map is provided (as per https://github.com/juxt/crux/issues/1579), then `err/node-out-of-sync` will be thrown
2. If tx-id isn't a Long then the `compare` will throw, although e.g. Doubles will be coerced
3. Negative tx-ids or "skipped" tx-ids (e.g. due to use of Kafka offsets rather than increments) will still erroneously return `true`, indicating that they are committed transactions (even though they aren't transactions at all), but this PR doesn't attempt resolve that. It seems an unlikely edge case
4. I think that covers all the cases

Aside: short of changing the implementation, perhaps inverting the API to `tx-failed?` would be more correct & intuitive.

This also removes a superfluous level of `cio/with-read-lock` introduced accidentally a few months ago: https://github.com/juxt/crux/commit/5aa08919c00c95b8567de7a4f91ede9d908f522c#diff-916b3a9365c6ef73a495a907307258d907bb8e5bced78616758ce70f0fab8e1eR127-R128

### DIFF
--- a/crux-core/src/crux/node.clj
+++ b/crux-core/src/crux/node.clj
@@ -126,14 +126,17 @@
         (merge crux-version
                (status (dissoc @!system :crux/node))))))
 
-  (tx-committed? [this {:keys [::tx/tx-id ::tx/tx-time] :as submitted-tx}]
+  (tx-committed? [this {:keys [::tx/tx-id] :as submitted-tx}]
     (cio/with-read-lock lock
-      (cio/with-read-lock lock
-        (ensure-node-open this)
-        (let [{latest-tx-id ::tx/tx-id, :as latest-tx} (api/latest-completed-tx this)]
-          (if (and tx-id (or (nil? latest-tx-id) (pos? (compare tx-id latest-tx-id))))
-            (throw (err/node-out-of-sync {:requested submitted-tx, :available latest-tx}))
-            (not (db/tx-failed? index-store tx-id)))))))
+      (ensure-node-open this)
+      (let [{latest-tx-id ::tx/tx-id, :as latest-tx} (api/latest-completed-tx this)]
+        (cond
+          (nil? tx-id) (throw (err/illegal-arg :invalid-tx {:tx submitted-tx}))
+
+          (or (nil? latest-tx-id) (pos? (compare tx-id latest-tx-id)))
+          (throw (err/node-out-of-sync {:requested submitted-tx, :available latest-tx}))
+
+          :else (not (db/tx-failed? index-store tx-id))))))
 
   (sync [this] (api/sync this nil))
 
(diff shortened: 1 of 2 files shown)


jarohen:
Cheers @refset :) 

Is there a use-case for someone legitimately submitting `nil` for `tx-id`? If not, reckon we should throw `err/illegal-arg` instead?

refset:
> Is there a use-case for someone legitimately submitting nil for tx-id? If not, reckon we should throw err/illegal-arg instead?

It seems possible that someone might want to routinely pipe the result of `latest-submitted-tx`, which is `nil` before the first transaction has been submitted, into `tx-committed?`.

jarohen:
To what end, though?

refset:
Some kind of programmatic monitoring, like "Give me a stream of all the failed transactions so I can figure out why they're failing and do something about it" - I realise this is ~mostly already covered through logging.

Or perhaps for some kind of stateless backoff & retry logic, i.e. where any node may need to re-submit a tx after a failed match operation appears on the log.

jarohen:
I don't think you'd get either from `latest-submitted-tx`, so `nil` is still an invalid parameter?

> Some kind of programmatic monitoring, like "Give me a stream of all the failed transactions so I can figure out why they're failing and do something about it" - I realise this is ~mostly already covered through logging.

You'd get your stream of transactions from `open-tx-log`, which are all valid transactions.

> Or perhaps for some kind of stateless backoff & retry logic, i.e. where any node may need to re-submit a tx after a failed match operation appears on the log.

You'd also have a valid tx in hand in this case.

Let's go with `err/illegal-arg`?

refset:
It's true, no-one will be polling `latest-submitted-tx` to generate such a stream :)

As an aside (and mentioned on the call just now), it's perhaps worth pointing out that anyone looking to scan from the original log (i.e. to include potentially failed transactions) can use the `open-tx-log` of the ingest client which necessarily has an unfiltered implementation:
https://github.com/juxt/crux/blob/63e99523b73478490618c99760f3e28e2e1acf33/crux-core/src/crux/ingest_client.clj#L22-L26
vs 
https://github.com/juxt/crux/blob/63e99523b73478490618c99760f3e28e2e1acf33/crux-core/src/crux/node.clj#L204-L227

> Let's go with err/illegal-arg?

Agreed!

PR #1583, opened 2021-07-29
Change logging to WARN for match failures (instead of DEBUG)

refset:
This brings match in line with transaction function and CAS logging.

### DIFF
--- a/crux-core/src/crux/tx.clj
+++ b/crux-core/src/crux/tx.clj
@@ -106,7 +106,7 @@
                                                tx-id)
         match? (= (c/new-id content-hash) (c/new-id v))]
     (when-not match?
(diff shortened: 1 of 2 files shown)


jarohen:
@refset: (as discussed on Slack) I'm inclined to keep this as DEBUG. I can't find a definitive guide as to what kind of messages should be what level, but IME it's often the case that INFO should be a good default level for folks to log across the board. WARN is for situations that, while they don't require immediate attention (ERROR or FATAL, depending on whether it's affected one operation or the whole system), should be investigated - I don't think a match failure warrants this? If match failures did became WARNs, we'd alsp potentially run the risk of users considering them too chatty, and turn WARN logging off for Crux.

A transaction function exception does warrant a WARN - transaction functions shouldn't throw exceptions in normal operation, and it's likely indicative of a problem in the code.

PR #1587, opened 2021-08-04
Tidy up Lucene 1.18 EAV changes

refset:
## Querying

In 1.18 we moved from AV storage to EAV storage for Lucene, so now adding a `distinct` step in the resolver fns is essential to avoiding a cartesian product of lookups.

This is essentially the same as what was discussed previously for the userspace limit fn https://github.com/juxt/crux/blob/master/crux-lucene/test/crux/lucene/extension_test.clj#L38-L54

Unfortunately we can't introduce the `distinct` within `l/pred-constraint` generically, since it won't work across the equality of the Lucene documents.

~~Alternatively we could revert the move to EAVs (and the removal of `exclusive-avs`).~~

## Indexing

I have attempted to re-use as many of the Lucene objects (Documents and Fields) where possible to avoid GC churn, although this seems to only offer very marginal to benefit performance, if any. Motivated by a suggestion in https://cwiki.apache.org/confluence/display/lucene/ImproveIndexingSpeed

~~I have stripped out field-crux-id and change updateDocument to addDocument since there is no longer any structural sharing.~~

~~I also moved `maybeRefreshBlocking` to the start of `search*`. This might slow down searches slightly but in the grand scheme of things is probably the better tradeoff.~~

~~These two changes alone saves 70% of indexing time on my local testing (approx. 120s -> 35s).~~

I also observed that use of pmap within `index!` shaves off a further few seconds, which suggests it _might_ be worth having a threadpool (but I've not added anything in this PR).
```clojure
(doall (pmap
            (fn add-doc [{e :crux.db/id, :as crux-doc}]
              (doseq  [[a v] (->> (dissoc crux-doc :crux.db/id)
                                  (mapcat (fn [[a v]]
                                            (for [v (cc/vectorize-value v)
                                                  :when (string? v)]
                                              [a v]))))
                       :let [doc (doto (Document.)
                                   ;; The actual term, which will be tokenized
                                   (.add (TextField. (keyword->k a), v, Field$Store/YES))
                                   ;; Used for wildcard searches
                                   (.add (TextField. field-crux-val, v, Field$Store/YES))
                                   ;; Used for eviction
                                   (.add (StringField. field-crux-eid, (->hash-str e), Field$Store/NO))
                                   ;; Used for wildcard searches
                                   (.add (StringField. field-crux-attr, (keyword->k a), Field$Store/YES)))]]
                (.addDocument ^IndexWriter index-writer doc)))
            (vals docs)))
```

### DIFF
--- a/crux-lucene/src/crux/lucene.clj
+++ b/crux-lucene/src/crux/lucene.clj
@@ -35,8 +35,6 @@
 (defn ^String ->hash-str [eid]
   (str (cc/new-id eid)))
 
-(defrecord DocumentId [e a v])
-
 (defn ^String keyword->k [k]
   (subs (str k) 1))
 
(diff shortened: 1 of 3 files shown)


jarohen:
Cheers @refset - could you also add a test that fails under the current code but passes under your proposed fix?

jarohen (changes requested):
Cheers @refset - could you also add a test that fails under the current code but passes under your proposed fix?

refset:
I can't think of a good test to write that's not a slowdown test - any suggestions?

Short of doing that though, I was able to prove it quite quickly:
image

```
"Elapsed time: 274.578581 msecs" ; updated resolver
"Elapsed time: 472811.735454 msecs" ; OOM on original resolver
14:37:44.728 [crux-lucene-fsync-1] WARN  crux.lucene - error during Lucene IndexWriter commit
java.lang.OutOfMemoryError: Java heap space
```

crux-lucene/src/crux/lucene.clj
```diff
@@ -238,7 +240,7 @@
                         (.add (StringField. field-crux-eid, (->hash-str e), Field$Store/NO))
```
jarohen: move `(->hash-str e)` up in the `doseq`, above the `[a v]` binding

jarohen: consider re-using Fields here, even just within a Crux document

jarohen: also consider bringing `(keyword->k a)` out to the batch à la `->content-idx-kvs`

jarohen:
> I can't think of a good test to write that's not a slowdown test - any suggestions?

```clojure
(t/is (= (count search-results)
         (count (resolve-search-results-a-v (c/->id-buffer :foo) index-snapshot db search-results))))
```

crux-lucene/src/crux/lucene.clj
```diff
@@ -226,10 +231,7 @@
                                  (for [v (cc/vectorize-value v)
                                        :when (string? v)]
                                    [a v]))))
-            :let [id-str (->hash-str (->DocumentId e a v))
```
jarohen, line 229: we do need this, because of duplication over time - but it could be a vector so that we're not tempted to sort the map

crux-lucene/src/crux/lucene.clj
```diff
@@ -70,6 +69,7 @@
   (user-data->tx-id (into {} (.getLiveCommitData index-writer))))
 
 (defn- search* [{:keys [^SearcherManager searcher-manager]}, ^Query q]
+  (.maybeRefreshBlocking searcher-manager)
```
jarohen: as discussed, let's err on the side of consistency here - we can consider a catching-up mode or similar.

jarohen:
Ideas from https://cwiki.apache.org/confluence/display/lucene/ImproveIndexingSpeed

* Add option to set RAM buffer size?
* Ensure we're not auto-committing

crux-lucene/src/crux/lucene.clj
refset: vectors aren't supported for IdToBuffer so I'm going with string concatenation instead

refset:
> Add option to set RAM buffer size?

I've added the config option, but it doesn't appear to have any impact. I experiment with the related `setRAMPerThreadHardLimitMB` also without success (though I didn't add a config option for this - ...should I?)

> Ensure we're not auto-committing

I think the note from https://cwiki.apache.org/confluence/display/lucene/ImproveIndexingSpeed on this is only relevant for much older versions. I can't obviously see that autoCommit (or equivalent) is still a thing.

refset:
Closing this in favour of #1589 

I will open a separate PR for other nice-to-have changes later:

- [ ] re-using Fields and Document, except unlike the commit in this PR, only add/remove the one changing field (just use existing structure mutably)
- [ ] add index-version to indexer protocol, takes a string for descriptive versioning ("eavindexer-1"), bump lucene index version(s) to 2
- [ ] configurable `search*` batch size - `{:keys [batch-size]}` opts map with two-arity versions
- [ ] microbench `DocumentId` change, e.g. current Record+Nippy, vs `id-str (str (count e-str) "." e-str (count k-str) "." k-str v)` (which is a more correct version that the commit in this PR)
- [ ] use byte array instead of str for `->hash ids`, probably only in place of e-str and id-str (looking at this PR)

PR #1589, opened 2021-08-05
Lucene fixes for 1.18 users

refset:
## During query

1. fix cardinality of AV resolver results

## During ingest

1. move maybeRefresh work into the fsync-loop thread
2. set `applyAllDeletes` to true


refset:
Should I add a note to the docs about the possible `start-node-with-disabled-refresh -> sync -> .close -> start-node-normal` initialisation pattern?

Issue #1337, opened 2020-12-21
evict then re-put to the same entity in the same transaction is broken


jarohen:
~put then evict: the evict doesn't always clear up the put~
evict then put: the evict sometimes clears the documents associated with the subsequent put

PR #1590, opened 2021-08-06
Fix Lucene put-then-evict indexing

refset:
It's a very minor edge case, but the current behaviour allows `put` documents from the same transaction to survive eviction in Lucene.

### DIFF
--- a/crux-lucene/src/crux/lucene.clj
+++ b/crux-lucene/src/crux/lucene.clj
@@ -261,11 +261,14 @@
   (->> tx-ops
        (map txc/conform-tx-op)
        (mapcat txc/flatten-tx-fn-ops)
-       (reduce (fn xf [acc {:keys [op docs] :as tx-op}]
-                 (if (= op :crux.tx/evict)
-                   (update acc :evicted-eids conj (:eid tx-op))
-                   (update acc :docs into docs)))
-               {:docs #{}
+       (reduce (fn xf [acc {:keys [op docs eid]}]
+                 (case op
+                   :crux.tx/put (update acc :docs into docs)
+                   :crux.tx/evict (-> acc
+                                      (update :evicted-eids conj eid)
+                                      (update :docs dissoc #(into {} (remove (comp #{eid} :crux.db/id val)) %)))
+                   acc))
+               {:docs {}
                 :evicted-eids #{}})))
 
 (defn- checkpoint-src [^IndexWriter index-writer]
--- a/crux-lucene/test/crux/lucene_test.clj
+++ b/crux-lucene/test/crux/lucene_test.clj
@@ -451,6 +451,27 @@
                   {:find '[?e]
                    :where '[[(text-search :name "Ivan") [[?e]]]]})))))
 
+(t/deftest test-put-then-evict
+  (submit+await-tx [[:crux.tx/put {:crux.db/id :foo :bar "baz"}]
+                    [:crux.tx/evict :foo]])
+  (t/is (= #{} (c/q (c/db *api*) '{:find [e]
+                                   :where [[e :crux.db/id]]})))
+
+  (with-open [search-results (l/search *api* "b*" {:default-field "bar"})]
+    (let [docs (iterator-seq search-results)]
+      (t/is (= 0 (count docs)))))
+
+  (submit+await-tx [[:crux.tx/put {:crux.db/id :foo :bar "baz"}]
+                    [:crux.tx/evict :foo]
+                    [:crux.tx/put {:crux.db/id :qux :bar "qaz"}]])
+
+  (t/is (= #{[:qux]} (c/q (c/db *api*) '{:find [e]
+                                         :where [[e :crux.db/id :qux]]})))
+
+  (with-open [search-results (l/search *api* "b*" {:default-field "bar"})]
+    (let [docs (iterator-seq search-results)]
+      (t/is (= 0 (count docs))))))
+
 (comment
   (do
     (import '[ch.qos.logback.classic Level Logger]


crux-lucene/src/crux/lucene.clj
```diff
@@ -262,15 +262,18 @@
                             (txc/conformed-tx-events->doc-hashes conformed-tx-events))]
     (reduce (fn [acc {:keys [op] :as tx-event}]
               (case op
-                :crux.tx/evict (-> acc (update :evicted-eids conj (:eid tx-event)))
+                :crux.tx/evict (let [evicted-eids (conj (:evicted-eids acc) (:eid tx-event))]
+                                 (-> (apply dissoc acc :docs evicted-eids)
```
jarohen: I don't think this does do what you want it to - try adding puts of other entities in the transaction and check whether those docs still end up in Lucene? i.e. put a, evict a, put b, query for b.

jarohen (changes requested):
Good spot @refset :clap: 

I don't think this is quite the full story, unfortunately. You're right in saying the current behaviour is incorrect in that it under-evicts, but I think this (bug aside) has the opposite problem - if you were to evict a document and re-put it, it wouldn't be re-added to Lucene. 

To fix this, we'd have to play the events in order, building up state as we go, which would mean combining `index!` and `evict!` on the protocol. Think we chatted about this before the last release, will see if I can dig out the discussion.

crux-lucene/src/crux/lucene.clj
refset: I just extended the test to try to capture your suggestion (which passes) - does it help?

refset:
Ah yes, you're right that the evict-then-put case isn't addressed by this solution. (Un)fortunately/coincidentally it seems like evict-then-put in #1337 isn't resolved still though, so this PR would at least make Lucene consistent with the main index store's (incorrect) behaviour.

jarohen (changes requested):
Hey @refset,

I've rebased this on the Lucene changes in master - turns out your test passes against current master without any changes to `crux.lucene` itself, I'm not sure why it would. :confused: I reckon, if anything, this means that the test isn't quite finding the bug, rather than the change on master having fixed the issue.

Here's the change you made to `transform-tx-events` in https://github.com/juxt/crux/blob/a138130f0d2456495c05c19a78ce88f0cf5d67a4/crux-lucene/src/crux/lucene.clj, just in case it gets GC'd:

```clojure
(defn- transform-tx-events [document-store tx-events]
  (let [conformed-tx-events (map txc/<-tx-event tx-events)
        docs (db/fetch-docs document-store
                            (txc/conformed-tx-events->doc-hashes conformed-tx-events))]
    (reduce (fn [acc {:keys [op] :as tx-event}]
              (case op
                :crux.tx/evict (let [evicted-eids (conj (:evicted-eids acc) (:eid tx-event))]
                                 (-> (apply dissoc acc :docs evicted-eids)
                                     (assoc :evicted-eids evicted-eids)))
                :crux.tx/fn (let [{:keys [args-content-hash]} tx-event
                                  ;; doc replaced by this point
                                  nested-events (-> (db/fetch-docs document-store #{args-content-hash})
                                                    (get args-content-hash)
                                                    :crux.db.fn/tx-events)
                                  {:keys [docs evicted-eids]} (transform-tx-events document-store nested-events)
                                  evicted-eids* (into (:evicted-eids acc) evicted-eids)]
                              {:docs (into (:docs acc) (apply dissoc docs :docs evicted-eids*))
                               :evicted-eids evicted-eids*})
                acc))
            {:docs docs
             :evicted-eids #{}}
            conformed-tx-events)))
```

Seems there are a couple of options:
* Confirm that the test is correctly passing - the change to master did miraculously and unintentionally fix the issue.
* Try to make the test fail again, and re-apply the fix to `transform-tx-ops`.

Cheers!

James

refset:
I can confirm that the test as-is _doesn't_ fail on master (the current master, anyway) because post-eviction-processed `:crux.tx/put` `docs` are _almost_ being indexed into Lucene, that look like  `{#crux/id "73b2c3e644442b3a2244551117b14f3af7f2e53e" {:crux.db/id #crux/id "0beec7b5ea3f0fdbc95d0dd47f3c5bc275da8a33", :crux.db/evicted? true}}`, except there are no strings for Lucene to index so it's ~technically okay, but not ideal.

In a somewhat similar pattern, successful `match` ops _are_ having their associated doc indexed into Lucene though, which is needless work. This is clear to see just by adding a `(prn op docs)` in `transform-tx-ops` and running some small tests manually.

The proposed fix now simply filters these ops/docs from Lucene's purview, for correctness and efficiency. I haven't tried to add a test to prove the old behaviour since they are both such transient effects.

p.s. could do `(comp (partial not= eid) ...` instead of  `(comp not #{eid} ...` :)

```clojure
(update {:docs {'a {:crux.db/id :a}
                'b {:crux.db/id :b}}}
        :docs
        #(into {} (filter (comp (partial not= :a) :crux.db/id val)) %))
;; => {:docs {b #:crux.db{:id :b}}}
```

jarohen (commented):
sorry for the bad steer here @refset - docs is a set

refset:
Aha, good spot!

Interestingly I think it is _currently_ (on master) returning a set of `MapEntry`s, which bizarrely works fine when `conform-tx-ops` is used for `index!` where `vals` gets called on it. For some reason I can only repro that approx. sequence and get a CCE when working with literals:

```clojure
(-> {:docs #{}}
    (update :docs into {'a {:crux.db/id :a}})
    :docs)
;;=> #{[a #:crux.db{:id :a}]}

;; works fine
(-> {:docs #{}}
    (update :docs into {'a {:crux.db/id :a}})
    :docs
    vals)
;;=> (#:crux.db{:id :a})

(vals #{['a {:crux.db/id :a}]})
;;=> java.lang.ClassCastException
```
I'll fix it in any case. I realised also that the removal should happen in the evict branch :)

jarohen (changes requested):
getting a test failure now?

refset:
(CI doing its job :sweat_smile: ) I didn't spot that the `acc`'s returned `:docs` does actually have to be a map to support the other indexers that do need to use the content hash, but it means we can use `dissoc` now though!

:bulb: It would be nice if Github had a "re-request review if CI passes" button

jarohen:
Cheers @refset :pray: 

As discussed, inadvertently fixed in parallel in https://github.com/juxt/crux/commit/5e8a0a6ae11ac6234b9b1f81aa50aa544a9d9741 because the documents are now attached from the in-tx document store, which has already replaced the put-then-evicted documents with tombstones - Lucene never sees the original document in this case.

Still good to have the test and the associated tidy-up - thanks!

Issue #1485, opened 2021-03-31
'S3 key not found: UnsafeBuffer{...}' log message


jarohen:
From @nivekuil, https://clojurians-log.clojureverse.org/crux/2021-01-11

> I'm puzzled by this log line: 
> `2021-01-11 03:40:32 2021-01-11T11:40:32.623Z 395282b8415d WARN [crux.s3:326] - S3 key not found: UnsafeBuffer{addressOffset=16, capacity=21, byteArray=[B@4652fc81, byteBuffer=null}`
> 
> it seems `crux.s3` has never been 326 lines long and `s3-key` is created by `str` so it shouldn't be an UnsafeBuffer in the log? 
> https://github.com/juxt/crux/blob/master/crux-s3/src/crux/s3.clj#L54

jarohen:
Adding this for history - we haven't managed to repro this since the initial report. Bugs fixed since include #1105 and #1387, so these may well have fixed the underlying issue. If anyone does run into this again, please do shout and we'll re-open the issue :pray:

refset:
This has now been reported by a different user running `21.06-1.17.1-beta` (which includes the fixes mentioned above), along with the other related WARN from `crux-s3`:

> There are 34 occurrences of "Error fetching S3 object" and 1 occurrence of "S3 key not found" for this service in the past weeks

Transaction functions are in play also, which I expect increases the probability of these incidents (due to the additional operations).

Whilst a resolution to this issue is outstanding, I suggest these WARNs should be treated like FATALs (i.e. cycle the node immediately & remove the indexes before restarting, and probably also delete all checkpoints in case if there's a chance that an inconsistent index could have be uploaded):
https://github.com/juxt/crux/blob/3e9f48b5ea9fa6815a5ef216ec65059db9219359/crux-s3/src/crux/s3.clj#L49-L51

It may also be prudent to avoid using s3 as the doc-store for the time being.

## Speculation

I believe the changes made in https://github.com/juxt/crux/pull/1470 _should_ halt the node on missing docs, but it seems the failure to return docs from `crux-s3` doesn't trigger the halting for some reason, and therefore all the user sees are WARNs in the logs (and subtly inconsistent nodes).

The "Error fetching S3 object" errors could have a variety of causes but the "S3 key not found" suggests that the strongly-consistent contract of `crux-s3` is being broken (i.e. https://aws.amazon.com/s3/consistency/)

PR #1602, opened 2021-08-18
Strict fetch-docs in transaction ingestion (#1485)

jarohen:
See #1485.

This ensures that everywhere in the transaction ingestion uses a stricter `fetch-docs` that fails if it can't find any of the documents.

I'm not 100% convinced about this one as yet - need to re-familiarise myself with the context around why `db/fetch-docs` can't fail on missing docs.

### DIFF
--- a/crux-core/src/crux/tx.clj
+++ b/crux-core/src/crux/tx.clj
@@ -44,6 +44,49 @@
 (s/def ::ingester-error #(instance? Exception %))
 (defmethod bus/event-spec ::ingester-error [_] (s/keys :req-un [::ingester-error]))
 
+(defn- tx-fn-doc? [doc]
+  (some #{:crux.db.fn/args
+          :crux.db.fn/tx-events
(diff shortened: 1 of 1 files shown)


jarohen:
History:

* Handling doc-stores being eventually consistent: https://github.com/juxt/crux/issues/1105 (https://github.com/juxt/crux/pull/1178). Blocked on ingest, caused #1387 
* https://github.com/juxt/crux/pull/1325/
* https://github.com/juxt/crux/issues/1387 (Discussion https://github.com/orgs/juxt/teams/crux-core/discussions/25, https://github.com/juxt/crux/pull/1470, Kafka https://github.com/juxt/crux/pull/1493)

I suspect that the change to Kafka in #1493 means that we could go (back) to `db/fetch-docs` throwing on missing docs, but this is a larger change that I'd want to consider more thoroughly. Will open a separate issue.

PR #1604, opened 2021-08-18
Improve lucene wildcard search performance on highly recurrent attr->val sets

kevinmershon:
At Ride Health we're using standardized keys reused across records with some DRY-violation for the purposes for record completeness, for example:

Passenger:
  - :crux.db/id 1
  - :person name first: John

Ride:
  - :crux.db/id 2
  - passenger id: 1
  - person name first: John

Because of the nested loop used in populating A/V/score over the lookup for documents, this means the search result space is artificially inflated to N^2 with a majority of the result space being documents looked up repeatedly. This is very inefficient. In our production environment, searching John was taking between 42 and 95 seconds to return all the 4074 matching documents.

With the optimization in this PR, the search time drops to 1008ms for the same results and scoring order.

### DIFF
--- a/crux-lucene/src/crux/lucene.clj
+++ b/crux-lucene/src/crux/lucene.clj
@@ -147,12 +147,23 @@
   "Given search results each containing a single A/V pair document,
   perform a temporal resolution against A/V to resolve the eid."
   [index-snapshot {:keys [entity-resolver-fn] :as db} search-results]
-  (mapcat (fn [[^Document doc score]]
-            (let [v (.get ^Document doc field-crux-val)
-                  a (keyword (.get ^Document doc field-crux-attr))]
-              (for [eid (doall (db/ave index-snapshot a v nil entity-resolver-fn))]
-                [(db/decode-value index-snapshot eid) v a score])))
-          search-results))
+  (->> search-results
+       (reduce (fn [m [^Document doc score]]
+                 (let [v (.get ^Document doc field-crux-val)
+                       a (keyword (.get ^Document doc field-crux-attr))]
+                   (if (get m [a v])
+                     (update m [a v] (fn [exist-score]
+                                       (min score exist-score)))
+                     (assoc m [a v] score))))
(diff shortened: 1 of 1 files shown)


kevinmershon:
Ah, you guys already handled this in the latest master. Closing.

refset:
Ah :) well thank you anyway @kevinmershon - it's useful to hear that you hit this also (given we have ~zero visibility on the shapes of queries that most users are actually running!). We'll have 1.18.1 released very soon, but in the meantime this fix is in the `1.18.1-beta1` preview release.

refset:
https://github.com/juxt/crux/pull/1604/files#diff-97fb172b1e6d0b14e8f78e6bdcb364d8fef607cf1ad68e42d97e6404744b9e9fR154-R157

```clojure
                   (if (get m [a v])
                     (update m [a v] (fn [exist-score]
                                       (min score exist-score)))
                     (assoc m [a v] score))))
```

Looking at your proposed fix, which does seem to do roughly the same thing as the change I made (in https://github.com/juxt/crux/pull/1589/files#diff-97fb172b1e6d0b14e8f78e6bdcb364d8fef607cf1ad68e42d97e6404744b9e9fR178), I believe we can forego the `min` step since duplicate matches against the exact same Lucene document _should_ always have the exact some score - though I'd be keen to hear if that assumption of mine/ours is wrong or there's some edge case you've foreseen.

kevinmershon:
@refset It actually should be max (max score ranked highest) 🤦 

That said, I think the solution you guys have is sufficient. It takes ~100ms longer than what I proposed, I think because the extra iterations in the first loop produce values that then get dropped with the distinct, but I'm not going to complain about a 100ms difference when previously we were seeing 2 orders of magnitude longer request times...

refset:
There might be some efficiency lost in the chunking, but the `(-> coll (map ...) (distinct ...))` part _should_ be similarly lazy/incremental, I think.

PR #1608, opened 2021-09-01
Add `nvd-check` to run in CircleCI.

### LINKED ISSUE #1606: Use `lein-nvd` to locate dependencies with CVEs
(thanks @vemv for the pointer :pray:)

https://github.com/rm-hull/lein-nvd

danmason:
(See [lein-nvd](https://github.com/rm-hull/lein-nvd))

Runs `lein nvd check` for every commit onto CI - uploads the html version of the report as an artifact.

Currently will fail the build every time currently due to the nvd-check. As per the lein-nvd options you can configure what level of severity causes a build to fail (or just make it never fail builds if you set that high enough).

Closes #1606

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,6 +121,19 @@ jobs:
       - store_test_results:
           path: "crux-test/target/test-results"
 
+  nvd-check:
+    docker:
+      - image: circleci/clojure:openjdk-8-lein
+    working_directory: ~/crux
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/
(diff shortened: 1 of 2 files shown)


jarohen:
Thanks @danmason, this is neat :pray:

* Yep, til we've had an initial bash at the dependencies, could you make it not fail the build? Will raise a card to fix the vulns, but don't want to get into the habit of ignoring a failing build.
* Could you add it to a separate profile? Think the plugin itself brings in a number of dependencies that we don't necessarily want in our `:dev` profile.

Cheers!

PR #1659, opened 2021-11-10
Try restoring Lucene from checkpoint

tatut:
Lucene currently does not actually try to restore from a checkpoint. Add the restore call.

Noticed this while investigating very slow startup in ECS. We have lots of transactions and this greatly speeds up startup time.


jarohen:
Cheers @tatut - have cherry-picked 73b0b95 :pray:

PR #1681, opened 2022-01-02
Feature/add java type safety

cartermyers:
This change came out of an issue I was experiencing detailed here: https://juxt-oss.zulipchat.com/#narrow/stream/194466-xtdb-users/topic/.E2.9C.94.20SQL.20Where.20clause.20errors/near/264782867

@refset helped me solve the problem, and he also submitted a PR here to verify the types are correct: #1667 .
That PR is helpful because it would've pointed out the problem to me sooner, but I prefer to have the types be coerced for easier use in Java/Kotlin. It's kind of a pain to create `PersistentVector`s every time you need one.

Thanks for taking a look at this PR. I'm not sure if this is up to the XTDB standards, so all comments are welcome :)

### DIFF
--- a/core/src/xtdb/api/XtdbDocument.java
+++ b/core/src/xtdb/api/XtdbDocument.java
@@ -5,10 +5,12 @@ import clojure.lang.IPersistentMap;
 import clojure.lang.Keyword;
 import clojure.lang.PersistentArrayMap;
 
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Objects;
+import java.net.URI;
+import java.net.URL;
+import java.util.*;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public final class XtdbDocument {
     private static final Keyword DB_ID = Keyword.intern("xt/id");
@@ -18,6 +20,7 @@ public final class XtdbDocument {
     private final IPersistentMap data;
 
     private XtdbDocument(Object id, IPersistentMap data) {
+        assertValidIdType(id);
         this.id = id;
         this.data = data;
     }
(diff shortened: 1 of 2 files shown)


core/src/xtdb/api/XtdbDocument.java
```diff
@@ -171,6 +175,38 @@ private static void assertNotReserved(Keyword key) {
         if (FN_ID.equals(key)) throw new IllegalArgumentException(":xt/fn is a reserved key");
     }
 
+    private static void assertValidIdType(Object id) {
+        final Stream<Class<?>> allowedIdTypes = Stream.of(
```
jarohen, line 179: Rather than creating this for every document, could we create a `private final static Set<..> ALLOWED_ID_TYPES` of these classes?

jarohen (changes requested):
Hey @cartermyers, thanks for submitting this PR, much appreciated :pray: 

Your casting change looks fine, but unfortunately there are some historical considerations in XTDB that make this kind of change less than straightforward. Specifically, we always have to consider documents that may already be in users' document stores, and (for the purposes of `match` operations) how documents inserted before a change compare to documents after a change. Ideally, Java lists and Clojure lists would be equivalent in this regard, but in the early days (when we only supported Clojure) this wasn't the case, so we can't rely on it :slightly_frowning_face: Safe to say, I'm particularly cautious about changes of this nature, as they often have unforeseen consequences!

We're currently looking into ways to solve this in a more general way, so that we don't have to worry about such things in the future - but that's a much wider change, which I wouldn't want to block getting your bug fixed :slightly_smiling_face: 

For now, I suspect the least intrusive way to fix your original issue would be for us to apply this cast when the SQL engine reads the document, so that we're not changing the structure of any documents already in users' XT instances. This should be a pretty quick change - I'll attach it to this PR.

Regarding the ID type safety change: good shout, cheers - once we're not creating the set for every document, this is good to merge.

Thanks again,

James

jarohen:
> This should be a pretty quick change - I'll attach it to this PR.

Ha, I should know better than to make such statements!

The reason we haven't done this previously is that we can't 'just' translate any Java lists into Clojure vectors for queries - Clojure makes the distinction between vectors and lists, a distinction we use in queries to differentiate between entity-attr-value (EAV/triple) clauses and predicates - we wouldn't know unambiguously which one to translate a Java list to.

With that in mind, I'd be tempted to create your SQL query using something like `XtdbDocument.factory(Clojure.read("{:xt/id <id>, :xtdb.sql/table-query {:find ..., :where ...}, ..."))` - let's continue on your Zulip thread :slightly_smiling_face: 

Apologies that this isn't more straightforward, we've certainly got some room for improvement on the Java integration.

James

cartermyers:
Thanks for the review @jarohen!

I was suspecting that breaking old implementations would be a problem, but I didn't consider it much to be honest. I thought that as long as the inserted document matches in the test that it would be sufficient, but probably not.

No worries on trying to prevent this in the specific SQL case. @refset's MR at #1667 is definitely enough for me. I thought I would open this to provide a general way to avoid more errors for Java users.

Looking forward to more Java improvements! Any Kotlin would be greatly appreciated too :)

refset:
Likewise thanks for the initial report and your feedback - it's also been great to lay out some of these thoughts here!

Whilst I'm here though, I'll add that XT doesn't currently attempt any type coercion before storing documents, so users can generally expect to rely on naively round-tripping arbitrary things via Nippy's serde without any unexpected changes to the types.

PR #1698, opened 2022-01-31
Schedule query timeout interrupt job before calling `open-q*`

### LINKED ISSUE #1654: Lucene module - wildcard-text-search doesn't respect query :timeout
I'm seeing the following query completely ignoring the :timeout option
```
(#xtdb.query_state.QueryState{:query-id "01dd4d16-0fb9-4786-9b71-0e489a089921",
                              :started-at #inst "2021-11-03T12:36:32.862-00:00",
                              :finished-at nil,
                              :status :in-progress,
                              :query {:find [?s ?e], :in [?term], :limit 1000, :timeout 1000, :where [[(wildcard-text-search ?term) [[?e ?v ?a ?s]]]]},
                              :error nil})
```

Note: @refset  has added this as a todo to #1599 

* This is with lmdb and ~1.6m documents
* The ?term that causes this is likely to match many documents (ie >10%)
* xtdb responds to new queries, but this query has now run for > 1hr ie unlikely to every complete
* htop shows `20   0 27.1G 18.0G 7139M R 99.3 56.8 54:54.05 xtdb-lucene-fsy`

refset:
* allows eager evaluation of long-running predicates (such as Lucene searches)
to be interrupted - closes #1654

Previously, the interrupt-job was only being scheduled _after_ the long-running predicate (e.g. `text-search`) had finished running (which happens eagerly when `open-q*` is called in the `with-open` bindings).

### DIFF
--- a/core/src/xtdb/query.clj
+++ b/core/src/xtdb/query.clj
@@ -1844,28 +1844,28 @@
       (entity-tx this index-snapshot eid)))
 
   (q* [this query args]
-    (with-open [res (xt/open-q* this query args)]
-      (let [result-coll-fn (if (some (normalize-query query) [:order-by :limit :offset]) vec set)
-            !timed-out? (atom false)
-            ^Future
-            interrupt-job (when-let [timeout-ms (get query :timeout (:query-timeout this))]
-                            (let [caller-thread (Thread/currentThread)]
-                              (.schedule interrupt-executor
-                                         ^Runnable
-                                         (fn []
-                                           (reset! !timed-out? true)
-                                           (.interrupt caller-thread))
-                                         ^long timeout-ms
-                                         TimeUnit/MILLISECONDS)))]
-        (try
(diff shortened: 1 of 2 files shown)


jarohen:
Good find @refset :clap: does that mean, though, that `open-q*` does some work on the query before it returns? and, if so, should that be the change? (i.e. making sure it only plans the query, no execution)

refset:
@jarohen good point - I expect the predicate->virtual-index transformation could (and perhaps _should_) happen lazily, and certainly would resolve the original issue too.

However, merging this change might be a good idea anyway, since it may be possible that the act of query planning itself can hit degenerative cases where it would be desirable to respect the timeout also.

jarohen (approved):
Ah, good call, yep. In which case, yeah, let's get this merged (once tests pass) and also fix up `q/query` :slightly_smiling_face:

PR #1705, opened 2022-02-08
Submit fn & submit match

naomijub:


samrose:
@naomijub are you still interested in this approach?

naomijub:
I have been researching what would solve my problem better, but I have my own grpc and capnproto xtdb implementations to work with.
Thanks

PR #1712, opened 2022-02-22
dead URL fix

The-Alchemist:


refset:
Ah, good spot @The-Alchemist, thank you for submitting the fix :slightly_smiling_face:

PR #1715, opened 2022-02-26
Allow setting a custom metrics registry

### LINKED ISSUE #1560: Using your own metrics registry
Some docs feedback/signpost for future users: it took me a while to figure out how get crux-metrics to use my own prometheus registry. This is because the docs don't have `:registry` listed as a parameter in the section here https://opencrux.com/reference/monitoring.html#_metrics_module because it's under `::sys/deps` instead of `::sys/args`, while it is present in the section below ("Prometheus reporter"), implying that you should set it there. However doing so entails using the pushgateway or http exporter. It turns out to just tell Crux to write to your own registry you have to do something like `:crux.metrics/metrics     {:crux.metrics/registry #ig/ref :app.metrics/dropwizard-registry}`, which I did not find clear from the docs.

edit: actually not sure that worked. Is it possible to do this at all? Maybe `crux.metrics/->registry` should take params?

edit2: ok, I think I got it. It looks like if a key is not in `::sys/args`, passing it will throw which is good. But `:registry` is actually not even in `::sys/args` in https://github.com/juxt/crux/blob/master/crux-metrics/src/crux/metrics/prometheus.clj, contra the docs. Also `::sys/deps` does not seem to be able to be set from the config. So I ended up patching `crux-metrics` to move `:registry` from `::sys/deps` to `::sys/args`.

nivekuil:
closes https://github.com/xtdb/xtdb/issues/1560

### DIFF
--- a/modules/metrics/src/xtdb/metrics.clj
+++ b/modules/metrics/src/xtdb/metrics.clj
@@ -6,8 +6,10 @@
             [xtdb.system :as sys])
   (:import [com.codahale.metrics Gauge Meter MetricRegistry Snapshot Timer]))
 
-(defn ->registry [_]
-  (dropwizard/new-registry))
+(defn ->registry {::sys/args {:registry {:doc       "Your own com.codahale.metrics.MetricRegistry"
+                                         :default   (dropwizard/new-registry)}}}
+  [{:keys [registry]}]
+  registry)
 
 (defn ->metrics {::sys/deps {:registry ::registry
                              :xtdb/node :xtdb/node


jarohen:
Hey @nivekuil :wave: hope you're keeping well!

Sorry for having dropped the ball on the original issue - it should be possible to pass any 1-arg function as `:xtdb/module`:

```clojure
(let [reg (xtdb.metrics.dropwizard/new-registry)]
  (with-open [node (xtdb.api/start-node {:xtdb.metrics/registry {:xtdb/module (constantly reg)}
                                         :xtdb.metrics/status-reporter {}})]
    (assert (identical? reg (get-in @(:!system node) [:xtdb.metrics/status-reporter :registry])))
    (xtdb.api/status node)))
```

Would this fix your issue?

Cheers,

James

nivekuil:
Nice, I didn't make the connection from the other modules that this could be done. Seems pretty elegant to me!

refset:
Thanks for the feedback here @nivekuil - 'elegant' is good enough for me :slightly_smiling_face:

This does however highlight that the documentation for XT's existing module system could/should be improved, although we're unlikely to invest energy in that soon. Anecdotally, switching to [Integrant](https://github.com/weavejester/integrant) may be on the cards for any future major release (2.x) as it seems to be able to handle ~all of the functionality needed, which wasn't clear to the team originally.

PR #1718, opened 2022-03-15
fixed typo

tommy-mor:


refset:
Good spot, thank you @tommy-mor !

PR #1720, opened 2022-03-21
Override tx time for import, #1517

### LINKED ISSUE #1517: Allowing tx-times from external systems to be preserved in Crux, e.g. for import
From @seeday on Slack, requesting the ability to explicitly specify tx-times when submitting transactions.

I doubt we'd be able to allow _any_ tx-time - if we were to receive out-of-order tx-times I'd guess this would break Crux's tx ingestion - but we might be able to allow tx-times to be explicitly specified so long as no later tx-time has yet been submitted.

jarohen:
resolves #1517 

Adding the ability to overload transaction times for a transaction as an overload to `submit-tx`: `(xt/submit-tx node [tx-ops...] {::xt/tx-time #inst "..."})`. In Java and Kotlin, this is added as a `withTxTime` method on the `Transaction.Builder`.

The transaction time must be no earlier than any transaction already in the system, and no later than the clock on the TxLog. If these constraints aren't adhered to, the system will abort the transaction, log a warning, and continue ingesting (option 2 on #1517).

Impacts:
* The format on the XT-provided TxLogs has changed - these are now maps rather than just the vector of operations. We must still be able to read from old-style tx-logs - this is covered by the migration test. This does mean that older versions of XT won't be able to read newer tx-logs.
* This is a breaking change for TxLog implementors - they should implement the extra overload. We could also consider introducing a second TxLog protocol, so that the same version of the third-party TxLog artifact can run against older and newer XT versions - I don't know how many people this would affect?

### DIFF
--- a/test/test/xtdb/fixtures/every_api.clj
+++ b/test/test/xtdb/fixtures/every_api.clj
@@ -33,7 +33,8 @@
                                    (with-meta {::embedded-kafka? true}))}
       #_(select-keys [:local-standalone])
       #_(select-keys [:local-standalone :remote])
-      #_(select-keys [:local-standalone :h2 :sqlite :remote])))
+      #_(select-keys [:local-standalone :h2 :sqlite :remote])
+      #_(select-keys [:local-kafka])))
 
 (def ^:dynamic *node-type*)
 
--- a/core/src/xtdb/api.clj
+++ b/core/src/xtdb/api.clj
@@ -16,6 +16,8 @@
 (s/def ::tx-time date?)
 (s/def ::tx (s/keys :opt [::tx-id ::tx-time]))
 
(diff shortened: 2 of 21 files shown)


test/test/xtdb/fixtures/every_api.clj
```diff
@@ -33,7 +33,8 @@
                                    (with-meta {::embedded-kafka? true}))}
       #_(select-keys [:local-standalone])
       #_(select-keys [:local-standalone :remote])
-      #_(select-keys [:local-standalone :h2 :sqlite :remote])))
+      #_(select-keys [:local-standalone :h2 :sqlite :remote])
+      (select-keys [:local-kafka])))
```
hraberg: Is this potentially committed by mistake?

jarohen: cheers :flushed:

PR #1721, opened 2022-03-21
Fix 1719 - rocksdb - warn about MALLOC_ARENA_MAX only when starting module

### LINKED ISSUE #1719: Only warn about MALLOC_ARENA_MAX when using RocksDB
A minor one, but we see:

`2022-03-21 09:48:50,080 WARN                                                      xtdb.api - MALLOC_ARENA_MAX not set, memory usage might be high, recommended setting for XTDB is 2`

when using LMDB.

It is probably easy to fix given the code [here](https://github.com/xtdb/xtdb/blob/529105575507e4d5d32595f8baea8bf153ada3ed/core/src/xtdb/api.clj#L249-L250).

Any known unknowns there? I might queue it up in my free time.

arichiardi:
Here is my patch, again, haven't tested it yet, still trying to figure that out (I left a message on Zulip :smile:).

### DIFF
--- a/core/src/xtdb/api.clj
+++ b/core/src/xtdb/api.clj
@@ -2,9 +2,7 @@
   "Public API of XTDB."
   (:refer-clojure :exclude [sync])
   (:require [clojure.spec.alpha :as s]
-            [clojure.tools.logging :as log]
             [xtdb.codec :as c]
-            [xtdb.io :as xio]
             [xtdb.system :as sys])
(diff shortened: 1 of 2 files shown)


jarohen:
Thanks! :pray: All looks good - CI run is [here](https://app.circleci.com/pipelines/github/xtdb/xtdb/2444/workflows/5496e1e3-e0b4-41a8-9636-8286f7a195bb).

James

PR #1746, opened 2022-04-22
Small fixes

green-coder:
PS: I made those edits only by reading the source code, need to test if it breaks things or not.


jarohen:
Thanks! :pray:

PR #1747, opened 2022-04-23
Update bitemporality.adoc

jeffp42ker:
Correct link path by removing `/posts/` and `.html` extension from link to "The Value of Bitemporality" blog post.

CLA sent


refset:
Thank you @jeffp42ker ! :pray:

PR #1756, opened 2022-05-17
fix(xtdb-kafka-connect): issue with parsing sink records in tx-mode

zirota:
Faced some issues with parsing sink records in tx-mode. For some context, a xtdb sink record in tx-mode is a submit-tx operation **vector** which includes the valid-time of the transaction. 

### JsonParseException not caught
Because `json/parse-string` is lazy, the `JsonParseException` in line 77 does not get thrown until further downstream and causing the exception to not be caught properly. This prevents `c/read-edn-string-with-readers` from running in the catch block.

### Incorrect assoc
Since a submit-tx operation is a vector, the resulting `doc` from `(record->edn record)` will be a vector.  `(assoc doc :xt/id id)` will fail because `doc` is not a valid map

### DIFF
--- a/modules/kafka-connect/src/xtdb/kafka/connect.clj
+++ b/modules/kafka-connect/src/xtdb/kafka/connect.clj
@@ -74,7 +74,7 @@
 
       (string? value)
       (try
-        (json/parse-string value true)
+        (doall (json/parse-string value true))
         (catch JsonParseException e
           (log/debug e "Failed to parse as JSON, trying EDN: " value)
           (c/read-edn-string-with-readers value)))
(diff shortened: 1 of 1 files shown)


zackteo:
We just discovered that there are issues with this for bulk transactions. We will be pushing a change to fix this

zirota:
Hi @refset I updated the PR. In my earlier change the tx-ops is not submitted correctly, because I am only submitting the first tx-op of any incoming transaction. A transaction in tx-mode can look something like this:
`
[[::xt/put {:xt/id 1} #inst "2019-01-21T05:47:26.853Z"]
[::xt/put {:xt/id 2} #inst "2019-01-21T05:47:26.853Z"]
[::xt/put {:xt/id 3} #inst "2019-01-21T05:47:26.853Z"]]
` 
But only the first tx-op will be submitted. 

This change will assume that every record after transformation is a valid transaction payload with all the tx-ops for submit-tx and thus we will call submit-tx for every record.

refset:
Hey @zirota thanks for contributing the initial work here and for the updates :)

I'm a bit out of context to understand all these changes currently, but it would be very helpful if you could add test cases that demonstrate the incorrect initial behaviours (if I checkout an old version of the code) _and_ the correct new behaviours, e.g. extending this test namespace: https://github.com/xtdb/xtdb/blob/e2f51ed99fc2716faa8ad254c0b18166c937b134/test/test/xtdb/kafka_connect_test.clj

Have you tried running/extending the tests in that namespace already?

refset:
Just noting that the CI build test (for `xtdb.kafka-connect-test/test-sink-task`) failed: https://app.circleci.com/pipelines/github/xtdb/xtdb/2542/workflows/ee8fcc64-0eeb-4b91-aff1-74414cbbbe86/jobs/9864?invite=true#step-103-1169 - I'll convert this PR to 'draft' in the meantime.

For anyone interested, there's some discussion also happening here: https://juxt-oss.zulipchat.com/#narrow/stream/251593-xtdb-dev/topic/Running.20xtdb.20tests

jarohen:
closing due inactivity, but please feel free to re-open :slightly_smiling_face:

PR #1757, opened 2022-05-20
Ingest pipelining

### LINKED ISSUE #1692: Improve ingestion performance by encoding using multiple threads
Currently, both the submission and indexing pipelines are single-threaded, and therefore the Nippy-encoding of documents, attributes and values is fully serial. However, for sufficiently large/deep/wide documents and long lists of transactions operations, using multiple threads would typically prove to be more efficient.

jonpither:
Closes https://github.com/xtdb/xtdb/issues/1692

Pipelines ingest operations:

- 15% - Fetching docs from doc-store for ingestion
- 15% - Encoding of doc KVs
- 65% - Processing of tx-ops and indexing to Rocks
- 5% - Stats

We have observed ~25% savings on ingestion times by parallelising the above operations. 

Potential issues:

- Code is made a little harder, for example `xtdb.tx/index-docs` knowing about docs and `encoded-docs`. I suggest we live with this for now, and a subsequent clean up / refactor - i.e. formalising and specifying what a _batch_ is - could be introduced later.
- The one inconsistency this code introduces currently, is stats recorded aborted KVs in attribute/value counts. We may want to tackle this as part of this PR.

TODOs:

- [x] Error handling (ensure Throwables aren't swallowed)
- [x] Make batch-size configurable (will not do in this PR)
- [x] Consider making stats consistent with previous approach
- [x] Use consistent ThreadPoolExecutor approach
- [x] Run Kondo (ns declaraton tidy up)
- [x] move common threadpool code to io
- [x] Test on ingest weather

### DIFF
--- a/core/src/xtdb/db.clj
+++ b/core/src/xtdb/db.clj
@@ -10,16 +10,18 @@
   (max-depth [this]))
 
 (defprotocol IndexStoreTx
-  (index-docs [this docs])
+  (index-docs [this docs encoded-docs])
   (unindex-eids [this eids])
   (index-entity-txs [this entity-txs])
   (commit-index-tx [this])
   (abort-index-tx [this]))
 
 (defprotocol IndexStore
+  (encode-docs [_ docs])
   (store-index-meta [this k v])
   (tx-failed? [this tx-id])
-  (begin-index-tx [index-store tx fork-at]))
+  (begin-index-tx [index-store tx fork-at])
+  (index-stats [this docs]))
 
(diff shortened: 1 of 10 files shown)


jonpither:
Do we need to make `tx-batch-size` configurable @jarohen ?

jonpither:
Converted to a draft whilst we ascertain better numbers.

jonpither:
Closing, in favour of #1762, which shows more promising performance.

PR #1759, opened 2022-05-24
fix(http-server): missing tx-time & valid-time in history diff ui

zirota:
This fixes the missing transaction time and valid time on UI when comparing the diff in the entity history


jarohen:
That's great, thanks! :pray:

PR #1760, opened 2022-05-25
fix(http-client): use *internal-http-request-fn* for api request

zirota:
Not to sure if the previous change was intended but I noticed that re-binding the *internal-http-request-fn* no longer works when requests are made via the remote-api-client.

This PR gives back the ability of overriding the *internal-http-request-fn* when making api-requests via the remote-api-client and fallback to using required juxt clj-http module.

### DIFF
--- a/modules/http-client/src/xtdb/remote_api_client.clj
+++ b/modules/http-client/src/xtdb/remote_api_client.clj
@@ -70,23 +70,25 @@
                       (throw error)
                       result))))
               (catch IOException _e))
+            (fn [opts]
+              (http/request opts))
             (fn [_]
               (throw (IllegalStateException. "No supported HTTP client found.")))))))))
 
 (defn- api-request-sync
   ([url {:keys [body http-opts ->jwt-token]}]
    (let [{:keys [body status] :as result}
-         (http/request (merge {:url url
-                               :method :post
-                               :headers (merge (when body
-                                                 {"Content-Type" "application/edn"})
-                                               (when ->jwt-token
-                                                 {"Authorization" (str "Bearer " (->jwt-token))}))
-                               :body (some-> body xio/pr-edn-str)
-                               :accept :edn
-                               :as "UTF-8"
-                               :throw-exceptions false}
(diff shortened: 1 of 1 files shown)


zirota:
any idea why the ci pipeline is failing? Did my change cause it to fail? 🤔

refset:
Happily it seems the CI error looks very unrelated :)
```clojure
Error encountered performing task 'nvd' with profile(s): 'base,system,user,provided,dev,nvd'
clojure.lang.ExceptionInfo: org.owasp.dependencycheck.exception.ExceptionCollection {:exceptions [#error {
 :cause nil
 :via
 [{:type org.owasp.dependencycheck.analyzer.exception.AnalysisException
   :message "Failed to request component-reports"
   :at [org.owasp.dependencycheck.analyzer.OssIndexAnalyzer analyzeDependency "OssIndexAnalyzer.java" 149]}
  {:type java.lang.NullPointerException
   :message nil
   :at [org.sonatype.ossindex.service.client.cache.DirectoryCache entryKey "DirectoryCache.java" 149]}]
 :trace
 [[org.sonatype.ossindex.service.client.cache.DirectoryCache entryKey "DirectoryCache.java" 149]
```

I'm not in context to understand your change right now, but having briefly looked - should it have a try-catch also (like the lines above)?

zirota:
@refset the use of `*internal-http-request-fn*` is to let users of the `remote-api-client` easily re-bind the http-client request function. (I was using it to add headers to the request)
  
It got changed to directly use the required http-client (`http/request`) by the namespace in v1.17.1: https://github.com/xtdb/xtdb/blob/docs-v1.17.1/crux-http-client/src/crux/remote_api_client.clj#L80

I did not add a try-catch because the `http/request` function was called directly (previously on line 81) without any try-catch. Also the try-catch above are presumably used to catch `IOExceptions` that occur when `requiring-resolve` fails such that it will cascade down until a valid http-client can be resolved

jarohen:
thanks @zirota! :pray: 

@refset: good spot, but fine in this case because we've got `:throw-exceptions false` specified.

PR #1762, opened 2022-06-01
Pipeline using KvStoreTx

### LINKED ISSUE #1692: Improve ingestion performance by encoding using multiple threads

jonpither:
This PR adds about 30-35 % ingest performance speed improvement, with #1773 , we are seeing a consistent 40% improvement.

- [x] reinstate use of `ExpandableByteBuffer` for content-hash/ecav during ingest (https://github.com/xtdb/xtdb/pull/1778 is now merged)

Notes:

- I expect this change will partially address https://github.com/xtdb/xtdb/issues/1509, when using Rocks.
- See #1779 for a subsequent minor change to make after this PR has landed, around re-using Attribute byte-buffers.
- https://github.com/avisi-apps/crux-xodus will not be impacted, but there could be ingest performance gains for this KV store (and any other), if it were to implement `KvStoreWithReadTransaction`.

As discussed with @jarohen, we may want to do some future refactoring, items that came up during the working of this PR:

- in `tx` the resolution of the tx-id could be done later in the pipeline rather than in the encoding step
- currently the code occupies a middle-ground, of fork not truly decorating index-store, but not being properly folded in either. 
- the wiring in `query` to make `with-tx` happen, is a bit gnarly, although it does build on an existing pattern.


core/src/xtdb/query.clj
```diff
@@ -2110,19 +2113,30 @@
...
+          delta-index-store-tx (db/begin-index-tx delta-index-store tx)
+          forked-index-store-tx (fork/->ForkedKvIndexStoreTx index-store, transient-kv, valid-time, tx-id, (atom #{}), delta-index-store-tx, nil)
+
+          in-flight-tx (-> (db/begin-tx (assoc tx-indexer :bus (reify xtdb.bus/EventSink
+                                                                 (send [_ _])))
+                                        tx)
+                           (assoc :index-store-tx forked-index-store-tx)
+                           (update :db-provider merge {:index-store forked-index-store-tx}))
+
+          conformed-tx-ops (map txc/conform-tx-op tx-ops)
```
jarohen, line 2131: it doesn't feel like the query engine should know about how to construct the ForkedKvIndexStoreTx - maybe some kind of helper in that ns?

core/src/xtdb/kv/mutable_kv.clj
```diff
@@ -42,19 +42,42 @@
   Closeable
   (close [_]))
 
-(deftype MutableKvStore [^NavigableMap db]
+(deftype MutableKvTx [^TreeMap db, ^NavigableMap db2]
```
jarohen: better names for these? what's the difference between `db` and `db2`?

core/src/xtdb/db.clj
```diff
@@ -66,13 +67,14 @@
     or, outside of f, complete the future returned from this function to stop the subscription."))
 
 (defprotocol InFlightTx
-  (index-tx-events [in-flight-tx tx-events prefetched-docs]
+  (index-tx-docs [in-flight-tx docs])
+  (index-tx-events [in-flight-tx tx-events]
     "prefetched-docs :: docs from the batch of transactions")
```
jarohen: this looks like an outdated comment now?

core/src/xtdb/kv/index_store.clj
```diff
@@ -170,9 +170,9 @@
   (^org.agrona.MutableDirectBuffer [b entity content-hash attr]
    (encode-ecav-key-to b entity content-hash attr mem/empty-buffer))
   (^org.agrona.MutableDirectBuffer [^MutableDirectBuffer b ^DirectBuffer entity ^DirectBuffer content-hash ^DirectBuffer attr ^DirectBuffer v]
-   (assert (or (zero? (.capacity attr)) (= c/id-size (.capacity attr)))
+   (assert (or (zero? (.capacity attr)) (<= c/id-size (.capacity attr)))
```
jarohen: OOI - why would these be anything apart from zero or `c/id-size`?

core/src/xtdb/kv/index_store.clj
```diff
@@ -460,35 +461,50 @@
...
+                                            (keys doc)))
+                                  {}
+                                  docs)]
+        (->> docs
+             (reduce (fn [acc doc]
+                       (let [e (:crux.db/id doc)]
+                         (reduce-kv (fn [acc k v]
+                                      (let [k-buf (get attr-key-bufs k)
+                                            stats-buf (or (get acc k-buf)
+                                        ;(kv/get-value transient-kv-snapshot k-buf)
```
jarohen: should this be either deleted or uncommented?

core/src/xtdb/kv/index_store.clj
```diff
@@ -1060,58 +1081,81 @@
...
+    (when kv-tx
+      (kv/abort-kv-tx kv-tx))
+    (let [{::xt/keys [tx-id tx-time]} tx
+          attr-bufs (->> (into #{} (mapcat keys) (vals docs))
+                         (into {} (map (juxt identity c/->id-buffer))))]
+
+      ;; we still put the ECAV KVs in so that we can keep track of what we need to evict later
+      ;; the bitemp indices will ensure these are never returned in queries
+
+      (->> (for [[content-hash doc] docs
```
jarohen: subjective nit: I tend to only use `->>` when it's a value I'm transforming through the thread, rather than because the value happens to be the last parameter each time - you kind of lose the side-effect here?

core/src/xtdb/tx.clj
```diff
@@ -342,32 +346,26 @@
...
 
     (log/debug "Transaction committed:" (pr-str tx))
 
     (let [{:keys [tx-events]} @!tx]
-      (bus/send bus (into {::xt/event-type ::indexed-tx,
-                           :submitted-tx tx,
-                           :committed? true
-                           ::txe/tx-events tx-events}
-                          (select-keys @!tx [:doc-ids :av-count :bytes-indexed])))))
+      (bus/send bus (merge {::xt/event-type ::indexed-tx,
```
jarohen: why the change from `into` to `merge` here? `into` uses transient collections under the hood, and no varargs, so is more performant than `merge` (which is essentially `reduce conj`)

core/src/xtdb/tx.clj
```diff
@@ -378,19 +376,17 @@
...
+    (let [index-store-tx (db/begin-index-tx index-store tx)
           document-store-tx (fork/begin-document-store-tx document-store)]
-      (->InFlightTx tx fork-at
+      (->InFlightTx tx
                     (atom :open)
-                    (atom {:doc-ids #{}
-                           :av-count 0
-                           :bytes-indexed 0
-                           :tx-events []})
+                    (atom {:tx-events [] :av-count 0 :bytes-indexed 0 :doc-ids #{}})
```
jarohen: subjective code style: I don't mind this being on one line, but if so I tend to put commas between the kvs to visually delineate them for the reader: `{:tx-events [], :av-count 0, ...}`

core/test/xtdb/fork_test.clj
```diff
@@ -30,6 +71,15 @@
     (t/is (= #{["Ivan"]}
              (xt/q db2 all-names-query)))
 
+    (t/is (= [#:xtdb.api{:tx-id 0, :doc {:name "Ivna", :xt/id :ivan}}
+	      #:xtdb.api{:tx-id 1, :doc {:name "Ivan", :xt/id :ivan}}]
```
jarohen: nit: indent

jarohen (changes requested):
nice work gents, a good speed boost :clap: a few changes below but nothing major.

as discussed, there are a few different interleaved changes here, so I'm not 100% sure I'll have picked up all of the potential impacts - if you're happy to proceed then that's cool. particularly, might be a familiarity thing, but I also feel less confident in my understanding of the forking code now that it's on a separate code path - let's consider refactoring this at a later date.

modules/rocksdb/src/xtdb/rocksdb.clj
```diff
@@ -24,31 +24,45 @@
...
-  (seek [this k]
+  (seek [_ k]
     (.seek i (mem/direct-byte-buffer k))
     (iterator->key i))
 
-  (next [this]
-    (.next i)
-    (iterator->key i))
+  (next [_]
+    (when (.isValid i)
```
jarohen, line 32: IIUC callers shouldn't be calling `next` if the iterator isn't valid, and I'm not sure what performance impact adding this on every call will have - do you know which caller is causing the issue here?

refset: I think we need to _always_ be checking `.isValid` before attempting to call `.next`. Based on:

1. https://github.com/facebook/rocksdb/issues/5611#issuecomment-1097435163
2. the "REQUIRES: isValid()" docstring https://javadoc.io/doc/org.rocksdb/rocksdbjni/latest/org/rocksdb/RocksIteratorInterface.html#next--
3.  https://github.com/facebook/rocksdb/issues/9306#issuecomment-997551091
4.  https://github.com/facebook/rocksdb/wiki/Iterator#error-handling

I can reproduce a JVM crash simply by changing the definition of the Rocks module with a bunch of extra `.next`s (and running a basic test in kv_test.clj):
```clojure
(defrecord RocksKvIterator [^RocksIterator i]
  kv/KvIterator
  (seek [this k]
    (.seek i (mem/direct-byte-buffer k))
    (iterator->key i))

  (next [this]
    (.next i)
    (.next i)
    (.next i)
    (.next i)
    (.next i)
    (.next i)
    (.next i)
    (.next i)
    (.next i)
    (.next i)
    (.next i)
    (iterator->key i))

  (prev [this]
    (.prev i)
    (iterator->key i))
 ...
 ```

It's not been a problem in practice only because in normal operation `.next` will keep returning things (i.e. continue being valid) across the codec boundaries, so this could only ever show up when calling `.next` on the final entry(s) in the final codec keyspace - which is the stats index: https://github.com/xtdb/xtdb/blob/7ec4828f48b3fb97ffd1baf4eda22857c9fcb48d/core/src/xtdb/codec.clj#L60

jonpither: Raise an a new issue @refset as it applies equally to master, and we can take care in a subsequent commit

core/src/xtdb/query.clj
jonpither: Currently 1index-store` depends on `fork` and not the other way around, against my own intuition (that `fork` should purely be a wrapper to `index-store`, with the latter being the main/critical path only). Pulling this fn out to the `fork` ns as you suggest, whilst seemingly reasonable, opens up a ns dependency battle, as then `fork` would need to depend on `index-store`. At this point, I'd prefer to land the PR, let it breathe, and then reflect.

jonpither: Currently index-store depends on fork and not the other way around, against my own intuition (that fork should purely be a wrapper to index-store, with the latter being the main/critical path only). Pulling this fn out to the fork ns as you suggest, whilst seemingly reasonable, opens up a ns dependency issue, as then fork would need to depend on index-store. At this point, I'd prefer to land the PR, let it breathe, and then reflect.

core/src/xtdb/kv/index_store.clj
jonpither: We're using an expandable byte buffer

jonpither: Good spot @jarohen. Since https://github.com/xtdb/xtdb/commit/898e7c992b57fe4a4b69713900c7945e5a7b4483 we can revert this assert to what it was doing - I have retested and made the revert.

jonpither:
Hi @jarohen,

Thanks for the review. I agree that the wiring could be approached with fresh eyes and a potential subsequent refactor - I've updated the description to cover some of our thinking, during working on the PR.

I haven't moved the logic from `with-tx` into `fork.clj`, as it's non-trivial to do, and I'd prefer we approach as a subsequent step, if that's agreeable.

Note, WRT to #1773, my preferred order is for this one go first.

Thanks again.

core/src/xtdb/tx.clj
jonpither: I've reverted back to `into`

PR #1765, opened 2022-06-08
Remove JNR backends

jonpither:
We no longer want to support JNR backends, given the trade-off in speed improvement vs maintainability. Performance of JNR backends is thought to be ~10%, which could be claimed elsewhere.


jarohen:
Hey @jonpither - quick one here, think we still pull in the JNR versions in the various `project.clj`s - do we need to?

jonpither:
Changes made @jarohen

PR #1773, opened 2022-06-20
Column families for temporal indexes

jonpither:
Sets up column families for temporal indexes. Minimizes temporal seek time during ingest.

This could be seen as a part 1 of column families. We could bring in more CFs and measure the respective performance gains. Note that each CF brings in a 64 mb by default mem-table.

Rebased on #1762

### DIFF
--- a/modules/rocksdb/src/xtdb/rocksdb.clj
+++ b/modules/rocksdb/src/xtdb/rocksdb.clj
@@ -41,9 +41,6 @@
 
 (set! *unchecked-math* :warn-on-boxed)
 
-(def ^:const column-family-defs [c/entity+vt+tt+tx-id->content-hash-index-id
-                                 c/entity+z+tx-id->content-hash-index-id])
-
 (defn- iterator->key [^RocksIterator i]
   (when (.isValid i)
     (mem/as-buffer (.key i))))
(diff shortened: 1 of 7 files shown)


modules/rocksdb/src/xtdb/rocksdb.clj
```diff
@@ -13,65 +13,98 @@
...
+                        WriteBatch WriteOptions Statistics StatsLevel
+                        ColumnFamilyOptions ColumnFamilyDescriptor ColumnFamilyHandle)))
+
+(defprotocol CfId
+  (->cf-id [this]))
+
+(extend-protocol CfId
+  (class (byte-array 0))
+  (->cf-id [this]
+    (first this))
```
jarohen: `(aget this 0)`, save going via Clojure's `seq` abstraction

modules/rocksdb/src/xtdb/rocksdb.clj
```diff
@@ -13,65 +13,98 @@
...
 
+(def ^:const column-family-defs [c/entity+vt+tt+tx-id->content-hash-index-id
+                                 c/entity+z+tx-id->content-hash-index-id])
+
 (defn- iterator->key [^RocksIterator i]
   (when (.isValid i)
     (mem/as-buffer (.key i))))
 
-(defrecord RocksKvIterator [^RocksIterator i]
+(defrecord RocksKvIterator [i is k->i]
```
jarohen: given this is low-level, I'd go for some lower-level, more performant primitives here:
* `defrecord` -> `deftype` (so you can use mutable fields)
* `is`: `(atom {})` -> `(HashMap.)`
* `i`: `(atom nil)` -> `^:unsynchronised-mutable ^RocksIterator i`

modules/rocksdb/src/xtdb/rocksdb.clj
```diff
@@ -13,65 +13,98 @@
...
+    (iterator->key ^RocksIterator @i))
 
   (value [this]
-    (mem/as-buffer (.value i)))
+    (mem/as-buffer (.value ^RocksIterator @i)))
 
   Closeable
   (close [this]
-    (.close i)))
+    (doseq [[_ i] @is]
```
jarohen: `(doseq [i (.values is)] ...)`

jonpither:
Changes made @jarohen. We have also made the generative test run for non-rocks implementations still.

jonpither:
There's an issue with entity history `asc` - it's broken. Tests were not picked up earlier, since fork-test / api-test only ran against the in-mem DB.

jonpither:
Made column families configurable as per @hraberg request.

PR #1774, opened 2022-06-20
Upgrade to RocksDB 7.3.1

jonpither:


jarohen:
thanks @jonpither - does this change the on-disk format at all? if so, could you add an index-version bump, and bump the Rocks-specific checkpoint version number too?

jonpither:
Changes made @jarohen

PR #1778, opened 2022-06-29
Fix codec/hex->id-buffer `to` arity

refset:
This fix is in support of #1762 where we would like to use a `ExpandableDirectByteBuffer` ThreadLocal but can't currently because it breaks `:local-kafka-transit` tests (which rely on `hex->id-buffer` to hydrate `#xtdb/id` reader IDs). This fix ensures that the correct thing is returned - i.e. a limited view buffer and not simply the supplied destination buffer.

I had a quick look to see whether any other `*->buffer` functions had a similar issue where the non-limited/sliced thing was being returned, but couldn't see any.

I also couldn't find any existing callers of `hex->id-buffer` that relied on the 2-arg `to` arity, so there should be no impact of the current behaviour (or indeed the fix) to existing users.

### DIFF
--- a/core/src/xtdb/codec.clj
+++ b/core/src/xtdb/codec.clj
@@ -695,7 +695,7 @@
   ([hex to]
    (.putByte ^MutableDirectBuffer to 0 id-value-type-id)
    (mem/hex->buffer hex (mem/slice-buffer to value-type-id-size hash/id-hash-size))
-   to))
+   (mem/limit-buffer to id-size)))
 
 (deftype EDNId [hex original-id]
   IdToBuffer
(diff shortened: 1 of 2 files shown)


jarohen (changes requested):
nice find @refset - mind fixing the test and hitting re-review?

refset:
It definitely pays to run a few more tests locally :slightly_smiling_face: thanks for the initial look @jarohen - I'm confident it's definitely ready now(!)

My misunderstanding was that the slice is actually necessary to account for the 1-byte prefix (i.e. to put the hex bytes at the correct offset, and not simply the start) - so the actual change is even simpler now.

PR #1780, opened 2022-06-30
Add Warning paragraph to streaming queries

FiV0:
cc @refset


refset:
Thanks for the prompt and input @FiV0 !

PR #1787, opened 2022-07-13
Note of warning about JDBC URL encoding.

jsulmont:
* Notify users about the need for correct URL encoding according to the driver they use.


refset:
Thanks @jsulmont !

Issue #1800, opened 2022-08-14
fetch-docs can send a huge number of requests causing ingestion to fail


nivekuil:
Testing 1.22.0-rc1, indexing fails in `strict_fetch_docs` with the exception indicating hundreds of mismatched hashes. These documents do exist in Cloudflare R2 so I think the huge batch of async requests may be getting rate limited, exacerbated by the new ingest pipelining. I tested this patch and it successfully boots, albeit slowly:
```patch
diff --git a/modules/s3/src/xtdb/s3.clj b/modules/s3/src/xtdb/s3.clj
index 76ffde07..ecd4c6a4 100644
--- a/modules/s3/src/xtdb/s3.clj
+++ b/modules/s3/src/xtdb/s3.clj
@@ -30,32 +30,36 @@
                (.get req)))))
 
 (defn ^:no-doc get-objects [{:keys [^S3Configurator configurator ^S3AsyncClient client bucket prefix]} reqs]
-  (->> (for [[path ^AsyncResponseTransformer response-transformer] reqs]
-         (let [s3-key (str prefix path)]
-           [path (-> (.getObject client
-                                 (-> (GetObjectRequest/builder)
-                                     (.bucket bucket)
-                                     (.key s3-key)
-                                     (->> (.configureGet configurator))
-                                     ^GetObjectRequest (.build))
-                                 response-transformer)
-
-                     (.handle (reify BiFunction
-                                (apply [_ resp e]
-                                  (if e
-                                    (try
-                                      (throw (.getCause ^Throwable e))
-                                      (catch NoSuchKeyException e
-                                        (log/warn "S3 key not found: " s3-key))
-                                      (catch Exception e
-                                        (log/warnf e "Error fetching S3 object: s3://%s/%s" bucket (str prefix path))))
-
-                                    resp)))))]))
-
-         (into {})
-         (into {} (keep (fn [[path ^CompletableFuture fut]]
-                          (when-let [resp (.get fut)]
-                            [path resp]))))))
+  (transduce
+   (map (fn [reqs]
+          (->> (for [[path ^AsyncResponseTransformer response-transformer] reqs]
+                 (let [s3-key (str prefix path)]
+                   [path (-> (.getObject client
+                                         (-> (GetObjectRequest/builder)
+                                             (.bucket bucket)
+                                             (.key s3-key)
+                                             (->> (.configureGet configurator))
+                                             ^GetObjectRequest (.build))
+                                         response-transformer)
+
[...]

PR #1808, opened 2022-08-26
Apply doc-count batching policy to transactions before pipelining

wotbrew:
## Problem

In #1762 pipelining was introduced during indexing from the golden stores. This improves IO resource utilisation and delivers a substantial improvement to ingest throughput.

However one implication is that document stores can receive a much higher set of ids to fetch a time, because now the documents for many transactions are fetched at once. There is some concern ( #1800 ) that the increased concurrent request volume may be triggering errors or breaching request-per-second limits.

## Solution

This PR then represents an early mitigation strategy that attempts to allow some transaction batching while putting _some_ limits on the number of fetches document stores will be requested to do in a single eager operation.

The mechanism added batches transactions according to number of referenced docs, so that many small transactions can benefit from a lot of batching together, but larger transactions will be issued in smaller batches.

Note that a single transaction can reference any number of documents, and so for large enough transactions this PR will not help - however _that issue_ will have been around pre #1762.

This represents a tactical, speculative change and may _not_ resolve the S3 / R2 problem raised in #1800. Further benchmarks are necessary to determine what impact the batching policy has on overall ingest throughput, though due to the lack of _any batching_ prior to #1762, I assume no performance regression.

## Configuration

A configuration variable is available to vary the ideal batch doc-count on the ingester: `:batch-preferred-doc-count`.
I would consider this variable a temporary tuning parameter we can use early on while testing but I would not consider it part of the configuration surface of XTDB.

### DIFF
--- a/core/src/xtdb/tx.clj
+++ b/core/src/xtdb/tx.clj
@@ -464,13 +464,16 @@
       :else
       true)))
 
-(defn ->tx-ingester {::sys/deps {:tx-indexer :xtdb/tx-indexer
+(defn ->tx-ingester {::sys/args {:batch-preferred-doc-count
+                                 {:default 1024
+                                  :spec ::sys/pos-int}}
+                     ::sys/deps {:tx-indexer :xtdb/tx-indexer
                                  :index-store :xtdb/index-store
                                  :document-store :xtdb/document-store
                                  :tx-log :xtdb/tx-log
                                  :bus :xtdb/bus
                                  :secondary-indices :xtdb/secondary-indices}}
-  [{:keys [tx-log tx-indexer document-store bus index-store secondary-indices]}]
+  [{:keys [tx-log tx-indexer document-store bus index-store secondary-indices batch-preferred-doc-count]}]
   (log/info "Started tx-ingester")
 
   (let [!error (atom nil)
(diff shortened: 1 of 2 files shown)


core/src/xtdb/tx.clj
```diff
@@ -489,7 +502,32 @@
...
+
+            ;; Applies a default transaction (fetch/encode) batching strategy that attempts to balance throughput and working set size
+            ;; (this will change if we can push this policy into each doc store)
+            ;; see docstring on tx-batch-preferred-doc-count
+            (batch-transactions [txs]
+              (lazy-seq
+                ((fn batch-txes [{:keys [batch, doc-hashes]
+                                  :or {batch [],
+                                       doc-hashes #{}}
+                                  :as buf} txs]
```
jarohen: style nit: try not to mix single-line params and multi-line params on the same line - it makes it harder to spot that there are in fact 2 params to this function

core/src/xtdb/tx.clj
```diff
@@ -489,7 +502,32 @@
...
+                   (cond
+                     (seq txs)
+                     (let [tx (first txs)
+                           new-doc-hashes (txc/tx-events->doc-hashes (::txe/tx-events tx))
+                           ;; union early so we get accurate doc cardinality (doc ids can be shared between tx)
+                           unioned-doc-hashes (into doc-hashes new-doc-hashes)]
+                       ;; if there is no room in the buffer for these docs, partition into a new batch buffer.
+                       (if (and buf (<= (long tx-batch-preferred-doc-count) (count unioned-doc-hashes)))
+                         (cons buf (batch-txes {:batch [tx], :doc-hashes (set new-doc-hashes)} (rest txs)))
```
jarohen: this function isn't actually lazy (beyond the initial invocation) because you're recursively calling `batch-txes` rather than `batch-transactions` here - guessing that's not the intention?

core/src/xtdb/tx.clj
```diff
@@ -464,6 +464,19 @@
       :else
       true)))
 
+;; introduced as a speculative change for 1.22 to avoid issues on ingest see #1800
+(def tx-batch-preferred-doc-count
```
jarohen: good to make this (temporarily) configurable - but let's pop it in as `::sys/args` on the ingester (can pair up)

jarohen (approved):

PR #1811, opened 2022-08-30
First cut of `get-valid-end-time`

FiV0:
This would allow to get valid-end-time in queries without resorting to creating future db instances
```clj
'{:find [?e ?end-time]
  :where [[?e :type :some-value2]
               [(get-valid-end-time ?e) ?end-time]]}
```
the predicate constraint also takes an optional second argument as default value if no `end-time` can be found. `nil` is returned if no end-time can be found an no default was specified. The `end-time` is currently defined as the next time the entity changes or is removed. 

Possible improvements:
- allow literals, currently only logic vars work
- currently another roundtrip (buffer -> value) is needed for accessing the `entity-history`. Another option is to seek with the `eid` buffer into the bitemp indices directly. This would possibly be faster and avoid iterating through the whole history of the entity. It would probably a bit less clean as there is no function/protocol to do so on the index snapshot.

Possible impact:
- The db `valid-time` had to be added the query-cache for this to work correctly.

### DIFF
--- a/core/src/xtdb/query.clj
+++ b/core/src/xtdb/query.clj
@@ -221,6 +221,16 @@
          :args (s/spec (s/cat :e-var logic-var?, :attr literal?, :not-found (s/? any?)))
          :return (s/? ::binding)))
 
+(defmethod pred-args-spec 'get-start-valid-time [_]
+  (s/cat :pred-fn #{'get-start-valid-time}
+         :args (s/spec (s/cat :e-var logic-var?))
+         :return logic-var?))
+
+(defmethod pred-args-spec 'get-end-valid-time [_]
+  (s/cat :pred-fn #{'get-end-valid-time}
+         :args (s/spec (s/cat :e-var logic-var?, :not-found (s/? any?)))
+         :return logic-var?))
+
 (defmethod pred-args-spec '== [_]
   (s/cat :pred-fn #{'==} :args (s/tuple some? some?)))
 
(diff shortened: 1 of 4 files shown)


refset:
Hey @FiV0 I just wanted to say thanks again for reflecting on your project requirements and suggesting the change here! I know @jarohen will offer more feedback in due course...but as you're probably aware it's a rather abnormally busy time for the core team over the next 2.5 weeks :sweat_smile: 

Also, noting that an alternative userspace workaround looks like (but requires extra `db` instances, as noted) https://gist.github.com/refset/8439ceebf3c1f08c614b0c818266a140

jarohen:
Hey @FiV0 - cheers for the PR. apologies for the delay in getting back to you - unless it's blocking you, would it be alright to run through it after SL? want to give it the time/brain-space it deserves :slightly_smiling_face:

FiV0:
> Hey @FiV0 - cheers for the PR. apologies for the delay in getting back to you - unless it's blocking you, would it be alright to run through it after SL? want to give it the time/brain-space it deserves slightly_smiling_face

Sounds good. It's not super urgent.

core/src/xtdb/query.clj
```diff
@@ -947,6 +952,22 @@
...
             (bind-binding return-type value-serde tuple-idxs-in-join-order (get idx-id->idx idx-id) (not-empty values))))))))
 
+(defmethod pred-constraint 'get-valid-end-time [_ {:keys [idx-id arg-bindings tuple-idxs-in-join-order valid-time] }]
+  (let [arg-bindings (rest arg-bindings)
+        [e-var not-found] arg-bindings
+        not-found? (= 2 (count arg-bindings))
+        e-result-index (.result-index ^VarBinding e-var)]
+    (fn pred-get-valid-end-time [{:keys [value-serde index-snapshot]} idx-id->idx ^List join-keys]
+      (let [e (.get join-keys e-result-index)
+            vs (->> (db/entity-history index-snapshot (db/decode-value value-serde e) :asc {})
```
jarohen: `db/entity-history` already takes various valid-time params - do you need a separate `drop-while` here?

jarohen (changes requested):
cheers @FiV0, nice one :pray: 

* There's a couple of things that may well Just Work™, but might be worth an explicit test:
  * Does the `?valid-time` logic var behave like any other logic var when it comes to unification? i.e. can you query for `:where [[(get-valid-end-time ?e) ?valid-time] [(get-valid-end-time ?e2) ?valid-time]]` and have it filter for `?e` and `?e2` having the same end valid-time?
  * Could you check this against a database that's back in transaction time? e.g. after your second change has been submitted and awaited, take out a database with tx-time from your first transaction. I'm not 100% sure we'd preserve repeatable queries here as it stands - do you get the same results for the first tx-time before and after the second transaction is submitted?
* Bikeshedding, but elsewhere we have `end-valid-time` params, so maybe this should be called `get-end-valid-time`?
* The addition of valid-time to the query cache is concerning - this essentially means that you'll need to query at exactly the same time (or manually specify the valid-time) to get a cache hit. Did you get any feeling for why this might have been required?
* meta (re pomegranate): on a more general note, we prefer to keep unrelated changes out of PRs - it's quite small on this occasion but it can get quite noisy.

Also, aware this is a 'first cut' so wouldn't have expected the below yet, but we'll want to do/consider them before we merge :slightly_smiling_face: 

* might we also want `get-start-valid-time`, for a more holistic API change?
* could you add it to the docs?

core/src/xtdb/query.clj
FiV0: Good point. I still need to drop one item in the case where the db `valid-time` coincides with the valid-time of the document. So the semantics of the `get-end-valid-time` are the next change of the entity according to the `db` except in the case where `valid-time` of the db an the document coincide.

test/test/xtdb/query_test.clj
```diff
@@ -1159,6 +1159,134 @@
...
+
+    (fix/submit+await-tx [[::xt/put {:xt/id :bar, :type :some-value2} later1 later2]])
+    (xt/sync *api*)
+
+    (t/testing "entity change"
+      (t/is (= #{[:bar later1]}
+               (xt/q (xt/db *api* now)
+                     '{:find [?e ?end-time]
+                       :where [[?e :type :some-value2]
+                               [(get-end-valid-time ?e) ?end-time]]})))
```
FiV0, line 1229: If I understand your comment concerning `back in transaction time` correctly. This test bit should already cover that. I added the `sync` call above it.

test/test/xtdb/query_test.clj
```diff
@@ -1159,6 +1159,134 @@
...
+
+    (t/testing "unification"
+      (t/is (= #{[:bar :toto] [:toto :bar]}
+               (xt/q (xt/db *api* now)
+                     '{:find [?e1 ?e2]
+                       :where [[?e1 :type]
+                               [?e2 :type]
+                               [(!= ?e1 ?e2)]
+                               [(get-end-valid-time ?e1) ?end-time]
+                               [(get-end-valid-time ?e2) ?end-time]]}))))))
```
FiV0, line 1290: Added the `unification` test above.

FiV0:
Concerning the other bits:
- [x] `get-valid-end-time` -> `get-end-valid-time`
- [x] addition of `get-start-valid-time`
- [x] removed `pomegranate`
- [x] addition to the docs
- IIRC I added the `valid-time` to the cache because I needed to pass it down to the pred contraint (via the ctx). Otherwise a query got cached and on later calls (different db instance) the valid-time would be off from the valid-time of the current db instance.

core/src/xtdb/query.clj
jarohen: paired - we tend to use `(inc valid-time)` for these cases

test/test/xtdb/query_test.clj
jarohen: paired on some tests to demo these cases

PR #1816, opened 2022-09-06
Fix LMDB MAP_FULL resizing race during store

wotbrew:
This PR the LMDB module has had some dependency bumps, including new natives for arm64 macs - which caused the tests to be runnable, but flaky.

TLDR; Flake seems to be due to a race condition in the `kv/store` implementation which allowed queued concurrent writes to all see a MAP_FULL error, and then go on each to increase the mapsize by a factor of two, very quickly resulting in segfault. 

## Notes on the `MAP_FULL` race

The pattern we use in the code is to double the size of the lmdb database when we receive a `MAP_FULL` on a kv/store call [here](https://github.com/xtdb/xtdb/blob/master/modules/lmdb/src/xtdb/lmdb.clj#L223). 

The (apparent) issue with the implementation is that multiple concurrent requests can race to resize the map at the same time (applying a factor), causing multiple doublings in a short amount of time, rather than an OOM error, this seems to result in a segfault. At least when running concurrent tests on my M1 mac.

This is made worse because of the exclusive lock on `increase-mapsize`, it means all pending transactions seeing `MAP_FULL` will queue up to increase the map size (and do so), regardless of whether it has already been increased by another thread. I observed a low <= 32 number of concurrent `store` operations could trigger this segfault.

It might be better to ensure at most one resize happens, and that pending operations (seeing the same `MAP_FULL`) are then retried. 

This PR attempts this by ensuring map resizes happen under an exclusive lock, so that only one of them gets to do so before pending writes are retried.

### DIFF
--- a/modules/lmdb/project.clj
+++ b/modules/lmdb/project.clj
@@ -14,9 +14,11 @@
   :dependencies [[org.clojure/clojure]
                  [org.clojure/tools.logging]
                  [com.xtdb/xtdb-core]
-                 [org.lmdbjava/lmdbjava "0.7.0"]
-                 [org.lwjgl/lwjgl "3.2.3" :classifier "natives-linux" :native-prefix ""]
-                 [org.lwjgl/lwjgl-lmdb "3.2.3" :classifier "natives-linux" :native-prefix ""]
-                 [org.lwjgl/lwjgl "3.2.3" :classifier "natives-macos" :native-prefix ""]
-                 [org.lwjgl/lwjgl-lmdb "3.2.3" :classifier "natives-macos" :native-prefix ""]
-                 [org.lwjgl/lwjgl-lmdb "3.2.3"]])
+                 [org.lmdbjava/lmdbjava "0.8.2"]
+                 [org.lwjgl/lwjgl "3.3.1" :classifier "natives-linux" :native-prefix ""]
+                 [org.lwjgl/lwjgl-lmdb "3.3.1" :classifier "natives-linux" :native-prefix ""]
+                 [org.lwjgl/lwjgl "3.3.1" :classifier "natives-macos" :native-prefix ""]
+                 [org.lwjgl/lwjgl-lmdb "3.3.1" :classifier "natives-macos" :native-prefix ""]
+                 [org.lwjgl/lwjgl "3.3.1" :classifier "natives-macos-arm64" :native-prefix ""]
+                 [org.lwjgl/lwjgl-lmdb "3.3.1" :classifier "natives-macos-arm64" :native-prefix ""]
+                 [org.lwjgl/lwjgl-lmdb "3.3.1"]])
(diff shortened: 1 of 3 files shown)


jarohen (changes requested):
@wotbrew any way we can refactor this to only use the mapsize-lock rather than two separate locks? the intention behind the mapsize-lock was (AFAIK) to do what you suggest in your comment - an exclusive lock so that writers and map-size-increasers don't run concurrently.

jarohen (changes requested):
hey @wotbrew, cheers for the updates :pray: 

* I don't quite understand why we need to split the transaction part into the two code paths, taking and not-taking the lock - to someone who isn't as familiar with the context it seems like this detail is being leaked? Could well be that it's required and I haven't had the lightbulb moment, so maybe it's best discussed? Naively, I'd be considering the normal transaction operation, the resizing, and some form of coordination as three separate components - maybe you've also considered this and it doesn't work, but it'd be good to understand why.
* subjective, probably, but I'm not mad keen on the refactoring to introduce 'CloseableBox' - doesn't seem like this change mandates it, I don't think the abstraction saves us a lot, and I'm not convinced it makes the code clearer? might be 'OO', but even in FP I like that resources with lifecycles, like txs and cursors, are specifically modeled as objects.

catch you tomorrow, if it's a good time :slightly_smiling_face:

wotbrew:
> * I don't quite understand why we need to split the transaction part into the two code paths, taking and not-taking the lock - to someone who isn't as familiar with the context it seems like this detail is being leaked? Could well be that it's required and I haven't had the lightbulb moment, so maybe it's best discussed? Naively, I'd be considering the normal transaction operation, the resizing, and some form of coordination as three separate components - maybe you've also considered this and it doesn't work, but it'd be good to understand why.

There is always more than one way to skin a cat, but the locking policy changes (both in ownership semantics, and read/write stamp) so that was the motivation, one with ownership (RAII / close), one without ownership (locks are on you). Being able to have a path without locks allows the cursor put to sit under a simple unconditional exclusive lock (write tx are exclusive via LMDB blocking anyway). Other pathways seemed to lead to lock promotion or more fields on the Transaction object (e.g mutable stamp to upgrade its privilege) 

> * subjective, probably, but I'm not mad keen on the refactoring to introduce 'CloseableBox' - doesn't seem like this change mandates it, I don't think the abstraction saves us a lot, and I'm not convinced it makes the code clearer? might be 'OO', but even in FP I like that resources with lifecycles, like txs and cursors, are specifically modeled as objects.

Early look at varying the lock policy necessitated more specificity/conditionality if it were part of something like LMDBTransaction, there may be some model where this may be clearer that I didn't see - so let's chat. Motivation for `closeable` then is simply`with-open` a-la-carte to sidestep the modelling problem, focusing on what was necessary mechanistically. Happy to delete if something different would be a better fit.

> catch you tomorrow, if it's a good time 🙂

Yep 👍

PR #1820, opened 2022-09-13
Reduce allocations in free-memory-loop.

sw1nn:
Whilst investigating performance issues with Second Chance cache I noticed that the free-memory-loop was allocating ~1k bytes/sec. 

This is excessive.

Two possible causes are noted
* multiple calls to `(Runtime/getRuntime)`
* (unnecessary) boxing of primitives and the creation of a clojure.lang.Ratio which is then cast immediately to a double. This is another occurrence of the issue fixed elsewhere in 9a36cad2e7316e4c8aa093af7a698292fe398033

This PR rearranges `free-memory-loop` to reduce allocations. With this change the allocation rate falls to ~24B/sec

### DIFF
--- a/core/src/xtdb/cache/second_chance.clj
+++ b/core/src/xtdb/cache/second_chance.clj
@@ -113,15 +113,16 @@
 (def ^:private ^:const free-memory-check-interval-ms 1000)
 
 (defn- free-memory-loop []
-  (try
-    (while true
-      (let [max-memory (.maxMemory (Runtime/getRuntime))
-            used-memory (- (.totalMemory (Runtime/getRuntime))
-                           (.freeMemory (Runtime/getRuntime)))
-            free-memory (- max-memory used-memory)]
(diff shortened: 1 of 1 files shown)


sw1nn:
On further reflection, I realised that the code should be possible without making _any_ allocations. It turns out the remaining allocation is caused by the use of `AtomicReference` for free-memory-ratio. 

In principle this ratio could be stored as an `AtomicLong` (range 0..100) and the conversion to double to use in the calculation in `move-to-cold-state`  reworked. This makes the code a little less clear and is less compelling that the initial change proposed here.

See https://github.com/sw1nn/xtdb/commit/b8bebb896c740802214b819060b4211eff6dcd5e

jarohen:
building [here](https://app.circleci.com/pipelines/github/jarohen/xtdb/2698/workflows/279b696b-59c5-43a6-acaf-e24ae390cfe3)

Issue #1818, opened 2022-09-12
Second Chance cache should not leak memory


refset:
Details below are from a dump of an application running normally, but for which an OOM was recently triggered (during long-running queries/exports, and possibly also concurrently with a checkpoint being created). This heat dump is from normal operation...

"Retained Size" descending:
![image](https://user-images.githubusercontent.com/456658/189623283-5201f1ba-d7a2-4fc9-8787-3e932cfd73c4.png)

"[Number of] Objects" descending:
![image](https://user-images.githubusercontent.com/456658/189623352-deb7cb5f-e122-46d1-9910-6ddc00e12e25.png)

Thank you for the report @FiV0

sw1nn:
The `cooling` queue is an [unbounded](https://github.com/xtdb/xtdb/blob/065f41945daed67c6742a5c480864eaa24f6e48d/core/src/xtdb/cache/second_chance.clj#L182) `LinkedBlockingQueue`, but items are [placed on the queue](https://github.com/xtdb/xtdb/blob/065f41945daed67c6742a5c480864eaa24f6e48d/core/src/xtdb/cache/second_chance.clj#L109) with `.offer` which at least implies there is an expectation in the design that the queue is bounded. I don't have any sage advice on how the cooling queue should be sized however

PR #1821, opened 2022-09-13
Apply back pressure to second chance writers when resizing

wotbrew:
It has been observed that the second chance cache may be using too much memory here #1818 

After reviewing the code I noticed that a cache can be full (and resizing) but no back pressure is applied to producers who may continue creating new cache entries.

I ran some experiments and observed that even with 2 threads, you could cause the cache to grow far beyond the `:cache-size` limit. (See examples below).

## Examples

I set up a quick test that filled the cache with integers over 10 seconds, and sampled the size of the hot cache every 10 milliseconds.

- `:cache-size` is `1024`
- `:adaptive-sizing?` is `false`

### 1 Thread
image

### 2 Threads
image

### 42 Threads
image

## Semaphore solution (initial choice) 

Force cache misses to await any pending resize when the cache is bigger than the max size, applying back pressure. A small buffer of `1024` allows a small number of misses to not block, if contention is not high. We reuse the resize semaphore to cause the blocking behaviour.

##  Counter solution

An alternative solution that was considered is to use an atomic counter initialised to max size, each put decrements the counter - if negative, the put must spin (and increment) until the counter is not negative. Evictions increment the counter.

This should bring about desired back pressure without risking the `.size` problem (see @sw1nn comment) and as far as co-ordination is concerned _should_ be very cheap - measurement pending.

### DIFF
--- a/core/src/xtdb/cache/second_chance.clj
+++ b/core/src/xtdb/cache/second_chance.clj
@@ -34,6 +34,29 @@
               (when-not (and (= next start) wrapped-around)
                 (recur next (if last true wrapped-around))))))))))
 
+(defn- adaptive-cache-max-size ^long [^long size ^double adaptive-break-even-level ^double free-memory-ratio]
+  (long (Math/pow size (+ adaptive-break-even-level free-memory-ratio))))
+
+(def ^:private ^AtomicReference free-memory-ratio (AtomicReference. 1.0))
+(def ^:private ^:const free-memory-check-interval-ms 1000)
+
+(defn- free-memory-loop []
+  (try
+    (while true
+      (let [max-memory (.maxMemory (Runtime/getRuntime))
+            used-memory (- (.totalMemory (Runtime/getRuntime))
+                           (.freeMemory (Runtime/getRuntime)))
+            free-memory (- max-memory used-memory)]
+        (.set free-memory-ratio (double (/ free-memory max-memory))))
+      (Thread/sleep free-memory-check-interval-ms))
+    (catch InterruptedException _)))
+
+(def ^:private ^Thread free-memory-thread
+  (do (when (and (bound? #'free-memory-thread) free-memory-thread)
+        (.interrupt ^Thread free-memory-thread))
+      (doto (Thread. ^Runnable free-memory-loop "xtdb.cache.second-chance.free-memory-thread")
+        (.setDaemon true))))
+
 (declare resize-cache)
 
 (definterface SecondChanceCachePrivate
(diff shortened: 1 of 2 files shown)


core/src/xtdb/cache/second_chance.clj
```diff
@@ -34,6 +34,29 @@
...
+            free-memory (- max-memory used-memory)]
+        (.set free-memory-ratio (double (/ free-memory max-memory))))
+      (Thread/sleep free-memory-check-interval-ms))
+    (catch InterruptedException _)))
+
+(def ^:private ^Thread free-memory-thread
+  (do (when (and (bound? #'free-memory-thread) free-memory-thread)
+        (.interrupt ^Thread free-memory-thread))
+      (doto (Thread. ^Runnable free-memory-loop "xtdb.cache.second-chance.free-memory-thread")
+        (.setDaemon true))))
```
FiV0, line 58: Some duplication from below.

wotbrew: Thanks 🙏 , copy paste error!

sw1nn (commented):
Not exactly specific to this change, but from [ConcurrentHashMap javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html) this comment seems pertinent.

> Bear in mind that the results of aggregate status methods including size, isEmpty, and containsValue are typically useful only when a map is not undergoing concurrent updates in other threads. Otherwise the results of these methods reflect transient states that may be adequate for monitoring or estimation purposes, but not for program control. 

There are many occurences in SCC of using aggregate status methods and in fact we have already [fallen foul of use of isEmpty](https://github.com/xtdb/xtdb/commit/25b7144031c588d0f40b14665686938e3dd9f589) for program control, where the SCC entered a tight loop in some circumstances

I have concerns that the proposed solution to apply backpressure may not work if the guarantees around `size()` on CHM are as documented.

wotbrew:
Thanks @sw1nn 🙏 , so using .size then is more of a heuristic or hint in this case. That said as a heuristic _it does seem_ to apply enough back pressure to keep the cache size almost entirely within the bounds over time - at least for my 42 threads banging on the cache all at once test! I am aware that is only one test however, and perhaps there are access scenarios that would cause problems.

I also have not yet performed the E2E perf tests (though I should get to that today), which may show an unacceptable performance regression.

As discussed on slack lets step back and talk about other solutions that might be a better fit before committing to this one due to the concern.

jarohen:
closing, feel free to re-open if/when we come back to it

PR #1827, opened 2022-09-28
Fix order-by for external sort

FiV0:
Currently the invocation of the `order-by` comparator  for external sorting is broken https://github.com/xtdb/xtdb/blob/ae34fd5b27aa33ea914df6bc4ab52c3e3d853c19/core/src/xtdb/io.clj#L130 as Comparators don't implement `IFn`.

I also made `default-external-sort-part-size` in `xtdb.io` dynamic which made testing easier and would allow dynamic adjustment of memory usage for certain `order-by` queries. This doesn't need to go in like this. We could also make this configurable on start up as the current default seems to be an issue on my current project.

### DIFF
--- a/core/src/xtdb/io.clj
+++ b/core/src/xtdb/io.clj
@@ -147,7 +147,7 @@
    (->> (new-merge-sort-priority-queue comp sorted-seqs)
         (merge-sort-priority-queue->seq))))
 
-(def ^:const default-external-sort-part-size (* 1024 1024))
+(def default-external-sort-part-size (* 1024 1024))
 
 (defn with-nippy-thaw-all* [f]
   (binding [nippy/*thaw-serializable-allowlist* #{"*"}]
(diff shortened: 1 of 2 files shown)


FiV0:
Another option is to do something like the following in `xtdb.io`
```clj
(defn- new-merge-sort-priority-queue-patched ^PriorityQueue [comp sorted-seqs]
  (let [sorted-seqs (remove empty? sorted-seqs)
        pq-comp (reify Comparator
                  (compare [_ [a] [b]]
                    (cond (fn? comp) (comp a b)
                          (instance? Comparator comp) (.compare comp a b)
                          :else (throw (ex-info "Don't know how to apply comparator" {:comp comp})))))]
    (doto (PriorityQueue. (max 1 (count sorted-seqs)) pq-comp)
      (.addAll sorted-seqs))))
```

PR #1829, opened 2022-10-01
Include awaited-tx info on transaction aborted exception

nivekuil:
Bit of a data loss incident with cloudflare r2, manual surgery is impossible without this info.

### DIFF
--- a/core/src/xtdb/node.clj
+++ b/core/src/xtdb/node.clj
@@ -58,7 +58,7 @@
       (future-cancel fut))
 
     (cond
-      ingester-error (throw (Exception. "Transaction ingester aborted." ingester-error))
+      ingester-error (throw (Exception. (str "Transaction ingester aborted on tx " (pr-str awaited-tx)) ingester-error))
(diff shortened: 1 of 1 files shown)


nivekuil:
I think this gives less information than I hoped, since the awaited tx on these is usually just the latest one

jarohen:
Hey @nivekuil - thanks for the PR! 

Yep, as you suspected, this unfortunately doesn't give the transaction that actually failed - only the one that you're awaiting (which may be the same transaction, of course).

We're certainly aware that ingester errors are fairly opaque at the moment - we currently have #1831 in flight. One thing we are considering is making the ingester error available as part of the status map - would this help in your use case?

James

nivekuil:
Ideally the doc store api would be given the transaction context so it could tell you what when wrong immediately (submit-docs would benefit too because it would let implementations be clever about locality) but I understand core2 is dramatically changing all that so not such how much you'd invest in it now. Anyway this one is not the right approach, closing.

Issue #1831, opened 2022-10-05
Improving user support on ingester errors


jarohen:
At the moment, in the unlikely event of an ingester error, we don't provide users with enough guidance about what's happened, why, or what to do next.

"For example, if the doc store has latency spikes and one request happens to hits a timeout, the node will left in a degraded state indefinitely" (as per #1854).

* include ingester-error in the API (maybe status map?) so that users can easily check for it (e.g. for healthchecks)
* link to documentation in the error message?
* consider node configuration to deterministically skip a troublesome transaction (needs to happen on all nodes - may require a reindex.)

wotbrew:
Closed by #1906

PR #1832, opened 2022-10-05
LMDB fairer lock synchronisation

### LINKED ISSUE #1826: LMDB "Transaction ingester aborted" problem in 1.22.0
After running some time, the node seems to stop ingesting.
Await tx causes exception.

The node in question is running on Linux in AWS ECS.
Starting a new instance will restore from checkpoint and indexing works again. 
Don't know the exact pattern that causes the initial problem.

Log has entries with the following exception:

```
java.lang.Exception: Transaction ingester aborted.
at xtdb.node$await_tx.invokeStatic(node.clj:60)
at xtdb.node$await_tx.invoke(node.clj:36)
at xtdb.node.XtdbNode.await_tx(node.clj:158)
at xtdb.node.XtdbNode.await_tx(node.clj:155)
... application code...
Caused by: java.lang.AssertionError: Assert failed: LMDB write lock timeout
(pos? stamp)
at xtdb.lmdb$acquire_write_lock.invokeStatic(lmdb.clj:39)
at xtdb.lmdb$acquire_write_lock.invokePrim(lmdb.clj)
at xtdb.lmdb.LMDBKv.store(lmdb.clj:238)
at xtdb.fork.ForkedKvIndexStoreTx.commit_index_tx(fork.clj:188)
at xtdb.tx.InFlightTx.commit(tx.clj:356)
at xtdb.tx$__GT_tx_ingester$fn__45494$txs_index_fn__45502.invoke(tx.clj:567)
at clojure.lang.AFn.applyToHelper(AFn.java:154)
at clojure.lang.AFn.applyTo(AFn.java:144)
at clojure.core$apply.invokeStatic(core.clj:665)
at clojure.core$apply.invoke(core.clj:660)
at xtdb.tx$__GT_tx_ingester$fn__45494$submit_job_BANG___45495$fn__45496.invoke(tx.clj:544)
at clojure.lang.AFn.call(AFn.java:18)
... 4 common frames omitted
```

The earliest related message seems to be:
```
level | ERROR
logger | xtdb.tx
message | Ingester error occurred
thread | xtdb-tx-processor-1

java.lang.AssertionError: Assert failed: LMDB write lock timeout
(pos? stamp)
at xtdb.lmdb$acquire_write_lock.invokeStatic(lmdb.clj:39)
at xtdb.lmdb$acquire_write_lock.invokePrim(lmdb.clj)
at xtdb.lmdb.LMDBKv.store(lmdb.clj:238)
at xtdb.fork.ForkedKvIndexStoreTx.commit_index_tx(fork.clj:188)
at xtdb.tx.InFlightTx.commit(tx.clj:356)
at xtdb.tx$__GT_tx_ingester$fn__45494$txs_index_fn__45502.invoke(tx.clj:567)
at clojure.lang.AFn.applyToHelper(AFn.java:154)
at clojure.lang.AFn.applyTo(AFn.java:144)
at clojure.core$apply.invokeStatic(core.clj:665)
at clojure.core$apply.invoke(core.clj:660)
at xtdb.tx$__GT_tx_ingester$fn__45494$submit_job_BANG___45495$fn__45496.invoke(tx.clj:544)
at clojure.lang.AFn.call(AFn.java:18)
at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
at java.base/java.lang.Thread.run(Thread.java:833)
```

After that all the await tx calls start failing.

wotbrew:
This PR _should_ fix #1826

## Problem

Release 1.22.0 introduced an additional write lock to avoid segfault when handling `MDB_MAP_FULL` errors (#1816). The write lock was used around all writes, as exclusivity of write transaction is necessary anyway for LMDB. This had an unintended consequence where the `acquire-write-lock` timeout was now used for ingester writes, where it would previously only be used for resizes.

Now, the timeout itself is fairly generous for writes, however due to the lack of fairness/write priority with the StampedLock mechanism, @tatut described running into ingester timeouts when ingesting nodes were also under high concurrent read load.

## Fix

This PR introduces a new locking mechanism in the java class `xtdb.lmdb.MapResizeSync`. It provides shared locks for open transactions, and exclusive 'resize' lock to deal with `MDB_MAP_FULL`.  It uses the base `AbstractQueueSynchronizer` to provide fairness and write priority to avoid in more situations reads blocking the ingester.

## Write lock timeouts

The PR as it stands _removes the write lock timeout_.  

I do not think there is a single good policy here, it depends ultimately on your write workload, but also how long at the limit your queries take, which is presumably open for many deployments.

I would welcome comments, not sure if a chesterton's fence situation. I understand that the timeout was introduced with #945 to provide observability of deadlock or otherwise blocking for too long. I think there may be better ways to achieve this without failing the ingester during resize.

I propose instead of timing out, we implement an ingester heartbeat (if it does not already exist) which will `WARN` if the ingester is blocked (for any reason, not just on lock acquisition). Perhaps as part of #1831. There is also deadlock/contention detection `ThreadMXBean` we can leverage to provide telemetry more holistically.

### DIFF
--- a/modules/lmdb/project.clj
+++ b/modules/lmdb/project.clj
@@ -11,6 +11,13 @@
 
   :scm {:dir "../.."}
 
+  :java-source-paths ["src"]
+  :javac-options ["-source" "8" "-target" "8"
+                  "-XDignore.symbol.file"
+                  "-Xlint:all,-options,-path"
+                  "-Werror"
+                  "-proc:none"]
+
   :dependencies [[org.clojure/clojure]
                  [org.clojure/tools.logging]
                  [com.xtdb/xtdb-core]
--- a/modules/lmdb/project.clj
+++ b/modules/lmdb/project.clj
@@ -1,23 +1,6 @@
 (defproject com.xtdb/xtdb-lmdb "<inherited>"
   :description "XTDB LMDB"
 
-  :plugins [[lein-parent "0.3.8"]]
-
-  :parent-project {:path "../../project.clj"
-                   :inherit [:version :repositories :deploy-repositories
-                             :managed-dependencies
-                             :pedantic? :global-vars
-                             :license :url :pom-addition]}
-
-  :scm {:dir "../.."}
-
-  :java-source-paths ["src"]
-  :javac-options ["-source" "8" "-target" "8"
-                  "-XDignore.symbol.file"
-                  "-Xlint:all,-options,-path"
-                  "-Werror"
-                  "-proc:none"]
-
   :dependencies [[org.clojure/clojure]
                  [org.clojure/tools.logging]
                  [com.xtdb/xtdb-core]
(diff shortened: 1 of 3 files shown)


modules/lmdb/project.clj
```diff
@@ -11,6 +11,13 @@
 
   :scm {:dir "../.."}
 
+  :java-source-paths ["src"]
```
jarohen, line 27: you'll need to pull in the javadoc and sources classifiers from another module btw (e.g. core) - now that you've got a public Java class Maven Central will expect javadoc and sources jars

PR #1833, opened 2022-10-10
Include documents indexed as part of tx-fn expansion in attribute stats

### LINKED ISSUE #1825: Attribute stats wrong in 1.22.0 version
I noticed that after upgrading to XTDB 1.22.0 and recreating the indexes, the attribute-stats are wrong.

In my case there were multiple attributes missing from attribute stats, but they could be found in queries and when pulling data.
Looks like it might be related to tx-ops generated by transaction functions.

(tested with tx-log, docs and indexes all in separate RocksDB kv-stores).

Discussion in clojurians slack https://clojurians.slack.com/archives/CG3AM2F7V/p1663659031894769

```clojure
(defn check-all-attrs [node]
  (let [doc-count (atom 0)
        attrs (atom #{})]
    (with-open [docs (xt/open-q (xt/db node)
                                '[:find ?e :where [?e :xt/id]])]
      (doseq [[d] (iterator-seq docs)
              :let [e (xt/entity (xt/db node) d)]]
        (swap! attrs into (keys e))
        (swap! doc-count inc)))
    {:docs @doc-count
     :attrs @attrs}))
```

The above check code returned attribute set that had more values than were in the `xt/attribute-stats`map.

wotbrew:
Fixes #1825

# Problem

Attribute stats are not calculated for documents introduced by transaction function expansion. Possibly a regression as part of the pipelined ingester work.

## Solution

This PR changes `db/index-tx-events` to return docs indexed during tx-event indexing (incl evaluation of transaction functions).

If the transaction is to be committed, these extra docs are sent (with the top-level ones) to the stats pipeline.

### DIFF
--- a/core/src/xtdb/db.clj
+++ b/core/src/xtdb/db.clj
@@ -77,7 +77,7 @@
 
 (defprotocol InFlightTx
   (index-tx-docs [in-flight-tx docs])
-  (index-tx-events [in-flight-tx tx])
+  (index-tx-events [in-flight-tx tx] "Returns a map of :committing? (true/false) representing whether the transactions were indexed, :indexed-docs (any new docs indexed during the expansion of tx fns)")
   (commit [in-flight-tx tx])
   (abort [in-flight-tx tx]))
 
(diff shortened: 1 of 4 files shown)


jonpither (approved):

jonpither:
LGTM

PR #1836, opened 2022-10-24
Fix non nil end-valid-time when start-valid-time is nil

### LINKED ISSUE #1835: Semantics of `nil` for `valid-time` and `end-valid-time`
I think `nil` for `valid-time` and `end-valid-time` should behave just like not being specified. This doesn't seem to be quite the case if the `start-valid-time` time is `nil`and `end-valid-time` not. 
```clj 
  (def node (xt/start-node {}))

  (xt/submit-tx node [[::xt/put {:xt/id 1 :v 0} nil #inst "2100"]])

  (xt/q (xt/db node)
        '{:find [(pull ?e [*]) ]
          :where [[?e :xt/id 1]]})
  ;; => #{}

  (xt/q (xt/db node #inst "2150")
        '{:find [(pull ?e [*]) ]
          :where [[?e :xt/id 1]]})
  ;; => #{[{:v 0, :xt/id 1}]}
```
It seems that in this case the `end-valid-time` is taken as `start-valid-time`.
```clj
  (xt/entity-history (xt/db node #inst "2150") 1 :asc)
  ;; => [#:xtdb.api{:tx-time #inst "2022-10-20T13:20:46.375-00:00",
  ;;                :tx-id 0,
  ;;                :valid-time #inst "2100-01-01T00:00:00.000-00:00",
  ;;                :content-hash
  ;;                #xtdb/id "36346e45c1e0dbc2fb2b1be69bc6ffaaee1d224f"}]
```

FiV0:
This is a potential fix for #1835. Mainly fixes the operation `[::xt/put doc nil #inst "2500"]`.

Things to consider: This now writes `nil` values for certain `start-valid-time`'s in the tx-log. Could this have any unintended consequences? Backwards compatibility issues?

### DIFF
--- a/core/src/xtdb/tx/conform.clj
+++ b/core/src/xtdb/tx/conform.clj
@@ -145,12 +145,12 @@
 
 (defmethod ->tx-event ::xt/put [{:keys [eid doc-id start-valid-time end-valid-time]}]
   (cond-> [:crux.tx/put (c/new-id eid) doc-id]
-    start-valid-time (conj start-valid-time)
+    (or start-valid-time end-valid-time) (conj start-valid-time)
     end-valid-time (conj end-valid-time)))
 
 (defmethod ->tx-event ::xt/delete [{:keys [eid start-valid-time end-valid-time]}]
   (cond-> [:crux.tx/delete (c/new-id eid)]
-    start-valid-time (conj start-valid-time)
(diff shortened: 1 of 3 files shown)


jarohen:
cheers @FiV0 :pray:

PR #1843, opened 2022-11-07
Reduce likelyhood of use-after-free segfault (LMDB/RocksDB)

### LINKED ISSUE #1803: Calling functions on a closing/closed node results in segfaults
Every now and then, my REPL crashes with a segmentation fault, when trying to create a `db` object for a LMDB-backed index store.

It happens from a thread started by a `ScheduledThreadPoolExecutor`.

I've seen similar segfaults with RocksDB.

Example logfiles generated when the segmentation fault happens:
* [hs_err_pid20189.log](https://github.com/xtdb/xtdb/files/9375714/hs_err_pid20189.log)
* [hs_err_pid63855.log](https://github.com/xtdb/xtdb/files/9375715/hs_err_pid63855.log)

wotbrew:
Currently, RocksDB and LMDB kv implementations will segfault if accessed after `.close`.

Resolves #1803.

In practice it is very hard for us to protect against racing with `.close` without locking when kv pointers are in use - so instead this PR offers a pragmatic solution for the immediate concern of trivial user-after-free.

This PR extends the index-store ThreadManager to track open snapshots in more methods that currently cause segfaults.

Additionally, the KvTxLog checks if-closed in a few more places to avoid trivial segfaults (such as on `last-submitted-tx`).

### DIFF
--- a/core/src/xtdb/kv/index_store.clj
+++ b/core/src/xtdb/kv/index_store.clj
@@ -674,36 +674,45 @@
     (latest-completed-tx-i i)))
 
 (defprotocol PThreadManager
-  (snapshot-opened [_ snapshot])
-  (snapshot-closed [_ snapshot]))
+  "An object that tracks unsafe kv usage (as tickets) and the threads that hold valid pointers to native kv handles (rocks/lmdb).
+
+  This tracking is necessary to avoid use after free segfaults.
+
+  Before opening a snapshot, call acquire-snapshot-ticket, when the snapshot is closed call .close on the ticket.
+
+  IMPORTANT:
+  When .closed the thread manager should attempt to interrupt threads that acquired said tickets. It is safe to interrupt these
+  threads only if valid pointers to index or kv snapshots are not passed between threads, otherwise arbitrary threads
+  could be interrupted by the thread manager."
+  (open-kv-ticket ^Closeable [thread-manager]))
 
 (deftype ThreadManager [^Map snapshot-threads
                         ^:volatile-mutable ^Semaphore closing-semaphore]
   PThreadManager
-  (snapshot-opened [this snapshot]
-    (locking this
+  (open-kv-ticket [thread-mgr]
+    (locking thread-mgr
       (when closing-semaphore
         (throw (IllegalStateException. "closing")))
-
-      (.put snapshot-threads snapshot (Thread/currentThread))))
-
-  (snapshot-closed [this snapshot]
-    (locking this
-      (.remove snapshot-threads snapshot)
-      (when closing-semaphore
-        (.release closing-semaphore))))
+      (let [ticket (reify Closeable
+                     (close [ticket]
+                       (locking thread-mgr
+                         (.remove snapshot-threads ticket)
+                         ;; use intentional field access to avoid closure over mutable field
+                         (when-some [^Semaphore closing-semaphore (.closing-semaphore thread-mgr)]
+                           (.release closing-semaphore)))))]
+        (.put snapshot-threads ticket (Thread/currentThread))
+        ticket)))
 
   Closeable
-  (close [this]
-    (let [open-thread-count (locking this
-                              (when-not closing-semaphore
-                                (set! (.closing-semaphore this) (Semaphore. 0))
-
-                                (let [open-threads (vals snapshot-threads)]
-                                  (doseq [^Thread thread (set open-threads)]
-                                    (.interrupt thread))
-                                  (count open-threads))))]
-
+  (close [thread-mgr]
+    (when-some [open-thread-count
+                (locking thread-mgr
+                  (when-not closing-semaphore
+                    (set! (.closing-semaphore thread-mgr) (Semaphore. 0))
+                    (let [open-threads (vals snapshot-threads)]
+                      (doseq [^Thread thread (set open-threads)]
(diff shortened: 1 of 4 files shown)


wotbrew:
As an additional pass prompted by a convo with @jarohen 🙏 , I've looked at the kv-index `ThreadManager` which is intended to protect against this cases like this when snapshots are known to be in use. It tracks threads that are working with open snapshots and interrupts them on `.close`.

It also contains a check that throws on close, however, it will do so _after it receives a snapshot pointer_, so I do not believe it helps in this case, where the segfault occurs on snapshot open. 

There are operations on snapshots that do are not covered by the ThreadManager (e.g `latest-completed-tx`), we may want to review these, though their lifetimes will typically be short, we could see use-after-free segfaults for similar reasons as the original issue.

I've spent an hour or so thinking about this, so perhaps I've missed something obvious.

## Proposal A - use ThreadManager, but not with an open snapshot

ThreadManager could be passed an arbitrary object instead of the snapshot itself, so we can say can opt into the safety it provides without having the need to open the snapshot first. The API does not actually require the 'snapshot' arg is a snapshot as far as I can tell, so this would work as is
(though we may want to consider changing names or comments so such usage is not confusing to readers in the future).

## Proposal B - all problems can be solved by another layer of indirection

Use a box on unsafe references such as RocksDB and the LMDB env pointer, inspired by rusts RefCell/C++ smart pointers.

If accesses to the Rocks/LMDB pointers were protected through the need to deref a box, we can ensure such a box always throws if the pointer may no longer be valid (due to .close).

Furthermore, such a box would permit tracking borrows, and as such, we may be able to do away with ThreadManager.

This would depend on the correct usage of the box, after all, nobody is going to check we don't just start throwing the unsafe references themselves around.

This will cost us a volatile deref on read in several places, though lower-level code can 'borrow' the raw pointer and use it unboxed within a scope, and makes explicit the safety dimension to the KV impls, where before they were somewhat intentionally unsafe.

---

As an aside, I am somewhat uncertain about the .interrupt behaviour (whether or not we make explicit the requirement to open, use and then close a snapshot on a single thread. Though I can imagine this is often the case in practice, this requirement conflicts with the seemingly intentional choice to allow transactions to cross thread boundaries in our LMDB wrapper. I can raise this as an issue if somebody agrees it might be a problem who has a better understanding of the intention behind these decisions.

wotbrew:
As discussed @jarohen I spent a bit of time looking at changing ThreadManager to re-name and document the open in terms of a ticket or sentinel rather than a snapshot.

e.g 

```clojure

(defprotocol PThreadManager
  "An object that tracks unsafe kv usage (as tickets) and the threads that hold valid pointers to native kv handles (rocks/lmdb).

  This tracking is necessary to avoid use after free segfaults.

  Additionally, when .closed the thread manager will attempt to interrupt threads that acquired said tickets. It is safe to interrupt these
  threads only as we do not expect valid pointers to index or kv snapshots to be passed between threads.

  Before opening a snapshot, call (open-kv-ticket), when the snapshot is closed call .close on the ticket."
  (open-kv-ticket ^Closeable [thread-manager]))

(deftype ThreadManager [^Map snapshot-threads
                        ^:volatile-mutable ^Semaphore closing-semaphore]
  PThreadManager
  (open-kv-ticket [this]
    (locking this
      (when closing-semaphore
        (throw (IllegalStateException. "closing")))
      (let [ticket (reify Closeable
                     (close [ticket]
                       (locking this
                         (.remove snapshot-threads ticket)
                         (when closing-semaphore
                           (.release closing-semaphore)))))]
        (.put snapshot-threads ticket (Thread/currentThread))
        ticket)))

```

I tried to document the protocol to clarify its now intended usage. Note the interrupt behaviour I've specified some semantics for the ticket-thread connection, which was uncomfortable to write and I do not think it is right. It is harder to frame this when this PThreadManager is not _only_ about 'threads holding open snapshots'.

This approach leads to calls like this: 

```clojure
  db/IndexSnapshotFactory
  (open-index-snapshot [_]
    (let [ticket (open-kv-ticket thread-mgr)]
      (try
        (new-kv-index-snapshot (kv/new-snapshot kv-store) true ticket cav-cache canonical-buffer-cache)
        (catch Throwable e
          (xio/try-close ticket)
          (throw e)))))
```

or 

```clojure
  db/LatestCompletedTx
  (latest-completed-tx [_]
    (with-open [_ (open-kv-ticket thread-mgr)]
      (latest-completed-tx kv-store)))
```

The goal then would be to employ guards like this to all sections where the kv store handle might be dereferenced unless already guarded by a ticket. Or rely on assumptions about the above layers (like the node itself and its api) to demonstrate safety in the absence of such guards.

The original bug report showed use after free on metadata/read paths rather than write so an argument could be made to only guard those sections and ignore the rest.

I feel nervous about this solution, the level of care I have to employ to get it right (one can imagine a naive version of open-index-snapshot without the careful exception/lifetime handling) and the amount of knowledge I have to have of how raw pointers are accessed leaking through the layers of abstraction on top of the kv. I understand that these concerns are subjective impressions!

That said _I think I can see something like this through_ that will provide at least better-than-now seg-fault protection.

Are you still satisfied with a use-thread-mgr type solution like this @jarohen? Is there a better way I can use the thread manager to avoid these use after free?

PR #1845, opened 2022-11-09
Add detail to `listen` docstring

### LINKED ISSUE #1844: `xtdb.api/listen` doc does not provide enough detail on behaviour
Remove "experimental" from listen API docstring and add further details: what I often look for in any listener adding functions is what I’m allowed to do in the handlers, like is it using some threadpool that could be exhausted if my listeners block or do something long running (thanks @tatut)

wotbrew:
This PR contains a new doc string for `listen`, resolves #1844.

Goals:

- Remove the experimental notice
- Remove a few ambiguities around threads and blocking
- Provide examples of what the `f` param is gonna get
- Provide some ideas for its utility

Some of this might be better delivered in the web documentation and a link. Further Improvement suggestions are welcome!
 
`listen` is a powerful function, I am sure the docs deserve further refinement.

### DIFF
--- a/core/src/xtdb/api.clj
+++ b/core/src/xtdb/api.clj
@@ -68,16 +68,54 @@
   node.")
 
   (listen ^java.lang.AutoCloseable [node event-opts f]
-    "Attaches a listener to XTDB's event bus.
+    "Attaches a listener to XTDB's event bus, the supplied callback `f` will be invoked as new events occur, receiving the event as the first argument.
 
-  `event-opts` should contain `:xtdb.api/event-type`, along with any other options the event-type requires.
+  Specify an event type `:xtdb.api/event-type` and options in the map `event-opts`
 
-  We currently only support one public event-type: `:xtdb.api/indexed-tx`.
-  Supplying `:with-tx-ops? true` will include the transaction's operations in the event passed to `f`.
+  For example, this listener prints indexed transactions:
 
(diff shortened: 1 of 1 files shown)


wotbrew:
Some things I noticed when investigating `listen` that I did not address super directly in the docs. If I'm correct I'd rather see if I can solve them so we do not need to document these properties:

- Each listener receives events on an unbounded (executor .submit) queue and does not provide back pressure, so if your callback is slow - events will build up and you may run out of memory.
-  Each listener starts a thread (via an Executor with one fixed thread) to process events. This means calling .close is necessary to stop and free resources acquired by that thread.  Could instead we use a cached thread pool or even possibly use the clojure.lang.Agent cached thread pool executor?  Using an agent or queue we can retain each listener receiving events in order without overlapping calls, and the property that we do not block or consume threads needed for 'real work' (incl other listeners). This would remove the 'resource management aspect of listening and so .close would only serve to detach the listener, GC can do the rest.

Are issues appropriate, has somebody already thought about this? Any thoughts are welcome 🙏!

wotbrew:
Discussed the above comments with @jarohen 🙏 

For both points one idea that was considered was allowing the specification of the executor, used for each listener, this would allow the user to vary both the queue and thread policy as required. 

Merging the above docstring, I would like to hammock the queue / policy a bit longer before committing to anything.

PR #1858, opened 2023-01-07
Add poll-wait-duration option for xtdb.api/open-tx-log #1834

### LINKED ISSUE #1834: open-tx-log with Kafka should tolerate delays longer than poll-wait-duration
If a Kafka log has high latency (e.g. a local node connecting to a remote data center), then `open-tx-log` should wait until the results are properly returned.

Currently XT will abort after `poll-wait-duration` is exceeded (e.g. returning an empty iterator). This can be trivially observed by lowering `poll-wait-duration`, and less-than-ideally mitigated by increasing `poll-wait-duration`.

Thanks for the report and help with the repro @alexisvincent :pray:

Notes:
1. open q: should open-tx-log (on the node, not the tx-log) ever finish?
2. Iterator doesn't really allow for IO exceptions - could consider exposing https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.html (a key class for Java9's Reactive Streams) - would require moving to Java11

jsulmont:
This PR addresses issue #1834.

* the third parameter of `xtdb.api/open-tx-log` can now be a map, which will be checked for the following keys:
  `[:with-ops? :kafka/poll-wait-duration]`

When `:kafka/poll-wait-duration` is not specified `:xtdb.api/open-tx-log` uses `poll-wait-duration` specified in `[:node-opts :xtdb/tx-log]` (defaulting to 1 second).

Please note that this addition is specific only to the Kafka module. All other storages can keep using a boolean as 3rd argument for `xtdb.api/open-tx-log`.

### DIFF
--- a/core/src/xtdb/db.clj
+++ b/core/src/xtdb/db.clj
@@ -61,7 +61,7 @@
 
 (defprotocol TxLog
   (submit-tx [this tx-events] [this tx-events opts])
-  (open-tx-log ^xtdb.api.ICursor [this after-tx-id]
+  (open-tx-log ^xtdb.api.ICursor [this after-tx-id opts]
    "Returns a serialized iterator (cursor) over transactions (see spec ::xtdb.api/tx) in the log after the given tx id.
 
    The cursor:
--- a/core/src/xtdb/kv/tx_log.clj
+++ b/core/src/xtdb/kv/tx_log.clj
@@ -94,7 +94,7 @@
   (latest-submitted-tx [_]
     (latest-submitted-tx kv-store))
 
-  (open-tx-log [this after-tx-id]
+  (open-tx-log [this after-tx-id _]
     (let [batch-size 100]
       (letfn [(tx-log [after-tx-id]
                 (lazy-seq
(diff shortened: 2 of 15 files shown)


core/src/xtdb/db.clj
```diff
@@ -61,7 +61,7 @@
 
 (defprotocol TxLog
   (submit-tx [this tx-events] [this tx-events opts])
-  (open-tx-log ^xtdb.api.ICursor [this after-tx-id]
+  (open-tx-log ^xtdb.api.ICursor [this after-tx-id opts]
```
wotbrew, line 64: Should we consider this breakage? Is TxLog our own internal SPI not intended for extension outside of XT? 

Any user tx log impls will be broken by an arity change here. One could have overloaded the existing arity, made the second arg after-tx-id _or_ a options map.

@jarohen what do you think?

jarohen (approved):
cheers @jsulmont! :pray:

core/src/xtdb/db.clj
jarohen: (as discussed on Slack) yep, happy with the breaking change here, let's go for 1.23.0 instead of 1.22.2 and pop a note in the release notes.

PR #1889, opened 2023-01-25
XTDB-1889 S3 transfer manager

jsulmont:
* added support for S3 transfer manager.

### DIFF
--- a/docs/administration/modules/ROOT/pages/checkpointing.adoc
+++ b/docs/administration/modules/ROOT/pages/checkpointing.adoc
@@ -89,11 +89,11 @@ EDN::
 
 Most Cloud providers offer some high-performance block storage facility (e.g.
 link:https://aws.amazon.com/ebs/[AWS EBS] or
-link:https://cloud.google.com/storage[Google Cloud Storage]. These volumes
+link:https://cloud.google.com/storage[Google Cloud Storage]). These volumes
 usually can be mounted on compute platforms (e.g. EC2) or containers (e.g. K8S).
 Often the Cloud provider also offers tooling to quickly snapshot those volumes
-(see link:https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSSnapshots.html[EBS snapshots]. A possible checkpoint strategy for XTDB nodes running in a cloud
-environment with such a facility, would consist in:
+(see link:https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSSnapshots.html[EBS snapshots]). A possible checkpoint strategy for XTDB nodes running in a cloud environment with such a facility, would consist in:
+
 * having `[:checkpointer :store :path]` pointing to a filesystem mounted on such
   a volume
 * having the joining XTDB node `[:checkpointer :store :path]` point to a
--- a/docs/administration/modules/ROOT/pages/checkpointing.adoc
+++ b/docs/administration/modules/ROOT/pages/checkpointing.adoc
@@ -87,14 +87,9 @@ EDN::
 
 === Note about using `FileSystem` Checkpoint Store
 
-Most Cloud providers offer some high-performance block storage facility (e.g.
-link:https://aws.amazon.com/ebs/[AWS EBS] or
-link:https://cloud.google.com/storage[Google Cloud Storage]). These volumes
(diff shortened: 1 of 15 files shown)


docs/administration/modules/ROOT/pages/checkpointing.adoc
```diff
@@ -84,3 +84,17 @@ EDN::
 
 == `FileSystem` Checkpoint Store parameters
 * `path` (required, string/`File`/`Path`/`URI`): path to store checkpoints.
+
+=== Note about using `FileSystem` Checkpoint Store
+
+Most Cloud providers offer some high-performance block storage facility (e.g.
```
jarohen: [adoc convention](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line) is one sentence per line rather than wrapping - I tend to then use `visual-line-mode` in emacs to soft-wrap them again.

modules/s3/src/xtdb/s3/checkpoint.clj
```diff
@@ -86,10 +149,14 @@
                                        :doc "S3 bucket"}
                               :prefix {:required? false,
                                        :spec ::s3/prefix
-                                       :doc "S3 prefix"}}}
-  [{:keys [^S3Configurator configurator bucket prefix]}]
+                                       :doc "S3 prefix"}
+                              :transfer-manager? {:required? false,
```
jarohen: one to discuss: is it worth us having two modules for this, rather than switching on this flag in a number of places?

e.g. so that a user would configure it as
```clojure
:checkpointer
{:xtdb/module `xtdb.checkpoint/->checkpointer
 :store {:xtdb/module `xtdb.s3.checkpoint/->xfmgr-cp-store
         :bucket "checkpoint-bucket"
         :prefix "checkpoint-dir"
         :configurator `s3-configurator}
 :approx-frequency (java.time.Duration/ofSeconds 30)}
```

modules/rocksdb/src/xtdb/rocksdb.clj
```diff
@@ -393,3 +393,10 @@
                                 :->column-family-handle ->column-family-handle})]
     (cond-> kv-store
       checkpointer (assoc :cp-job (cp/start checkpointer kv-store {::cp/cp-format cp-format})))))
+
```
jarohen: unrelated?

modules/s3/src/xtdb/s3.clj
```diff
@@ -57,7 +59,7 @@
                           (when-let [resp (.get fut)]
                             [path resp]))))))
 
-(defn ^:no-doc list-objects [{:keys [^S3Configurator configurator ^S3AsyncClient client bucket prefix]}
+(defn ^:no-doc list-objects [{:keys [^S3Configurator _configurator ^S3AsyncClient client bucket prefix]}
```
jarohen: nit: can just kill it if it's not used

docs/storage/modules/ROOT/pages/aws-s3.adoc
```diff
@@ -121,3 +124,51 @@ Clojure::
...
+link:https://aws.amazon.com/blogs/developer/introducing-crt-based-s3-client-and-the-s3-transfer-manager-in-the-aws-sdk-for-java-2-x/[CRT-based S3 client] in order to gain its full benefit.
+The following snippet
+
+[tabs]
+====
+Clojure::
++
+[source,clojure]
+----
+(defn- s3-configurator [_]
```
jarohen, line 179: can we give users a utility for this? maybe (see below comment) if we make a different module for the xfer-mgr we can give it a different default for S3 configurator too

jarohen (commented):
Cheers @jsulmont, looking good :pray: some minor feedback below, let's catch up

docs/storage/modules/ROOT/pages/aws-s3.adoc
jsulmont: Ok --  do you reckon we should use the CRT S3 Async client by default? That would be the only sensible configuration.  But we would  then have to embark the CRT jar in project.clj (module/s3/project.clj).
Not sure it's the right thing to do. The doc gives an example of how to use `S3Configurator` to configure a CRT Client with options.
I'd think someone having to deal with this would refer to the AWS CRT S3 Async Client documentation. No?

PR #1893, opened 2023-01-31
Jan/checkpoint cleanup

### LINKED ISSUE #1873: Ensuring that we don't leave incomplete checkpoints on disk
* Both upload and restore
* Both when a checkpoint errors, or if the node is closed during the proces

jsulmont:
fixes #1873 

* will cleanup any leftover on the CP storage (S3 FileSystem) after a failed checkpoint upload.

* will cleanup `[:index-store :kv-store :db-dir]`  when the download of a checkpoint fails (S3 & Filesystem).

### DIFF
--- a/core/src/xtdb/checkpoint.clj
+++ b/core/src/xtdb/checkpoint.clj
@@ -50,7 +50,6 @@
               (->> pr-str (log/info "Uploaded checkpoint:")))
             (swap! success inc)
             (finally
-              (prn [:DIR (type dir)])
               (xio/delete-dir dir)
               (when-not (pos? @success)
                 (cleanup-checkpoint store opts)
--- a/core/src/xtdb/checkpoint.clj
+++ b/core/src/xtdb/checkpoint.clj
@@ -153,19 +153,23 @@
         cp)))
 
   (download-checkpoint [_ {::keys [cp-path]} dir]
-    (let [to-path (.toPath ^File dir)]
+    (let [to-path (.toPath ^File dir)
+          success (atom 0)]
       (when-not (or (not (Files/exists to-path (make-array LinkOption 0)))
                     (empty? (Files/list to-path)))
         (throw (IllegalArgumentException. "non-empty checkpoint restore dir: " to-path)))
 
       (try
         (sync-path cp-path to-path)
+        (swap! success inc)
         (catch Exception e
-          (xio/delete-dir to-path)
           (throw (ex-info "incomplete checkpoint restore"
                           {:cp-path cp-path
                            :local-dir to-path}
-                          e))))))
+                          e)))
+        (finally
+          (when-not (pos? @success)
+            (xio/delete-dir dir))))))
 
   (upload-checkpoint [_ dir {:keys [tx cp-at ::cp-format]}]
     (let [from-path (.toPath ^File dir)
--- a/core/src/xtdb/checkpoint.clj
+++ b/core/src/xtdb/checkpoint.clj
@@ -77,7 +77,6 @@
           ses (Executors/newSingleThreadScheduledExecutor (xio/thread-factory "xtdb-checkpoint"))]
       (letfn [(run [[time & more-times]]
                 (try
-
                   (checkpoint {:approx-frequency approx-frequency,
                                :dir checkpoint-dir,
                                :src src,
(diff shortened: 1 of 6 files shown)


dev/dev.clj
```diff
@@ -106,8 +134,9 @@
                                   :kafka-config :kafka-config}}}})
 
 
-;; swap for `embedded-kafka-config` to use embedded-kafka
-(ir/set-prep! (fn [] standalone-config))
+;; swap for `embedded-kafka-config`  to use embedded-kafka
+(ir/set-prep! (fn [] checkpoint-fs-config))
```
jarohen: fine that you've popped an example config in here, thanks - but could we preserve the `standalone-config` as the default?

core/src/xtdb/checkpoint.clj
```diff
@@ -142,24 +152,29 @@
...
-    (let [to-path (.toPath ^File dir)]
+    (let [to-path (.toPath ^File dir)
+          success (atom 0)]
       (when-not (or (not (Files/exists to-path (make-array LinkOption 0)))
                     (empty? (Files/list to-path)))
         (throw (IllegalArgumentException. "non-empty checkpoint restore dir: " to-path)))
 
       (try
         (sync-path cp-path to-path)
+        (swap! success inc)
```
jarohen: I don't think you need any state in here - IIUC you're essentially doing this?

```clojure
(try
  (upload-checkpoint!)
  (catch Throwable t
    (cleanup-checkpoint!)
    (throw (ex-info "incomplete checkpoint" {...} t))))
```

jarohen: subjective style note (feel free to keep it as it is, it works perfectly fine) but I'd probably have the cleanup included in the contract of each `upload-checkpoint` rather than introduce a separate protocol function?

It's quite a common pattern for us to have various functions shaped like the following, so that most functions have an implicit contract that, if they go wrong, the caller can trust that they don't leave anything behind:

```clojure
(try
  ;; do the thing
  (catch Throwable t
    ;; clean up
    (throw t)))
```

jarohen: maybe different on the download case because the caller 'owns' the temporary directory, and is instructing the cp-store to use it - i.e. the cleanup's the same for all of the implementations

test/src/xtdb/fixtures/checkpoint_store.clj
```diff
@@ -3,7 +3,22 @@
             [xtdb.fixtures :as fix]
             [xtdb.checkpoint :as cp]
             [clojure.java.io :as io]
-            [clojure.test :as t]))
+            [clojure.test :as t])
+  (:import  [java.nio.file NoSuchFileException CopyOption Files FileVisitOption LinkOption Path]
+            java.nio.file.attribute.FileAttribute))
+
+(defn- sync-path-throw
```
jarohen: duplicate fn?

test/src/xtdb/fixtures/checkpoint_store.clj
```diff
@@ -52,3 +67,41 @@
...
+          :let [to-path (.resolve to-root-path (str (.relativize from-root-path from-path)))]]
+    (cond
+      (Files/isDirectory from-path (make-array LinkOption 0))
+      (Files/createDirectories to-path (make-array FileAttribute 0))
+
+      (Files/isRegularFile from-path (make-array LinkOption 0))
+      (Files/copy from-path to-path ^"[Ljava.nio.file.CopyOption;" (make-array CopyOption 0))))
+  (throw (Exception. "broken!")))
+
+(defn test-checkpoint-broken-store-failed-download
```
jarohen, line 71: `deftest`?

jarohen (changes requested):
Cheers @jsulmont! :pray: few minor points to address but otherwise all good :smile:

test/src/xtdb/fixtures/checkpoint_store.clj
jsulmont: well the test is defined in `xtdb/checkpoint-test` (test-fs-checkpoint-store-failed-download).
Just being consistent with an existing test defined above that does just that.

core/src/xtdb/checkpoint.clj
jsulmont: When I added  `cleanup-checkpoint` to the protocol, I had in mind to indeed enrich the protocol, e.g., having also a `valid-checkpoint?` method (and perhaps other), to allow us to implement some checkpoint management procedure which would operate in an abstract manner (i.e., regardless the kind of store it is using).

PR #1895, opened 2023-02-08
XTDB HTTP Health Check module

### LINKED ISSUE #1867: Nodes are getting killed if checkpoint restore takes too long
We currently wait for the checkpoint to be downloaded and restored before the node starts up, but for large checkpoints this means that we don't respond promptly to healthchecks, causing the node to be killed, another node started, sadness loop.

jsulmont:
* fix #1867 

This PR implements an optional `http-health-check` module that brings up a mini HTTP server servicing basic routes (TBD). 

The goal  is for this health-check to be usable by e.g., K8S, ECS, ...

The `http-health-check` module starts early on (precisely before `[:xtdb/index-store :kv-store]`), so that it can track potentially long-lasting conditions (e.g., index recovering from checkpoint) happening before the node created.


docs/extensions/modules/ROOT/pages/health-server.adoc
```diff
@@ -0,0 +1,105 @@
...
+{:event "end-restore",
+ :thread "nREPL-session-8139ad0e-476e-455e-9579-810d6d4e5f06",
+ :clock 10.406734083,
+ :event-type "healthz",
+ :timestamp "2023-02-10T06:16:33.190Z",
+ :namespace "xtdb.checkpoint"}
+----
+The `:clock` attribute is a number of seconds (accurate to the nanosecond) and reasonably monotonic.  Internally the various XTDB components communicate via a lightweight bus.
+
+// TODO finish this write-up once I've sorted the event question.
```
jarohen, line 24: still applicable?

modules/http-health-check/README.adoc
```diff
@@ -0,0 +1,21 @@
+= xtdb-http-server
+
+This module offers an HTTP server that allows you to send transactions and run queries via a REST API.
+
+.deps.ednxxx
```
jarohen: xxx?

modules/http-health-check/project.clj
```diff
@@ -0,0 +1,24 @@
...
+                             :pedantic? :global-vars
+                             :license :url :pom-addition]}
+
+  :scm {:dir "../.."}
+
+  :dependencies [[org.clojure/clojure]
+                 [com.xtdb/xtdb-core]
+                 [ring/ring-core "1.9.6"]
+                 [ring/ring-jetty-adapter "1.9.6"]
+                 [compojure "1.7.0"]]
```
jarohen: We use the reitit stack in the main HTTP server - any reason to differ here?

Also, we're hosted on Maven Central so we can only use other deps already on Central (i.e. not on Clojars) - we have mirrored most of the HTTP stack already which you can likely reuse here - see the http-server project.clj.

resources/logback-test.xml
```diff
@@ -4,7 +4,7 @@
       <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
     </encoder>
   </appender>
-  <root level="WARN">
+  <root level="INFO">
```
jarohen: just to check - was this intentional?

core/src/xtdb/bus.clj
```diff
@@ -36,6 +36,15 @@
 (s/def ::event (s/merge (s/keys :req [::xt/event-type])
                         (s/multi-spec event-spec ::xt/event-type)))
 
+(defmacro ->event
+  [event-type msg]
+  `{::xt/event-type ~event-type
+    :namespace ~(str *ns*)
```
jarohen, line 42: do we necessarily want to couple these events to the actual XT namespaces? it makes what's essentially an impl detail part of the public API.

although maybe they already are, because of the module keys in the config. :thinking:

modules/http-health-check/src/xtdb/http_health_check.clj
```diff
@@ -0,0 +1,56 @@
...
+
+(defrecord HTTPServer [^Server server ^Closeable listener events options]
+  Closeable
+  (close [_]
+    (.stop server)
+    (.close listener)))
+
+(def ^:const default-server-port 7000)
+
+(defonce base-time (System/nanoTime))
```
jarohen, line 24: should this be part of the node startup, rather than `defonce`? the node's lifecycle isn't necessarily tied to the JVM (although most people probably use it that way) - it's technically a component. we have had questions in the past related to multi-tenanted XT nodes - users that manage multiple XT nodes that come and go within one JVM process, for data isolation purposes.

jarohen: also, do we need a separate timestamp and nano-time? I understand you probably want to use nano-time for elapsed time since the start of the node, but maybe an Instant (with microseconds, in practice) has enough granularity for both?

modules/http-health-check/README.adoc
```diff
@@ -0,0 +1,21 @@
+= xtdb-http-server
```
jarohen: this file looks like it's meant to be for the health check, but is a straight copy of the HTTP one?

jarohen (changes requested):
cheers @jsulmont - looking good, minor feedback as always :smile: aware you said you're still working on the docs so I haven't reviewed those in detail, but they're looking promising.

modules/http-health-check/src/xtdb/http_health_check.clj
jsulmont: It doesn't really matter when this base-time is initialised as long as it's only initialised once per XT (works fine when multiple XT nodes share the same JVM). Indeed that `base-time` is just used as a ... base time against which subsequent calls to `System/nanoTime` will be related. 
The `:clock` attribute, is (hopefully) monotonic -- I find a double easier to look at, than a timestamp.
Note that the `event` map already has a `:timestamp` attribute, which receives whatever `java.util.Date.` will return (that is, accurate to the millisecond). 
This is entirely subjective and I am happy to get rid of it altogether -- please let me know.
PS note that Criterium  is using `System/nanoTime`.

core/src/xtdb/bus.clj
jsulmont: I thought that would make the life of users easier. The HTTP serves a `/healthz/ns/:ns` route that allows to fetch events of the said namespace. Also a GET on `/healthz` simply returns the `last` event of each namespace.
XT namespaces encapsulate related parts (e.g., `node` `checkpoint` `tx` ...) and I thought it'd be nice for a user to be able to inspect them. A good example, is the checkpointing, which runs its own thread -- yet is able to seriously impact the overall function of the node. 

Also this macro allows for categories of events, likewise a topic, to which the health-check can subscribe

resources/logback-test.xml
jsulmont: Nope sorry!

modules/http-health-check/project.clj
jsulmont: No reason, other than I've never used the reitit stack and found compojure way.

Also the dependencies embarked by the http-server are *heavy*, most of them totally useless for this tiny web server; 

Happy to move to reitit let me know. 🤷🏻

core/src/xtdb/bus.clj
jarohen: Yep - I agree with the categorisation, for the reasons you say, it was more about enforcing the coupling of categories to the impl namespaces. we could maybe use the namespace of the event-type keyword instead?

(just ideas/speculation, would be fine to merge as-is)

modules/http-health-check/src/xtdb/http_health_check.clj
jarohen: > I find a double easier to look at, than a timestamp.
yep, I understand - my comment above was suggesting that you could do this with `(- now base-instant)` (or, rather, the appropriate java.time method)

not a big deal - again, fine to merge as-is

core/src/xtdb/bus.clj
jsulmont: @jarohen I thought about it (in fact, I am doing that for internal events (i.e,  events where ::xt/event-type is not :healthz)
However for it to work correctly  that would require the event to be a namespaced kw  and I thought that would be too much of a restriction.

modules/http-health-check/project.clj
jarohen: > No reason, other than I've never used the reitit stack and found compojure way.

cool - in that case, yeah, let's go with consistency

> Also the dependencies embarked by the http-server are heavy, most of them totally useless for this tiny web server

this is a fair point, and good to consider dependency weight - but if the user's using an XT-provided Docker image it'll already contain most of them, and chances are the startup time introduced by these deps will be pretty insignificant compared to restoring a checkpoint

jsulmont: Thank you! I promised I'll get into reitit 👍🏻

jarohen (approved):
one cheeky `prn`, otherwise good to go - cheers @jsulmont! :pray:

PR #1898, opened 2023-02-10
Remove unneeded where clause to optimise query

### LINKED ISSUE #1896: `latest-submitted-tx` should avoid scanning the MariaDB tx-log
There has a been a report of running `sync` with MariaDB (via `xtdb-jdbc`) taking much longer than necessary because the lookup isn't using SQL indexes correctly (and the index-store is otherwise caught up). This requires further investigation.

zackteo:
Remove unneeded where clause to optimise query #1896

### DIFF
--- a/modules/jdbc/src/xtdb/jdbc.clj
+++ b/modules/jdbc/src/xtdb/jdbc.clj
@@ -183,7 +183,7 @@
     (tx-sub/handle-polling-subscription this after-tx-id {:poll-sleep-duration (Duration/ofMillis 100)} f))
 
   (latest-submitted-tx [_]
-    (when-let [max-offset (-> (jdbc/execute-one! pool ["SELECT max(EVENT_OFFSET) AS max_offset FROM tx_events WHERE topic = 'txs'"]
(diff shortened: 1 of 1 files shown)


jarohen:
Hey @zackteo, cheers for the PR 🙏

I'll defer to @wotbrew as he's in context - but I wonder if this works if you're currently inserting docs - maybe you'll get a document row as your latest-submitted-tx?

(ideally we'd have transactions and docs in two separate tables, obviously, but unfortunately the table structure existed before we made the document store split, and we chose to not impose a golden store migration on users at the time)

Thanks,

James

zackteo:
Hey @jarohen, cheers :)

Oops, you are right! This PR makes the assumption that `docs` come in before `txs` so it is more of a hack. I think I best defer this back to @wotbrew and the rest of the XTDB team for a better solution for this.

Shall we close this PR then?

Thanks!
Zachary

wotbrew:
Thanks for the PR @zackteo 🙏, but yes I'm working on a separate fix for #1896 - closing for now.

PR #1902, opened 2023-02-15
RFC: :xt/fn validation

wotbrew:
RFC (tests will fail until I confirm this has merit, they can be fixed but this is a breaking change)

At looking at solving #1831 I was surprised to find we do not validate `:xt/fn` so a user can supply any form, quoted function or not. The supplied form will be eval'd later as part of ingest where the user might discover an error.

i.e `[[::xt/put {:xt/id :boom, :xt/fn '(println "boom")}]]` is a valid transaction, and if you later `[[:xt/fn :boom]]` you will see the print on eval.

This PR then provides immediate feedback on submit in the following cases:

- invalid fn form, i.e a clojure function object, non-list and so on. I use the `:xtdb.api/fn` spec, which I've extended to support destructuring bindings.
- error on compilation such as symbol resolution (this is achieved by running a speculative eval and catching any exceptions).

Security wise the early eval opens a vector where you no longer need to collide with an `:xt/id` that will later be called for RCE. I would like to have that conversation, I think perhaps the early eval could be a dev-only flag. The spec _should_ ensure only fn forms of course.

Possible breaking change if users are doing funky stuff like using combinators (comp) (constantly) etc, or running code as part of definition (let over lambda).

One idea to avoid breakage and (some) security concerns would be to add a specialised op for definition so it does not overload put.

### DIFF
--- a/core/src/xtdb/api.clj
+++ b/core/src/xtdb/api.clj
@@ -23,9 +23,12 @@
 (s/def ::evicted? boolean?)
 (s/def :crux.db.fn/args (s/coll-of any? :kind vector?))
 
+(s/def ::fn-db-arg (s/or :sym symbol? :destructuring map?))
+(s/def ::fn-arg (s/or :sym symbol? :destructuring (s/or :map map? :vec vector?)))
+
 (s/def ::fn
   (s/cat :fn #{'fn}
-         :args (s/coll-of symbol? :kind vector? :min-count 1)
+         :args (s/and vector? (s/cat :db ::fn-db-arg :args (s/* ::fn-arg)))
          :body (s/* any?)))
 
 (defprotocol PXtdb
--- a/core/src/xtdb/tx/conform.clj
+++ b/core/src/xtdb/tx/conform.clj
@@ -1,5 +1,6 @@
 (ns ^:no-doc xtdb.tx.conform
-  (:require [xtdb.api :as xt]
+  (:require [clojure.spec.alpha :as s]
+            [xtdb.api :as xt]
             [xtdb.codec :as c]
             [xtdb.db :as db]
             [xtdb.error :as err]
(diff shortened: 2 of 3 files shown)


jarohen:
Hey @wotbrew - I like the idea, in theory :slightly_smiling_face: but yeah, the only way to reliably check it without false positives is to eval it, and we don't particularly want to do that for the reasons you've suggested. hmm :thinking: 

(thinking out loud) 
* I'm guessing this has come about because it's not straightforward to see why a tx-fn's failed on the node itself? maybe there's ways we can fix that problem instead?
* We could offer this validation as an opt-in util function? Downside being that users may not know it exists...
* Otherwise, does `clojure.core/fn` have a fn-spec already that we could re-use?

refset:
Related https://github.com/xtdb/xtdb/issues/1672

djtango:
@jarohen 

>Otherwise, does clojure.core/fn have a fn-spec already that we could re-use?

Here is a fragment of some work I incidentally have done relating to this:

```clj
(ns validate-clojure
  (:require [clojure.core.specs.alpha :as s.clj]
            [clojure.spec.alpha :as s]
            [nyaya.transformation-create.clojure.input-expression :as-alias in-exp]))

(s/def ::fn-definition (:args (s/get-spec 'clojure.core/fn)))

;; NOTE for some reason it's very difficult to write one single s/cat spec for
;; the (fn ...) form without also using separate functions which loses spec
;; details so split it up into ::fn-structure and ::fn-definition
(s/def ::fn-structure (s/cat :fn-sym #{'fn 'clojure.core/fn}
                             :body (s/+ any?)))

(defn eval-error? [exception]
  (= "EvalReader not allowed when *read-eval* is false."
     (-> exception Throwable->map :cause)))

(defn- read-clojure [s]
  (try
    (binding [*read-eval* false] ;; disable eval via #=(...)
      (read-string s))
    (catch Exception e
      (cond (eval-error? e)
            (throw (ex-info "Eval is disabled" {:error e}))
            :else
            (throw (ex-info "Invalid clojure expression supplied" {:error e}))))))

(defn valid-fn? [fn-str]
  (let [fn-statement (read-clojure fn-str)
        fn-structure-failures (s/explain-data ::fn-structure fn-statement)
        fn-definition-failures (s/explain-data ::fn-definition (rest fn-statement))]
    (cond (and (empty? fn-structure-failures)
               (empty? fn-definition-failures))
      true
      (some? fn-structure-failures)
      (throw (ex-info "invalid-clojure-not-a-fn"
                      {:failures fn-structure-failures
                       :fn-str fn-str}))
      (some? fn-definition-failures)
      (throw (ex-info "invalid-clojure-bad-fn-definition"
                      {:failures fn-definition-failures
                       :fn-str fn-str})))))
```

jarohen:
let's reopen when we come back to it

PR #1903, opened 2023-02-15
->xtdb-event-sink

### LINKED ISSUE #1901: implement `->xtdb-handler` in the `health-check-module`
* Often XTDB is embedded in an application implementing a web-server, itself running in a Docker container managed in e.g. ECS or K8S (e.g.,  [Site](https://github.com/juxt/site) ). 
* These applications often implement their own health-check but currently have no way to detect some XTDB internal conditions (for example a long lasting index recovery from checkpoint etc.)
* `->xtdb-handler` would allow such applications to leverage XTDB health-check into their own handler,

jsulmont:
addresses #1901


jarohen:
closing for now, let's re-open if we return to it later

PR #1908, opened 2023-03-02
Improved error policy in the face of document store failure during transaction processing

wotbrew:
## Situation

Storage errors (e.g unavailability) will cause the ingester to fail during fetch-docs. If we encounter an error fetching outside of a transaction function query or entity call - we will see the ingester panic and stop processing. If storage fails during transaction function processing, the tx fn is skipped, leading to non-deterministic outcomes for the database in a clustered setting (concurrent execution). 

Examples of errors:

- Connection pool saturated JDBC
- Servers down / outage
- Networking DNS, Firewall etc (Kafka,JDBC,S3)
- Too many requests / slow down (quota enforcement)

As errors relating to storage unavailability are likely to be among the most likely in production, this makes XTDB quite brittle to flakey network conditions, limits or outages of storage.

## Solution

Goals:

- Provide automatic recovery from IO exceptions
- Limit drift possibility with non-deterministic transaction function failures

For exceptions thrown by the doc store, implement a exponential back off & retry policy, so that XTDB automatically recovers from storage errors during transaction processing. This avoids an ingester panic in many cases where the node will simply recover in a short amount of time.

The starting policy will be to retry on all java.lang.Exception unless they contain a non retryable [`:cognitect.anomalies/category`](https://github.com/cognitect-labs/anomalies) in their ex-data.

The same back-off behaviour should be used for doc store errors thrown during transaction functions (rather than skipping), to avoid inconsistency if nodes encounter errors while processing the nodes concurrently. Transaction functions will still be skipped if a non-document-store exception is raised.

The user will be notified of failed retries in the logs. This may later extend to bus messages.

## Additional bug fixes

- Interrupting a transaction function now aborts the ingester rather than skipping the transaction, this was necessary to do as part of this PR to avoid blocking node `.close` if the ingester is in an exp-backoff loop.


wotbrew:
Working pain point is around correct failure behaviour for unclassified storage failures during transaction functions. The solution proposed will cause affected transactions to loop forever. This might demand tooling to help users overcome the implications in production. It may not be an actual problem because all failures will be callee/unavailability type problems where the solution is always wait, do some admin to $doc-store or reconfigure and restart the node all together.

Alternatives have their own problems, skipping bad due to node drift if it is a recoverable failure, panic bad due to outage if it is recoverable

wotbrew:
On reflection I do not think it is a problem

> It may not be an actual problem because all failures will be callee/unavailability type problems where the solution is always wait

wotbrew:
I need to check the implications of strict fetch in transaction functions - I don't think it is right, as missing is a possibility in that context. This brings up questions of whether it is possible to protect against inconsistency due to missing docs

PR #2482, opened 2023-04-23
Document query timeout default

daveliepmann:
It's nice for the query docs to mention the default, and how the default can be configured.


daveliepmann:
I am not able to test right now whether that second commit's link syntax is correct

wotbrew:
Committed as 7d4bdf02acc7f6187a13474e9992e524c7fb38d3.

I'll deploy the docs with the next release (happening shortly)

Thanks @daveliepmann 🙏 .

Issue #2155, opened 2022-06-30
Support for JDK17


jarohen:
Requires some [JDK module shenanigans](https://arrow.apache.org/docs/java/install.html) for Arrow, as well as probably a few other fixes.

We currently support JDK11 as a minimum, so let's ensure we either continue to work on JDK11, or explicitly decide to bump the min version to 17.

(I'm not 100% sure all the JVM flags required to make 17 happy existed in 11 - it might need different startup params/`clj` aliases depending on the version)

Issue #2154, opened 2022-06-29
Evaluate `util/set-value-count` workaround


jarohen:
We have a `util/set-value-count` workaround to bypass Arrow's `setValueCount` - however, this likely means that we aren't [idiomatically](https://arrow.apache.org/docs/java/vector.html#vector-life-cycle) using `setValueCount` - i.e. a single call once you've finished writing to the vector.

Composite Arrow vectors then propagate this call down to their child vectors, and, further up, the VectorSchemaRoot propagates its `setRowCount` to its own children.

Its reason for existence is because (unsurprisingly) DUV. DUV's `setValueCount` scans through its type-buffer in order to calculate the value-count for each of its child vectors. Given that we call `setValueCount` at the end of every ingest transaction, linearly scanning through the whole DUV every time meant an O(n^2) operation.

Thing is, because we're essentially stopping the propagation of the set-value-count call when we hit a DUV, this itself introduces a number of required workarounds, and no doubt as-yet-undiscovered bugs. e.g. we have to ensure that we're setting value counts on each individual vector as we go, rather than one call to the top-level at the end. 

I wonder whether it's possible to get DUV to keep a running set of value-counts for each of its child vectors whenever a type-id is set - then, all it would need to do on the call to `setValueCount` is propagate these counts on. Obviously, this is then cached data, so needs to be updated wherever it's possible to update a type-id (and has the usual cached-data risk of going stale).

jarohen:
Fixed the issue in DUV, so we can start to remove these workarounds when we see them.

jarohen:
been removed throughout now

PR #2488, opened 2023-04-24
Fix null src-vector's for struct and list writers. 

### LINKED ISSUE #2393: CCE with structs in outer-joins
The problem is probably symmetric with right-outer-join and likely more fundamental then just outer-joins. There seems to be a problem of casting the nil vectors to certain other arrow type vectors
```clj
(t/deftest test-left-outer-with-structs
  (t/is (= nil #_{:res [{{:a 12, :b 12, :c 2} 1, {:a 12, :b 12, :c 0} 1, {:a 0, :b nil, :c nil} 1}
                        {{:a 12, :b 12, :c 2} 1, {:a 100, :b 100, :c 3} 1, {:a 12, :b 12, :c 0} 1}]
                  :col-types '{a :i64, b [:union #{:null :i64}], c [:union #{:null :i64}]}}
           (-> (tu/query-ra [:left-outer-join '[{a b}]
                             [::tu/blocks
                              [[{:a 12}, {:a 0}]
                               [{:a 12}, {:a 100}]]]
                             [::tu/blocks
                              [[{:b 12, :c {:foo 1}}, {:b 2, :c {:foo 1}}]
                               [{:b 12, :c {:foo 2}}, {:b 100, :c {:foo 2}}]]]]
                            {:preserve-blocks? true, :with-col-types? true})
               (update :res (partial mapv frequencies))))))
```

kind of blocking xtdb/xtdb#2352

FiV0:
Fixes #2393

### DIFF
--- a/core/src/main/clojure/xtdb/vector/writer.clj
+++ b/core/src/main/clojure/xtdb/vector/writer.clj
@@ -60,7 +60,7 @@
       (.setTypeId parent-duv parent-pos type-id)
       (.setOffset parent-duv parent-pos pos)
 
-      ;; TODO (#252) do we still need this here?
+      ;; TODO (#2155) do we still need this here?
       (.setValueCount parent-duv (inc parent-pos))
 
       (.startValue inner-writer)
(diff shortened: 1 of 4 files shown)


core/src/main/clojure/xtdb/vector/writer.clj
```diff
@@ -60,7 +60,7 @@
       (.setTypeId parent-duv parent-pos type-id)
       (.setOffset parent-duv parent-pos pos)
 
-      ;; TODO (#252) do we still need this here?
+      ;; TODO (#2155) do we still need this here?
```
FiV0, line 63: I didn't quite understand how this relates to #2155.

core/src/main/clojure/xtdb/vector/writer.clj
```diff
@@ -60,7 +60,7 @@
...
       (.setValueCount parent-duv (inc parent-pos))
```
FiV0, line 64: I did not quite understand how this relates to #2155.

jarohen: #2154?

jarohen (approved):
nice one @FiV0 :clap:

PR #2538, opened 2023-05-10
Add Azure Blob Storage object store #2298

### LINKED ISSUE #2298: Support Azure Blob Storage
Want to handle a few things under this ticket:
- Implementing Azure Blob Storage as an objectstore. 
  - Implementing the ObjectStore interface itself.
  - Handling any configuration related to authentication to the blob storage containers.
  - Implementing any integrant related config to be able to use it as a component.
  - Adding any relevant tests against the BlobObjectStore.
- Creating an XTDB Azure account.
- Setting up a test/benchmark container for Azure Blobs
- Setting up any relevant access control to the containers such that we can test against them.
- Setting up any containerization to run a TPCH benchmark against azure blob storage on Azure itself.

danmason:
Resolves #2298

Implements a new object store using Azure Blob Storage, alongside:
- All of the related integrant config to add it to a node.
- Including the module within various places - namely the xtdb standalone deps and dev dependencies. 
- Documentation in the form of a README in the module.
- A test file making use of the object-store tests.
   - Expects ENV vars to be present - these have been added to the CI.  
- A dev file for locally playing around with it under `dev/azure-testing`

## TODO

- At a later point - want to benchmark this/containerize this for Azure, when the current benchmarking story has been better fleshed out.

## Local Benchmarking Results
Ran a benchmark against TPCH with a 0.05 scale-factor. This was done by (temporarily) adding a new node type to the "bench.clj" (local-azure) that used local-directory-log and azure/blob-object-store. Can see some of the files/folders/chunks appear on the Azure blob storage container, and got back some timings (keeping in mind it's being ran locally on an xps15 with 64GB RAM):

```
'submit-docs', 35137.653094ms
'sync', 278904.935517ms
'finish-chunk', 27307.571926ms
('ingest', 341356.655297ms)

'cold-queries', 51346.988333ms
'hot-queries', 8876.262816ms
```

### DIFF
--- a/modules/azure/README.adoc
+++ b/modules/azure/README.adoc
@@ -28,7 +28,7 @@ Authentication for the components in the module is done via the https://learn.mi
 
 ## Azure Blob Storage Object Store
 
-We can swap out the implementation of the objedct store with one based on Azure Blob Storage. To do so, we add the `xtdb.azure/blob-object-store` key within the integrant config for our node, alongside any config:
+We can swap out the implementation of the object store with one based on Azure Blob Storage. To do so, we add the `xtdb.azure/blob-object-store` key within the integrant config for our node, alongside any config:
 ```clojure
 {
   ...
--- a/modules/azure/README.adoc
+++ b/modules/azure/README.adoc
@@ -1,0 +1,61 @@
+# Azure Module
+
+Within our XTDB node, we can make use of Azure Services for certain purposes. Currently, we can:
+
+* Use *Azure Blob Storage* as the XTDB Object Store.
+
+### Project Dependency 
+
+In order to use any of the azure services, you will need to include a dependency on the xtdb.azure module.
+
+_deps.edn_
+```
+com.xtdb.labs/xtdb-azure {:mvn/version "2.0.0-SNAPSHOT"}
+```
+
+_pom.xml_
+```
+<dependency>
+    <groupId>com.xtdb.labs</groupId>
+    <artifactId>xtdb-azure</artifactId>
+    <version>2.0.0-SNAPSHOT</version>
+</dependency>
+```
+
+### Authentication
+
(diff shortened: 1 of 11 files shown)


modules/azure/README.adoc
```diff
@@ -0,0 +1,61 @@
...
+</dependency>
+```
+
+### Authentication
+
+Authentication for the components in the module is done via the https://learn.microsoft.com/en-us/java/api/com.azure.identity.defaultazurecredential?view=azure-java-stable[*DefaultAzureCredential*] class - you will need to setup authentication using any of the methods listed within the Azure documentation to be able to make use of the operations inside the modules.
+
+## Azure Blob Storage Object Store
+
+We can swap out the implementation of the objedct store with one based on Azure Blob Storage. To do so, we add the `xtdb.azure/blob-object-store` key within the integrant config for our node, alongside any config:
```
jarohen: thai-po

src/test/clojure/xtdb/azure_test.clj
```diff
@@ -0,0 +1,34 @@
...
+                                                :prefix (str "xtdb.azure-test." (UUID/randomUUID))}}
+                    ig/prep
+                    ig/init)]
+        (try
+          (binding [*obj-store* (::azure/blob-object-store sys)]
+            (f))
+          (finally
+            (ig/halt! sys)))))))
+
+;; TODO: Throws an error when trying to return an IllegalStateException - classcast issue?
```
jarohen: still a TODO?

jarohen (approved):

src/test/clojure/xtdb/azure_test.clj
danmason: Nope - fixed that, it was because I forgot to actually "throw" the error :sweat_smile:

PR #2540, opened 2023-05-15
typo(http-server): small correction on route description

iomonad:
Page is generated here: https://docs.xtdb.com/clients/http/openapi/1.23.1/#/paths/~1_xtdb~1db/get


jarohen:
cheers!

PR #2550, opened 2023-05-19
Change `tx-id` type to integer in http client docs

Donnype:
I think I found a typo in the http client docs suggesting the `tx-id` parameter should be a date, while it should actually be an integer.


refset:
Good spot! Thanks for the PR :pray:

PR #2558, opened 2023-06-08
Add support for select *, see ADR: 0041

### LINKED ISSUE #2086: Asterisk support, §7.12
Now that we have distinct tables and table metadata, we should be able to support:

Unqualified `*` where only a single unambiguous correlation is in scope.

`SELECT * FROM users u`

Qualified `x.*` where more than one correlation is in in scope.

`SELECT u.*, o.* FROM users u, orders o`

The latter case brings into question naming collisions, if we take the example query, where tab1 and tab1 have identical col names.

`SELECT new_tab.* FROM (SELECT tab0.*, tab1.* FROM tab0, tab1) AS new_tab`

In postgres, this returns all columns named as they are on the base tables (so duplicate col names), the implication of this is that trying to select `new_tab.col0` for example, must throw an ambiguous col ref error.

SQLite instead renames columns in left -> right order, so the columns from tab1 become `col0:1` etc.

MySQL opts to throw a `Duplicate column name` error instead.

That being said, at the top level `SELECT u.*, o.* FROM users u, orders o` both MySQL and Postgres return unrenamed duplicate cols.

TBD: What does the spec say?

mbutlerw:
Adds support for qualified and unqualified `*`

```
SELECT * FROM TRA, TRB

SELECT *.TRB FROM TRA, TRB
```

Change pushes the generation of unique column names where duplicates exist down to the projected-columns attribute of query specification, any `outer-name`s generated this way are then forwarded by table-primary projected cols as an `inner-name` in the case of subqueries etc. enabling `build-query-specification` to use the correct column names during the plan step. If we wanted to return the original column names this information could be forwarded with the outer projection and then used in combination with remove-names (probably dropping the outer-projection) to do a single clojure rename to a duplicate name supporting format.

### DIFF
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -28,10 +28,11 @@
            xtdb.api.protocols.TransactionInstant
            xtdb.buffer_pool.IBufferPool
            xtdb.ICursor
+           xtdb.live_chunk.ILiveTableWatermark
            (xtdb.metadata IMetadataManager ITableMetadata)
            xtdb.operator.IRelationSelector
            (xtdb.vector IIndirectRelation IIndirectVector)
-           xtdb.watermark.IWatermark))
+           (xtdb.watermark Watermark IWatermark IWatermarkSource)))
 
 (s/def ::table symbol?)
 
@@ -47,6 +48,7 @@
 #_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
 (definterface IScanEmitter
   (tableColNames [^xtdb.watermark.IWatermark wm, ^String table-name])
+  (allTableColNames [^xtdb.watermark.IWatermark wm])
   (scanColTypes [^xtdb.watermark.IWatermark wm, scan-cols])
   (emitScan [scan-expr scan-col-types param-types]))
 
@@ -415,6 +417,12 @@
     (.temporalRootsSource watermark)
     (util/instant->micros (:current-time basis))))
 
+(defn tables-with-cols [basis ^IWatermarkSource wm-src ^IScanEmitter scan-emitter]
+  (let [{:keys [tx, after-tx]} basis
+        wm-tx (or tx after-tx)]
+    (with-open [^Watermark wm (.openWatermark wm-src wm-tx)]
+      (.allTableColNames scan-emitter wm))))
+
 (defmethod ig/prep-key ::scan-emitter [_ opts]
   (merge opts
          {:metadata-mgr (ig/ref ::meta/metadata-manager)
@@ -430,7 +438,19 @@
                                (.columnTypes)
                                keys)])))
 
+    (allTableColNames [_ wm]
+      (merge-with
+        set/union
+        (update-vals
+          (.allColumnTypes metadata-mgr)
+          (comp set keys))
+        (update-vals
+          (some-> (.liveChunk wm)
+                  (.allColumnTypes))
+          (comp set keys))))
+
     (scanColTypes [_ wm scan-cols]
+
       (letfn [(->col-type [[table col-name]]
                 (let [normalized-table (util/str->normal-form-str (str table))
                       normalized-col-name (util/str->normal-form-str (str col-name))]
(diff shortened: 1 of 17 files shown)


core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -430,7 +438,15 @@
                                (.columnTypes)
                                keys)])))
 
+    (allTableColNames [_ wm]
```
jarohen, line 441: this looks like it returns col-types rather than col-names?

if it were to return types, it's best to merge them with `types/merge-col-types` rather than just `merge`, but tbh I'd probably go with taking keys and `set/union`ing?

core/src/main/clojure/xtdb/sql.clj
```diff
@@ -11,12 +11,18 @@
            java.util.HashMap
            org.apache.arrow.memory.BufferAllocator))
 
+(defn parse-query
+  [query]
+  (binding [r/*memo* (HashMap.)]
```
jarohen, line 16: worth binding this to `(or r/*memo* (HashMap.))` so that you don't use a different memo for parse + analyse in the compile case?

(might have to explicitly def `r/*memo*` to `nil` so that it's not `Var$Unbound`, if it's not already?)

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -1293,33 +1293,18 @@
...
-                     (fn [projection-col unique-unqualified-column-name]
-                       (assoc projection-col :unique-unqualified-column-name unique-unqualified-column-name))
-                     projection
-                     (generate-unique-column-names (map #(unqualified-projection-symbol %) projection)))
-        qualified-projection (vec (for [{:keys [qualified-column unique-unqualified-column-name] :as column} projection
-                                        :let [derived-column (:ref (meta column))]]
+        qualified-projection (vec (for [{:keys [qualified-column outer-name inner-name] :as column} projection
+                                        :let [derived-column (:ref (meta column))
+                                              outer-projection-symbol (unqualified-projection-symbol
+                                                                        (if outer-name (assoc column :identifier outer-name) column))]]
```
jarohen: nit: these could be `cond->` - `(cond-> column outer-name (assoc :identifier outer-name))`. might seem like golfing but I find it a bit more obviously intention-revealing - 'return me the column, with outer-name if there is one'

notes/adr/0041-select-star-dup-cols.adoc
```diff
@@ -0,0 +1,91 @@
+= 41. SQL SELECT * / Duplicate Columns
+
+Date: 2023-05-16
+
+== Status
+
+Proposed
```
jarohen: accepted?

core/src/main/clojure/xtdb/sql/analyze.clj
```diff
@@ -542,20 +563,23 @@
...
                                    (distinct)))))]
         [(reduce
-          (fn [acc {:keys [identifier index]}]
-            (conj acc (cond-> (with-meta {:index (count acc)} {:table table})
-                        identifier (assoc :identifier identifier)
-                        (and index (nil? identifier)) (assoc :original-index index)
-                        correlation-name (assoc :qualified-column [correlation-name identifier]))))
-          []
-          projections)]))
+           (fn [acc {:keys [identifier index outer-name]}]
```
jarohen: aware the `reduce` was here before so dw about changing it now, but if we only need `(count acc)` we could use `map-indexed` instead. again, might seem like golfing but because `map-indexed` is less powerful than `reduce` you're telling the reader that they don't need to worry about different elements affecting each other, and that the result is 1-to-1 with the input - whereas, with `reduce`, anything could happen :slightly_smiling_face:

notes/adr/0041-select-star-dup-cols.adoc
```diff
@@ -0,0 +1,91 @@
+= 41. SQL SELECT * / Duplicate Columns
```
jarohen, line 1: pedantry: adoc idiom is [one sentence per line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line) :slightly_smiling_face:

jarohen (approved):
cheers @mbutlerw, nice one :clap: subjective feedback as always but otherwise looks good to go without further review

PR #2580, opened 2023-06-20
Round-tripping and operations on decimals

### LINKED ISSUE #2044: Handling BigDecimal-like vectors in the expression engine
Vectors: `DecimalVector`, `Decimal256Vector`

* See xtdb/xtdb#1961 for the main checklist
* You'll also need to consider how this fits into the `types/arrow-type-hierarchy`, and likely adapt a lot of the `expr/codegen-call` arithmetic operations to cope with (e.g.) adding a Float8 to a Decimal (including the type such an operation should return)
* This is definitely the trickier of the xtdb/xtdb#1961 sub-tasks! (but I don't intend to create a 'good-second-issue' label)

FiV0:
Short circuiting #2044. Handling of decimals generally in a dynamic engine seems to be quite a rabbit hole. 
This PR essentially allows operation on decimals for precision 38 and scale 19. Anything outside of this range will currently throw an exception. Anything coming out of the engine will also have that precision and scale set.

Some useful information:
- https://github.com/apache/arrow-rs/issues/2387 - only useful link I could find of why arrow actually stores precision
- [IEEE-754 2008](http://www.dsc.ufcg.edu.br/~cnum/modulos/Modulo2/IEEE754_2008.pdf#%5B%7B%22num%22%3A281%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C110.2%2C720%2C0%5D) - expression evaluation between different types (doesn't actually say much)
- SQL standard 6.27 <numeric value expression> - if both operands are exact numeric (decimal, ints)  -> (implementation dependent) exact numeric, if one type is approximate numeric (float, double) -> (implementation dependent) approximate numeric  
- Postgres (http://sqlfiddle.com/#!17/8756f8/2) and Clojure (`(type (+' (double 1.0) (bigdec 1.0)))`) seem to favor Double over Decimal    

Stuff still todo/decide:
- ~correctly support operations between other types and decimal~ this currently follows the same behavior as Clojure (fewest changes). One could think about promoting an operation on `:f32` and `:decimal` to `:f64`, but the SQL spec does not require it. 
- How should we deal with decimals that still fit into an Arrow Decimal (128bit) but fall outside the above range (`1E+35M` , `1E-35M`)
- How to deal with things that don't fit into an Arrow Decimal ? Over- and Underflow?

### DIFF
--- a/core/src/main/clojure/xtdb/types.clj
+++ b/core/src/main/clojure/xtdb/types.clj
@@ -10,13 +10,15 @@
            (java.time Duration Instant LocalDate LocalTime Period ZoneId)
            java.util.concurrent.ConcurrentHashMap
            java.util.function.Function
-           (org.apache.arrow.vector BitVector DateDayVector DateMilliVector IntervalDayVector IntervalMonthDayNanoVector IntervalYearVector TimeMicroVector TimeMilliVector TimeNanoVector TimeSecVector TimeStampMicroTZVector TimeStampMilliTZVector TimeStampNanoTZVector TimeStampSecTZVector ValueVector VarBinaryVector VarCharVector)
+           (org.apache.arrow.vector BitVector DateDayVector DateMilliVector DecimalVector Decimal256Vector IntervalDayVector IntervalMonthDayNanoVector IntervalYearVector TimeMicroVector TimeMilliVector TimeNanoVector TimeSecVector TimeStampMicroTZVector TimeStampMilliTZVector TimeStampNanoTZVector TimeStampSecTZVector ValueVector VarBinaryVector VarCharVector)
            (org.apache.arrow.vector.complex DenseUnionVector FixedSizeListVector ListVector StructVector)
            (org.apache.arrow.vector.holders NullableIntervalDayHolder NullableIntervalMonthDayNanoHolder)
            (org.apache.arrow.vector.types DateUnit FloatingPointPrecision IntervalUnit TimeUnit Types$MinorType UnionMode)
-           (org.apache.arrow.vector.types.pojo ArrowType ArrowType$Binary ArrowType$Bool ArrowType$Date ArrowType$Duration ArrowType$FixedSizeBinary ArrowType$FixedSizeList ArrowType$FloatingPoint ArrowType$Int ArrowType$Interval ArrowType$List ArrowType$Null ArrowType$Struct ArrowType$Time ArrowType$Time ArrowType$Timestamp ArrowType$Union ArrowType$Utf8 Field FieldType)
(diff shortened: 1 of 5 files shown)


core/src/main/clojure/xtdb/vector/writer.clj
```diff
@@ -140,6 +141,21 @@
         (writeNull [_ _] (.setNull arrow-vec (.getPositionAndIncrement wp)))
         (writeLong [_ days] (.setSafe arrow-vec (.getPositionAndIncrement wp) (* days 86400000)))))))
 
+(extend-protocol WriterFactory
+  DecimalVector
+  (->writer [arrow-vec]
+    (let [wp (IWriterPosition/build (.getValueCount arrow-vec))]
+      (reify IVectorWriter
+        (getVector [_] (doto arrow-vec (.setValueCount (.getPosition wp))))
```
jarohen: this now doesn't need to setValueCount - think this changed while your branch was in flight

core/src/main/clojure/xtdb/types.clj
```diff
@@ -564,7 +573,8 @@
   (-> col-type-hierarchy
       (derive :i8 :i16) (derive :i16 :i32) (derive :i32 :i64)
       (derive :f32 :f64)
-      (derive :int :f64) (derive :int :f32)))
+      (derive :int :f64) (derive :int :f32)
+      (derive :int :decimal) (derive :uint :decimal)))
```
jarohen, line 582: This one's made me pause for thought... :thinking: we sure we can do this?

Theoretically, it's probably true - all ints could be treated as decimals - but I think the expression engine might then expect to be able to read a value from an IntVector (primitive) and use it as a `BigDecimal` object without any cast required.

jarohen (commented):
Nice one @FiV0 :clap: 

One question - when you (e.g.) `(.setScale 1M 10)` you get back `1.0000000000M`, so are you in fact getting back bigdecs with extra zeroes? (would be hidden by the equality check) - and, if so, does it matter?

I guess the alternative would be to parameterise the `scale` on our decimal col-type (like we do for the timestamps), but I guess this has other consequences, like needing separate vectors/casts for bigdecs that happen to differ in scale...

Happy to go with your decision on this one - but regardless of what you choose, mind writing up your decision and rationale on the PR so that we remember it when we come back to future work on bigdecs?

Cheers! :pray:

FiV0:
The liberal widening of `int` -> `decimal` and `decimal` -> `float` in the widening-hierarchy is only possible because we are mainly using polymorphic clojure functions in the expression engine. The cases where `Math/*` is used it just works out of the box because `Math/pow` and `Math/log` correctly coerce the arguments to double. The cases of `greatest/least` (`Math/min` / `Math/max` under the hood) only work because 2-arugments invocations get replaced by calls to `<` and `>`. So this part definitely needs to change when expanding these cases.

PR #2586, opened 2023-06-23
Checking for changes on the checkpointer

### LINKED ISSUE #2581: Nodes shouldn't create a checkpoint if there's already one in the checkpoint store at the latest-completed-tx
Need to consider potential impacts on lifecycle policies here - e.g. if someone has a lifecycle policy to delete old checkpoints after 7 days and no transactions in that time, a naive impl might result in an empty checkpoint bucket.

danmason:
Resolves #2581 

NOTE: We will want to be clear about this and warn around retention policies within the **release notes**.

Updates the behaviour of the checkpointer module - checking for changes:
- We check for changes since the last checkpoint, only making a new checkpoint if the latest completed`tx-id` is different to that on the checkpoint itself. 
- Added a test to this to `checkpoint_test`.
- Added some docs under `parameters` on the checkpointing document (a warning around the new behaviour & lifecycle policies)

Follow-up work:
- Adding lifecycle policies to the checkpointing module itself

### DIFF
--- a/docs/administration/modules/ROOT/pages/checkpointing.adoc
+++ b/docs/administration/modules/ROOT/pages/checkpointing.adoc
@@ -18,6 +18,8 @@ The desired frequency of checkpoints can be set using `approx-frequency`.
 
 The default lifecycle of checkpoints is unbounded - XTDB will not attempt to clean up old checkpoints. Users will typically want to define a retention policy that best fits their requirements and create means of implementing that policy that is external to XTDB.
 
+WARNING: Current behaviour of the checkpointer is such that we only make a checkpoint if we have processed new transactions since the previous checkpoint. As such - you will need to be careful with any lifecycle policies set on your checkpoints - if no new transactions are handled for a period, no new checkpoints will be made, and you may risk having no checkpoints and having to replay from the Transaction Log!
+
 == Setting up
 
 You can enable checkpoints on your index-store by adding a `:checkpointer` dependency to the underlying KV store:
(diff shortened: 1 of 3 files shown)


docs/administration/modules/ROOT/pages/checkpointing.adoc
```diff
@@ -77,18 +77,20 @@ EDN::
...
 
+WARNING: If **check-for-changes?** is set to true, you will need to be aware of any lifecycle policies set on your checkpoints  - if no new checkpoints are made before checkpoints are cleaned up, you may risk having no checkpointing!
+
 == `FileSystem` Checkpoint Store parameters
 * `path` (required, string/`File`/`Path`/`URI`): path to store checkpoints.
 
 === Note about using `FileSystem` Checkpoint Store
 
 Most Cloud providers offer some high-performance block storage facility (e.g. link:https://aws.amazon.com/ebs/[AWS EBS] or link:https://cloud.google.com/storage[Google Cloud Storage]).
-These volumes usually can be mounted on compute platforms (e.g. EC2) or containers (e.g. K8S).
```
danmason: Not sure what happened here - will fix this..

jarohen (commented):
Looks good @danmason :clap: On reflection, I'd prefer we didn't merge this until we've sorted the retention policy - I think it'd be weird to release this one on its own. For example, as part of the retention policy change, we may decide that we then don't need a separate `check-for-changes?` flag - so let's firm up the spec for that one :slightly_smiling_face:

danmason:
As discussed - will make this default behaviour, and add a specific warning note to the docs.

jarohen (approved):

PR #2591, opened 2023-06-26
WIP: Add retention policy settings to checkpointer

### LINKED ISSUE #2589: Ability to specify checkpoint retention policy
Currently, we advise that users create (e.g.) an S3 retention policy to clear out old checkpoints. Unfortunately, with #2586, this could lead to the S3 lifecycle policy removing all checkpoints if there aren't any new transactions within the retention period.

So, let's bring the retention policy within XT:

```clojure
{:xtdb/checkpointer {:store {...}
                     :retention-policy {:retain-newer-than "PT7D"
                                        :retain-at-least 2}}}
```

* `:retention-policy` is optional here.
* When we upload a checkpoint, we then enforce the retention rules. If a checkpoint is _both_ older than the `:retain-newer-than` (if provided) and there are more than `:retain-at-least` more recent checkpoints, the node should delete the checkpoint.
* To be considered: how should this work with cp-formats?
  * We want to eventually remove _all_ checkpoints of previous cp-formats (regardless of the `:retain-at-least` parameter)
  * The checkpointer may not necessarily know which format is older - maybe the checkpoint-store can return the checkpoints in a sorted order?

danmason:
Resolves #2589 

Adds `retention-policy` settings to the checkpointer:
- Takes two options, retain-newer-than and retain-at-least
- When we upload a checkpoint, we then enforce the retention rules:
  - If we specify only `retain-at-least`, then we only keep that many checkpoints.
  - If we specify only `retain-newer-than`, then we only keep checkpointers newer than the provided time. 
  - If we specify both - we keep at least `retain-at-least` checkpoints and the rest we decide whether or not to keep based on their age.
- Add tests for the above cases:
  - Both specified
  - Only `retain-at-least`.
  - Only `retain-newer-than` 
- Add a test for `retain-at-least` with a filesystem checkpoint store to sanity check behaviour at a higher level.  
- Adds documentation around configuring the retention policy.

### DIFF
--- a/core/src/xtdb/checkpoint.clj
+++ b/core/src/xtdb/checkpoint.clj
@@ -1,9 +1,11 @@
 (ns xtdb.checkpoint
   (:require [clojure.tools.logging :as log]
+            [clojure.spec.alpha :as s]
             [clojure.java.io :as io]
             [xtdb.api :as xt]
             [xtdb.io :as xio]
             [xtdb.bus :as bus]
+            [xtdb.error :as err]
             [xtdb.system :as sys])
   (:import [java.io Closeable File]
            java.net.URI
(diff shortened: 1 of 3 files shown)


jarohen:
grand, cheers @danmason :pray:

PR #2595, opened 2023-06-27
Ensure cleanup-checkpoint on checkpoint stores deletes the checkpoint metadata

### LINKED ISSUE #2593: Ensure `cleanup-checkpoint` on checkpoint stores deletes the checkpoint metadata file
Currently, all implementations of `cleanup-checkpoint` currently seem to **not** delete the checkpoint metadata file (`checkpoint-<tx-id>-<cp-at>.edn`). As such, failed checkpoints will not be fully deleted and there will be lingering `.edn` files. Unsure on the exact problems this may cause, though as we use these checkpoint files within `available-checkpoints` we should not be keeping them for deleted/non-reachable checkpoints. 

Additionally, the `s3-transfer-manager` checkpoint store will require an implementation of `cleanup-checkpoint`.

danmason:
Resolves #2593 

Within this PR:
- Ensure `cleanup-checkpoint` on the filesystem checkpoint store deletes everything (including the metadata file), add test for this.
- Test that the s3 checkpoint store deletes all files on `cleanup-checkpoint`:
  - Ensure we explicitly call to delete the metadata file first. 
- Add `cleanup-checkpoint` to s3 transfer manager, test behaviour:
  - Similar to it's implementation of `available-checkpoints` - this is copied from the s3 checkpoint store.

### DIFF
--- a/modules/s3/src/xtdb/s3/checkpoint.clj
+++ b/modules/s3/src/xtdb/s3/checkpoint.clj
@@ -75,15 +75,20 @@
                                                                ::cp/checkpoint-at cp-at}))})
         cp)))
 
-  (cleanup-checkpoint [this  {:keys [tx cp-at]}]
-    (let [s3-dir (format "checkpoint-%s-%s" (::xt/tx-id tx) (xio/format-rfc3339-date cp-at))
-          files (as-> (s3/list-objects this {:path s3-dir :recursive? true}) $
-                  (keep (fn [[type arg]]
-                          (when (= type :object)
-                            arg)) $)
-                  (vec $)
-                  (conj $ s3-dir))]
-      (s3/delete-objects this files)))
+  (cleanup-checkpoint [this {:keys [tx cp-at]}]
+    (let [s3-dir (format "checkpoint-%s-%s" (::xt/tx-id tx) (xio/format-rfc3339-date cp-at))]
+
+      ;; Delete EDN file first
+      (s3/delete-objects this [(str s3-dir ".edn")])
+
+      ;; List & delete directory contents
+      (let [files (as-> (s3/list-objects this {:path s3-dir :recursive? true}) $
+                    (keep (fn [[type arg]]
+                            (when (= type :object)
+                              arg)) $)
+                    (vec $)
+                    (conj $ s3-dir))]
+        (s3/delete-objects this files))))
 
   Closeable
   (close [_]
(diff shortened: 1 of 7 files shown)


modules/s3/src/xtdb/s3/checkpoint_transfer_manager.clj
```diff
@@ -112,6 +112,16 @@
                                                                ::s3-dir (str s3-dir "/")
                                                                ::cp/checkpoint-at cp-at}))})
         cp)))
+  
+  (cleanup-checkpoint [this {:keys [tx cp-at]}]
```
jarohen, line 116: does this also need to delete the edn file first, so that no nodes download a corrupted checkpoint?

danmason: I am a little unclear on what you mean - the current behaviour of both this and the same implementation of `cleanup-checkpoint` on the other s3 checkpoint store is that they delete everything _including_ the edn file?

danmason: Or would you prefer that it splits out the edn file from the list and calls to delete that separately?

jarohen: @danmason yeah, would prefer that we confirm the EDN file's been deleted successfully before deleting anything else - we can't necessarily rely on its position in the list meaning it's deleted first, for example

danmason: Can certainly do that :)

modules/s3/src/xtdb/s3/checkpoint.clj
```diff
@@ -75,15 +75,20 @@
...
-  (cleanup-checkpoint [this  {:keys [tx cp-at]}]
-    (let [s3-dir (format "checkpoint-%s-%s" (::xt/tx-id tx) (xio/format-rfc3339-date cp-at))
-          files (as-> (s3/list-objects this {:path s3-dir :recursive? true}) $
-                  (keep (fn [[type arg]]
-                          (when (= type :object)
-                            arg)) $)
-                  (vec $)
-                  (conj $ s3-dir))]
-      (s3/delete-objects this files)))
+  (cleanup-checkpoint [this {:keys [tx cp-at]}]
```
danmason, line 78: @jarohen Updated the cleanup-checkpoint implementation on both of the s3 checkpoint stores and added a basic sanity test to ensure that edn metadata file is deleted first and even if the rest fails you cannot list the checkpoint as available :)

jarohen (approved):
cheers @danmason - good for a squash-n-merge :pray:

PR #2596, opened 2023-06-27
Backport 2.x `$` document match syntax back to 1.x

jarohen:
Intention is to get early user feedback on the new `$` ('match') syntax in the 2.x branch by backporting it to 1.x, and also make the migration path easier (as, if the pilot is successful, users can start to write queries using match syntax before the migration). 

Syntax is the same as 2.x, except without the table and temporal opts parameters:
```clojure
{:find [a c]
 :where [;; full long-hand
         ;; don't need `{:xt/id eid}` if `eid` not used
         ($ [{:xt/id eid, :a a, :b "literal-b", :c c}])

         ;; sugar - [a c] in the vec is equivalent to {:a a, :c c}
         ($ [{:b "literal-b"} a c])

         ;; sugar - drop the vec if it's a singleton map
         ($ {:a a, :b "literal-b", :c c})]}

;; any of the above one-liners are equivalent to:

{:find [a c]
 :where [[eid :a a]
         [eid :b "literal-b"]
         [eid :c c]]}
```

See the TPC-H diff in the PR for more in-depth/less contrived examples

Implemented as a sugar in `xtdb.query/normalise-clause`

TODO:
* if the approach is agreed, will write up the docs. will of course make clear that it's an alpha/subject-to-change addition.
* port some more tests over. TPC-H is a reasonable stress test, but wouldn't hurt to port a few more.

### DIFF
--- a/core/src/xtdb/query.clj
+++ b/core/src/xtdb/query.clj
@@ -54,10 +54,41 @@
 (defn- aggregate? [x]
   (contains? (methods aggregate) x))
 
+(s/def ::triple-value (some-fn logic-var? literal?))
+
 (s/def ::triple
   (s/and vector? (s/cat :e (some-fn logic-var? c/valid-id? set?)
                         :a (s/and c/valid-id? some?)
-                        :v (s/? (some-fn logic-var? literal?)))))
+                        :v (s/? ::triple-value))))
+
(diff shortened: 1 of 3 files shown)


jarohen:
closing due XTQL

PR #2605, opened 2023-06-29
Content and temporal WAL for Covering Index Approach

### LINKED ISSUE #2572: Ingest to write out content + temporal WAL
(we already have quite a lot of this - spec for completeness)

For each chunk, we'll upload two WAL files - content and temporal.

* Temporal file will just be puts/deletes; content file will contain the evictable data.
* Files to be aligned - one row in the content file to one row in the temporal file put vector.
 
  e.g. if temporal is 'put row 0', 'delete row 0', 'put row 1', the content file will contain two rows - 0 and 1. These correspond to indices in the put vector.
* WAL files to contain rows for all tables (i.e. not split by table) - see the tx-producer schema.
* Arrow files to be split into pages of 1Ki rows - as soon as a transaction takes us over this threshold, we close the page.
* The schema of the content file won't be able to be finalised until the end of the chunk, so it's not possible to write this out as we go (because Arrow requires the final schema ahead-of-time).
* Live WALs able to be rolled back if a transaction aborts

These files are to be considered a golden source - if we were to ever completely change the indexing strategy, the whole state should be able to be reconstructed from these two files alone. 

Apart from that, I doubt an awful lot will read from these files - maybe an external secondary index (e.g. for populating a full-text search index) - so users may choose to store these files in a less-accessible storage class.

Current state is pretty much this, but the content files are currently split by table and then by column - we move all the columns/tables into one content file per chunk.

FiV0:
This adds a WAL for content and temporal data. The puts in the temporal file are aligned with the content file.
See also #2572.

Still todo/decide:
- [x] where the files should go? (currently `chunk-00/temporal-log.arrow`)
   - keep them there for now
- [x] ~sql erase doesn't show up in the temporal log~ was related to bug #2607 
- [x] ~how iid's are represented when they hit those two logs?~
  - just saw that there is already a new fn `->iid` in the code base. Will adapt to that.
- [x] ~Avoid row-copier recreation in content-log~
  - Not really possible because sql ops create rels per op. Maybe something to revisit.
- [x] ~Update not properly logging non updated content into the content log~

Main changes:
- temporal log indexer -
   - just a simple sequence of transactions (`tx-id`, `system-time`, `tx-ops` (DenseUnion list))
   - the concatenation of `puts` in the the transactions aligns with the content in the content-log
- content log indexer
  - one row per `put` (also SQL insert or update)
  - the schema is not fixed here but rather a DUV where each leg (identified by name) specifies the table 
- the main struggle was to square the writing out of a dynamically changing schema using our `IIndirectRelation` and `IVectorWriter` abstractions. 
  - there is now a hack in `writers.clj` for DUV's which deals specially when the duv legs are named different to the col-type default. That is to make working with fixed named schemas easier.
  - The other bit that this PR revealed is that if one wants to write from an indirect relation into a vector (struct or duv) or vice versa the two corresponding interfaces don't combine very nicely. We decided here to go from struct vector to irel. There now is a function `struct-writer->rel-writer` which does just that. The issue is that one might still need to call `endRow` on a higher irel writer of which the struct-writer is a child so one can not easily just encapsulate the rel-writer of the struct. Maybe something to revisit.


core/src/main/clojure/xtdb/indexer/temporal_log_indexer.clj
```diff
@@ -0,0 +1,153 @@
...
+    (reify ILogIndexer
+      (startTx [_ tx-key]
+        (.writeLong tx-id-wtr (.tx-id tx-key))
+        (vw/write-value! (.system-time tx-key) system-time-wtr)
+
+        (.startList tx-ops-wtr)
+        (reify ILogOpIndexer
+          (logPut [_ iid row-id app-time-start app-time-end]
+            (.startStruct put-wtr)
+            (.writeBytes put-iid-wtr (ByteBuffer/wrap (long->byte-hash iid)))
```
FiV0: I currently still do this hacky long->byte-hash conversion. What should the iid be that comes in here.

core/src/main/clojure/xtdb/indexer/content_log_indexer.clj
```diff
@@ -0,0 +1,114 @@
...
+        !block-row-count (AtomicInteger.)
+        !transient-block-row-count (AtomicInteger.)]
+
+    (reify IContentIndexer
+      (startTx [_]
+        (reify IContentOpIndexer
+          (logPut [_ table-name table-vec doc-offset]
+            (let [field (types/->field table-name types/dense-union-type false)
+                  table-wrt (.writerForField document-wtr field)
+                  copier (.rowCopier table-wrt table-vec)]
```
FiV0: The row copier here could probably at least be hashed per tx.

core/src/main/clojure/xtdb/indexer.clj
```diff
@@ -343,27 +362,31 @@
...
-                               util/end-of-time-μs)]
-                (when (> valid-from valid-to)
-                  (throw (err/runtime-err :xtdb.indexer/invalid-valid-times
-                                          {:valid-from (util/micros->instant valid-from)
-                                           :valid-to (util/micros->instant valid-to)})))
-
-                (.logPut log-op-idxer iid row-id valid-from valid-to)
-                (.indexPut temporal-idxer iid row-id valid-from valid-to new-entity?)))))))))
+          ;; TODO revisit to check if copying to content log can be done more efficiently (see also sql-update)
+          (with-open [^ValueVector content-vec (vw/irel->struct-vec allocator content-rel)]
```
FiV0: @jarohen This is kind of hack. In the content-log I would like to have a vec which I get in `put` but not here. Open for suggestions, but we can also revisit once things stablilize.

core/src/main/clojure/xtdb/indexer/content_log_indexer.clj
```diff
@@ -0,0 +1,114 @@
...
+(defn- ->log-obj-key [chunk-idx]
+  (format "chunk-%s/content-log.arrow" (util/->lex-hex-string chunk-idx)))
+
+(defmethod ig/prep-key :xtdb.indexer/content-log-indexer [_ opts]
+  (merge {:allocator (ig/ref :xtdb/allocator)
+          :object-store (ig/ref :xtdb/object-store)}
+         opts))
+
+(def content-log-schema
+  (Schema. [(types/->field "documents" types/dense-union-type false)]))
```
FiV0, line 43: @jarohen The schema here is:
- documents (duv)
  - table-name (duv-leg name)
    - documents for that table also a `duv`
  
~One could potentially also add the tables as top-level vectors to the schema~
Then we lose the alignment.

FiV0: Anyways one could potentially lay it out a bit differently. (`StructVector` with unions pushed down etc...)

core/src/main/clojure/xtdb/indexer.clj
```diff
@@ -410,53 +443,58 @@
...
+                                             (.forEachRemaining (reify Consumer
+                                                                  (accept [_ block-root]
+                                                                    (let [col (iv/->direct-vec (.getVector ^VectorSchemaRoot block-root col-name))]
+                                                                      (doto ^IRowCopier (->live-row-copier col)
+                                                                        (.copyRow idx))))))))))))))
+                        (util/rethrowing-cause)))
+
+                  (.logPut log-op-idxer iid new-row-id valid-from valid-to)
+                  ;; TODO scan does not currently pass through eid, so not possible to do eid -> iid
+                  (.logPut temporal-log-op-idxer (long->byte-hash iid) new-row-id valid-from valid-to)
```
FiV0: The `xt$id` column doesn't come through here with the in-rel. So I can't do `(->iid eid)`. Should that be changed?

core/src/main/clojure/xtdb/indexer.clj
```diff
@@ -384,24 +408,32 @@
         (.findFirst)
         (.getAsInt))))
 
-(defn- ->sql-update-indexer ^xtdb.indexer.SqlOpIndexer [^IMetadataManager metadata-mgr, ^IBufferPool buffer-pool
-                                                        ^ILogOpIndexer log-op-idxer, ^ITemporalTxIndexer temporal-idxer
-                                                        ^ILiveChunkTx live-chunk]
+;; TODO replace the use of this with ->iid in sql update, delete and evict
```
jarohen: still to do, I'd guess? :wink:

core/src/main/clojure/xtdb/indexer.clj
jarohen: I'd lean more on the `IRowCopier` interface here rather than a separate `.logPut` on the content-log - live-chunk is probably more of an inspiration than the existing log indexer for this one. Let's pair :slightly_smiling_face:

jarohen (changes requested):
Hey @FiV0 (and @mbutlerw too) - nice work gents :clap: 

Broadly speaking this is on the money - as you say, let's pair up on a few of the more nuanced areas and I reckon we can get this landed.

Separately, I think we can now remove the original log indexer - it should be possible to read the same values from this one - but let's land this PR first and then see what _really_ still relies on the old one, no doubt it's more than I think...

core/src/main/clojure/xtdb/indexer/temporal_log_indexer.clj
```diff
@@ -0,0 +1,165 @@
...
+         opts))
+
+(def ^:private nullable-inst-type [:union #{:null [:timestamp-tz :micro "UTC"]}])
+
+(def temporal-log-schema
+  (Schema. [(types/col-type->field "tx-id" :i64)
+            (types/col-type->field "system-time" types/temporal-col-type)
+            (types/->field "tx-ops" types/list-type false
+                           ;; a null value indicates an abort
+                           (types/->field "$data" (ArrowType$Union. UnionMode/Dense (int-array (range 3))) true
```
jarohen: DUV's can't be null :sob:

core/src/main/clojure/xtdb/vector/writer.clj
```diff
@@ -604,9 +619,16 @@
...
                             n)
-            col-type (types/field->col-type (.get child-fields n))]
-        (aset copier-mapping src-type-id (.rowCopier (.writerForType dest-col col-type)
-                                                     (.getVectorByType src-vec src-type-id)))))
+            ^Field child-field (.get child-fields n)
+            child-field-name (.getName child-field)
+            col-type (types/field->col-type child-field)]
+        (aset copier-mapping src-type-id
+              ;; HACK to make things work for named duv legs
+              (if-not (str/starts-with? child-field-name (types/col-type->field-name col-type))
```
jarohen: there's a col-type->duv-leg(?) function that we likely could/should use here instead of the prefix of the col-name

core/src/main/clojure/xtdb/indexer/temporal_log_indexer.clj
FiV0: Move it up to `tx-ops`

core/src/main/clojure/xtdb/indexer.clj
FiV0: Tracked in #2611.

core/src/main/clojure/xtdb/indexer.clj
FiV0: Replaced with `nil-iid`s. Tracked in #2611.

core/src/main/clojure/xtdb/vector/writer.clj
FiV0: As discussed: removed the field-name indexing and did the comparison of field-name and `(col-type->field-name)`.

core/src/main/clojure/xtdb/indexer.clj
FiV0: See main PR comment.

jarohen (approved):

PR #2606, opened 2023-06-29
Updates to the Azure Blobs module

### LINKED ISSUE #2602: Azure blob store should roundtrip documents containing UUIDs
Context via: https://github.com/xtdb/xtdb/issues/1422#issuecomment-1612631604

> Ran into this issue yesterday when configuring the document-store to use Azure Blob Storage. And here is an example to replicate the behaviour https://github.com/philippkueng/xtdb-1422/blob/main/src/core.clj - I've also tried rocksdb and jdbc and both of them work just fine. It's only Azure failing consistently when inserting values of type UUID.

### LINKED ISSUE #2603: Azure NIO checkpoint + doc store should be tested & documented
The Google Storage module uses the NIO abstraction, which means it can be easily adapted for Azure also.

danmason:
Fixes #2602 and resolves #2603.

- Re-implement the document store in terms of the Azure API
  - Using the default credential chain of said API
  - Essentially a port of core2 object store ported to v1 document stores.
- Explicitly test for #2602 - seems to work fine with the new document store.
- Implement and test a new implementation of `checkpoint-store` 
- Update docs for both the azure blobs document store and for the new checkpoint store
  - Point to Azure docs with regards to setting up and auth

### DIFF
--- a/modules/azure-blobs/project.clj
+++ b/modules/azure-blobs/project.clj
@@ -14,7 +14,19 @@
   :dependencies [[org.clojure/clojure]
                  [org.clojure/tools.logging]
                  [com.xtdb/xtdb-core]
-                 [pro.juxt.clojars-mirrors.clj-http/clj-http "3.12.2"]]
+                 [com.azure/azure-storage-blob "12.22.0"]
+                 [com.azure/azure-identity "1.9.0"
+                  :exclusions [net.minidev/json-smart
+                               com.nimbusds/lang-tag]]
+
+                 ;; remove misc dependency warnings
+                 [com.azure/azure-core "1.39.0"]
+                 [com.azure/azure-core-http-netty "1.13.3"]
+                 [net.java.dev.jna/jna "5.13.0"]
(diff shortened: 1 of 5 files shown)


modules/azure-blobs/project.clj
```diff
@@ -14,7 +14,19 @@
   :dependencies [[org.clojure/clojure]
                  [org.clojure/tools.logging]
                  [com.xtdb/xtdb-core]
-                 [pro.juxt.clojars-mirrors.clj-http/clj-http "3.12.2"]]
+                 [com.azure/azure-storage-blob "12.22.0"]
+                 [com.azure/azure-identity "1.9.0"
+                  :exclusions [net.minidev/json-smart
+                               com.nimbusds/lang-tag]]
+
+                 ;; remove misc dependency warnings
```
jarohen, line 22: these belong in `:managed-dependencies` these days :slightly_smiling_face:

jarohen (approved):
cheers @danmason :pray:

PR #2609, opened 2023-06-30
Add getObjectRange request to the object store protocol

### LINKED ISSUE #2579: Nodes able to make partial buffer-pool/object-store requests (byte ranges)
We deliberately make the files in remote storage significantly larger than the page size (e.g. 128MiB compared to 64KiB), in part so that it's easier to make larger writes atomic, but also to cut down on the number of files in larger databases (to reduce the task of understanding what's there).

We likely won't be able to justify downloading that file for a low-latency request, though - so 

* [ ] let's expose an API on the buffer-pool and object-store to buffer/download a partial file, using a byte range
  * the buffer pool will need to handle various cases: where it's got the whole file and a user asks for a page (slice the file?), where it's got a page and the user asks for the file (download the file and evict the single page?), etc.
* [ ] let's store the byte ranges of each individual page (Arrow record batch) in the trie files - this is accessible through the Arrow footer at the end of the leaf files.
* [ ] the scan operator might want to make a decision about whether it'd prefer to download the whole file, if it's going to be downloading above a certain proportion of all the pages anyway. 
  * maybe a decision for the buffer pool, if the scan operator can batch up its requests? e.g. it could give a list of ranges, and maybe something the object store (/the mirror) knows the total size of the file?

In the object store, this will need to be implemented for the in-memory, file-system and S3 object stores (via the [Range](https://docs.aws.amazon.com/whitepapers/latest/s3-optimizing-performance-best-practices/use-byte-range-fetches.html) header).

wotbrew:
Linked to #2579

This PR includes a .getObjectRange implementation for the memory, filesystem, s3 and azure blob object stores. JDBC works but is not optimised, given the anaemic state of JDBC at the moment I figured this could be forgiven.

Boundary cases currently must be dealt with for every implementation to meet the tests. Zero-length and negative indexes are not permitted due to inconsistencies in implementation behaviour. This might later change.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -182,6 +182,8 @@ dependencies {
     testImplementation("cheshire", "cheshire", "5.11.0")
     testImplementation("org.xerial", "sqlite-jdbc", "3.39.3.0")
     testImplementation("org.clojure", "test.check", "1.1.1")
+    // for AWS profiles (managing datasets)
+    devImplementation("software.amazon.awssdk", "sts", "2.16.76")
 }
 
 if (hasProperty("fin")) {
--- a/core/src/main/clojure/xtdb/object_store.clj
+++ b/core/src/main/clojure/xtdb/object_store.clj
@@ -5,6 +5,7 @@
             [clojure.spec.alpha :as s])
   (:import java.io.Closeable
            java.nio.ByteBuffer
+           (java.nio.channels FileChannel$MapMode)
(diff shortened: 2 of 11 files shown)


jarohen (approved):
cheers @wotbrew :pray: could you ensure your Azure impl doesn't introduce back any reflection warnings? (conscious Finn's change went in concurrently)

wotbrew:
Superseded by #2617

PR #2612, opened 2023-07-03
Remove log-indexer

FiV0:
This removes all references to the old log-indexer after #2605. 
- Updates all test check files of the indexer to work against temporal/content log instead of the old log indexer files
- Removes all reference to log-indexer in the indexer
- Updates the temporal-log interfaces from `ILogIndexer2` -> `ILogIndexer`


jarohen:
awesome, cheers @FiV0 :pray:

PR #2613, opened 2023-07-04
Pass `xt/id` through to indexer in sql ops

### LINKED ISSUE #2611: Pass `xt$id` through to SQL `update`,`delete` and `erase` in indexer.
Currently the relations passed into `sql-update`, `sql-delete` and `sql-erase` don't contain `xt$id`. This will be needed to create the `iid` once #2605 goes in.

FiV0:
Closes #2611. Potential followup remove `iid` and possibly `row-id` from getting passed through once the old temporal indexer goes.


jarohen:
cheers @FiV0! :pray:

PR #2621, opened 2023-07-07
Use the google cloud storage API for the google cloud storage docs store/checkpoint store

### LINKED ISSUE #1865: Using GCP APIs for doc-store/checkpoint-store
We currently use the NIO adapter for GCP (for original ease of implementation)

Using official GCP APIs would give us more control over/insight into error handling, checkpoint upload/download and verification etc. 

(related to #1864)

### LINKED ISSUE #2616: Ensure google cloud checkpoint store has proper behaviour
Testing google cloud checkpoint store against recent changes/tests we've done against the other store:

- Ensure it always deletes the metadata file first
- Add any new tests we've added to the other checkpoint stores

danmason:
Resolves #1865, and Resolves #2616

Modeled off of the updated Azure blobs module - utilizing some similar tricks. 
- Metadata files are prefixed with `metadata-` as the list function is recursive (no configuration to prevent this, as far as I can tell, so we use the same prefix behaviour as azure blobs).
- Authentication handled as it was before (default google auth)
- Added extra tests for cleanup checkpoint behaviour based on those used against the other checkpoint stores  
- Updated docs

### DIFF
--- a/modules/google-cloud-storage/src/xtdb/google/cloud_storage.clj
+++ b/modules/google-cloud-storage/src/xtdb/google/cloud_storage.clj
@@ -62,14 +62,15 @@
          vec
          (run! deref)))
 
-  (fetch-docs [this docs]
+  (fetch-docs [this ids]
     (xio/with-nippy-thaw-all
       (reduce
-       #(if-let [doc (get-blob this %2)]
-          (assoc %1 %2 (nippy/thaw doc))
-          %1)
+       (fn [docs id]
+         (if-let [doc (get-blob this id)]
+           (assoc docs id (nippy/thaw doc))
+           docs))
        {}
-       docs))))
+       ids))))
 
 (defrecord GoogleCloudStorageCheckpointStore [^Storage storage-service bucket-name prefix]
   cp/CheckpointStore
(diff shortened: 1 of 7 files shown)


modules/google-cloud-storage/src/xtdb/google/cloud_storage.clj
```diff
@@ -1,14 +1,182 @@
...
+           (future
+             ;; Put blob - key set to id, blob is nippy compressed document byte array
+             (put-blob-bytes this id (nippy/freeze doc))))
+         vec
+         (run! deref)))
+
+  (fetch-docs [this docs]
+    (xio/with-nippy-thaw-all
+      (reduce
+       #(if-let [doc (get-blob this %2)]
```
jarohen: nit: for these anon-fns, when you've got multiple arguments and you're using them multiple times, I tend to use `fn` with named params rather than `#()`

modules/google-cloud-storage/test/xtdb/google/cloud_storage_test.clj
```diff
@@ -1,27 +1,145 @@
...
+    (try
+      (let [storage (-> (StorageOptions/newBuilder)
+                        (.setProjectId project-id)
+                        (.build)
+                        (.getService))
+            bucket (.get storage test-bucket (into-array Storage$BucketGetOption []))]
+        (when (.exists bucket (into-array Bucket$BucketSourceOption []))
+          (f)))
+      (catch StorageException e
+        (if (= 401(.getCode e))
```
jarohen: nit: `when-not`?

jarohen (approved):
couple of minor nits but otherwise good to go, cheers @danmason! :pray:

PR #2625, opened 2023-07-10
`cleanup-checkpoint` on NIO filestores

### LINKED ISSUE #2623: Ensure call to `cleanup-checkpoint` on NIO filesystems works as expected
- Currently, the call to `cleanup-checkpoint` does not work against NIO filesystems 
- This is caused by the usage of `io/file` operations, which do not support NIO
- Want to update and explicitly test this against an NIO filestore (such as google cloud)

danmason:
Resolves #2623 

Fixes and tests the previous issue of NIO file stores not deleting files on `cleanup-checkpoint`:
- Accomplishes this by using direct Java interop using Paths, rather than clojure.java.io (which does not support NIO)
- Adds some new helper functions around paths to `xtdb.io` - `path-exists?` and `delete-path`. 
- Adds a new test file (originally failing, now passing) against NIO filesystem checkpoint store (ran via lein profiles against gcloud nio)


jarohen (approved):
cheers @danmason 🙏

PR #2629, opened 2023-07-13
Implement google cloud storage object store

### LINKED ISSUE #2628: Google Cloud Storage as an Object Store
Related to #1952 

Can copy across the implementation of document store and the various calls we're making to the GCP API from the recent core1 changes.

danmason:
Resolves #2628 

Implements a new object store using Azure Blob Storage, alongside:
- All of the related integrant config to add it to a node.
- Including the module within various places - namely the xtdb standalone deps and dev dependencies.
- Documentation in the form of a README in the module.
- A test file making use of the object-store tests.
  - Expects various bits of auth & the specified bucket to be present
- Updates to `object-store-test/test-range`:
  - Remove assertion where we expect an error if starting position > object length (this is _undefined_ behaviour, we don't necessarily expect an error to be thrown, thanks @wotbrew !)
  - Add an assertion that the method throws an `IllegalStateException` if the object doesn't exist/is not found)  

As a note on the behaviour:
- I'm _fairly sure_ using reader on google cloud blobs shouldnt read the whole file - [https://javadoc.io/doc/com.google.cloud/google-cloud-storage/latest/com/google/cloud/storage/Storage.html#reader-com.google.c[…]BlobSourceOption...-](https://javadoc.io/doc/com.google.cloud/google-cloud-storage/latest/com/google/cloud/storage/Storage.html#reader-com.google.cloud.storage.BlobId-com.google.cloud.storage.Storage.BlobSourceOption...-)
- It's using `ReadableByteChannel` under the hood, which seems to read byte by byte, so I'm assuming should have the kinda behaviour we're looking for.
- Was following https://cloud.google.com/storage/docs/samples/storage-download-byte-range initially, though have had to dip into `ReadableByteChannel` itself to read the content into a `ByteBuffer`

### DIFF
--- a/modules/google-cloud/README.adoc
+++ b/modules/google-cloud/README.adoc
@@ -1,0 +1,61 @@
+# Google Cloud Module
+
+Within our XTDB node, we can make use of Google Cloud Services for certain purposes. Currently, we can:
+
+* Use *Google Cloud Storage* as the XTDB Object Store.
+
+### Project Dependency 
+
+In order to use any of the Azure services, you will need to include a dependency on the xtdb.azure module.
+
+_deps.edn_
+```
+com.xtdb.labs/xtdb-google-cloud {:mvn/version "2.0.0-SNAPSHOT"}
+```
+
+_pom.xml_
+```
+<dependency>
+    <groupId>com.xtdb.labs</groupId>
+    <artifactId>xtdb-google-cloud</artifactId>
+    <version>2.0.0-SNAPSHOT</version>
+</dependency>
+```
+
+### Authentication
+
+Authentication for both the document store and checkpoint store components within the module is handled via Google’s "Application Default Credentials" - see the https://github.com/googleapis/google-auth-library-java/blob/main/README.md#application-default-credentials[*relevant documentation*] to get set up. You will need to setup authentication using any of the methods listed within the Azure documentation to be able to make use of the operations inside the modules. 
+
+## Google Cloud Storage Object Store
+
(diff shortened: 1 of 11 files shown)


modules/google-cloud/README.adoc
```diff
@@ -0,0 +1,61 @@
+# Google Cloud Module
+
+Within our XTDB node, we can make use of Google Cloud Services for certain purposes. Currently, we can:
+
+* Use *Google Cloud Storage* as the XTDB Object Store.
+
+### Project Dependency 
+
+In order to use any of the Azure services, you will need to include a dependency on the xtdb.azure module.
```
jarohen: did you copy paste this?! :scream: 

(j/k, obv, but looks like this file needs a find/replace)

modules/google-cloud/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
...
+        }
+    }
+}
+
+dependencies {
+    api(project(":api"))
+    api(project(":core"))
+
+    api("com.google.cloud", "google-cloud-storage", "2.23.0")
+    api("com.google.guava", "guava", "31.0.1-jre")
```
jarohen: you may not need guava in Gradle, if it's just for dependency resolution - Gradle tends to be much better at resolving dependencies

jarohen (approved):
cheers @danmason! just a find-replace on the docs and this is good to merge

modules/google-cloud/README.adoc
danmason: So close.... I guess my find and replace missed a couple :scream:

modules/google-cloud/build.gradle.kts
danmason: :eyes: That's useful to know - what would be a good way of confirming that? Attempting to run the tests without it, perhaps?

PR #2633, opened 2023-07-17
Local Filename caching for S3 

### LINKED ISSUE #2578: Nodes can subscribe to files created/deleted in the object store, to reduce LIST requests
Once a database grows above a certain threshold, we'll likely have many files on remote storage, to the point where it may be inefficient (not to mention uneconomical) to list the remote storage every time we need to check whether there's a new file (particularly as we'll need to paginate beyond 1000 files on S3, for example). 

We'd prefer to mirror the file listing locally by subscribing to updates, and then query a local cache rather than making requests of the remote store.

* API to be designed - although we may well just want this to live entirely within the object store for now rather than being exposed through its API (i.e. assuming that we'll mirror the whole object store).
* Consider the startup race condition - we'll want to list the remote store initially and also subscribe to updates, but we likely can't do these atomically. We don't want to miss files in the gap between the list and the subscription kicking in, but subscribing and then listing might mean we get an update for a file that's already there.
  * tbc, but I think we can assume that a deleted file is never re-put, if that helps?
* At a later date (out of scope for now) we may well want the creation of a file to (e.g.) stop the current node to stop background processing an index file (if another node's already done that work), or the creation of a chunk-finished marker to advance the node (if it's running slowly).

To be implemented for:
* the in-memory object store (an atom watcher, I'd guess)
* the file-system object store (using the `FileWatcher` API, I'd guess)
* S3. In this case, as a starter-for-ten, we could pipe S3 create/delete notifications to an SNS topic, and then (on node startup) create an SQS queue to subscribe to the SNS topic (deleting the queue on node shutdown).
  * If you do go this route, create a CloudFormation template in the XT repo that users can use for the S3 bucket, SNS topic and S3 events, along with an IAM role for nodes to create/delete SQS queues, and subscribe them to the topic.
  * May be a separate card...

danmason:
Resolves #2578

Currently implemented:
- The general filename caching approach, used on s3 object stores 
  - We use a `NavigableSet` under the hood for filename caching purposes.
  - We run a function against this to start handling the list - firstly, we subscribe to file changes (how this is handled is up to the implementation), then we do an "init" function which lists the current files in the object store, and then we handle the stream of file changes by either adding/removing from the cache.
  - List functions on object store essentially become either return the set as a vector, or doing some filtering based on prefixed and then returning all results which contain the prefix (a helper function is available for this)    
- S3 related setup:
  - Created cloudformation templates (one for regular use, one specifically for our testing), that set up:
    - An SNS topic  
    - An S3 bucket, sending notifications to the SNS topic on object creation & deletion.
    - An SNS topic policy that allows the S3 bucket to send messages to the topic. 
    - An XTDB IAM Role, with all relevant permissions for S3, SNS, and SQS (for the creation and usage of queues in code) 
    - S3 object store now requires `topic-arn` for the SNS topic to be passed in, alongside all the permissions of the XTDB IAM User
    - We firstly setup our sqs queue subscribed to the SNS topic alongside all relevant permissins.
    - We then do our 'init' behaviour, which uses s3 list-objects to get all current bucket contents.    
    - We then start handling the notifications from the sqs queue - polling for messages in a loop (running on a thread):
- Some updates to the object store tests to ensure that they fit new behaviour.
   - Essentially, these have to wait for a while while we catch up the filename list internally. 
- Added a README to S3 containing info on config/using cloudformation.

### DIFF
--- a/core/src/main/clojure/xtdb/file_list.clj
+++ b/core/src/main/clojure/xtdb/file_list.clj
@@ -1,0 +1,8 @@
+(ns xtdb.file-list
+  (:require [clojure.string :as string])
+  (:import java.util.NavigableSet))
+
+(defn list-files-under-prefix [^NavigableSet file-name-cache prefix]
+  (->> (.tailSet ^NavigableSet file-name-cache prefix)
+       (take-while #(string/starts-with? % prefix))
+       (into [])))
--- a/modules/s3/README.adoc
+++ b/modules/s3/README.adoc
@@ -30,8 +30,6 @@ The paramaterized stack expects the following to be provided:
 
 * The name of S3 Bucket that will be used by the created object store bucket and referenced in associated resources
 ** If unset, uses a default value of `xtdb-object-store`
-* The name of the XTDB IAM user to create.
-** If unset, uses a default value of `xtdb-user`
 
 ### Created Resources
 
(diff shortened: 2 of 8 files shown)


jarohen:
@danmason as discussed, let's use AWS roles for the user-facing CF template rather than an explicit user :slightly_smiling_face:

danmason:
@jarohen Have switched over to an IAM role, and having assumed it locally - all works as we'd hope :) Updated the README, also

PR #2637, opened 2023-07-19
Add stagnant log flushing process

### LINKED ISSUE #2548: If a tx isn't included in a chunk before the retention period on the log runs out, it may get lost
* All the nodes must agree on when to flush the live index.

One potential solution here would be for the nodes to put a 'let's all finish the chunk now if 8912 is still the latest tx' message on the tx-log, if they've got a chunk that's been idle for over ~a few hours? then, when that gets indexed, all the nodes flush the live index?

wotbrew:
Resolves #2548

Adds a new per-node process (`xtdb.stagnant-log-flusher`) that, if the log has has not caused any chunks to be flushed within a default period of 4 hours, will emit a message that causes any outstanding indexer writes to be flushed to durable storage.

The goal is to avoid data loss if the log has a retention policy set and sees no new activity within that period (transactions could then only exist in the transient live chunk). 

We solve this problem by checking periodically if we have flushed anything new since the last run, If we have not, we write a `:flush-index-if-log-not-moving` tx op to the log containing the latest transaction id. The indexer, receiving this message will then flush any blocks and chunks to storage, provided the latest tx-id remains the same.

The process accepts a `:duration` configuration key against the `:xtdb.stagnant-log-flusher/flusher` integrant component to change the timing of its heartbeats.

### DIFF
--- a/core/src/main/clojure/xtdb/indexer.clj
+++ b/core/src/main/clojure/xtdb/indexer.clj
@@ -6,7 +6,7 @@
             [xtdb.error :as err]
             xtdb.indexer.internal-id-manager
             xtdb.indexer.live-index
-            xtdb.live-chunk
+            [xtdb.live-chunk :as live-chunk]
             [xtdb.metadata :as meta]
             xtdb.object-store
             [xtdb.operator :as op]
@@ -42,7 +42,7 @@
            xtdb.operator.IRaQuerySource
            (xtdb.operator.scan IScanEmitter)
            (xtdb.temporal ITemporalManager ITemporalTxIndexer)
-           (xtdb.vector IRowCopier IVectorReader IVectorWriter RelationReader)
+           (xtdb.vector IRowCopier IVectorReader RelationReader)
            (xtdb.watermark IWatermark IWatermarkSource)))
 
 (set! *unchecked-math* :warn-on-boxed)
@@ -51,7 +51,9 @@
 (definterface IIndexer
   (^xtdb.api.protocols.TransactionInstant indexTx [^xtdb.api.protocols.TransactionInstant tx
                                                    ^org.apache.arrow.vector.VectorSchemaRoot txRoot])
-  (^xtdb.api.protocols.TransactionInstant latestCompletedTx []))
+  (^xtdb.api.protocols.TransactionInstant latestCompletedTx [])
+  (^xtdb.api.protocols.TransactionInstant latestCompletedChunkTx [])
+  (^void forceFlush [^xtdb.api.protocols.TransactionInstant txKey ^long expected-last-chunk-tx-id]))
 
 (defprotocol Finish
   (^void finish-block! [_])
@@ -528,6 +530,7 @@
                   ^ILiveIndex live-idx
 
                   ^:volatile-mutable ^TransactionInstant latest-completed-tx
+                  ^:volatile-mutable ^TransactionInstant latest-completed-chunk-tx
 
                   ^:volatile-mutable ^IWatermark shared-wm
                   ^StampedLock wm-lock]
@@ -620,13 +623,23 @@
               (finally
                 (.unlock wm-lock wm-lock-stamp)))))
 
-        (while (.isBlockFull live-chunk)
+        (while (live-chunk/block-full? live-chunk)
           (finish-block! this))
 
         (when (.isChunkFull live-chunk)
           (finish-chunk! this))
 
         tx-key)))
+
+  (forceFlush [this tx-key expected-last-chunk-tx-id]
+
+    (when (= (:tx-id latest-completed-chunk-tx -1) expected-last-chunk-tx-id)
+      (while (pos? (.blockRowCount live-chunk))
+        (finish-block! this))
+
+      (finish-chunk! this))
+
+    (set! (.-latest_completed_tx this) tx-key))
 
   IWatermarkSource
   (openWatermark [this tx-key]
@@ -659,6 +672,7 @@
                 (.unlock wm-lock wm-lock-stamp)))))))
 
   (latestCompletedTx [_] latest-completed-tx)
+  (latestCompletedChunkTx [_] latest-completed-chunk-tx)
 
   Finish
   (finish-block! [this]
@@ -676,6 +690,8 @@
           (finally
             (.unlock wm-lock wm-lock-stamp))))
 
+      (log/debug "finished block")
+
       (catch Throwable t
         (clojure.tools.logging/error t "fail")
         (throw t))))
@@ -684,7 +700,6 @@
     (let [chunk-idx (.chunkIdx live-chunk)]
       (.registerNewChunk temporal-mgr chunk-idx)
       @(.finishChunk live-chunk latest-completed-tx)
-
       (.finishChunk live-idx chunk-idx))
 
     (let [wm-lock-stamp (.writeLock wm-lock)]
@@ -694,6 +709,7 @@
           (.close shared-wm))
 
         (.nextChunk live-chunk)
+        (set! (.-latest_completed_chunk_tx this) latest-completed-tx)
 
         (finally
           (.unlock wm-lock wm-lock-stamp))))
@@ -722,13 +738,14 @@
 
   (let [{:keys [latest-completed-tx]} (meta/latest-chunk-metadata metadata-mgr)]
 
-    (Indexer. allocator object-store scan-emitter temporal-mgr internal-id-mgr
-              ra-src live-chunk live-index
+    (->Indexer allocator object-store scan-emitter temporal-mgr internal-id-mgr
+               ra-src live-chunk live-index
 
-              latest-completed-tx
+               latest-completed-tx
+               latest-completed-tx
 
-              nil ; watermark
-              (StampedLock.))))
+               nil ; watermark
+               (StampedLock.))))
 
 (defmethod ig/halt-key! :xtdb/indexer [_ ^AutoCloseable indexer]
   (.close indexer))
--- a/core/src/main/clojure/xtdb/stagnant_log_flusher.clj
+++ b/core/src/main/clojure/xtdb/stagnant_log_flusher.clj
@@ -1,0 +1,96 @@
+(ns xtdb.stagnant-log-flusher
+  (:require [clojure.tools.logging :as log]
+            [xtdb.log :as xt-log]
+            [xtdb.indexer]
+            [xtdb.node :as node]
+            [xtdb.tx-producer]
+            [juxt.clojars-mirrors.integrant.core :as ig]
+            [xtdb.util :as util])
+  (:import (java.nio ByteBuffer)
+           (java.nio.channels ClosedByInterruptException)
+           (java.time Duration)
+           (java.util.concurrent ExecutorService Executors ThreadFactory TimeUnit)
+           (xtdb.indexer IIndexer)
+           (xtdb.log Log)))
+
+;; see https://github.com/xtdb/xtdb/issues/2548
+
+(defmethod ig/prep-key ::flusher
+  [_ opts]
+  (merge {:indexer (ig/ref :xtdb/indexer)
+          :log (ig/ref :xtdb/log)}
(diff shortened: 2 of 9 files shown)


core/src/main/clojure/xtdb/stagnant_log_flusher.clj
```diff
@@ -0,0 +1,111 @@
...
+(defmethod ig/init-key ::flusher
+  [_ {:keys [^ITxProducer tx-producer
+             ^IIndexer indexer
+             duration
+             ;; callback hook used to control timing in tests
+             ;; receives a map of the :last-flush, :last-seen tx-keys
+             on-heartbeat]
+      :or {duration #time/duration "PT4H"
+           on-heartbeat (constantly nil)}}]
+  (let [exr-tf (reify ThreadFactory
```
jarohen: we have something similar in util, `->prefix-thread-factory` - suitable here?

core/src/main/clojure/xtdb/stagnant_log_flusher.clj
```diff
@@ -0,0 +1,111 @@
...
+     :task (.scheduleAtFixedRate exr f 0 (.toMillis ^Duration duration) TimeUnit/MILLISECONDS)}))
+
+(defmethod ig/halt-key! ::flusher [_ {:keys [^ExecutorService executor, task]}]
+  (future-cancel task)
+  (.shutdown executor)
+  (let [timeout-secs 10]
+    (when-not (.awaitTermination executor timeout-secs TimeUnit/SECONDS)
+      (log/warnf "flusher did not shutdown within %d seconds"))))
+
+(comment
```
jarohen, line 74: nit: worth shifting this to `dev`?

src/test/clojure/xtdb/stagnant_log_flusher_test.clj
```diff
@@ -0,0 +1,174 @@
...
+  "Change if tolerances change and tests need more time (such as slower CI machines), used for `spin`."
+  500)
+
+(defmacro spin [expr] `(spin-until *spin-ms* ~expr))
+
+(defn each-fixture [f]
+  (let [log-nses '[xtdb.stagnant-log-flusher
+                   xtdb.indexer]
+        old-levels (mapv tu/get-log-level! log-nses)]
+    (try
```
jarohen: if `util/with-log-level` isn't suitable here, maybe we could figure out what we can do to make it suitable?

core/src/main/clojure/xtdb/indexer.clj
```diff
@@ -614,11 +624,25 @@
               (finally
                 (.unlock wm-lock wm-lock-stamp)))))
 
-        (while (.isBlockFull live-chunk)
-          (finish-block! this))
+        (let [force-flush-req (some? @force-flush-tx-id)
```
jarohen: nit: `when-let`?

core/src/main/clojure/xtdb/indexer.clj
```diff
@@ -614,11 +624,25 @@
...
+            (log/info "received flush instruction: no newer transactions found, will flush chunk.")
+
+            force-flush-req
+            (log/info "received flush instruction: newer transaction found, so not flushing chunk."))
+
+          (when (or (.isBlockFull live-chunk)
+                    force-flush-match)
+            (loop []
+              (finish-block! this)
+              (when (.isBlockFull live-chunk)
```
jarohen: this `loop` is to keep flushing blocks if we happen to get a large transaction in that's bigger than our block size - if memory serves, `finish-block!` might not leave the live-chunk empty. so I'd assume you'll want to keep flushing blocks til it's completely empty here?

(could definitely have benefited from a comment here, apologies)

jarohen (changes requested):
cheers @wotbrew :slightly_smiling_face: :pray: 

couple of things to consider here:
* I'm not sure it's sufficient to just check the latest completed tx - we might get a pattern of transactions in that just about keep that check happy, but a chunk still hasn't been flushed within the retention period (because it's not full). Think we have to check against the latest tx in the previous chunk, say?
* Are there risks in making this a tx-op, that can be submitted with other tx-ops? I'm not sure of the semantics if a user were to include this in the middle of their tx, for example - would it continue indexing the rest of the tx in the next chunk? we could say 'a user'll never submit this', I guess.

  Admittedly I'm out of context, but I did wonder about whether this might be more straightforward as a replacement for the whole tx - the ingester could then (e.g.) call a different function on the indexer, and the rest of the tx-processing in `index-tx` wouldn't need to know about it. (aware this is subjective design, happy to go with your impl if you'd prefer.)

core/src/main/clojure/xtdb/indexer.clj
wotbrew: Yes I think we need a way to tell non-empty. I would look at something like blockRowLimit and rowCount methods (removing isBlockFull), but I would welcome your guidance as these are not familiar interfaces to me.

src/test/clojure/xtdb/stagnant_log_flusher_test.clj
wotbrew: I did at one point have more nses in the list, believe me it was actually uglier! Now it is 2 tis probably ok for `with`as it stands. If we want to make it work nicely for `n` levels we can make it take multiple levels. What about pair style: 

```clojure 
(tu/with-log-levels
  {xtdb.stagnant-log-flusher :info 
   xtdb.indexer :debug}
  (do-stuff))
```

Not sure what you think?
  
Again I try to not change shared utils unless essential as part of an unrelated PR, but if you are happy I can change it. (I would argue for separate issue, refactor task)

core/src/main/clojure/xtdb/indexer.clj
wotbrew: I use the match variable lower down in the `or`'s guarding block/chunk finishing, if you want to provide a rewrite happy to adopt it

core/src/main/clojure/xtdb/stagnant_log_flusher.clj
wotbrew: Not exactly sure what you mean, could you clarify?

wotbrew:
> cheers @wotbrew 🙂 🙏
> 
> couple of things to consider here:
> 
> * I'm not sure it's sufficient to just check the latest completed tx - we might get a pattern of transactions in that just about keep that check happy, but a chunk still hasn't been flushed within the retention period (because it's not full). Think we have to check against the latest tx in the previous chunk, say?
> * Are there risks in making this a tx-op, that can be submitted with other tx-ops? I'm not sure of the semantics if a user were to include this in the middle of their tx, for example - would it continue indexing the rest of the tx in the next chunk? we could say 'a user'll never submit this', I guess.
>   Admittedly I'm out of context, but I did wonder about whether this might be more straightforward as a replacement for the whole tx - the ingester could then (e.g.) call a different function on the indexer, and the rest of the tx-processing in `index-tx` wouldn't need to know about it. (aware this is subjective design, happy to go with your impl if you'd prefer.)

Good points, certainly the first one is a big misconception/logical flaw in my approach (one point Rich Hickey problem pyramid)! 

I'll look at changing to meet the expectations laid out above 👍, sounds better to me.

core/src/main/clojure/xtdb/indexer.clj
jarohen: yeah - think this is an example of unnecessarily hiding the data that you mentioned the other day. I'd be happy with that change :+1: `blockRowLimit` then needn't live with the live-chunk, think that might already be configuration on the indexer...?

core/src/main/clojure/xtdb/stagnant_log_flusher.clj
jarohen: we have a `dev` namespace with this kind of code - there are a few system definitions in there that folks can use to test certain things. would this system config be appropriate for standalone-config, for example? or maybe a separate config map?

core/src/main/clojure/xtdb/indexer.clj
jarohen: it's more to cut down on the number of names required - but tbh I shouldn't golf :golf: 

```clojure
(let [force-flush-tx-id @force-flush-tx-id
      force-flush-match (and force-flush-tx-id (= force-flush-tx-id last-tx-id))]
  (cond
    force-flush-match
    (log/info "received flush instruction: no newer transactions found, will flush chunk.")

    force-flush-tx-id
    (log/info "received flush instruction: newer transaction found, so not flushing chunk."))

  ...
)
```

src/test/clojure/xtdb/stagnant_log_flusher_test.clj
jarohen: yep, pair style would work for me. :+1:

jarohen: it might also be helpful as a test fixture, on reflection:

```clojure
(defn with-log-level [ns level f]
  ...)

(t/use-fixtures :each (tu/with-log-level ns1 l1) (tu/with-log-level ns2 l2) your-each-fixture)
```

...

although maybe only in this case, because you happen to be using it in `:each`... I think in other use cases we only apply it in certain tests within a namespace

core/src/main/clojure/xtdb/stagnant_log_flusher.clj
wotbrew: Oh yea, the sys config is difficult. The problem with sub-configurations is there is a large number of them I suppose. I suspect it's easier to start with a full node config and then partially configure it.

wotbrew: Something like: 

```clojure
(def sys
    (-> (node/node-system {::flusher {:duration #time/duration "PT5S"}})
        (ig/prep)
        (ig/init [::flusher])))
```

wotbrew: But I'd need to add the system constructor to do the defaulting and so on that node/start-node does (or it gets ugly again?)

core/src/main/clojure/xtdb/indexer.clj
wotbrew: Did this, but I put the rowsPerBlock on live-chunk, (indexer didn't have it, and to me it seems the live-chunk should be transparent about it) let me know if that is a problem.

src/test/clojure/xtdb/stagnant_log_flusher_test.clj
wotbrew: oh-no-please-god-no.gif 

Sorry, memories of point-free fixture hell.

For these, honestly I would use a trick with a dynamic var because then you can denest and still clean up. Hold my beer? EDIT: I have something on a shelf if you want to observe a bit of heresy

core/src/main/clojure/xtdb/indexer.clj
```diff
@@ -618,14 +626,24 @@
...
 
+  (forceFlush [this tx-key expected-last-chunk-tx-id]
+
+    (when (= (:tx-id latest-completed-chunk-tx -1) expected-last-chunk-tx-id)
+      (while (pos? (.blockRowCount live-chunk))
+        (finish-block! this))
+
+      (finish-chunk! this))
+
+    (set! (.-latest_completed_tx this) tx-key))
```
jarohen, line 642: what's this `set!` doing? shouldn't the indexer already be maintaining this?

core/src/main/clojure/xtdb/indexer.clj
```diff
@@ -516,6 +518,11 @@
 
       (.indexPut temporal-tx iid row-id system-time-µs util/end-of-time-μs true))))
 
+(defn- get-force-flush-tx-id [^DenseUnionVector tx-ops-vec tx-op-idx]
```
jarohen: dead?

core/src/main/clojure/xtdb/indexer.clj
jarohen: oh, I get it, the flush message is essentially a tx in its own right.

PR #2638, opened 2023-07-19
Adapting scan for live trie and trie logs

### LINKED ISSUE #2574: Scan operator to read live log tries for point/point queries
As a first pass before #2575, let's just have the scan operator reading the live log tries.

* Live log tries to be watermarked
* When the scan operator spots a point/point query, on a node with no persisted chunks, it reads the live log trie instead of using the current cursors.

FiV0:
This brings in a first version of a valid-point cursor that merges live-trie as well as not yet merged trie logs. 
The rough order of operation is:
- get relevant trie logs + live trie data
- merge leaf data sorted by iid (breaking ties via newer chunks, ie greater system time)
- simple temporal resolution for valid points queries
- vector alignment for columns going out of `:scan`
- apply any scan predicates.

Stuff still todo:
- [x] ~some issue with absent columns~ was a now fixed issue with `:union-all` 
- [ ] cleanup + more efficient merging (could do that in next step)
- [x] ~metadata checking (is there actually any data in this chunk for this table ?)~ for later

### DIFF
--- a/core/src/main/clojure/xtdb/indexer.clj
+++ b/core/src/main/clojure/xtdb/indexer.clj
@@ -73,7 +73,6 @@
 (defn- ->iid ^bytes [eid]
   (if (uuid? eid)
     (util/uuid->bytes eid)
-
     (let [^bytes eid-bytes (cond
                              (string? eid) (.getBytes (str "s" eid))
                              (keyword? eid) (.getBytes (str "k" eid))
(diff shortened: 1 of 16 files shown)


jarohen:
superseded by #2642 - we'll still be able to take a lot of inspiration from this branch when it comes to #2575

PR #2639, opened 2023-07-26
Basic external sort for order-by

### LINKED ISSUE #1939: Query engine able to cope with larger-than-memory intermediate results
- [x] Join (spill hash-map to disk)
- [x] Order by (external sort)
- [ ] Group by (although less important if we assume the number of groups will be small)

FiV0:
This creates a first version of an external sort for order by. The rough algorithm is as follows:
- Consume batches until a chunk is full and then flush that chunk to disk.
- Do k-way-merges (k configurable) of these chunk files until we hit one file.
- The k-way-merge flushes out sorted chunks for the next k-way-merge. So overall this means we should never have more then (k+1) chunks in memory. These (k+1) chunks can of course still vary greatly based on the number of columns that are projected out.
- From the one sorted file we consume chunks as we move along (batch chunking still todo).

There is no memory mapping of these files as ArrowRecordBatches should assure that we always have only one chunk per file in memory.
 
Todo / possible improvements
- [ ] ~batch chunking~ to test if we actual need it
- [x] cleanup files as we merge
- [x] ~can object (re)creation (readers, comparators, copiers) be avoided?~ to some extent
- [ ] ~optimisations of all sorts~ (probably good for another ticket)

### DIFF
--- a/core/src/main/clojure/xtdb/operator/order_by.clj
+++ b/core/src/main/clojure/xtdb/operator/order_by.clj
@@ -1,15 +1,27 @@
 (ns xtdb.operator.order-by
-  (:require [clojure.spec.alpha :as s]
+  (:require [clojure.java.io :as io]
+            [clojure.spec.alpha :as s]
             [xtdb.expression.comparator :as expr.comp]
             [xtdb.logical-plan :as lp]
+            [xtdb.types :as types]
             [xtdb.util :as util]
+            [xtdb.vector.reader :as vr]
             [xtdb.vector.writer :as vw])
-  (:import (java.util Comparator)
+  (:import (clojure.lang IPersistentMap PersistentVector)
+           (java.io File OutputStream)
+           (java.nio.channels Channels)
+           java.nio.file.Path
+           (java.util HashMap PriorityQueue)
+           (java.util Comparator)
            (java.util.function Consumer ToIntFunction)
            java.util.stream.IntStream
-           org.apache.arrow.memory.BufferAllocator
+           (org.apache.arrow.memory BufferAllocator)
+           (org.apache.arrow.vector ValueVector VectorSchemaRoot)
+           (org.apache.arrow.vector.ipc ArrowStreamReader ArrowStreamWriter)
+           [org.apache.arrow.vector.ipc ArrowStreamReader]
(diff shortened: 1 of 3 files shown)


core/src/main/clojure/xtdb/operator/order_by.clj
```diff
@@ -59,25 +81,211 @@
...
+      (letfn [(load-next-batch []
+                (if (.loadNextBatch ^ArrowStreamReader (.reader this))
+                  (with-open [out-rel (vr/<-root (.read-root this))]
+                    (.accept c out-rel)
+                    true)
+                  (do
+                    (io/delete-file (.sorted-file this))
+                    (set! (.consumed? this) true)
+                    false)))]
+        (if (nil? reader)
```
jarohen: nit: invert `if`

core/src/main/clojure/xtdb/operator/order_by.clj
```diff
@@ -59,25 +81,211 @@
...
+                      (inc file-idx)))
+             new-batches))
+         (inc batch-idx))))))
+
 (deftype OrderByCursor [^BufferAllocator allocator
                         ^ICursor in-cursor
-                        order-specs]
+                        ^IPersistentMap static-fields
+                        order-specs
+                        ^:unsynchronized-mutable ^Boolean consumed?
```
jarohen: `^:boolean`

core/src/main/clojure/xtdb/operator/order_by.clj
```diff
@@ -59,25 +81,211 @@
...
+                      (.add pq i)))
+                  ;; spill next chunk
+                  (when (< ^int *chunk-size* (.getPositionAndIncrement rel-wrt-cnt))
+                    (.syncRowCount rel-wrt)
+                    (.writeBatch wrt)
+                    (.clear rel-wrt)
+                    (.setPosition rel-wrt-cnt 0)
+                    (doseq [i (range k)]
+                      (aset copiers i (.rowCopier rel-wrt (aget rel-rdrs i))))))))
+            ;; spill remaining parts
```
jarohen: "spill remaining rows"/"spill last batch"?

jarohen (approved):

Issue #2655, opened 2023-08-01
Fix trie scan cursors for consistent `xt/system-to`


FiV0:
The system-to coming from a live-trie is always `NULL` and will be set when coming from a 4R trie. The behavior should be consistently set.

FiV0:
Test that period predicates work correctly:
`period_predicates.test`
```
query ITT nosort
SELECT t1.xt$id, t1.start, t1.end FROM t1 WHERE t1.SYSTEM_TIME CONTAINS PERIOD (t1.start, t1.end)
-----
2
2100-01-01
3000-01-01
```

refset:
Hey @FiV0 do you think this is still relevant at all?

PR #2661, opened 2023-08-03
Range queries for trie cursors

### LINKED ISSUE #2257: Should the temporal index yield zero-width periods?
see (e.g.) https://github.com/xtdb/core2/blob/e4953c89276dc62bf4e293801d1d83a28e806c65/test/core2/api_test.clj#L255-L286

### LINKED ISSUE #2649: Scan serves temporal range queries using the tries

FiV0:
This adds two new row-pickers:
- A RangePoint row-picker for scanning valid-times with a fixed system-time.
- A RangeRange row-picker for anything that doesn't fall into point/point or range/point.

Some implications of this change:
- certain tests needed to be updated because we are not consistently returning `system-to` (see #2655)
- As we are doing the rectangle slicing differently now some tests output changed significantly (see sl-demo.test)
- Some tests got zero-width system-time results removed as I am currently just overcoloring things. So zero-width won't show up in a result set.
- Before this we were applying some temporal predicates (for example `(<= xt$system_from ?sys_time)`) when actually doing temporal resolution. This is only applied as a post filter.
- The different Consumers of the leaf need to share some state (like should the next `iid` be skipped etc..). This is currently encapsulated in some `picker-state` but could probably be shared in some more transparent manner to the consumers.

Todo:
- [x] refactor into one cursor? (only the selection differs) 
- [x] issues with system-time boundaries 
- [x] erase later than specified sys-time
- [x] a basis + some predicates can not necessarily represented with a 8 coordinates.


src/test/clojure/xtdb/node_test.clj
```diff
@@ -8,17 +8,14 @@
 
 (t/deftest test-multi-value-insert-423
   (letfn [(expected [tt]
-            {[(util/->zdt #inst "2024-01-01") (util/->zdt util/end-of-time), tt tt]
-             "Happy 2024!",
-
-             ;; weird? zero-width system-time so won't normally show up
-             [(util/->zdt #inst "2024-01-01") (util/->zdt #inst "2026-01-01"), tt tt]
+            {;; FIXME once a decision stands on zero-width intervals #2257
```
jarohen: if your change happens to remove zero-width intervals, that's great - and feel free to tick off #2257

src/test/clojure/xtdb/as_of_test.clj
```diff
@@ -84,22 +83,24 @@
                 :xt/system-from tt2
                 :xt/system-to end-of-time-zdt}]
 
-    (t/is (= [replaced-v0-doc v1-doc]
-             (tu/query-ra '[:scan {:table xt_docs}
-                            [xt/id version
-                             xt/valid-from xt/valid-to
-                             xt/system-from xt/system-to]]
-                          {:node tu/*node* :default-all-valid-time? true}))
+    ;; FIXME nil system-to
```
jarohen: is this one to fix before merge?

src/test/clojure/xtdb/sql/temporal_test.clj
```diff
@@ -12,15 +12,15 @@
...
+           (set (query-at-tx
+                 "SELECT foo.last_updated FROM foo"
+                 tx2))))
 
-    (is (= [{:last_updated "tx1"} {:last_updated "tx1"} {:last_updated "tx2"}]
-           (query-at-tx
-            "SELECT foo.last_updated FROM foo FOR ALL SYSTEM_TIME"
-            tx2)))))
+    (is (= [ {:last_updated "tx1"} {:last_updated "tx1"} {:last_updated "tx2"}]
+           (sort-by :last_updated (query-at-tx
```
jarohen: nit: could pop this in the query if duplicates are important

src/test/resources/xtdb/sql/logic_test/direct-sql/identifier-case-insensitivity.test
```diff
@@ -71,19 +71,19 @@ cat
 query TI
 SELECT T1.col1, AVG(t1.col2) FROM t1 GROUP BY T1.col1
 ----
+30
```
jarohen, line 74: as @mbutlerw mentioned, there's a sorting construct within the SLT runner that's probably more suitable here

jarohen: ah, I see now you've done it for the others 🙂

src/test/resources/xtdb/sql/logic_test/direct-sql/period_predicates.test
```diff
@@ -36,13 +36,6 @@ SELECT t1.xt$id, t1.start, t1.end FROM t1 WHERE PERIOD (t1.start, t1.end) PRECED
 2000-01-01
 2020-01-01
 
-query ITT nosort
```
jarohen, line 39: at a glance this one still seems relevant?

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -528,39 +605,363 @@
...
-(deftype PointPointCursor [^BufferAllocator allocator, arrow-leaves
-                           ^Iterator merge-tasks, ^ints leaf-idxs, ^ints current-arrow-page-idxs
-                           col-names, ^Map col-preds, ^longs temporal-timestamps,
-                           params]
+                      (constrain-bounds valid-from valid-to))))))))))))
+
+(deftype Interval [^long start, ^long end, ^long sys-from])
+
+(defn- intersect [^Interval i1 ^Interval i2]
+  ;; TODO strict checks ?
```
jarohen: could you resolve these to-dos?

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -528,39 +605,363 @@
...
+              (.clear !ranges)
+              (duplicate-ptr prev-iid-ptr current-iid-ptr))
+
+            (if (= :evict (.getLeg op-rdr idx))
+              (duplicate-ptr skip-iid-ptr current-iid-ptr)
+              (let [sys-from (.getLong sys-from-rdr idx)]
+                (when (<= sys-from system-time)
+                  (case (.getLeg op-rdr idx)
+                    :put
+                    (let [i1 (->Interval (.getLong put-valid-from-rdr idx) (.getLong put-valid-to-rdr idx) sys-from)
```
jarohen: by calling `->Interval` here I _think_ you're boxing up the two values - whereas you don't with `Interval.` - happy to be corrected on that though 🙂

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -528,39 +605,363 @@
...
+                              sys-to (.sys-to r)]
+                          (when (and (<= valid-from-lower valid-from)
+                                     (<= valid-from valid-from-upper)
+                                     (<= valid-to-lower valid-to)
+                                     (<= valid-to valid-to-upper)
+                                     (<= sys-from-lower sys-from)
+                                     (<= sys-from sys-from-upper)
+                                     (<= sys-to-lower sys-to)
+                                     (<= sys-to sys-to-upper)
+                                     ;; FIXME
```
jarohen: still to fix?

jarohen (changes requested):
Cheers @FiV0 - a few minor pointers to address but then we'll be good to merge 🙂

src/test/resources/xtdb/sql/logic_test/direct-sql/period_predicates.test
FiV0: This has to do with the `nil` system-to issue. The predicate doesn't work on that. I have added this removal to #2655.

PR #2665, opened 2023-08-04
Add filename caching to Azure object store

### LINKED ISSUE #2643: Local filename caching setup for Azure Blobs Object Store
- Similarly to the implementation within S3 (see #2633) - we want to set up local filename caching for list requests within the  Azure Blobs Object Store.
- This will likely be quite similar to S3 - we will need to make an initial list request and then react to all creation/deletion events in the bucket.
- We can push container events to Azure Event Grid and then subscribe to the event grid, filtering out the messages we need - see [Reacting to Blob storage events](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-event-overview)
- For infrastructure set up (including the blob container) - we will want some equivalent to the cloudformation template within AWS. This can be done in Azure using "Azure Resource Manager" templates: https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/overview#template-deployment

danmason:
# Main PR contents

Resolves #2643 

Implements local filename caching for Azure - alongside a template for all the related necessary infrastructure:
- Created a parameterized azure resource template, that sets up:
  - A new storage account
  - A container to use as the object store.
  - An event grid system topic, subscribed to blobcreate/blobdelete events on the container.
  - A `standard` tier service-bus namespace.
  - A service-bus topic for the service-bus namespace.
  - A subscription on the eventgrid system topic for the service-bus topic.
  - A custom role definition for all of the necessary permissions for XTDB to use the above.  
  - Azure blob object store now expects the `servicebus-namespace` and `servicebus-topic-name` to be passed into config, and the process is expected to have all the permissions of the custom role
  - We firstly create a new subscription to the service bus topic.
  - We then do our 'init' behaviour, which uses azure list-objects to get all current bucket contents.
  - We create a service bus processor for the topic subscription - handling messages/errors. 
- Updates to the azure test to ensure that they fit new listing behaviour, and all works as expected.
- Update the Azure README with setup info

Also included is: 
- A fix to the `get-object` error behaviour to ensure that the object store throws expected errors in `get-range`.
- Pin the Azure module to a specific azure-http-netty version, to not pull in a newer transitive dependency and break other XTDB functions.

# Initial implementation and why/how it had to change

A note of the other approach I tried here, and what I had to in the end - more details below:

Initially, I wanted to set this up using _just_ [**Service Bus Queues**](https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-queues-topics-subscriptions#queues) (as this would allow me to use the BASIC tier of ServiceBus, and pricing to be scale-to-zero). 
- I setup all of the required to create everything needed within the code:
  - Created a queue via the ServiceBusAdministrationClient (using a new queue name for each - suffixed with a unique short id).
  - Created a subscription to the eventgrid system topic via the EventGrid ResourceManager class.
  - Also created all of the code necessary to clean both of these up.
- The servicebus processor client pointed at the created queue.
- This would process messages fine - it worked very similarly to how it does now.

We encountered a problem, however. There was a test that did the following:
- Start an object store (1)
- Add a few things to the object store
- Query that the list function works - this always passed, could always see messages going through.
- Close object store (1).
- Start a new object store (2), same settings as before
- Call to list objects - can see the previous elements added to the previous object store.
- DELETE one of the previous elements.
- Call to list objects - should process the delete, should be one less element.

Now - that delete event never seem to came on the second object store. Even if we waited for up to a minute, it never seemed to come through. As a note, this should have been a brand new service bus queue created with a new subscription. This issue _also_ broke things _between_ tests - ie, if I ran all tests, they would fail, whereas individually they were fine.

If we waited some amount of time between closing the first object store and creating the second, all seemed to work fine! In practice, this was around a minute or so, at which case the new object store would be created, its new queue and subscription created, and it would process the delete as expected.

There's a more extended investigation/set of notes on the [thread in the JUXT slack](https://juxt.slack.com/archives/G018EA25K71/p1691075790519189), but to summarize, on looking at azure level logging, this is what I believe the problem to be:
-  Under the hood of the servicebus queue, they use the AMQP protocol - I could see the creation of AMQP connections/channels in the logs (see [**here**](https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-amqp-protocol-guide) for more info)
- Looking at the logs, it seems like in the case where we didn't wait between creating object stores, AMQP reuses the original connection made for the first queue. I suppose it may well be a retry thing - if it temporarily gets disconnected it wants to connect.
- As such - no messages ever come in on the old queue/connection, and it simply doesn't receive anything.
- In the case where we wait between object store creation - it makes a **new** AMQP connection, which is why I believe it succeeds.

From what I can tell - there's not a lot that can be done with the AMQP setup as a high level service bus API user. The most I could do was switch to using a [different AMQP transport type](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.servicebus.transporttype?view=azure-dotnet), though this had no impact from what I could tell. I also tried switching off of the [**ServiceBusProcessorClient**](https://learn.microsoft.com/en-us/java/api/com.azure.messaging.servicebus.servicebusprocessorclient?view=azure-java-stable) to using the [**ServiceBusReceiverClient**](https://learn.microsoft.com/en-us/java/api/com.azure.messaging.servicebus.servicebusreceiverclient?view=azure-java-stable) - and again, I saw no change.

What **did** fix the problem, in the end, was switching over from using Service Bus Queues to using a [**Service Bus Topic**](https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-queues-topics-subscriptions#topics-and-subscriptions) that is subscribed to the Event Grid system topic, and creating 'subscriptions' to that in the code and using those with the processor client. In THAT case, all seems to be well - at the expense of having to use the 'STANDARD' pricing tier for Service Bus.

### DIFF
--- a/modules/azure/README.adoc
+++ b/modules/azure/README.adoc
@@ -1,11 +1,11 @@
-# Azure Module
+= Azure Module
 
 Within our XTDB node, we can make use of Azure Services for certain purposes. Currently, we can:
 
 * Use *Azure Blob Storage* as the XTDB Object Store.
 * Use *Azure Eventhubs* as the Log.
 
-### Project Dependency 
+== Project Dependency 
 
 In order to use any of the Azure services, you will need to include a dependency on the xtdb.azure module.
 
@@ -23,11 +23,11 @@ _pom.xml_
 </dependency>
 ```
 
-### Authentication
+== Authentication
 
 Authentication for the components in the module is done via the https://learn.microsoft.com/en-us/java/api/com.azure.identity.defaultazurecredential?view=azure-java-stable[*DefaultAzureCredential*] class - you will need to setup authentication using any of the methods listed within the Azure documentation to be able to make use of the operations inside the modules.
 
-## Azure Blob Storage Object Store
+== Azure Blob Storage Object Store
 
 We can swap out the implementation of the object store with one based on Azure Blob Storage. To do so, we add the `xtdb.azure/blob-object-store` key within the integrant config for our node, alongside any config:
 ```clojure
@@ -39,7 +39,40 @@ We can swap out the implementation of the object store with one based on Azure B
 }
 ```
 
-### Parameters
+=== Azure Resource Manager Stack
+
+In order to handle the various bits of Azure infrastructure required to use Azure Blob Storage as an XTDB object store, we provide a link:azure-resource-manager/azure-stack.json[parameterized 'Azure Resource Manager' stack] to setup everything that you should need.
+
+==== Inputs to the stack
+
+The paramaterized stack expects the following to be provided:
+
+* The location to build the resources -- Defaults to the location of the resource group you're deploying the template in.
+* The name of the storage account to create -- Defaults to `xtdbstorageaccount`
+* The pricing tier/type of storage account to use -- Defaults to `Standard_LRS`
+* The name of the storage container to use as the XTDB object store (and used by associated resources) -- Defaults to `xtdb-object-store`
+* The name of the EventGrid system topic we use for container events -- Defaults to `xtdb-storage-account-system-topic`
+* The name to use for the Service Bus namespace -- Defaults to `xtdb-storage-account-eventbus`
+* The name to use for the Service Bus topic created in the namespace above -- Defaults to `xtdb-storage-bus-topic`
+* The name of the custom XTDB role to create with permissions for the created resources -- Defaults to `xtdb-role`
+
+
+==== Created Resources
+
+The Azure Resource Manager stack will create the following resources within the specified resource group in Azure:
+
+* An https://learn.microsoft.com/en-us/azure/storage/common/storage-account-overview[Azure Storage Account]
+* A https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction#containers[Storage Account Container] within said storage account to use as the object store.
+* An https://learn.microsoft.com/en-us/azure/event-grid/system-topics[EventGrid System Topic], subscribed to blob creation/deletion events on the container.
+* A `standard` tier https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-messaging-overview#namespaces[Service Bus Namespace]:
+** A https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-queues-topics-subscriptions#topics-and-subscriptions[Service Bus Topic] for the Service Bus namespace.
+* An https://learn.microsoft.com/en-us/azure/event-grid/concepts#event-subscriptions[Event Subscription] on the EventGrid System Topic for the Service Bus Topic.
+* A https://learn.microsoft.com/en-us/azure/role-based-access-control/custom-roles[custom role definition] for all of the necessary permissions for XTDB to use the above:
+** Using the created storage container (Get, Put, Delete and List).
+** Reading messages sent to EventGrid.
+** Creating subscriptions on the Service Bus topic.
+
+=== Parameters
 
 These are the following parameters that can be passed within the config for our `xtdb.azure/blob-object-store`:
 [cols="1,1,2,1"]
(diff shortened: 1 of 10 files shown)


modules/azure/README.adoc
```diff
@@ -39,7 +39,40 @@ We can swap out the implementation of the object store with one based on Azure B
...
+* The name of the storage container to use as the XTDB object store (and used by associated resources) -- Defaults to `xtdb-object-store`
+* The name of the EventGrid system topic we use for container events -- Defaults to `xtdb-storage-account-system-topic`
+* The name to use for the Service Bus namespace -- Defaults to `xtdb-storage-account-eventbus`
+* The name to use for the Service Bus topic created in the namespace above -- Defaults to `xtdb-storage-bus-topic`
+* The name of the custom XTDB role to create with permissions for the created resources -- Defaults to `xtdb-role`
+
+
+==== Created Resources
+
+The CloudFormation stack will create the following resources within the specified resource group in Azure:
```
jarohen: copy pasta! :smile:

modules/azure/README.adoc
jarohen, line 1: I'd probably re-order this file a little - it seems you talk first about the object-store, then the stack, then the params for the object store. Worth moving the stack to below the params?

modules/azure/src/main/clojure/xtdb/azure/file_watch.clj
```diff
@@ -0,0 +1,88 @@
...
+                           (mapv (fn [^BlobItem blob-item]
+                                   (subs (.getName blob-item) (count prefix)))))]
+    (.addAll file-name-cache filename-list)))
+
+(defn mk-short-uuid []
+  (subs (str (UUID/randomUUID)) 0 8))
+
+(defn setup-topic-subscription [{:keys [^TokenCredential azure-credential servicebus-namespace servicebus-topic-name]}]
+  (let [servicebus-admin-client (-> (ServiceBusAdministrationClientBuilder.)
+                                    (.credential (format "%s.servicebus.windows.net" servicebus-namespace)
```
jarohen, line 29: is this a different credential to the 'default credential provider'?

modules/azure/src/main/clojure/xtdb/azure/file_watch.clj
```diff
@@ -0,0 +1,88 @@
...
+                             (.buildProcessorClient))]
+
+      ;; Start processing messages from the queue
+    (log/info "Watching for filechanges from container " container)
+    (.start processor-client)
+
+      ;; Return all closeable opts from the function
+    (assoc closable-opts :processor-client processor-client)))
+
+(defn watcher-close-fn [{:keys [^ServiceBusAdministrationClient servicebus-admin-client ^ServiceBusProcessorClient processor-client servicebus-topic-name subscription-name]}]
```
jarohen: contrary to my advice from last week, given that you now have something closable this probably _is_ the time for an object :wink: 

In terms of naming, we tend to prefix these functions with `open-` - i.e. `open-file-list-watcher`, which would then return (probably) a `reify AutoCloseable`

src/test/clojure/xtdb/azure_test.clj
```diff
@@ -42,20 +45,49 @@
...
 
+(def wait-time-ms 5000)
+
 (t/deftest ^:azure list-test
-  (let [os (object-store (random-uuid))]
-    (os-test/test-list-objects os)))
+  (with-open [os (object-store (random-uuid))]
+    (os-test/test-list-objects os wait-time-ms)))
+
+(t/deftest ^:s3 list-test-with-prior-objects
```
jarohen: copy pasta :stuck_out_tongue:

jarohen (changes requested):
cheers @danmason :pray:

modules/azure/src/main/clojure/xtdb/azure/file_watch.clj
danmason: Yeah, needs to point to the service bus namespace

jarohen: more that this seems to require a `TokenCredential` from the user?

modules/azure/README.adoc
danmason: :+1: Seems reasonable, yeah :) README is still pretty work in progress, need to talk about how to auth and whatnot anyway, but I think parameters followed by stack makes sense (so long as we clearly point at the stack from the head of the file so people see it)

modules/azure/src/main/clojure/xtdb/azure/file_watch.clj
danmason: Oh sorry - was confused by the quesiton, I think a 'TokenCredential' is just the generic class for all of the auth classes, what is being based in here is the default credential created in the azure module?

modules/azure/src/main/clojure/xtdb/azure/file_watch.clj
danmason: Agreed - this is much cleaner. May want to do similar for S3, though I'm sure there may be some other refinements/changes to make there anyway so can probably do a followup for that..

jarohen (approved):
cheers @danmason, good to go :slightly_smiling_face: 

* When you squash down the commits, could you just leave the commit messages that are relevant to the chosen approach?
* Could you add notes to the PR description about the other approach, and what blocked it?

cheers Dan!

PR #2674, opened 2023-08-08
Unify scan row pickers

### LINKED ISSUE #2662: "Smarter" temporal resolution for trie cursor

FiV0:
This PR tries to do all the temporal resolution via one algorithm. The algorithm works for the most general case (range/range) it essentially stops early (point/point) or does less work (range/point) for the less "heavy" use cases. @jarohen has come up with this idea. The picture below tries to give an idea of the algorithm. We are playing through the "events" of an iid in reverse system-time order. This means that when we have processed that event and it is not covered by anything before (in system time) it can immediately be written to the result set. 
   

signal-2023-08-08-131133_002

In the example above we first process e3. So we push out `{:value e3 :valid-time [1 4] : system-time [3 eot]}`. 
Next we process e2: For the valid-time `[3 4]` we have "ceiling" of 3 of system-time and  for valid-time `[4 6]` no system time constraints.
So we push out  `{:value e2 :valid-time [3 4] : system-time [2 3]}` and `{:value e2 :valid-time [4 6] : system-time [2 eot]}`.
Finally we process e1 where through the same process we obtain `{:value e1 :valid-time [2 3] : system-time [1 3]}` and `{:value e1 :valid-time [3 5] : system-time [1 2]}`. If any of the events match a point-point query we stop this process early. In case of a range/point query we are only filling in "holes" in the valid-time timeline and are not "overwriting" newer events with older ones on the valid-time timeline (updating the ceilings).

Todo:
- [x] ~Understand why update creates zero-width intervals~ See #2678
- [x] Comprehensive test.


mbutlerw (approved):

FiV0:
Closing in favor of #2686

PR #2685, opened 2023-08-10
Add google cloud file watching

### LINKED ISSUE #2677: Local filename caching setup for Google Cloud Object Store
- Similarly to the implementation within S3 (see #2633) and Azure (see #2665) - we want to set up local filename caching for list requests within the Google Cloud Object Store
- This will likely be quite similar to the others - we will need to make an initial list request and then react to all creation/deletion events in the bucket.
- Will need to research what equivalent to cloudformation google cloud uses (https://cloud.google.com/deployment-manager/docs/configuration/templates/create-basic-template)
- Seems we can publish operations on blobs to Google Pub/Sub: https://cloud.google.com/storage/docs/pubsub-notifications

danmason:
Resolves #2677 

Implements local filename caching for Google Cloud - alongside a template for all the related necessary infrastructure (TODO):
  - Google cloud object store now expects the `pubsub-topic` to be passed into config
  - We create a new subscription to the Pub/Sub topic
  - We then do our 'init' behaviour, which uses google cloud list-objects to get all current bucket contents.
  - We create a 'Subscriber' for the subscription - handling messages from the subscription. 
- Added a google deployment resource configuration file (see here https://cloud.google.com/deployment-manager/docs/configuration), using a configurable template:
  - Sets up the container 
  - Sets up the pubsub topic
  - Sets up the notifications to the pubsub topic, 
  - Setup a custom role with all relevant permissions for the above
- Update the README with setup info. 
- Updates to the google cloud test to ensure that they fit new listing behaviour, and all works as expected.
- Fix cleanup bug with subscriber/processor - ensure SubscriptionAdminClient cleaned up properly.

### DIFF
--- a/modules/azure/src/main/clojure/xtdb/azure/file_watch.clj
+++ b/modules/azure/src/main/clojure/xtdb/azure/file_watch.clj
@@ -64,7 +64,7 @@
                                                                                    file-url (:url msg-data)
                                                                                    file (when (string/starts-with? file-url base-file-url)
                                                                                           (subs file-url (count base-file-url)))]
-                                                                               (log/info (format "Message received, performing %s on file %s" event-type file))
+                                                                               (log/debug (format "Message received, performing %s on file %s" event-type file))
                                                                                (when (and event-type file)
                                                                                  (cond
                                                                                    (= event-type :create) (.add file-name-cache file)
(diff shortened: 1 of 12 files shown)


jarohen:
cleanup bug:
* throws a warning on close. intermittent. may have caused a test fail (check logs)

auth:
* uses a service account - local users have to assume the service account role

PR #2688, opened 2023-08-11
IID fast path

### LINKED ISSUE #2648: Scan fast-path for queries that filter by entity ID

FiV0:
This PR adds a fast path option for `scan` which contains an equal iid predicte (ie ` {xt/id (= xt/id #uuid "f83b7a34-ab34-47e4-ae98-ea8ec5339332")}`. We first find the leaves that potentially could contain that iid and then binary search the start of the iid runs in the leaves. We do a selection of the iid runs before passing them to the `LeafMergeQueue`. In this PR we are only doing literals, we will revisit for in params.

One potential improvement which is not done here is to not at all use a `LeafMergeQueue`. As the iid runs come from different tries we could essentially just consume them newest to oldest leaf. 

- [x] better boundaries between temporal resolution logic and trie walking
- [x] correctly check for iid predicates
- [x] how to do `ByteBuffer` and `ArrowBufPointer` comparison

### DIFF
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -17,10 +17,11 @@
             xtdb.watermark)
   (:import (clojure.lang IPersistentMap MapEntry)
            [java.lang AutoCloseable]
-           (java.util ArrayList Arrays Iterator LinkedList List ListIterator Map)
+           java.nio.ByteBuffer
+           (java.util ArrayList Iterator LinkedList List ListIterator Map)
            (java.util.function IntConsumer)
-           org.apache.arrow.memory.ArrowBuf
-           org.apache.arrow.memory.BufferAllocator
+           (java.util.stream IntStream)
+           (org.apache.arrow.memory ArrowBuf BufferAllocator)
            [org.apache.arrow.memory.util ArrowBufPointer]
            (org.apache.arrow.vector VectorLoader VectorSchemaRoot)
            xtdb.api.protocols.TransactionInstant
@@ -30,7 +31,7 @@
            (xtdb.metadata IMetadataManager)
            xtdb.object_store.ObjectStore
            xtdb.operator.IRelationSelector
-           (xtdb.trie ArrowHashTrie ArrowHashTrie$Leaf HashTrie LeafMergeQueue LeafMergeQueue$LeafPointer LiveHashTrie$Leaf)
+           (xtdb.trie ArrowHashTrie ArrowHashTrie$Leaf LeafMergeQueue LeafMergeQueue$LeafPointer LiveHashTrie$Leaf)
            (xtdb.vector IRelationWriter IRowCopier IVectorReader IVectorWriter RelationReader)
            (xtdb.watermark IWatermark IWatermarkSource Watermark)))
 
@@ -398,6 +399,31 @@
                                  system-from
                                  nil))))))))))
 
+(defn iid-selector [^ByteBuffer iid-bb]
+  (reify IRelationSelector
+    (select [_ allocator rel-rdr _params]
+      (with-open [arrow-buf (util/->arrow-buf-view allocator iid-bb)]
+        (let [iid-ptr (ArrowBufPointer. arrow-buf 0 (.capacity iid-bb))
+              ptr (ArrowBufPointer.)
+              iid-rdr (.readerForName rel-rdr "xt$iid")
+              value-count (.valueCount iid-rdr)]
+          (if (pos-int? value-count)
+            ;; lower-bound
+            (loop [left 0 right (dec value-count)]
+              (if (= left right)
+                (if (= iid-ptr (.getPointer iid-rdr left ptr))
+                  ;; upper bound
+                  (loop [right left]
+                    (if (or (>= right value-count) (not= iid-ptr (.getPointer iid-rdr right ptr)))
+                      (.toArray (IntStream/range left right))
+                      (recur (inc right))))
+                  (int-array 0))
+                (let [mid (quot (+ left right) 2)]
+                  (if (<= (.compareTo iid-ptr (.getPointer iid-rdr mid ptr)) 0)
+                    (recur left mid)
+                    (recur (inc mid) right)))))
+            (int-array 0)))))))
+
 (deftype TrieCursor [^BufferAllocator allocator, arrow-leaves, ^Iterator merge-tasks
                      col-names, ^Map col-preds, ^longs temporal-timestamps,
                      params, ^IPersistentMap picker-state]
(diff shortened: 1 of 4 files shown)


jarohen (approved):

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -398,6 +399,34 @@
                                  system-from
                                  nil))))))))))
 
+(def ^:private neg-or-zero-int? (complement pos-int?))
```
jarohen: this will box - worth inlining and using primitive operators

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -398,6 +399,34 @@
...
+
+(defn- iid-selector [^ByteBuffer iid-bb]
```
jarohen: I'd be tempted by a specific unit test for this function, like we did for the event-resolver - quite functional, and plenty of potential for an off-by-one here

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -398,6 +399,34 @@
...
+  (reify IRelationSelector
+    (select [_ _allocator rel-rdr _params]
+      (let [ptr (ArrowBufPointer.)
+            iid-rdr (.readerForName rel-rdr "xt$iid")
+            value-count (.valueCount iid-rdr)]
+        ;; lower-bound
+        (loop [left 0 right (dec value-count)]
```
jarohen: think this might be broken for an empty rel? worth a unit test.

i.e. left 0, right -1 then doesn't hit your `(= left right)` case to fall out of the loop.

also nit: don't mind this being on one line, but prefer a comma to visually separate the two bindings: `[left 0, right (dec value-count)]`

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -398,6 +399,34 @@
...
+          (if (= left right)
+            (if (zero? (util/compare-nio-buffers-unsigned iid-bb (util/arrow-buf-pointer->bb (.getPointer iid-rdr left ptr))))
+              ;; upper bound
```
jarohen: I don't know if it's worth binary searching again for the upper bound - if there are fewer than log2(N) entries per entity on average it'll be quicker to scan from the lower bound

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -398,6 +399,34 @@
...
+(defn- iid-selector [^ByteBuffer iid-bb]
+  (reify IRelationSelector
+    (select [_ _allocator rel-rdr _params]
+      (let [ptr (ArrowBufPointer.)
+            iid-rdr (.readerForName rel-rdr "xt$iid")
+            value-count (.valueCount iid-rdr)]
+        ;; lower-bound
+        (loop [left 0 right (dec value-count)]
+          (if (= left right)
+            (if (zero? (util/compare-nio-buffers-unsigned iid-bb (util/arrow-buf-pointer->bb (.getPointer iid-rdr left ptr))))
```
jarohen: I'd probably go the other way here - rather than turning each of the off-heap values into a byte-buf, you can turn the 'needle' IID into an ABP at the start of the function?

core/src/main/clojure/xtdb/trie.clj
```diff
@@ -176,3 +176,29 @@
                                              {:ordinal ordinal, :leaf node})))))}])))]
 
     (vec (trie-merge-tasks* (map #(some-> ^HashTrie % (.rootNode)) tries) []))))
+
+(defn- bucket-for [^bytes iid level]
+  (let [level-offset-bits (* HashTrie/LEVEL_BITS (inc level))
+        level-offset-bytes (/ (- level-offset-bits HashTrie/LEVEL_BITS) Byte/SIZE)]
+    (bit-and (bit-shift-right (get iid level-offset-bytes) (mod level-offset-bits Byte/SIZE)) HashTrie/LEVEL_MASK)))
+
+(defn iid-trie-merge-task [tries ^bytes iid]
```
jarohen, line 185: hoping that maybe when we reintroduce metadata this'll naturally Just Work™ as a metadata selector...?

(i.e. this is fine for now, just that it might be temporary)

core/src/main/clojure/xtdb/util.clj
```diff
@@ -117,6 +117,18 @@
     (.position bb 0)
     bb))
 
+(defn byte-buffer->uuid [^ByteBuffer bb]
+  (UUID. (.getLong bb 0) (.getLong bb 1)))
+
+(defn arrow-buf-pointer->bb [^ArrowBufPointer ptr]
```
jarohen, line 123: this might well go if you go for the iid-buf -> ABP suggestion above, but if not, IIRC you can call `(.nioBuffer (.getBuf ptr) (.getOffset ptr) (.getLength ptr))`

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -608,6 +648,15 @@
                                               (expr/->expression-relation-selector select-form {:col-types col-types, :param-types param-types})))
                            (into {}))
 
+            iid-bb (when-let [eid-select (or (get selects 'xt/id) (get selects 'xt$id))]
```
jarohen: if you could split this out into a function, you could perhaps test that individually rather than relying on the `with-redefs`?

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -608,6 +648,15 @@
...
+                     (let [eid (nth eid-select 2)]
+                       (when (and (= '= (first eid-select)) (s/valid? ::lp/value eid))
```
jarohen: this seems a little brittle - although I understand you may be waiting to see how the param/multiple IIDs plays out. I wonder whether it's worth us taking inspiration from how the metadata works - i.e. turning it into an expr and checking that, rather than as a form? :thinking:

core/src/main/clojure/xtdb/util.clj
jarohen: suspect this might go if you opt for the iid-buf -> ABP suggestion above; but, if not, IIRC this could be `(.nioBuffer (.getBuf ptr) (.getOffset ptr) (.getLength ptr))` - i.e. `nioBuffer` also has a 3-arg arity.

src/test/clojure/xtdb/operator/scan_test.clj
```diff
@@ -386,6 +387,60 @@
...
+  (let [before-uuid #uuid "00000000-0000-0000-0000-000000000000"
+        search-uuid #uuid "80000000-0000-0000-0000-000000000000"
+        after-uuid #uuid "f0000000-0000-0000-0000-000000000000"]
+    (with-open [node (node/start-node {})]
+      (xt/submit-tx node [[:put :xt-docs {:xt/id before-uuid :version 1}]
+                          [:put :xt-docs {:xt/id search-uuid :version 1}]
+                          [:put :xt-docs {:xt/id after-uuid :version 1}]])
+      (xt/submit-tx node [[:put :xt-docs {:xt/id search-uuid :version 2}]])
+
+      (with-redefs [trie/iid-trie-merge-task (fn [& _args] (throw (UnsupportedOperationException. "")))]
```
jarohen: as mentioned above, I might be tempted to check the col-pred fn rather than `with-redefs`...?

jarohen (changes requested):
thanks @FiV0! few suggestions below, but otherwise, looking forward to getting this merged :slightly_smiling_face:

PR #2712, opened 2023-08-22
Less Clojure stdlib in event-resolver

jarohen:
Now done in two independent passes: one to yield the output rows, one to update the state - means we can use fewer intermediate data structures (mostly Clojure lazy seqs).

I did try to combine them into one pass, but it essentially seemed like each Allen interval then required different treatment 😅

### DIFF
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -18,7 +18,7 @@
   (:import (clojure.lang IPersistentMap MapEntry)
            [java.lang AutoCloseable]
            java.nio.ByteBuffer
-           (java.util ArrayList Iterator LinkedList List ListIterator Map)
+           (java.util ArrayList Iterator LinkedList List Map)
            (java.util.function IntConsumer)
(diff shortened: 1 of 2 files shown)


core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -210,66 +202,74 @@
...
-  (^xtdb.operator.scan.EventResolver [range-range?] (event-resolver range-range? (LinkedList.)))
-
-  (^xtdb.operator.scan.EventResolver [range-range?, ^List !ranges]
-   (let [!overlapping-ranges (LinkedList.)]
-     (reify EventResolver
-       (nextIid [_]
-         (.clear !ranges))
-
-       (resolveEvent [_ idx valid-from valid-to system-from rc]
-         (assert (ranges-invariant !ranges))
```
FiV0: Could we keep this assertion? Gets compiled away in distributed code anyway, right?

jarohen: Fair, yep - although I'm not sure it would, given we distribute as source - would have to check.

PR #2715, opened 2023-08-23
Fix LeafMergeQueue for non-consecutive ordinals

### LINKED ISSUE #2714: Fix `LeafMergeQueue` when ordinals are not consecutive

FiV0:
A leaf merge task might contain ordinals (the log tries participating in the merge task)  that are not consecutive. Say for example log-tries with ordinals 1 and 3 are participating in the merge task. This would create issues with the indexing into the readers in the `LeafMergeQueue`. This PR addresses that.

### DIFF
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -570,10 +570,10 @@
                                                                      (.select rel-rdr (.select iid-pred allocator rel-rdr params))))))
                 merge-q (->merge-queue loaded-leaves merge-task)
                 ^"[Ljava.util.function.IntConsumer;"
-                row-pickers (into-array IntConsumer
-                                        (for [{:keys [rel-rdr]} loaded-leaves]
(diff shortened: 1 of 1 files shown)


core/src/main/java/xtdb/trie/LeafMergeQueue.java
```diff
@@ -35,31 +36,44 @@ public String toString() {
 
     private final byte[] path;
     private final IVectorReader[] rdrs;
+    private final HashMap<Integer, Integer> ordinalToIndex;
     private final PriorityQueue<LeafPointer> pq;
```
FiV0: Realized during Standup that we could also just maintain a map from <ordinal, ivectorreader> and that that is probably simpler.

FiV0: Addressed.

jarohen (changes requested):
Nice one @FiV0, good spot 👏

We probably don't want to be doing map lookups here if we can avoid it - would a sparse array work instead?

FiV0:
@jarohen which of the two map lookups? I am assuming the Clojure one.

PR #2722, opened 2023-08-25
Reinstate pushdown blooms

### LINKED ISSUE #2718: Add pushdown blooms back into new metadata support

FiV0:
This PR reinstates the pushdown blooms. There are also some changes to the join operators. We now create the RHS (probe side) lazily. The main reason for the move to lazy is so that `scan` can use the pushdown blooms when creating the cursor. We also need to pass down as state some dynamic vars which will no longer be bound when doing things lazily. The dynamic vars will be removed in a later step.

### DIFF
--- a/core/src/main/clojure/xtdb/operator/set.clj
+++ b/core/src/main/clojure/xtdb/operator/set.clj
@@ -289,8 +289,8 @@
     (util/try-close recursive-cursor)
     (util/try-close base-cursor)))
 
-(def ^:dynamic ^:private *relation-variable->col-types* {})
-(def ^:dynamic ^:private *relation-variable->cursor-factory* {})
+(def ^:dynamic *relation-variable->col-types* {})
+(def ^:dynamic *relation-variable->cursor-factory* {})
 
 (defmethod lp/emit-expr :relation [{:keys [relation]} _opts]
   (let [col-types (*relation-variable->col-types* relation)]
--- a/src/test/clojure/xtdb/operator/scan_test.clj
+++ b/src/test/clojure/xtdb/operator/scan_test.clj
@@ -12,7 +12,7 @@
            xtdb.operator.IRelationSelector
            (xtdb.operator.scan RowConsumer)))
 
-(t/use-fixtures :each tu/with-mock-clock tu/with-allocator)
+(t/use-fixtures :each tu/with-mock-clock tu/with-allocator tu/with-node)
 
 (t/deftest test-simple-scan
   (with-open [node (node/start-node {})]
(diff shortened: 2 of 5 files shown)


core/src/main/clojure/xtdb/logical_plan.clj
```diff
@@ -127,10 +127,7 @@
         {:keys [col-types ->cursor stats]} (f left-col-types right-col-types)]
```
FiV0: Should the `unery-expr` above also pass to the the style of the `binary-expr`?

src/test/clojure/xtdb/operator/scan_test.clj
```diff
@@ -610,3 +610,26 @@
...
+                           [:put :xt-docs {:xt/id :bar, :col "bar"}]])
+  (tu/finish-chunk! tu/*node*)
+  (xt/submit-tx tu/*node* [[:put :xt-docs {:xt/id :toto, :col "toto"}]])
+  (tu/finish-chunk! tu/*node*)
+
+  (let [!page-idxs-cnt (atom 0)
+        old-filter-trie-match scan/filter-trie-match]
+    (with-redefs [scan/filter-trie-match (fn [& args]
+                                           (let [res (apply old-filter-trie-match args)]
+                                             (prn res)
```
jarohen: cheeky prn :wink:

core/src/main/clojure/xtdb/operator/set.clj
```diff
@@ -99,8 +99,8 @@
   (lp/binary-expr (lp/emit-expr left args) (lp/emit-expr right args)
     (fn [left-col-types right-col-types]
       {:col-types (union-col-types left-col-types right-col-types)
-       :->cursor (fn [_opts left-cursor right-cursor]
-                   (UnionAllCursor. left-cursor right-cursor))})))
+       :->cursor (fn [_opts ->left-cursor ->right-cursor]
+                   (UnionAllCursor. (->left-cursor) (->right-cursor)))})))
```
jarohen: these should likely have `with-close-on-catch` too

core/src/main/clojure/xtdb/logical_plan.clj
jarohen: tbh, given how little `binary-expr` now does for join, I'd maybe have opted for join bypassing it entirely, so that the set operators can continue to use the current version.

context is that previously (before we had `with-close-on-catch`) this did way more - imagine inlining `with-close-on-catch` in all of the call-sites :sweat_smile:

PR #2731, opened 2023-08-28
IID fast path for non sequential in params

### LINKED ISSUE #2691: Scan fast-path for queries that filter by entity ID via in params

FiV0:
This PR adds faster IID lookup for in params when the params are not a collection or a relation. Collections and relations are joined with a scan. To push down these IID selections would require a lot more work which I think is currently not warranted.

### DIFF
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -623,11 +623,18 @@
                              :current-iid-ptr (ArrowBufPointer.)}
                       (at-point-point? scan-opts) (assoc :point-point? true))))))
 
-(defn selects->iid-byte-buffer ^ByteBuffer [selects]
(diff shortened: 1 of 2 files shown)


core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -623,11 +623,18 @@
...
-        (trie/->iid eid)))))
+      (when (= '= (first eid-select))
+        (cond
+          (s/valid? ::lp/value eid)
+          (trie/->iid eid)
+
+          (s/valid? ::lp/param eid)
+          (let [eid-rdr (.readerForName params-rel (name eid))]
+            (when (= 1 (.valueCount eid-rdr))
+              (trie/->iid (.getObject eid-rdr 0)))))))))
```
FiV0, line 637: Still quite brittle here. I could also make this work for multiple IIDs here, but it seemed like a lot more work.

jarohen (approved):

Issue #1916, opened 2023-03-10
Lucene indexing should not halt


refset:
An underlying `FileLock` error is happening (quite possibly due to non-JVM OS activity) but it is causing an error that can't be recovered from without manually deleting the lock file and restarting the node.

This symptom only surfaced _after_ upgrading from `1.20.0` to `1.22.2-rc1` (and there was a minor Lucene bump in the intervening changes).

For more information and speculation about causes, see https://juxt-oss.zulipchat.com/#narrow/stream/194466-xtdb-users/topic/Lucene.20indexer.20fail.20leads.20to.20need.20to.20fully.20reindex

```
2023-02-16 22:49:46.047+0000 [WARN ][xtdb-lucene-refresh-1][] xtdb.lucene: error during Lucene IndexWriter commit
org.apache.lucene.store.AlreadyClosedException: this IndexWriter is closed
    at org.apache.lucene.index.IndexWriter.ensureOpen(IndexWriter.java:877)
    at org.apache.lucene.index.IndexWriter.ensureOpen(IndexWriter.java:891)
    at org.apache.lucene.index.IndexWriter.nrtIsCurrent(IndexWriter.java:5241)
    at org.apache.lucene.index.StandardDirectoryReader.doOpenFromWriter(StandardDirectoryReader.java:289)
    at org.apache.lucene.index.StandardDirectoryReader.doOpenIfChanged(StandardDirectoryReader.java:268)
    at org.apache.lucene.index.StandardDirectoryReader.doOpenIfChanged(StandardDirectoryReader.java:258)
    at org.apache.lucene.index.DirectoryReader.openIfChanged(DirectoryReader.java:173)
    at org.apache.lucene.search.SearcherManager.refreshIfNeeded(SearcherManager.java:156)
    at org.apache.lucene.search.SearcherManager.refreshIfNeeded(SearcherManager.java:58)
    at org.apache.lucene.search.ReferenceManager.doMaybeRefresh(ReferenceManager.java:176)
    at org.apache.lucene.search.ReferenceManager.maybeRefresh(ReferenceManager.java:225)
    at xtdb.lucene$refresh.invokeStatic(lucene.clj:319)
    at xtdb.lucene$refresh.invoke(lucene.clj:318)
    at xtdb.lucene$refresh_loop$fn__291867.invoke(lucene.clj:327)
    at xtdb.lucene$refresh_loop.invokeStatic(lucene.clj:325)
[...]

PR #2743, opened 2023-08-31
Add :log-iw option to the lucene store

wotbrew:
Provides to the IndexWriter an instruction to log `IW` InfoStream messages (with the supplied log-level)

Hope is this will permit further debugging of #1916


wotbrew:
Goal will be to see this message: `infoStream.message("IW", "hit tragic " + tragedy.getClass().getSimpleName() + " inside " + location);` in onTragicEvent

This will help provide evidence for the 'error causing close' hypothesis

jarohen (approved):

PR #2746, opened 2023-08-31
Add page-idx-pred to merge-plan and fix it

FiV0:
Not an ideal solution, but I didn't like the idea that absence (`nil`) indicates taking the page. Happy to do it differently as well.


core/src/main/clojure/xtdb/compactor.clj
```diff
@@ -71,6 +72,8 @@
 
           (.end trie-wtr))))))
 
+(def ^:private constant-all-page-bitmap (RoaringBitmap/flip (RoaringBitmap.) 0 (bit-shift-left 1 16)))
```
FiV0, line 75: My understanding of how many pages there ever can be in a trie. Not sure if that holds.

Issue #2736, opened 2023-08-30
Prevent double-loading of trie buffers on scan?


jarohen:
I think we load trie-file buffers twice during a scan - once to check their metadata (to get `trie-file->page-idxs` in scan), and then once to calculate the merge plan (`trie/table-merge-plan`)

Haven't measured its impact on performance, but guessing it's not optimal 🙂

FiV0:
Also worth rethinking if we don't want some toplevel metadata per trie to check if need to load the trie at all.

FiV0:
Completed in #2753.

PR #2747, opened 2023-08-31
Revert #2726

FiV0:
Just depends on #2746. Can o/w go in immediately.


FiV0:
Converting this back to draft as I think it's best to rethink this one in the context of filtering before checking metadata and also #2736.

FiV0:
Closing in favor of #2753 or some change in a similar direction.

Issue #2752, opened 2023-09-01
Is the MetadataManager still warranted?


FiV0:
With the new index structure the metadata is tied to the trie files which need to be loaded anyway when doing a scan. This means the boundary between metadata loading and data loading becomes fuzzy. It's unlikely that the metadata manager then manages the buffers independently of scan. That then begs the question if we should just have a metadata ns that helps us with obtaining the metadata, but does no longer contain a stateful metadata manager object. The only purpose such an object would still serve is the caching of metadata.

jarohen:
This does maintain a cache now, so it lives 🙂

PR #2753, opened 2023-09-01
Reorganize trie loading

### LINKED ISSUE #2736: Prevent double-loading of trie buffers on scan?

FiV0:
This is solves #2736 and puts iid page selection in front of metadata checking. We are also slimming down the MetadataManager (related ticket #2752) by essentially it mainly acting as a cache. The metadata checking also gets moved to `->merge-plan`. We are mainly trying to delay any checks so we can decide whether or not we actually need to do a check. 
The iid fast path now works as follows:
- We get an iid from the query
- We construct a path predicate 
- When calculating the merge plan we first check the path predicate to prune the trie as much as possible.
- If we get to a leaf we do the metadata checks (reason for for delaying the metadata checks).
- If everything passes one gets one merge-task.

The general case is similar except that there is no pruning happening and the `path-predicate` will always return true.

TODO:
- [ ] Worth checking if for non-selective scans this makes things slower.

### DIFF
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -19,11 +19,10 @@
   (:import (clojure.lang IPersistentMap MapEntry)
            java.nio.ByteBuffer
            (java.util Iterator LinkedList List Map)
-           (java.util.function IntConsumer)
+           (java.util.function IntConsumer IntPredicate)
            (java.util.stream IntStream)
            (org.apache.arrow.memory BufferAllocator)
            [org.apache.arrow.memory.util ArrowBufPointer]
-           [org.roaringbitmap RoaringBitmap]
            [org.roaringbitmap.buffer MutableRoaringBitmap]
            xtdb.api.protocols.TransactionInstant
            xtdb.buffer_pool.IBufferPool
@@ -31,7 +30,7 @@
            (xtdb.metadata IMetadataManager ITableMetadata)
            xtdb.object_store.ObjectStore
            xtdb.operator.IRelationSelector
-           (xtdb.trie LeafMergeQueue$LeafPointer)
+           (xtdb.trie HashTrie LeafMergeQueue$LeafPointer)
            (xtdb.vector IRelationWriter IRowCopier IVectorReader IVectorWriter RelationReader)
            (xtdb.watermark ILiveTableWatermark IWatermark IWatermarkSource Watermark)))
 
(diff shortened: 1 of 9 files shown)


core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -62,29 +61,24 @@
...
-    page-idxs))
+    (let [^ITableMetadata table-metadata (.tableMetadata metadata-manager trie-rdr trie-file)
+          metadata-rdr (.metadataReader table-metadata)
+          bloom-rdr (-> (.structKeyReader metadata-rdr "columns")
+                        (.listElementReader)
+                        (.structKeyReader "bloom"))]
+      (reify IntPredicate
+        (test [_ page-idx]
+          (boolean
+           (when (.test page-idx-pred page-idx)
```
FiV0: I don't know if this "predicate stacking" has any performance implications.

jarohen: I'd guess not - if it's used often enough chances are the JVM will be able to inline it. As always, though, see what the profiler says :slightly_smiling_face:

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -62,29 +61,24 @@
 
 (def ^:dynamic *column->pushdown-bloom* {})
 
-(defn- filter-pushdown-bloom-page-idxs [^IMetadataManager metadata-manager buf-key ^String col-name ^RoaringBitmap page-idxs]
+(defn filter-pushdown-bloom-page-idx-pred ^IntPredicate [^IMetadataManager metadata-manager ^String col-name
+                                                         {:keys [trie-file ^RelationReader trie-rdr] :as _trie-match} ^IntPredicate page-idx-pred]
   (if-let [^MutableRoaringBitmap pushdown-bloom (get *column->pushdown-bloom* (symbol col-name))]
     ;; would prefer this `^long` to be on the param but can only have 4 params in a primitive hinted function in Clojure
```
jarohen: does this comment apply any more?

core/src/main/clojure/xtdb/operator/scan.clj
jarohen: oh, I see, it's not just two predicates, it's used within a `reduce`. same probably still applies unless we're talking hundreds?

jarohen: if you're concerned, rather than a reduce here, I'd create an IntPredicate for each (i.e. map) and then write a single 'and' intpredicate (if it doesn't already exist) which tests all of them in order (short-circuiting if possible)

jarohen (approved):
nice one @FiV0, cheers! :pray:

core/src/main/clojure/xtdb/operator/scan.clj
FiV0: Good point. `IntPredicate` has a `(.and ...)` method.

PR #2774, opened 2023-09-15
Permit loading of trie leaf buffers batch at time

### LINKED ISSUE #2694: merge set storage iterator integrated with scan
Following up from: #2660 

Try to get a PR together proposing a scan integration with a merge set iterator. 

The iterator itself and the underlying algorithm can be slow to start with, should be seen as a first pass

wotbrew:
Provides (naive for now) buffer pool functions to get a record batch and arrow footer. These are integrated into scan such that batches can later be loaded using range requests by the buffer pool.

Note:

- For now, the ArrowLoaders still exist for compaction.
- LeafPointers are still used for the algorithmic sorting but their scope is limited they are constructed by the MergeQueue when it wraps the task readers
- there is no explicit iterator / spliterator, it turns out the memory management angle is not as expected due to management being offloaded to a cursor shared VectorSchemaRoot cache.

Surprisingly early testing shows no degradation in performance

TPCH  0.05

| Test | PR2774 | 2.X |
| - | - | - |
| HOT | 115682.257417ms | 133288.293041ms |
| COLD | 117183.607208ms | 135290.006541ms | 

Resolves #2694

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="ExternalStorageConfigurationManager" enabled="true" />
+  <component name="FrameworkDetectionExcludesConfiguration">
+    <file type="web" url="file://$PROJECT_DIR$" />
+  </component>
   <component name="ProjectRootManager" version="2" languageLevel="JDK_17" default="true" project-jdk-name="java-17-openjdk" project-jdk-type="JavaSDK" />
 </project>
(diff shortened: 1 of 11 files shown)


jarohen (commented):
hey @wotbrew, it's not really clear to me what you're aiming for with this refactoring PR...? guessing it might be clearer as part of the wider changes, i.e. with ordinals/leaf-ptrs/leaf-loaders removed. will have another read through though.

btw feel free to pop some of the contained changes straight in as distinct commits btw, that might make the PR more focused and hence easier to read - e.g. the child allocator change (:+1:), the compactor fix, xray

FiV0 (approved):
LGTM

PR #2784, opened 2023-09-20
LocalObjectStore interface

### LINKED ISSUE #2772: Intermittent issue with slow buffer pool/object store
I only ran into this a couple of times but it sometimes seems the object store seems to hang for a while on some calls. Queries that normally take milliseconds all of sudden take seconds.
Selection_017

FiV0:
This resolves #2772. The problem is that the object store interface was created with a remote object store in mind. So that a call like `(.getObject object-store k cache-path)` would actually copy the object from remote storage to the machine. When working with the file-system object store this just does unnecessary copies locally which show up in the profiler. The PR tries to avoid the unnecessary copying by introducing a temporary solution for the case described above. @wotbrew mentioned that these interfaces will change significantly with his IO work and that the case of a local file system object store might then completely be dealt with by the buffer-pool.

### DIFF
--- a/core/src/main/clojure/xtdb/buffer_pool.clj
+++ b/core/src/main/clojure/xtdb/buffer_pool.clj
@@ -22,10 +22,12 @@
   (^java.util.concurrent.CompletableFuture getRangeBuffer [^String k ^int start ^int len])
   (^boolean evictBuffer [^String k]))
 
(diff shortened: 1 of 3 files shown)


FiV0:
Closing in favor of #2787.

PR #2821, opened 2023-10-09
Locally cache file names against XTDB object

### LINKED ISSUE #2783: Incorrect query results using S3 object store
It appears using the s3 object store causes incorrect results until the node is restarted

Same code works as expected on an in-memory object store 

Something to do with the file watching?

--- 

Repro:

```clojure
  (def node
    (node/start-node {:xtdb.s3/object-store {:bucket "xtdb-s3-test",
                                             :prefix (str "test-" (random-uuid))
                                             :sns-topic-arn "arn:aws:sns:eu-west-1:199686536682:xtdb-object-store-iam-test-bucket-events"
                                             :configurator (reify S3Configurator)}}))

  (doseq [tx (->> (for [i (range 1e6)] [:put :ints {:xt/id i, :n i}])
                  (partition-all 512))]
    (xt/submit-tx node tx))
  
  ;; should be 100k, observed: 78,400
  (time (count (xt/q node "select * from ints")))
```

Adding a finish chunk makes things more interesting: 

```clojure 
  (def node
    (node/start-node {:xtdb.s3/object-store {:bucket "xtdb-s3-test",
                                             :prefix (str "test-" (random-uuid))
                                             :sns-topic-arn "arn:aws:sns:eu-west-1:199686536682:xtdb-object-store-iam-test-bucket-events"
                                             :configurator (reify S3Configurator)}}))

  (doseq [tx (->> (for [i (range 1e6)] [:put :ints {:xt/id i, :n i}])
                  (partition-all 512))]
    (xt/submit-tx node tx))

  (tu/finish-chunk! node)

  ;; now returns 0, e.g finish-chunk! stops me from seeing any of the data
  (time (count (xt/q node "select * from ints")))
```
If the node is restarted after finish-chunk! however, the correct query results are returned. I think the re-initialisation of the file listing fixes the problem.

danmason:
Addresses #2783 

# In this PR

Removes the lag between writes/reads of file names on remote object stores by writing to the local filename cache as soon as we PUT/DELETE an object. This will ensure the filename is available locally on the java object immediately.

Done in this PR:
- Add a set of notes to the dev README to do auth for tests on remote object store implementations.
- Add/remove to cache within put/delete calls of remote object stores:
  - Done for S3 object store
  - Done for Azure object store
  - Done for Google Cloud object store.
- Update object store tests to remove unnecessary calls to wait (keeping it when testing with multiple object stores - ie, where the filename list will be impacted by the file watching)
  - Updated S3 tests with this behaviour.
  - Update Azure tests with this behaviour.
  - Update Google Cloud tests with this behaviour.

## Concerns around operation duplication

The current way of handling this re-uses the existing filename cache to add/remove files. How this works:
- When an object is put on/deleted from the remote object store, it will firstly be added/removed to the local filename cache.
- The object will be put on/deleted from the remote object store - this will trigger a notification to be sent which will be handled by the individual file watchers.
- The file watcher will add/remove from the local filename cache upon receipt of a notification.

What this means is that there will be duplication of adds/removes for a local object - ie, the call to **add** to the local filename cache will happen twice - first when the call to PUT is made, secondly when the notification for the PUT operation is handled by the file watcher. The same will happen for removals. 
- Generally speaking, this will not do anything - ie, adding a filename twice will be a no-op on the second call.
- There is, however, a certain side case:
  - If a file is **removed** from the local cache before the remote PUT operation is handled, there may be a case where a file is removed from the local filename cache and then re-added.
  - This would then get removed from the filename cache as the DELETE operation propagates through the system.
- This may well be an acceptable situation? By my understanding, it would be pretty odd to immediately remove an object after writing it. \In general, it would be good to understand when an object gets removed. Usual handling of remote operations should be fairly quick, so this would likely be in the magnitude of a few seconds? It is, however, one to keep in mind.

### DIFF
--- a/dev/README.adoc
+++ b/dev/README.adoc
@@ -45,6 +45,37 @@ You do not need to install Gradle to develop XT2 - there is a Gradle https://doc
 ** `./gradlew jdbc-test` / `./gradlew kafka-test`
 ** `docker-compose down`
 
+=== Auth for testing cloud based resources 
+
+.*S3* (ensure all of the below are done on a consistent region):
+* Need to ensure the `s3-stack` is setup via CloudFormation
+** See the README for more info here.
+* Need to log in to the AWS CLI - using `aws configure sso` 
+* Need to assume the role on the CLI - `aws sts assume-role --role-arn <ARN of XTDBIamRole> --role-session-name <session-name> --profile <SSO profile>`
+* Set `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` & `AWS_SESSION_TOKEN` from the output 
+  of the above, then set AWS_REGION=<region for the cloudformation>
+* Start the Gradle REPL and connect to it, to have all of the AWS creds available.
+
+.*Azure*
+* Ensure the Azure stack, `azure-resource-manager/azure-stack.json` stack is setup via the Azure deployment manager.
+** See the README for more info here.
+* Ensure a user/app registration is created with the create `xtdb-custom-role`.
+* Create an access token for said user/app registration.
+* Set `AZURE_CLIENT_ID` & `AZURE_CLIENT_SECRET` from the created access token. 
+* Set `AZURE_TENANT_ID` based on the tenant id on which you created the user/app registration/  
+* Set `AZURE_SUBSCRIPTION_ID` based on whichever subscription you created the stack on.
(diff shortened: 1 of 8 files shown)


jarohen (approved):
Cheers @danmason, lgtm!

As mentioned on Slack, we're not likely to GC files on the obj-store without a long grace period, so I'd be prepared to accept this edge case :slightly_smiling_face:

PR #2822, opened 2023-10-09
Scan to start of iids relevant for specific merge task

### LINKED ISSUE #2185: CI builds on forks shouldn't require our Slack credentials
We're only really interested in Slack notifications for when xtdb/core2 master fails - any other time, chances are, we're looking out for it (e.g. PRs get a status tick).

### LINKED ISSUE #2815: Larger table scans return duplicate rows
Discovered during s3 testing, but happens for any object store / log.

Smaller chunk sizes causes more duplicates, if data is ingested in larger transactions e.g 512 will not always return duplicates.

```clojure
  (with-open [node (node/start-node {})]
    (->> (for [i (range 1e6)]
           [:put :ints {:xt/id (hash i)}])
         (partition-all 32)
         (run! #(xt/submit-tx node %)))
    (count (xt/q node "select ints.xt$id from ints")))
=> 1050692
```

FiV0:
Fixes #2815. Does a binary search for the start of a potential IID run (via the `path` of a merge task) on ScanDataRowPointer initialization.

### DIFF
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -280,10 +280,8 @@
                                                      iid-pred (.select (.select iid-pred allocator data-rdr params)))
                           data-row-ptr (ScanDataRowPointer. leaf-rdr
                                                             (-> (copy-row-consumer out-rel leaf-rdr col-names)
-                                                                (wrap-temporal-bounds temporal-bounds)))]]
-              (while (and (< (.getIndex data-row-ptr) (.rowCount data-row-ptr))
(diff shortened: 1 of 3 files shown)


jarohen (approved):
Looks good :slightly_smiling_face: just a couple of things:
* looks like this does a binary search but the PR description says it doesn't? (the good way around, at least :smile:)
* might be performance bikeshedding, but we could save ourselves the binary search if the first row is after the path (I'd assume quite a common case)? maybe worth a benchmark at a later date...

in any event, feel free to merge as-is without further review :slightly_smiling_face:

PR #2829, opened 2023-10-11
Use mergeSort for logs instead of compacting the whole trie in scan

### LINKED ISSUE #2735: Scan should sort-merge logs rather than calling compactLogs
Currently scan compacts the entire live trie before reading to ensure all data is moved from log -> data and is in IID -> sys-time order. 

Compacting logs may also create extra levels etc. all of which is wasted work.

It may be more efficient for scan to simply sort-merge each leafs data and log fields as it accesses them.

FiV0:
This merge sorts the logs into the remaining data on per leaf basis. Should preform better in most cases but especially for sparse queries. One could potentially think about ways to only sort this data once and make use of that work for subsequent queries or the next compacting call in the indexer.

### DIFF
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -257,8 +257,7 @@
           (.load loader rb)
           (vr/<-root vsr))))
 
-    :live (let [{:keys [^RelationReader rel-rdr, ^LiveHashTrie$Leaf node]} leaf-arg]
-            (.select ^RelationReader rel-rdr (.data ^LiveHashTrie$Leaf node)))))
+    :live (:rel-rdr leaf-arg)))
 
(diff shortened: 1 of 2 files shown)


core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -379,7 +378,11 @@
                                                                    :trie-key trie-key}]))))
 
                                   LiveHashTrie$Leaf
-                                  (recur true (conj leaves [:live {:node trie-node, :rel-rdr (.liveRelation live-table-wm)}])))
+                                  (recur true (conj leaves
+                                                    [:live {:rel-rdr
+                                                            (->>
```
jarohen: minor style: I probably wouldn't thread here - the thing you're threading isn't the thing being transformed

jarohen (approved):

PR #2832, opened 2023-10-12
Add functions for multipart upload for S3 object store

### LINKED ISSUE #2823: Support multipart uploads on the S3 Object Store
To help address handling of large blocks/files (see #2814), we want to chunk our larger uploads on the object store.

Within S3, we can upload large files using [**Multipart uploads**](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html). 

Add a set of multipart related functions to the S3 object store that can be used elsewhere in the system.

danmason:
Resolves #2823 

## In this PR

- Add a new interface to the object store namespace, `IMultipartUpload`. This has three functions:
  - `uploadPart` to send a new part to the multipart upload.
  - `complete` to finish a mulitpart upload and verify all parts have been submitted.
  - `abort` to cancel a multipart upload. 
- Add a new method to the `ObjectStore` protocol:
  - `startMultipart`, which initiates a multipart upload and returns an instance of `IMultipartUpload`
- Implement the above within the `s3` object store:
  - `startMultipart` calls to initiate a multipart upload in S3, returns the upload id and creates an s3 defrecord `MultipartUpload` which implements `IMultipartUpload`.
  - Implements all of the functions of IMultipartUpload using corresponding functions on the S3 API - completed parts being stored/added to an `ArrayList` and part number being derived from the number of items in the list.
  - On completing an upload, calls a `complete-fn` passed from the object store. This adds the initial key to the filename cache. 
- Also adds a new field to the S3 object store -  `multipart-minimum-part-size`. This might be useful for callers to decide part boundaries.
- Test all of the above for S3 - starting a multipart, aborting a multipart, uploading parts, verifying parts are uploaded as expected.

### DIFF
--- a/modules/s3/src/main/clojure/xtdb/s3.clj
+++ b/modules/s3/src/main/clojure/xtdb/s3.clj
@@ -1,26 +1,23 @@
 (ns xtdb.s3
   (:require [clojure.spec.alpha :as s]
             [clojure.string :as string]
-            [xtdb.object-store :as os]
-            [xtdb.util :as util]
+            [juxt.clojars-mirrors.integrant.core :as ig]
             [xtdb.file-list :as file-list]
+            [xtdb.object-store :as os]
             [xtdb.s3.file-list :as s3-file-watch]
-            [juxt.clojars-mirrors.integrant.core :as ig])
-  (:import [java.lang AutoCloseable]
-           [java.io Closeable]
+            [xtdb.util :as util])
+  (:import [java.io Closeable]
+           [java.lang AutoCloseable]
            [java.nio ByteBuffer]
-           [java.util NavigableSet]
+           [java.util ArrayList List NavigableSet]
            [java.util.concurrent CompletableFuture ConcurrentSkipListSet]
            [java.util.function Function]
            [software.amazon.awssdk.core ResponseBytes]
            [software.amazon.awssdk.core.async AsyncRequestBody AsyncResponseTransformer]
-           [software.amazon.awssdk.services.s3.model DeleteObjectRequest GetObjectRequest HeadObjectRequest NoSuchKeyException PutObjectRequest]
            [software.amazon.awssdk.services.s3 S3AsyncClient]
-           [xtdb.object_store ObjectStore]
+           [software.amazon.awssdk.services.s3.model AbortMultipartUploadRequest CompleteMultipartUploadRequest CompletedPart CompletedMultipartUpload CreateMultipartUploadRequest CreateMultipartUploadResponse DeleteObjectRequest GetObjectRequest HeadObjectRequest NoSuchKeyException PutObjectRequest UploadPartRequest UploadPartResponse]
+           [xtdb.object_store ObjectStore IMultipartUpload]
            [xtdb.s3 S3Configurator]))
-
-;; TODO:
-;; Need to consider - local cache vs file watcher, dont want to re-add/re-remove the same file?
 
 (defn- get-obj-req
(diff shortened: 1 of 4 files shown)


modules/s3/src/main/clojure/xtdb/s3.clj
```diff
@@ -48,6 +46,72 @@
...
+
+(defn multipart-upload
+  [{:keys [^S3AsyncClient client bucket prefix]} k ^ByteBuffer buf]
+  (let [prefixed-key (str prefix k)
+        initiate-request (-> (CreateMultipartUploadRequest/builder)
+                             (.bucket bucket)
+                             (.key prefixed-key)
+                             ^CreateMultipartUploadRequest (.build))]
+    (-> (.createMultipartUpload client initiate-request)
+        (util/then-compose (fn [^CreateMultipartUploadResponse initiate-response]
```
danmason: Thinking we could potentially include some try logic in this then-compose block which, on an exception, calls to abortMultipartUpload?

danmason: It does seem like multipart operations do not automatically get cleaned up - thats a bucket level config: https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html

We could probably add that into our S3 stack - but might be better to allow XTDB to cleanup failed multipart uploads itself.

wotbrew: Yes I wasn't aware it was opt-in, I thought it was opt-out.  I agree if we are not a bring-your-own-bucket database and you must configure it using cloudformation (or replicate our template manually) then I think we can include a 24hr rule.

Even with the rule I think it'd be worth adding interrupt/exception aborts if it is quite easy to do so.

modules/s3/src/main/clojure/xtdb/s3.clj
```diff
@@ -48,6 +46,72 @@
...
+                                                                            buf-limit (min byte-buffer-size
+                                                                                           (+ current-pos chunk-size))
+                                                                            chunk (-> buf-dup
+                                                                                      (.position ^Integer buf-position)
+                                                                                      (.limit ^Integer buf-limit))
+                                                                            content-length (- buf-limit buf-position)
+                                                                            _ (log/debug (str "Uploading part number: " part-number))
+                                                                            _ (log/debug (str "Buffer chunk: " chunk))
+                                                                        ;; Waits for upload part to finish before starting next one
+                                                                            ^UploadPartResponse upload-part-response (.join
```
wotbrew: I think we need to get rid of the .join here to avoid thread starvation. I am not sure exactly how the async IO in the SDK is achieved but it is possible we will choke an IO thread pool or a default forkjoin pool depending on how system is configured. We probably want to do the coordination using thenCompose/thenComposeAsync ensuring our function returns a future. Lets talk this through? I might have it wrong somehow.

Basically we want to chain in the cpu bound work in thenCompose, thenApply blocks etc

I realise this breaks the one-at-a-time sequencing, but we ought to be able to recover that with some future api methods.

wotbrew: Like if we block the .putObject thread that is less of a problem, though we probably still should not as the contract with callers is currently non blocking. I think I need to sanity check my understanding of CompletableFuture.thenCompose

wotbrew: I think thenCompose fns are pushed into the future if it has not completed and the pool will run it.

jarohen (approved):

PR #2834, opened 2023-10-16
WIP: Add object-store.arrow-writer

wotbrew:
Mostly speculative while awaiting #2832


wotbrew:
Used a file based approach in #2848

PR #2836, opened 2023-10-16
Splitting the buffer-pool into in-memory, local and remote

jarohen:
The algorithm for buffer-pooling differs significantly depending on whether the object are stored in-memory, local or remote - we've already seen issues like #2772 where treating a local object-store (OS) like a remote object-store leads to performance issues in the former.

This PR splits the buffer pool (BP) into in-memory, local and remote implementations. 

* The in-memory essentially inlines the in-memory OS (with the latter being relegated to a test fixture now); local does the same but with the file-system store, and the current BP implementation becomes the remote one. 
* The remote BP is the only one that requires an object-store implementation. i.e. the user configures one of the following:
  * in-memory BP (default)
  * local BP (specifying a path)
  * remote BP (additionally configuring an object store)
* All put/list requests now go to the BP - nothing hits the OS directly apart from the BP.
* I removed `evictBuffer` as the behaviour would now be unclear between the various BPs (and nothing was calling it anyway). At some point we'll want to reintroduce this, but at that point I'd like to be clear that it's effectively the same as `deleteObject` (i.e. irretrievable).

### DIFF
--- a/core/src/main/clojure/xtdb/object_store.clj
+++ b/core/src/main/clojure/xtdb/object_store.clj
@@ -59,25 +59,56 @@
 (defn obj-missing-exception [k]
   (IllegalStateException. (format "Object '%s' doesn't exist." k)))
 
+(defn get-path ^java.nio.ByteBuffer [^Path root-path, ^String k]
+  (let [from-path (.resolve root-path k)]
+    (when-not (util/path-exists from-path)
+      (throw (obj-missing-exception k)))
+
+    (util/->mmap-path from-path)))
+
+(defn get-path-range ^java.nio.ByteBuffer [^Path root-path, ^String k, ^long start, ^long len]
+  (ensure-shared-range-oob-behaviour start len)
+
+  (let [from-path (.resolve root-path k)]
+    (when-not (util/path-exists from-path)
+      (throw (obj-missing-exception k)))
+
+    (with-open [in (util/->file-channel from-path #{:read})]
+      (.map in FileChannel$MapMode/READ_ONLY start (max 1 (min (- (.size in) start) len))))))
+
+(defn put-path [^Path root-path, ^String k, ^ByteBuffer buf]
+  (let [buf (.duplicate buf)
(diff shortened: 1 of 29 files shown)


core/src/main/clojure/xtdb/object_store.clj
```diff
@@ -61,102 +59,56 @@
...
-
-  )
+  ([^Path root-path, ^String dir]
+   (let [dir (.resolve root-path dir)]
+     (when (Files/exists dir (make-array LinkOption 0))
+       (with-open [dir-stream (Files/newDirectoryStream dir)]
+         (vec (sort (for [^Path path dir-stream]
+                      (str (.relativize root-path path))))))))))
 
 (deftype FileSystemObjectStore [^Path root-path, ^ExecutorService pool]
```
danmason, line 105: Presumably this is now our remote filesystem store implementation (thus why we're using all the calls to `Files` and whatnot)?.

src/test/clojure/xtdb/object_store_test.clj
```diff
@@ -35,9 +39,57 @@
 (defn get-bytes [^ObjectStore obj-store k start len]
   (byte-buf->bytes @(.getObjectRange obj-store k start len)))
 
-(defn in-memory ^Closeable []
-  (->> (ig/prep-key ::os/memory-object-store {})
-       (ig/init-key ::os/memory-object-store)))
+(deftype InMemoryObjectStore [^NavigableMap os]
```
danmason, line 42: Is there much reason to implement this? I'm assuming it would be here if there were some references to the  in memory object store in os-test but feels like they should be cleaned up?

danmason:
Couple of comments/questions from me, but otherwise looking great!

src/test/clojure/xtdb/object_store_test.clj
jarohen: yep - IIRC main user for this would be the remote buffer pool test, to check that it's actually writing the objects down correctly

PR #2839, opened 2023-10-18
Add multipart support to Azure Object Store

### LINKED ISSUE #2838: Support multipart/block uploads on the Azure Object Store
To help address handling of large blocks/files (see https://github.com/xtdb/xtdb/issues/2814), we want to chunk our larger uploads where possible.

To this end, we've added an optional interface for object stores, `SupportsMultipart` - this contains one method, `startMultipart`, that returns a `IMultipartUpload` implementing object that can upload parts, complete/abort the multipart upload. We wish to implement this with Azure functions

### Implementing in Azure

For Azure - there is the BlockBlobClient, the process of uploading in parts/blocks (from https://github.com/Azure/azure-storage-php/issues/170#issuecomment-458385176)

>   For Azure Storage, there are 2 steps:
> 
>   - Do not need to initialize a multipart uploading
>   - Upload block (like S3's part) to a blockblob with the blockID
>    - Commit the blocklist to save the blockblob with the blockIDs

Looking at the Java API, we create a BlockBlobClient, can call to upload, which will upload an individual block, or 'part', and then when we finish/complete a block we commit the list of blocks. Seems pretty analogous to what we have - I think entirely possible to treat it the same. (Initializing becomes 'create the block blob client for the given blob')

In essence, the implementation from the Azure ObjectStore's perspective:
- Implement `SupportsMultipart` interface - `startMultipart` will create a BlockBlobClient and pass back a `MultipartUpload` object that implements `IMultipartUpload` 
  - `uploadpart` will upload a block to the block blob.
  - `complete` will commit the blocklist
  - `abort` will cancel/delete the block blob.

danmason:
Resolves #2838 

## In this PR

Implements `SupportsMultipart` interface in the Azure Object store - using [**Azure's BlockBlobs**](https://learn.microsoft.com/en-us/rest/api/storageservices/understanding-block-blobs--append-blobs--and-page-blobs#about-block-blobs)
- Implement the interface within the `azure` object store:
  - `startMultipart` calls to create a BlockBlobClient, which is passed into the azure defrecord of `MultipartUpload` (which implements `IMultipartUpload`).
  - Multipart upload acts on the BlockBlobClient:
    - `uploadPart` stages a new block at a Base64 ID made from the 'part number', and stores the staged blocks ids in a list
    - `complete` calls to `commitBlockList` on the object with the above IDs, which will write out a file consisting of the blocks onto Azure.
    - `abort` calls to `commitBlockList` on the object with an empty list - creating an empty object and removing all staged blocks. We then delete the subsequent file (if it exists).   
  - On completing an upload, calls a `complete-fn` passed from the object store. This adds the initial key to the filename cache. 
- Adds the `multipart-minimum-part-size` field to the object store -  useful for callers to decide part boundaries.
  - This is set to 0 for Azure as there is no minimum block size. 
- Test all of the above for azure - starting a multipart, aborting a multipart, uploading parts, verifying parts are uploaded as expected.

### DIFF
--- a/modules/azure/src/main/clojure/xtdb/azure.clj
+++ b/modules/azure/src/main/clojure/xtdb/azure.clj
@@ -7,8 +7,7 @@
             [xtdb.azure.object-store :as os]
             [xtdb.azure.file-watch :as azure-file-watch]
             [xtdb.log :as xtdb-log]
-            [xtdb.util :as util]
-            [clojure.reflect :as reflect])
+            [xtdb.util :as util])
   (:import (com.azure.core.credential TokenCredential)
            (com.azure.core.management AzureEnvironment)
            (com.azure.core.management.profile AzureProfile)
--- a/modules/azure/src/main/clojure/xtdb/azure.clj
+++ b/modules/azure/src/main/clojure/xtdb/azure.clj
@@ -7,7 +7,8 @@
             [xtdb.azure.object-store :as os]
             [xtdb.azure.file-watch :as azure-file-watch]
             [xtdb.log :as xtdb-log]
-            [xtdb.util :as util])
+            [xtdb.util :as util]
+            [clojure.reflect :as reflect])
   (:import (com.azure.core.credential TokenCredential)
            (com.azure.core.management AzureEnvironment)
            (com.azure.core.management.profile AzureProfile)
@@ -40,6 +41,9 @@
   (s/keys :req-un [::storage-account ::container ::servicebus-namespace ::servicebus-topic-name]
           :opt-un [::prefix]))
 
+;; No minimum block size in azure
+(def minimum-part-size 0)
+
 (defmethod ig/init-key ::blob-object-store [_ {:keys [storage-account container prefix] :as opts}]
   (let [credential (.build (DefaultAzureCredentialBuilder.))
         blob-service-client (cond-> (-> (BlobServiceClientBuilder.)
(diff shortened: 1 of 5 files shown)


modules/azure/src/main/clojure/xtdb/azure.clj
```diff
@@ -7,7 +7,8 @@
             [xtdb.azure.object-store :as os]
             [xtdb.azure.file-watch :as azure-file-watch]
             [xtdb.log :as xtdb-log]
-            [xtdb.util :as util])
+            [xtdb.util :as util]
+            [clojure.reflect :as reflect])
```
danmason: TODO: Remove this

modules/azure/src/main/clojure/xtdb/azure/object_store.clj
```diff
@@ -53,7 +54,44 @@
...
+       (.add !staged-block-ids block-id))))
+
+  (complete [_]
+    (CompletableFuture/completedFuture
+     (do
+       (.commitBlockList block-blob-client !staged-block-ids)
+       ;; Run passed in on-complete function (adds the key to the filename cache)
+       (on-complete))))
+  
+  (abort [_]
```
danmason, line 80: Doesn't seem to be a direct way to delete uncomitted blocks other than this - using the same idea as from this StackOverflow answer: https://stackoverflow.com/questions/68662051/how-to-delete-uncommitted-blocks-in-azure-blob-storage

danmason: (They will be automatically deleted after 7 days if not cleaned up manually, but I think we'd prefer to explicitly remove these failed blobs?)

jarohen (approved):
`^:s3` -> `^:azure`; otherwise, nice one, cheers @danmason! :clap:

Issue #2483, opened 2023-04-24
Case/punctuation sensitivity/normalisation for nested data.


FiV0:
See #2481 for the ticket about toplevel columns.

Currently the case/punctuation normalisation only happens for toplevel columns (keys). To make things consistent this should likely also be the case for nested data. For example
```clj
(xt/submit-tx node [[:put :docs {:xt/id 1 :foo {:a/b "foo"}}]])
(xt/q node "SELECT docs.foo.a$b FROM docs")
```

PR #2841, opened 2023-10-19
Switch `file-name-cache` to work with `java.nio.Path` #2482 #2483

### LINKED ISSUE #2842: Ensure remote cloud object stores (S3, Azure, Google Cloud) do not return recursive keys from `listObjects`
Need to update the implementation of `file-list/list-keys-under-prefix` to ensure that it only returns top level files and directories.

### LINKED ISSUE #2843: Rework `file-name-cache` to store and use `java.nio.Path` objects
Following on from #2841, we are doing a lot of string based filtering and sorting which would likely be made simpler/easier by the use of `java.nio.Path` instead of just storing the file names. 

Would like to prospectively rework our usages of `file-name-cache` to store and use `Path` - we would need to update anything that calls to add/remove to/from/list the filename cache in the remote object stores (the object stores themselves and their file watchers). These are currently being used by the remote object stores:
  - S3
  - Azure
  - Google Cloud

Likely want to pull any shared functions/transformations to the `xtdb.file-list` namespace, and will need to update `list-files-under-prefix`. Listing files should continue to return strings - so will need to transform the Paths out.

danmason:
Resolves #2842 and Resolves #2843

- Updates the `xtdb.file-list` namespace to add a set of functions which handle adding, removing and listing the `file-name-cache` using `java.nio.Path` objects. 
- Updates all usages of the `file-name-cache` to use these functions, so everywhere we use the cache we are handling `Path` objects for the sake of consistency. 
- Add some tests for the "list with prefix" behaviour - should work as explained below (and in line with #2842)

A benefit to switching over the usages of `file-name-cache` to all work with functions defined in `file-list` is that we can change the representation of the `file-name-cache` however we want under the hood - all that callers need to know is that they pass in and get back out strings representing the file-name/object key.

## Impacted object stores

This change will impact all object stores that use the `file-name-cache` for listing objects - this is currently:
- S3 Object Store
- Azure Object Store
- Google Cloud Object Store

Have tested these with their test suites (particularly the various list tests, including the new tests for prefix behaviour and the `multiple-object-store-list-test` to ensure that the `file_watch` functions are adding/removing stuff correctly).

## Expected prefix behaviour

Given a set of filenames with the following structure:

```
foo
  foo/example1
  foo/example2
  foo/bar/example3
  foo/bar/example4
  foo/baz/example5
```  

Previoius behaviour of calling `listObjects` with prefix `foo` on object stores that were using `list-files-under-prefix` would return the full nested list, ie:
```
foo/example1 foo/example2 foo/bar/example3 foo/bar/example4 foo/baz/example5
```

What we now return is a shallow list - only returning files and directories at foo level, ie:
```
foo/example1 foo/example2 foo/bar foo/baz
```

### DIFF
--- a/core/src/main/clojure/xtdb/file_list.clj
+++ b/core/src/main/clojure/xtdb/file_list.clj
@@ -1,8 +1,29 @@
 (ns xtdb.file-list
-  (:require [clojure.string :as string])
-  (:import java.util.NavigableSet))
+  (:require [clojure.java.io :as io])
+  (:import [java.nio.file Path]
+           [java.util NavigableSet]))
+
+(defn file-name->path [^String file-name]
+  (.toPath (io/file file-name)))
+
+(defn add-filename [^NavigableSet file-name-cache ^String file-name]
+  (.add file-name-cache (file-name->path file-name)))
+
+(defn add-filename-list [^NavigableSet file-name-cache file-name-list]
+  (.addAll file-name-cache (mapv file-name->path file-name-list)))
+
+(defn remove-filename [^NavigableSet file-name-cache ^String file-name]
+  (.remove file-name-cache (file-name->path file-name)))
+
+(defn list-files [^NavigableSet file-name-cache]
+  (mapv str file-name-cache))
 
 (defn list-files-under-prefix [^NavigableSet file-name-cache prefix]
-  (->> (.tailSet ^NavigableSet file-name-cache prefix)
-       (take-while #(string/starts-with? % prefix))
-       (into [])))
+  (let [prefix-path ^Path (file-name->path prefix)
+        prefix-depth (.getNameCount prefix-path)]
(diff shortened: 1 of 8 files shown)


jarohen (changes requested):
hey @danmason - couple of weird results here when you've got files with common prefixes:

```clojure
(-> (doto (java.util.TreeSet.)
      (.add "foo/bar/baz")
      (.add "foo/bara/baz"))
    (list-files-under-prefix "foo/bar"))
;; => ("foo/bar/baz" nil)

(-> (doto (java.util.TreeSet.)
      (.add "foo/bar/baz")
      (.add "foo/bara/baz"))
    (list-files-under-prefix "foo/bar/"))
;; => (nil)
```

jarohen (approved):

Issue #2814, opened 2023-10-05
.putObject does not support large files very well


wotbrew:
We have not yet decided exactly how large files should be allowed to become - we are doing some work to make sure we can better estimate file sizes here: #2816. 

Should we decide to store large files (1GB+) it might be worth solving problems related to memory usage and network timeouts.

There are two main sources of .putObject calls, L0 writes as when flushing a chunk during indexing - and compaction which generates L1+ files. 

We should be less concerned (I think) about L0 writes as they will likely be bounded by a limit described here #2816, which per-table is unlikely to cause significant memory or timeout issues.

## Memory

.putObject currently takes a nio ByteBuffer. This assumes that either we have the whole object in memory or memory mapped. If we are not memory mapping this may cause larger buffers to exhaust available memory, or put undue pressure on memory-starved queries.

Compaction currently writes to ByteArrayOutputStreams whose arrays are then realised as heap ByteBuffers to be sent to storage via .putObject, this requires at best the whole object to be realised in heap or 2X the memory to be allocated due to the copy semantics of .toArray on ByteArrayOutputStream. 

We could choose to:

- Require a memory mapped buffer, or have it be our accepted way of writing large files
  - requires writing the whole buffer up front, memory map or not
  - upload can be partitioned and sent in parts in parallel if using multipart
  - sdk should be able to work on mapped memory directly and avoid copies (this assumption needs testing)
  - multipart not strictly necessary if timeouts policy can be gotten right
  - implies a file when memory map is used

- Require a file e.g putFile rather than putObject
  - requires writing the whole file up front 
  - requires disk space
  - might not need multipart
  - cloudsdk can normally work with file handles directly requires us to write very little of our own code
  - latencies involved will make file open/close and even copy potentially not that important for small buffers
  - would still need to look at timeout policies and test timeouts do not need scaling the size of the file 
  - may provide fewer opportunities to tune parallelism
  - may need to remember to clean up temp files

- Expose an OutputStream or nio.WritableByteChannel 
  - pipelining and limited parallelism possible 
  - do not need to realise whole stream up front 
  - might be relatively more sophisticated, probably requires multipart

## Timeouts

With larger files, it seems possible to get S3 timeout exceptions depending on network speed. To reproduce, use a slow connection and/or `:put` large rows to cause file sizes to increase.

.putObject protocol does not explicitly define timeout policy, but if you have a 200MB ByteBuffer to put - given a timeout a caller has few options to resolve themselves.

[...]

PR #2848, opened 2023-10-23
Add channelled writes to BufferPool

wotbrew:
This PR:

- Adds ArrowFileWriter / WriteableByteChannel options to enable streaming writes
  - Multipart is used when write is large enough 
- Removes FileSystemObjectStore (responsibility sits entirely within buffer pool)
- Caches are warmed with writes, assumption: written buffers will be likely future reads
- Handful of new tests to cover at least in part remote multipart branches, which would not be exercised by existing test
- Removes buffer pool range request implementation as it is not used, support exists on object stores, can be re-added when used. 

This PR does not:

- Integrate streaming writes with compactor/indexer (will be a separate PR)
- Test record batch cuts (alignment of record batches to part boundaries), due to unknown importance of feature.
- Test failure behaviour

The use of multipart for large files should partially or entirely solve #2814, pending testing with large compactions.

I would like to follow up with further tests but prefer to submit for review now.


jarohen:
superseded by #2851

PR #2851, opened 2023-10-23
Channelled writers for the buffer pool, objects

### LINKED ISSUE #2814: .putObject does not support large files very well

wotbrew:
This PR:

- Adds ArrowFileWriter / WriteableByteChannel options to enable streaming writes
  - Multipart is used when write is large enough 
- Removes FileSystemObjectStore (responsibility sits entirely within buffer pool)
- Caches are warmed with writes, assumption: written buffers will be likely future reads
- Handful of new tests to cover at least in part remote multipart branches, which would not be exercised by existing test

This PR does not:

- Integrate streaming writes with compactor/indexer (will be a separate PR)
- Test record batch cuts (alignment of record batches to part boundaries), due to unknown importance of feature.
- Test failure behaviour

The use of multipart for large files should partially or entirely solve #2814, pending testing with large compactions.

I would like to follow up with further tests but prefer to submit for review now.

### DIFF
--- a/core/src/main/clojure/xtdb/buffer_pool.clj
+++ b/core/src/main/clojure/xtdb/buffer_pool.clj
@@ -2,10 +2,12 @@
   (:require [clojure.string :as str]
             [clojure.tools.logging :as log]
             [juxt.clojars-mirrors.integrant.core :as ig]
-            [xtdb.object-store :as os]
-            [xtdb.util :as util])
+            [xtdb.object-store :as object-store]
+            [xtdb.util :as util]
+            [xtdb.object-store :as os])
   (:import java.io.Closeable
            (java.nio ByteBuffer)
+           (java.nio.channels FileChannel$MapMode)
            [java.nio.file FileSystems FileVisitOption Files LinkOption Path]
            [java.util Map NavigableMap]
            [java.util.concurrent CompletableFuture ConcurrentSkipListMap]
@@ -63,6 +65,15 @@
          (record-cache-hit cached-buffer)
          cached-buffer))))
 
+  (getRangeBuffer [_ k start len]
+    (object-store/ensure-shared-range-oob-behaviour start len)
+    (CompletableFuture/completedFuture
+     (when k
+       (when-let [buffer (when-let [^ArrowBuf full-buffer (cache-get buffers k)]
+                           (.slice full-buffer start len))]
+         (record-cache-hit buffer)
+         buffer))))
+
   (putObject [_ k buf]
     (.put buffers k (util/->arrow-buf-view allocator buf))
     (CompletableFuture/completedFuture nil))
@@ -103,6 +114,16 @@
 
     (util/->mmap-path from-path)))
 
+(defn get-path-range ^ByteBuffer [^Path root-path, ^String k, ^long start, ^long len]
+  (os/ensure-shared-range-oob-behaviour start len)
+
+  (let [from-path (.resolve root-path k)]
+    (when-not (util/path-exists from-path)
+      (throw (os/obj-missing-exception k)))
+
+    (with-open [in (util/->file-channel from-path #{:read})]
+      (.map in FileChannel$MapMode/READ_ONLY start (max 1 (min (- (.size in) start) len))))))
+
 (defn put-path [^Path root-path, ^String k, ^ByteBuffer buf]
   (let [buf (.duplicate buf)
         to-path (.resolve root-path k)]
@@ -142,6 +163,23 @@
          (let [nio-buffer (get-path root-path k)
                create-arrow-buf #(util/->arrow-buf-view allocator nio-buffer)
                [_ buf] (cache-compute buffers k create-arrow-buf)]
+           buf)))))
+
+  (getRangeBuffer [_ k start len]
+    (object-store/ensure-shared-range-oob-behaviour start len)
+
+    (CompletableFuture/completedFuture
+     (when k
+       (if-let [cached-buffer (or (cache-get buffers [k start len])
+                                  (when-let [^ArrowBuf cached-full-buffer (cache-get buffers k)]
+                                    (.slice cached-full-buffer start len)))]
+         (do
+           (record-cache-hit cached-buffer)
+           cached-buffer)
+
+         (let [nio-buffer (get-path-range root-path k start len)
+               create-arrow-buf #(util/->arrow-buf-view allocator nio-buffer)
+               [_ buf] (cache-compute buffers [k start len] create-arrow-buf)]
            buf)))))
 
   (putObject [_ k buf] (CompletableFuture/completedFuture (put-path root-path k buf)))
@@ -220,6 +258,30 @@
                     (record-io-wait start-ns)
                     (let [create-arrow-buf #(util/->arrow-buf-view allocator nio-buffer)
                           [_ buf] (cache-compute buffers k create-arrow-buf)]
+                      buf)))))))))
+
+  (getRangeBuffer [_ k start len]
+    (object-store/ensure-shared-range-oob-behaviour start len)
+    (if (nil? k)
+      (CompletableFuture/completedFuture nil)
+      (let [cached-full-buffer (cache-get buffers k)
+
+            cached-buffer
+            (or (cache-get buffers [k start len])
+                (when ^ArrowBuf cached-full-buffer
+                  (.slice cached-full-buffer start len)))]
+
+        (if cached-buffer
+          (do
+            (record-cache-hit cached-buffer)
+            (CompletableFuture/completedFuture cached-buffer))
+          (let [start-ns (System/nanoTime)]
+            (-> (.getObjectRange object-store k start len)
+                (util/then-apply
+                  (fn [nio-buffer]
+                    (record-io-wait start-ns)
+                    (let [create-arrow-buf #(util/->arrow-buf-view allocator nio-buffer)
+                          [_ buf] (cache-compute buffers [k start len] create-arrow-buf)]
                       buf)))))))))
 
   (putObject [_ k buf] (.putObject object-store k buf))
--- a/core/src/main/clojure/xtdb/buffer_pool.clj
+++ b/core/src/main/clojure/xtdb/buffer_pool.clj
@@ -250,8 +250,7 @@
    ^ObjectStore remote-store
    ^FileChannel file-channel
    tmp-path
-   ^String k
-   allow-file-deletion]
+   ^String k]
   WritableByteChannel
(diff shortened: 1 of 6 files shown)


core/src/main/clojure/xtdb/object_store.clj
```diff
@@ -61,118 +52,3 @@
...
-  (putObject [_this k buf] (util/completable-future pool (put-path root-path k buf)))
-  (listObjects [_this] (list-path root-path))
-  (listObjects [_this dir] (list-path root-path dir))
-  (deleteObject [_this k] (util/completable-future pool (util/delete-file (.resolve root-path k))))
-
-  Closeable
-  (close [_this]
-    (util/shutdown-pool pool)))
-
-(derive ::file-system-object-store :xtdb/object-store)
```
danmason: I think the reason we kept the file-system object store was for remote buffer pool + remote file stores?

wotbrew: Not sure about this, wouldn't we use the disk buffer pool? We need to guarantee for correctness atomicity of writes in the single node case, I reckon correct atomicity would make it ok for multiple processes to bang on the same disk buffer pool path, files are immutable...

danmason (approved):
All seems sensible to me - as long as we're happy removing the filesystem object store and the getRange calls.

core/src/main/clojure/xtdb/object_store.clj
jarohen: I'm not sure about this - won't the networked obj-store behave differently from the local BP? e.g. local BP knows it's entirely in control of the directory, networked obj-store will likely need to do some kind of put-to-a-temp-file then copy dance?

core/src/main/clojure/xtdb/buffer_pool.clj
```diff
@@ -35,258 +48,424 @@
...
+  (close [this]
+    (.close file-channel)
+    (let [file-path (.resolve disk-store k)]
+      (util/create-parents file-path)
+      ;; see #2847
+      (util/atomic-move tmp-path file-path)
+      (util/then-apply (.getBuffer bp k) util/close)))
+  (write [_ src] (.write file-channel src)))
+
+(defrecord DiskOnlyBufferPool
```
jarohen: naming is obv hard, but I'm not sure about this change - it's not disk-only (it has a memory cache, for example) and I liked that this previously specified 'local', to clearly differentiate between this and the remote version.

core/src/main/clojure/xtdb/buffer_pool.clj
```diff
@@ -35,258 +48,424 @@
...
+                  (let [nio-buffer (util/->mmap-path path)
+                        create-arrow-buf #(util/->arrow-buf-view allocator nio-buffer)
+                        [_ buf] (cache-compute memory-store k create-arrow-buf)]
+                    buf))))))))
+  (listObjects [_]
+    (with-open [dir-stream (Files/walk disk-store (make-array FileVisitOption 0))]
+      (vec (sort (for [^Path path (iterator-seq (.iterator dir-stream))
+                       :when (Files/isRegularFile path (make-array LinkOption 0))]
+                   (str (.relativize disk-store path)))))))
+  (listObjects [_ dir]
```
jarohen, line 236: can we reinstate the blank lines removed between different methods here? without the whitespace I find my eyes don't have anything to lock on to when scanning the code

core/src/main/clojure/xtdb/buffer_pool.clj
```diff
@@ -35,258 +48,424 @@
...
+(deftype RemoteWritableChannel
+  [^IBufferPool bp
+   allocator
+   memory-store
+   ^Path disk-store
+   ^ObjectStore remote-store
+   ^FileChannel file-channel
+   tmp-path
+   ^String k
+   allow-file-deletion]
```
jarohen: do we use `allow-file-deletion`?

jarohen (changes requested):
thanks for the PR @wotbrew :pray: couple of things, maybe better discussed in a quick chat:
* let's talk about the removal of the file-based OS, probably more to this than's best discussed in back-n-forth text
* it seems like intercepting at the level of channel is maybe making things harder than they need to be? for example, we could do this behaviour in `ArrowFileWriter/.end` and then use the vanilla channels.
* openly, I'm finding this PR a little harder to review than perhaps it could be - is there any reason to swap the order of the buffer-pools in the file, for example?

jarohen (changes requested):
thanks for the PR @wotbrew  :pray: couple of things, maybe better discussed in a quick chat:
* let's talk about the removal of the file-based OS, probably more to this than's best discussed in back-n-forth text
* it seems like intercepting at the level of channel is maybe making things harder than they need to be? for example, we could do this behaviour in `ArrowFileWriter/.end` and then use the vanilla channels.
* openly, I'm finding this PR a little harder to review than perhaps it could be - is there any reason to swap the order of the buffer-pools in the file, for example?
  * edit: I missed, for example, the change from a `ConcurrentSkipListMap` to a `TreeMap` - was this intentional?

core/src/main/clojure/xtdb/object_store.clj
wotbrew: I don't think it will behave differently, we make it a requirement of a local bp to be able to share a data-dir with other processes. The acceptance tests will then include the concurrent checks over a networked filesystem.

need for atomicity of write (never seeing partial files, or making them discoverable), and the immutability property are our friends.

Also not a problem to revert a delete if the team is still not sure.

wotbrew: A specialisation may be possible for performance to have a single process directory with a lock so we can manage more in memory - however, this is not how the current disk buffer pool works, it defers to the disk for listings and so on.

I would add this specialisation if the one multi-process supporting buffer pool is not sufficient.

core/src/main/clojure/xtdb/buffer_pool.clj
wotbrew: Ha! this is a sneaky feature, it seemed reasonable to have this in non object version due to implementation sharing. Here it denotes whether or not the disk cache size should be the same as the virtual memory limit (or mappings limit). 

In my opinion, this may be a value add (which could be extended to a separate disk limit), but honestly, it snuck in as I refactored from the newer non-object base.

wotbrew:
> * it seems like intercepting at the level of channel is maybe making things harder than they need to be? for example, we could do this behaviour in `ArrowFileWriter/.end` and then use the vanilla channels.

Yes, I saw that but the metadata writes transit files currently, if we change that we can do more in .end.

> * openly, I'm finding this PR a little harder to review than perhaps it could be - is there any reason to swap the order of the buffer-pools in the file, for example?

Apologies I started the refactor from my previous PR, so things have moved and I didn't go and move them back - made things messier but wanted to get this in ASAP>

>   * edit: I missed, for example, the change from a `ConcurrentSkipListMap` to a `TreeMap` - was this intentional?

We use a lock for all buffers access AFAIK, this came in recently for correct memory management IIRC. Used least fancy navigable - tests pass? Not sure if we need it?

jarohen:
merged, cheers @wotbrew :pray:

PR #2854, opened 2023-10-25
reader and copier tests

FiV0:
Some tests that mainly test copier behavior and add some better error throwing in case of type mismatches.

### DIFF
--- a/core/src/main/clojure/xtdb/vector/writer.clj
+++ b/core/src/main/clojure/xtdb/vector/writer.clj
@@ -460,6 +460,14 @@
           (.writeNull absent-writer))
         (throw (UnsupportedOperationException. "populate-with-absents needs a nullable or union underneath!"))))))
 
+
+(defn- check-field-types [^FieldType expected-field-type ^FieldType given-field-type]
+  (or (and (= #xt.arrow/type :null (.getType expected-field-type))
+           (= #xt.arrow/type :null (.getType given-field-type)))
+      (when-not (= expected-field-type given-field-type)
+        (throw (IllegalStateException. (str "Field type mismatch: " (pr-str {:expected expected-field-type
+                                                                             :given given-field-type})))))))
+
 (extend-protocol WriterFactory
   ListVector
   (->writer* [arrow-vec notify!]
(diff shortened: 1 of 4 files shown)


core/src/main/clojure/xtdb/vector/writer.clj
```diff
@@ -460,6 +460,14 @@
           (.writeNull absent-writer))
         (throw (UnsupportedOperationException. "populate-with-absents needs a nullable or union underneath!"))))))
 
+
+(defn- check-field-types [^FieldType expected-field-type ^FieldType given-field-type]
+  (or (and (= #xt.arrow/type :null (.getType expected-field-type))
+           (= #xt.arrow/type :null (.getType given-field-type)))
```
FiV0, line 466: This is a bit of weird one. I sometimes get `<FieldType :null null>` and `<FieldType :null not-null>` which then clash here.

FiV0: cc @jarohen

jarohen: suspect that's a case of something indiscriminately calling `FieldType/notNullable` when it should probably be checking that the arrow-type is not `:null`

core/src/main/clojure/xtdb/vector/writer.clj
```diff
@@ -488,22 +496,26 @@
...
+                          (do
+                            (.startList this-wtr)
+                            (let [start-idx (.getElementStartIndex src-vec src-idx)
+                                  end-idx (.getElementEndIndex src-vec src-idx)]
+                              (dotimes [n (- end-idx start-idx)]
+                                (.copyRow inner-copier (+ start-idx n))))
+                            (.endList this-wtr)))
+                        pos))))
+                :else
+                (throw (IllegalStateException.
```
jarohen: nit: this is an `err/illegal-arg` - it's that the argument was completely wrong, rather than the argument being wrong for the state that the object's in

jarohen (approved):
cheers @FiV0! :pray:

core/src/main/clojure/xtdb/vector/writer.clj
FiV0: Done.

PR #2857, opened 2023-10-25
Readers/writers documentation namespace

FiV0:
This adds an interactive namespace to explain our readers / writers to somebody joining the project.

### DIFF
--- a/src/dev/clojure/arrow_readers_writers_doc.clj
+++ b/src/dev/clojure/arrow_readers_writers_doc.clj
@@ -1,0 +1,265 @@
+(ns arrow-readers-writers-doc
+  (:require [xtdb.vector.writer :as vw]
+            [xtdb.vector.reader :as vr]
+            [xtdb.types :as types])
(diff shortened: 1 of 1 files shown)


danmason:
I think this serves as a nice intro to the types at play and what we're using under the hood, especially to me given how new it all is to me, though a few questions:
- In our "before reading" section we say it's worth being somewhat familiar with the concepts of Arrow. I did have a read through the linked document, primarily the first few sections (Overview at the top, "Terminology", and first few subsections of "Physical Memory Layout"). What level of familiarity are we expecting going into this?
- We say we introduce "readers" here - are we using the term interchangably with a copier, or have I misunderstood something? Seems like primarily in that section we're handling something that copies between writers? Is that what we mean by a reader?
- These particular terms, "reader", "writer" and "copier" - are they primarily what we've been calling them, or is this arrow's terminology? Might be worth our own little glossary as a comment at the top, maybe.

I think on a practical level this serves as a nice crash course - though it may be the case I would need to dive a bit more into arrow's docs to slot all of these things together :)

PR #2861, opened 2023-10-27
Setup for `nightly-test` job

### LINKED ISSUE #2859: Setup basic nightly scheduled test run on 2.x branch
Want to set up a nightly test run to automatically run some of the tests that are currently only being ran manually, such that we can catch any overnight regressions/breakages.

- Add a "nightly-test" task to gradle to run tests we usually do not run with the regular test suite. 
- Setup a scheduled github action to run on a (weeknight) nightly basis.
  - This should run the "nightly-test" task on gradle  
- Setup any credentials/environment variables necessary for testing - long running s3 credentials, azure credentials, etc. This may require some setup on the services themselves, which we should do our best to capture into runnable templates.

As a follow on, I'm expecting/hoping we can add some more high level end-to-end style tests that run against various node configurations, but the above makes sense as a solid incremental step to add the setup that we will require and to ensure everything we currently have is at least being tested _somewhere_.

### Prior Art

We already have a night SLT flow running: https://github.com/xtdb/xtdb/actions/workflows/slt.yml

Aiming to keep the new nightly test workflow separate (such that we can easily see results of test runs, and perhaps even alert to the dev channel), but shall certainly aim to take inspiration from the above and take note of any special behaviour/fixes it has to use. Currently:
- It has to run from master and check out 2.x
- There's a check, `if: github.repository == 'xtdb/xtdb'` so that it doesn't run on forks.
- Currently, said `if` check runs at such a point that it cancels the workflow midrun, so the list on forks looks like the following:  https://github.com/mbutlerw/xtdb/actions/workflows/slt.yml

Plan is to use the above as a jumping off point, and dig through the actions documentation to see if any improvements/changes can be made. Any improvements we make to the nightly we can also copy over to the SLT run.

danmason:
Sets up an initial version for a nightly test job, with a number of other bits of setup in the background

- Creates the `nightly-test` task in gradle - this currently is running the `s3`, `azure` and `google-cloud` tests which are not tested in the regular `test` task. There are others we could potentially run here (`jdbc`, `kafka`, `timescale`), but I have left them out for now.
  - Seems to be a few issues with the docker-compsoe that will likely need to be sorted out.
  - There are also discussions around removing `jdbc` object store, so might not be so useful to run that regardless.
- Makes a number of small fixes to existing code/tests:
  - Comments out some known problematic tests in `object_store`/azure.  
  - Fixes an issue with the google cloud custom role missing some permissions, flagged by running the test.
  - Fixes a bug in the file-list `add-filename` function (soon to be irrelevant, but helps make the tests green).
  - Increases timeout in one of the s3 tests to make it run more consistently on CI
- Set up various credentials/auth on the cloud providers themselves, and have added relevant vars/secrets to the repo for setting up auth.
- Added a new asciidoc file under `dev` that goes through steps of setting up credentials for github actions, as well as including some template for cloudformation IAM user. This also talks about the general process of making changes to the nightly test job.

In addition to the above, there is a commit that adds the nightly test workflow to the repo. As a note - since this will **have** to live on `master` to be able to run, it's likely not worth actually including on the `2.x` branch, nor this PR, but I have included it for the sake of being able to give this a once-over. Once we're happy with the changes here, will pull out the workflow file commit and make a PR to add that to `master`.

The nightly test has been ran on this branch (by making some small edits to the file) - so will need to be later verified it schedules/runs correctly when pushed to master.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -59,6 +59,12 @@ allprojects {
             }
         }
 
+        tasks.create("nightly-test", Test::class) {
+            useJUnitPlatform {
+                includeTags("kafka", "docker", "jdbc", "s3", "azure", "google-cloud")
+            }
+        }
+
         tasks.withType(Test::class) {
             jvmArgs = defaultJvmArgs + sixGBJvmArgs
         }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -61,7 +61,7 @@ allprojects {
 
         tasks.create("nightly-test", Test::class) {
             useJUnitPlatform {
-                includeTags("kafka", "docker", "jdbc", "s3", "azure", "google-cloud")
+                includeTags("s3", "google-cloud", "azure")
             }
         }
 
--- a/dev/NIGHTLY_CI_SETUP.adoc
+++ b/dev/NIGHTLY_CI_SETUP.adoc
@@ -1,0 +1,78 @@
+= Nightly CI Setup
+
+Within our nightly test build, we run a number of tests:
+
+* Tests against various cloud providers - S3, Azure, Google Cloud
+* Tests involving kafka & tests involving postgres - setup via docker-compose
+
+When it comes to testing against the cloud providers, we require authentication for each of those handled on the CI. We will pass in everything we need to the environment variables/secrets of Github a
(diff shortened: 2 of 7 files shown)


jarohen (approved):
cheers @danmason :pray:

PR #2862, opened 2023-10-30
refactor XTQL compiler to decouple planning the query from executing it

jarohen:
@mbutlerw: `QueryOpts` naturally fell out of this one, but I'm aware there's probably a reason behind its introduction - if you had grander plans for it, shout, and I can probably just reinstate it?


mbutlerw (approved):
QueryOpts was introduced in a world where args worked more like `:in` args within unify, as this would require us to pass around `QueryArgs` during planning.

Now we've moved entirely to scalar params, I think handling them as you've done here is acceptable.

PR #2865, opened 2023-10-30
col-type -> field in operators and metadata

FiV0:
This passes from col-types to fields in the operators as well as the metadata. The boundary to col-types is mostly the EE. There are some places where I kept col-types because it was very much intertwined with the EE, for example `IProjectionSpec` as well as aggregates.


core/src/main/clojure/xtdb/operator/project.clj
```diff
@@ -106,30 +107,36 @@
...
+                                      (for [[p-type arg] projections]
+                                        (case p-type
+                                          :column (->identity-projection-spec arg (get inner-fields arg))
+                                          :row-number-column (let [[col-name _form] (first arg)]
+                                                               (->row-number-projection-spec col-name))
+                                          :rename (let [[to-name from-name] (first arg)]
+                                                    (try
+                                                      (->rename-projection-spec to-name from-name (get inner-fields from-name))
+                                                      (catch Throwable t
+                                                        (prn from-name)
```
jarohen: cheeky prns

jarohen (approved):
nice one @FiV0 :pray: 

As always, could you give your PRs a quick scan for things like `prn`s and (I assume) inadvertent formatting changes - apart from that, good to go :slightly_smiling_face:

PR #2897, opened 2023-11-03
Updates to the docker container/directory

danmason:
Numerous updates to the docker config to create `xtdb-standalone-ea` - and getting that set up to deploy on pushes to `2.x`:
- Ensure s3 module included in created uberjar.
- Fixed up/updated some of the local standalone docker config, bringing it all up to the present and getting it running locally with the minimal default config we want.
  - Updated config for buffer pool.
  - Removing all but the http-server.
  - Renamed to `local_xtdb.edn` 
- Looked at the various scripts/docs in the docker file - removed manual deployment script and documentation.
- Wrote up a workflow step to build the uberjar, dockerfile, and push to the repository
- Add `deploy-docker` step to the build workflow, running only on pushes to `2.x` on the new repo
  - Builds and tags the `xtdb/xtdb-standalone-ea:latest` image, and pushes it to GHCR  
  - Uses docker buildx, builds to multiple platforms (as per prior deploy script)

Have tested the deploy-docker step - built image [**here**](https://github.com/xtdb/xtdb/pkgs/container/xtdb-standalone-ea)


refset:
@danmason One of the things we had figured out previously was the use of `buildx` for doing multi-arch builds - per `docker buildx build --push --platform linux/arm/v7,linux/arm64/v8,linux/amd64` (in the deleted `docker-build-and-push.sh` file). Does this new approach solve that also?

danmason:
@refset I didnt particularly aim to handle this here - so probably not by default. Hopefully should be possible with some configuration to the `build-and-push` job?

Seems like we should be able to pass a `platforms` key? https://docs.docker.com/build/ci/github-actions/multi-platform/

danmason:
@refset Have pushed another commit to handle docker buildx and build for the multiple platforms you mentioned in your message.

jarohen (approved):
thanks @danmason, nice one :clap: 

slight bikeshed: I think 'local' and 'standalone' are saying the same thing? if so, shall we just go for 'standalone' (à la XT1?)

danmason:
> slight bikeshed: I think 'local' and 'standalone' are saying the same thing? if so, shall we just go for 'standalone' (à la XT1?)

Yeah - I was wondering about the name. We could definitely call it xtdb-standalone - do you think we should mark it as `2.x` somehow in the name?

PR #2903, opened 2023-11-06
coalescing table-args into params

FiV0:
This removes the special treatment of table-args. Parameters passed to the table operator must now be a struct list vector.


jarohen (commented):
nice one :clap:

PR #2908, opened 2023-11-07
xtql `table` operator

### LINKED ISSUE #2869: Add `table` operator to XTQL
Name TBC, XTQL needs to provide some way for users to generate tables from values

FiV0:
This adds a table operator which allows for 
```clj
(table [{:first-name "Ivan"} {:first-name "Petr"}] [first-name])
;; =>
[{:first-name "Ivan"} {:first-name "Petr"}]
```
and
```clj
(xt/q tu/*node* '(table $ivan+petr [name])
  {:args {:ivan+petr [{:name "Ivan"} {:name "Petr"}]}})
;; =>
[{:name "Ivan"} {:name "Petr"}]
```
Some things to decide:
- [ ] - no bindings option which results in all columns being projected (might be confusing in unify)
- [ ] - currently no unification in out bindings spec (filtering as well as same logic var unification)

### DIFF
--- a/api/src/main/java/xtdb/query/Query.java
+++ b/api/src/main/java/xtdb/query/Query.java
@@ -365,4 +365,34 @@ public sealed interface Query {
         return new Offset(length);
 
     }
+    final class Table implements Query, UnifyClause {
+        public final List<Map<String, Expr>> documents;
+        public final Expr.Param param;
+        public final List<OutSpec> bindings;
+
+        private Table(List<Map<String, Expr>> documents, Expr.Param param, List<OutSpec> bindings) {
+            this.documents = documents;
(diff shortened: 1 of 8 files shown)


FiV0:
@mbutlerw The Java AST constructor has overloads in this case. I don't know if that is bad or if that should be dealt with differently. There was no previous example for a behavior like that.

mbutlerw (approved):

jarohen (commented):
nice one @FiV0 :clap: 

I might be tempted to have two different `Table` AST classes for this - it seems like throughout the code you're trying to differentiate between the two cases. Both can still have a `table` builder method, and render as `(table ...)`, but it gives you the ability to switch on type instead.

Issue #2891, opened 2023-11-02
Deployed documentation for the XT2 HTTP API


jarohen:
* OpenAPI, available through any XT HTTP server (is there a 'well-known' path for these?)
* Generated, deployed with, and linked from the documentation site

## Plan (Dan M):

Current plan here for serving this:

- We have some `openapi.json` file we serve from the HTTP server.
  - This lives under the `resources` directory in the http-server.
  - We serve this file from an appropriate route on the server.
- We edit this file manually with any changes - treating it as a proper OpenAPI spec.
  - Can include whatever examples we want here.
- We have a symbolic link from the `docs` directory to this file.
- We serve a documentation site for the `openapi.json` file using Amplify:
  - We extend the yarn setup to have a script to create a static documentation page.
    - Can use a number of tools for this - have prior used `redoc-cli` for this.
    - We return the generated page as part of the `dist` directory (ie, the docs directory)
  - We run the above script as part of the amplify build - as the file is included in `dist` directory it should be deployed.

danmason:
As discussed - the pipeline to update thsi & deploy this via docs-live onto the docs site is now available, even if the contents are not finalized. Shall close this - will likely need separate ticket for contents (as and when appropriate)

PR #2928, opened 2023-11-14
Returning OpenAPI doc from http server, build documentation site via Amplify

danmason:
See #2891

Some discussion required here - in particular, what SHOULD the `openapi.json` contain? Should this include examples for `transit+json`, or only for regular json? Does the HTTP server need to support `application/json`? 

I think my questions are mainly around actual content of the openapi.json and the HTTP server in general:
- Currently only seems to accept/return transit+json, presumably because thats what the client is using.
- I think in practise, if we want to document the HTTP API and expect people to use it (which I think we probably do? Useful for non clojure usage) - it probably needs to support MORE than transit+json and that is probably what should be documented in the openapi spec?

In essence, I think most of the 'plumbing' of #2891 is mostly done, but have questions on actual potential contents here and on how we want the HTTP server to behave in general

## Done so Far

- Wrote a basic `openapi.json` file based on the generate swagger docs for the routes.
- Return said static file from the server.
- Create symlink in docs to this file.
- Add `redoc-cli` and bindings to `yarn`- build documentation page from `openapi.json`.
- Do the above as part of the amplify build.

## Update

Given that we're happy with the pipeline/main body of the changes,going to merge this and push to `2.x` and `docs-live`. Can edit the `openapi.yaml` later as and when it's appropriate :)


jarohen:
Cheers @danmason :pray:

Yep, this is pretty much what I was looking for from this block of work. Was aware that we only have Transit for now - the more important thing is that you added an E2E s.t. when we do have fuller JSON support, the docs pipeline's already in place.

If we're to edit this by hand, I'd have a slight preference for YAML (apologies if I didn't mention this...?) but no worries if everything else expects JSON.

danmason:
Migrated from `openapi.json` to `openapi.yaml` - I'm going to merge this on 2.x and then update `docs-live` :)

PR #2937, opened 2023-11-16
Adapt http client/server to encode xt errors correctly

### LINKED ISSUE #2932: Error serialization for HTTP client
When one hits an error with the http client the response is just completely useless. I think we can do better than that and add something that gives at least some hint of what is going on.

### LINKED ISSUE #2934: Http node silently swallowing errors
```clj 
  (def http-node (client/start-client "http://localhost:3000"))

  (xt/q http-node '(-> (table [{}] []) 
                       (with {:foo (/ 1 0)})))
  ;; => []
```

FiV0:
This changes a little how we encode errors in our client / server modules. There are two types of errors:
- IllegalArgument (we know there is an issue before we run anything): The body of that request just becomes the `xtdb.IllegalArgumentException`. We decode this accordingly on the client side for `4XX` repsonse codes. Essentially anything in the 4XX range should be decodable for us.
- RuntimeError (i.e. division by zero): We potentially only hit this when some results have been produced. I handled this by encoding the error into the result set and only throwing the error when result set is consumed on the client side. ~This is a problem if the results sets could actually contain RuntimeError's as values (for example in the `xt/txn` table, although currently encoded as `clj-form`).~ could only ever be nested 
- Anything else (essentially anything we don't (yet) have good error handling for, i.e. holes in the EE) we try to encode as a `clojure.lang.ExceptionInfo`.

### DIFF
--- a/http-client-clj/src/main/clojure/xtdb/client.clj
+++ b/http-client-clj/src/main/clojure/xtdb/client.clj
@@ -3,7 +3,6 @@
             [juxt.clojars-mirrors.hato.v0v8v2.hato.client :as hato]
             [juxt.clojars-mirrors.hato.v0v8v2.hato.middleware :as hato.middleware]
             [juxt.clojars-mirrors.reitit-core.v0v5v15.reitit.core :as r]
-            [xtdb.api :as xt]
             [xtdb.api.protocols :as xtp]
             [xtdb.error :as err]
             [xtdb.transit :as xt.transit])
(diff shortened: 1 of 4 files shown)


http-client-clj/src/main/clojure/xtdb/client.clj
```diff
@@ -61,19 +60,32 @@
       (throw (NoSuchElementException.)))
     (let [el (.next-el this)]
       (set! (.next-el this) nil)
-      el))
+      (if (instance? xtdb.RuntimeException el)
+        (throw el)
+        el)))
```
FiV0, line 65: @jarohen don't know if you see a problem with this approach.

http-client-clj/src/main/clojure/xtdb/client.clj
```diff
@@ -61,19 +60,32 @@
...
+    (cond
+      (= 200 status)
+      (parse-body (fn [rdr] (TransitResultSet. body rdr nil)))
+
+      ;; This should be an error we know how to decode
+      (<= 400 status 499)
+      (parse-body transit/read)
+
+      ;; Error will be thrown by error handler as is
+      :else nil)))
```
FiV0: I don't know how to best handle other cases here. For the 5XX range should I also expect some `tranist+json`?

PR #2947, opened 2023-11-20
Increase wm-cnt on startTx

### LINKED ISSUE #2800: Memory leak on ingest in live-index
This does not happen every time this test is run. I was not able to reproduce it with smaller ingestion size
```clj 
(defn- random-maps [n]
  (let [nb-ks 5
        ks [:foo :bar :baz :toto :fufu]]
    (->> (repeatedly n #(zipmap ks (map str (repeatedly nb-ks random-uuid))))
         (map-indexed #(assoc %2 :xt/id %1)))))

(deftest incoming-puts-dont-cause-memory-leak-2800
  (let [node-dir (io/file "test/incoming-puts-dont-cause-memory-leak")
        node-opts {:node-dir (.toPath node-dir)
                   :instant-src InstantSource/SYSTEM
                   :rows-per-chunk 16}]
    (dotimes [_ 5]
      (with-open [node (tu/->local-node node-opts)]

        (doseq [tx (->> (random-maps 32)
                        (map #(vector :put :docs %))
                        (partition-all 16))]
          (xt/submit-tx node tx))))))

```
results sometimes in
```
1. Caused by java.lang.IllegalStateException
   Memory was leaked by query. Memory leaked: (504832)
   Allocator(live-index) 0/504832/48329728/9223372036854775807
   (res/actual/peak/limit)
```

It seems to be especially not happen on first run.

### LINKED ISSUE #2930: Shutdown behaviour interaction `live-index` / `indexer`.
Assume a period where the system is only processing `puts` (all queries have finished and no more watermark counts are "open" for the `live-index`).  We are not taking out any watermarks on the live-index when processing `puts`. So the `live-index` will try to shut down cleanly (no watermark open), but the `indexer` is still processing puts. This results in a memory leak in the `live-index`.

Downstream issue of this one: #2800

FiV0:
This is a fix for #2930 and more concretely #2800. Whenever a transaction starts we also increase the wm-cnt on the live-index to not shut it down prematurely. After this fix the `indexer` might still run into race conditions where the `live-index` is already shutting down, but the `indexer` tries to call `startTx`. That would currently result in a toplevel indexer error, but should no longer result in memory leaks.

### DIFF
--- a/core/src/main/clojure/xtdb/indexer.clj
+++ b/core/src/main/clojure/xtdb/indexer.clj
@@ -562,7 +562,7 @@
                      (pr-str tx-key)
                      (pr-str latest-completed-tx))
 
-          (let [live-idx-tx (.startTx live-idx tx-key)]
+          (util/with-open [live-idx-tx (.startTx live-idx tx-key)]
             (add-tx-row! row-counter live-idx-tx tx-key
                          (err/illegal-arg :invalid-system-time
                                           {::err/message "specified system-time older than current tx"
@@ -625,7 +625,7 @@
                         (log/debug e "aborted tx")
                         (.abort live-idx-tx))
 
-                      (let [live-idx-tx (.startTx live-idx tx-key)]
+                      (util/with-open [live-idx-tx (.startTx live-idx tx-key)]
                         (add-tx-row! row-counter live-idx-tx tx-key e)
                         (.commit live-idx-tx)))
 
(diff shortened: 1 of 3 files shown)


FiV0:
This actually doesn't seem to fix it as the test I added is failing on CI. :thinking:

FiV0:
The whole analysis is incorrect.

PR #2949, opened 2023-11-20
Cloudformation battery included template

### LINKED ISSUE #2889: CloudFormation template for users to run a standalone XT2 Docker image on AWS
To consider:
* Where to deploy this? ECS/EKS?
  * We'd likely recommend deploying on instances with fast instance-local storage (e.g. `i3.*`) - worth checking what Fargate now supports.
  * K8s 'stateful sets' seem pretty close to our ideal scenario - can this be effectively replicated on ECS or is it worth us Just™ recommending k8s? (cross-cloud benefits here too, obv). Does EKS support them?
  * What other considerations are there?

### LINKED ISSUE #2896: Batteries-included CloudFormation template for reference AWS deployment
For users to be able to Just™ run this CloudFormation template, supplied a small number of required parameters (e.g. S3 bucket name, node count, instance type etc), and get a working multi-node XTDB cluster.

related to #2895, #2889

danmason:
See #2896 

Adds a `cloudformation` directory with a number of templates to be used in conjunction to create a full stack on AWS for XTDB (S3 + Kafka XTDB nodes, setting up all the dependent infrastructure and running on ECS/EC2 instances, connectable via an application loadbalancer)

# Templates

There are a total of 5 templates, set up in roughly the following order:

## `xtdb-vpc` 

Sets up a VPC to be used by the other infrastructure.

**Setup infrastructure:**
- Two public subnets
- Two private subnets
- Internet gateway
- Public & private routing tables, route associations, etc
- Security group

**Outputs:**
- `VpcId`: ID of the VPC created by the stack
- `PublicSubnets`: IDs of the both of the public subnets created within the VPC
- `PrivateSubnets`: IDs of the both of the private subnet created within the VPC
- `SecurityGroupId`: ID of the security group created for the VPC

## `xtdb-s3` 

Doesn't depend on the VPC - sets up all S3 infrastructure to be used by XTDB: 

**Inputs:**
- `S3BucketName` - Desired name of the bucket which will contain the XTDB Object Store.

**Setup infrastructure:**
- An SNS topic that acts as a bucket changelog (used by the XTDB file watcher) 
- An S3 bucket for the object store that publishes changes to the SNS topic. 
- Relevant roles to allow the bucket to send notifications to SNS.
- A managed policy for all S3/SNS/SQS permissions.  

**Outputs:**
- `BucketName`: passing back out the same input parameter 
- `SNSTopicArn`: The ARN of the SNS Topic with the S3 notifications
- `S3AccessPolicyArn`:  ARN of the managed policy that has relevant SNS and S3 permissions, and ability to create SQS queues

## `xtdb-msk` (dependent on `xtdb-vpc`)

Sets up MSK for use as the XTDB log

**Inputs:**
- `VpcId`: ID of the VPC to host the MSK cluster on (from `xtdb-vpc`)
- `SecurityGroupId`: Group ID of the security group to be used by the ECS nodes (from `xtdb-vpc`)
- `MSKClusterName`: User desired name of the MSK kafka cluster used as the XTDB Kafka Log - defaults to `xtdb-cluster-kafka`
- `MSKVolumeSize`: The size in GiB of the EBS volume for the data drive on each broker node of the kafka cluster - defaults to `100`

**Setup infrastructure:**
- Two private subnets, a route table, route table associations, and a security group for the subnets to be used exclusively by MSK.
- An MSK cluster with relevant settings: 
  - Name set by `MSKClusterName` parameter.
  - Kafka version `2.6.2`  
  - Each broker running `kafka.t3.small`, with `MSKVolumeSize` of EBS storage
  - Setup to be used unauthenticated (though only accessible by other things on the VPC)
- A managed policy for all S3/SNS/SQS permissions.  

**Outputs:**
- `MSKClusterArn`: ARN of the MSK cluster created by the stack 
- `MSKClusterName`:  Name of the created MSK cluster
- `MSKAccessPolicyArn`:  ARN of the managed policy that has relevant MSK permissions

## `xtdb-alb` (dependent on `xtdb-vpc`)

Sets up application load balancer and all relevant infrastructure for that to route traffic across the ECS service

**Inputs:**
- `VpcId`: ID of the VPC to host the MSK cluster on (from `xtdb-vpc`)
- `SecurityGroupId`: Group ID of the security group to be used by the ECS nodes (from `xtdb-vpc`)
- `PublicSubnets`: List of public subnets to host the load balancer on (from `xtdb-vpc`)

**Outputs:**
- `TargetGroupArn`: ARN of the Target Group created for the nodes
- `LoadBalancerArn`: ARN of the Applicaiton Load Balancer created for the nodes
- `LoadBalancerUrl`: The Load-Balanced XTDB node URL - 'http://${ECSALB.DNSName}:3000'

## `xtdb-ecs` (dependent on `xtdb-vpc`,  `xtdb-msk`, `xtdb-s3` & `xtdb-alb`)

Sets up XTDB on ECS, using EC2 instances. 

**Inputs (grouped in the template):**
- Parameters/resources from other stacks:
  - `SecurityGroupId`: Group ID of the security group to be used by the ECS nodes (from `xtdb-vpc`)
  - `PublicSubnets`: List of public subnets to host the load balancer on (from `xtdb-vpc`)
  - `TargetGroupArn: ARN of the Target Group created for the nodes (from `xtdb-alb`)
  - `LoadBalancerArn: ARN of the  Application Load Balancer created for the nodes (from `xtdb-alb`)
  - `S3BucketName`: Name of the S3 bucket to use as the XTDB object store (from `xtdb-s3`)
  -  `SNSTopicArn`: The ARN of the SNS Topic which the S3 sends notifications to (from `xtdb-s3`)
  - `S3AccessPolicyArn`: ARN of the managed policy offering access to all the S3 permissions necessary for the object store (from `xtdb-s3`)
  - `MSKBootstrapServers Comma separated list containing all Kafka bootstrap server URLs from MSK (needs to be grabbed manually from the MSK cluster info)
  - MSKAccessPolicyArn: ARN of the managed policy offering access to all the MSK permissions (from `xtdb-msk`)
- ECS Configuration:
  - `ClusterName`: Name of the desired ECS cluster -  defaults to `xtdb-cluster`
  -  `EC2InstanceType`: EC2 Instance Type used for ECS Service - defaults to `i3.large` (storage optimized)
  - `DesiredCapacity`: Number of EC2 instances to launch in your ECS cluster / XTDB node tasks to run - defaults to `1`
  - `ImageId:` Used to grab an ECS Optimized Image from SSM Parameter Store - defalts to `/aws/service/ecs/optimized-ami/amazon-linux-2023/recommended/image_id` 

**Setup infrastructure:**
- Sets up an ECS cluster. 
- Sets up a cloudwatch log group for the ECS tasks.
- Sets up an IAM role to be used by the EC2 instances that has access to everything it needs:
    - Makes use of a number of ManagedPolicies, namely the ones from the `s3` stack, `msk` stack and the aws `AmazonEC2ContainerServiceforEC2Role`
- Creates a launch configuration to setup ECS enabled EC2 instances using the image provided from the parameters:
  - AMI id grabbed from SSM parameter store using the `ImageId` parameter.
  - Instance type of these set by parameter, `EC2InstanceType`
  - Has some user data needed to set up the instance as an ECS container, and to mount the NVME drives.  
- Sets up an autoscaling group to create the `DesiredCapacity` amount of EC2 instances.
- Sets up an ECS task for the XTDB node container: 
  - Expects to run on `EC2`, has the relevant capability.
  - NetworkMode: 'bridge' 
  - Uses the `xtdb-aws-ea` docker image.
  - Adds a `xtdb-storage` volume pointed to a location within the EC2's mounted NVME drive, then uses this as a mountpoint within the container definition.  
  - Has a `MemoryReservation` to set a minimum memory limit - should allow the container to use up to the amount of memory on the instanceit runs on.
  - Passes through the various environment variables expected by the node startup, from the cloudformation template:
    - `KAFKA_BOOTSTRAP_SERVERS` - the list of bootstrap server URLs from MSK
    - `XTDB_TOPIC_NAME` - set to xtdb-log
    - `XTDB_S3_BUCKET` - the name of the created S3 bucket.
    - `XTDB_SNS_TOPIC` - the ARN of the created SNS topic.
  - Exposes port 3000 (which the HTTP server runs on)
  - Outputs logs to the created log group. 
- Creates an ECS service:
  - Launch type set to EC2
  - Desired count set to the parameter `DesiredCapacity`
  - Load balancer set up on container port 3000 using the setup application load balancer/target group.
  - Uses the **PlacementConstraint** of `distinctInstance` to ensure one task is ran per EC2 instance. 
  
# Considerations

A few things that may be a bit weird/contentious and I think might require some thoughts/opinions:
- Networking - currently just make it publicly available via a load balancer. Is this desired?
- The ECS service setup, in particular:
  - `DesiredCapacity` being used to configure both how many instances we want, and how many tasks/nodes to run (since we have 1 task per instance) - might well want to rethink this.
  - Partially set up the above to allow ECS tasks to set their minimum memory reservation and then use as much memory as they have available - it **might** be the case we want to set memory hard limits and allow users to configure those (can then tune these however they please based on what EC2 instance image they run) 

# EDIT: Issues with the XTDB http client

I was testing it locally, trying to send some messages through the XTDB client with the load balancer URL (http://ecsalb-1380902940.eu-west-1.elb.amazonaws.com/) - ran into some issues, seems it's giving me back 400s.. Currently not entirely sure why, seems to be somehow related to the use of `hato` with the loadbalancer.

Body comes back `nil`, status `400`, and headers seem pretty strange:
```
{"connection" "keep-alive", "content-length" "54", "content-type" "text/html;charset=iso-8859-1", "date" "Tue, 28 Nov 2023 16:37:10 GMT", "server" "Jetty(11.0.18)"}
```

Wrote up a number of comments of various investigations into this issue - in the end, could see the following:
- When we connected via hato directly to ec2, it works fine.
- Calls to the ALB endpoint from the browser, from CURL, and from clj-http - all worked fine.
- When we connected via hato to the ALB endpoint:
   - It seems to get directed to the target group - ie, actually hits the server.
   - The server then seems to return a 400 - though never logs anything, indicating the server rejects the client request.
   - The loadbalancer then passes back out the 400 response.

Since the message hit the server, looked at Jetty's Debug logs:
- The message sent when connecting directly via ec2 passed successfully, performing a h2c upgrade, no HTTP2Settings header.
- The message sent from the loadbalancer failed, and caused jetty an exception, had a HTTP2Settings header it could not parse.

Explicitly setting the http client to `http-1.1` seemed to fix the issue - so a commit to do so was pushed to `2.x`. Will likely want to explicitly re-instate `http-2` at a later point.


danmason:
There is a fair bit of setup here - so it's a fairly chunky PR 

A few things that may be a bit weird/contentious and I think might require some thoughts/opinions:
- Networking - currently just make it publicly available via a load balancer. Is this desired?
- Passing through the broker list/bootstrap servers - given that this is NOT available on the cloudformation attributes of the MSK cluster, I had to find a way to grab these via a lambda & cloudformation custom resource. 
- The ECS service setup, in particular:
  - `DesiredCapacity` being used to configure both how many instances we want, and how many tasks/nodes to run (since we have 1 task per instance) - might well want to rethink this.
  - Partially setup the above to allow ECS tasks to set their minimum memory reservation and then use as much memory as they have available - might be the case we want to set memory hard limits and allow users to configure those (can then tune these however they please based on what EC2 instance image they run)

danmason:
Another thing to note (just seen this and confirmed it) - with the current setup of one task per instance and instance count = task count - rolling deployments do not just work by default, as it goes to place the new task and fails. You need to manually scale the autoscaling group up (ie, increase the amount of instances) to allow a rolling deployment.

Suspect if we want to stick with the current approach we'll have to do a little bit more config (scalable targets, maybe?) https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-applicationautoscaling-scalabletarget.html

cloudformation/xtdb_vpc.yml
```diff
@@ -0,0 +1,88 @@
...
+      RouteTableId: !Ref PublicRouteTable
+      SubnetId: !Ref PublicSubnet1
+
+  PublicSubnet2RouteTableAssociation:
+    Type: AWS::EC2::SubnetRouteTableAssociation
+    Properties:
+      RouteTableId: !Ref PublicRouteTable
+      SubnetId: !Ref PublicSubnet2
+
+  SecurityGroup:
```
jarohen, line 94: let's tighten this up if we can - assume we'll only want HTTPS from 0.0.0.0/0 and full access only _within_ the security group?

cloudformation/xtdb_msk.yml
```diff
@@ -0,0 +1,161 @@
...
+      - IpProtocol: tcp
+        FromPort: 9094
+        ToPort: 9094
+        SourceSecurityGroupId: !Ref SecurityGroupId
+      - IpProtocol: tcp
+        FromPort: 9092
+        ToPort: 9092
+        SourceSecurityGroupId: !Ref SecurityGroupId
+  
+  MSKSelfIngressAllowRule:
```
jarohen, line 50: does this need a separate resource, or can it be included under `MSKSecurityGroup` with the others?

cloudformation/xtdb_vpc.yml
```diff
@@ -0,0 +1,88 @@
...
+  InternetGateway:
+    Type: AWS::EC2::InternetGateway
+
+  InternetGatewayAttachment:
+    Type: AWS::EC2::VPCGatewayAttachment
+    Properties:
+      InternetGatewayId: !Ref InternetGateway
+      VpcId: !Ref VPC
+
+  PublicSubnet1:
```
jarohen: `PublicSubnetOne`, for consistency with the MSK one?

cloudformation/xtdb_s3.yml
```diff
@@ -0,0 +1,100 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: A stack for setting up XTDB S3 object store and associated resources (SNS topics for changes, XTDB user)
+
+Parameters:
+  S3BucketName:
+    Type: String
+    Default: xtdb-cluster-bucket
```
jarohen: We can't really give a default for the XTDB bucket name, as these have to be globally unique - i.e. as soon as one user uses the default nobody else can

cloudformation/xtdb_alb.yml
```diff
@@ -0,0 +1,79 @@
...
+      Name: ECSALB
+      Scheme: internet-facing
+      LoadBalancerAttributes:
+        - Key: idle_timeout.timeout_seconds
+          Value: '30'
+      Subnets: !Ref PublicSubnets
+      SecurityGroups:
+        - !Ref SecurityGroupId
+
+  ECSTG:
```
jarohen: `ECSTargetGroup`?

cloudformation/xtdb_alb.yml
```diff
@@ -0,0 +1,79 @@
...
+      VpcId: !Ref VpcId
+
+  ALBListener:
+    Type: 'AWS::ElasticLoadBalancingV2::Listener'
+    Properties:
+      DefaultActions:
+        - Type: forward
+          TargetGroupArn: !Ref ECSTG
+      LoadBalancerArn: !Ref ECSALB
+      Port: '3000'
```
jarohen: this the external port? if so, 443?

cloudformation/xtdb_msk.yml
```diff
@@ -0,0 +1,161 @@
...
+    Default: xtdb-cluster-kafka
+    Description: Enter the desired name of the MSK kafka cluster used as the XTDB Kafka Log
+
+  MSKVolumeSize:
+    Type: Number
+    Default: 100
+    Description: The size in GiB of the EBS volume for the data drive on each broker node of the kafka cluster.
+
+Resources:
+  PrivateSubnetMSKOne:
```
jarohen: For MSK you create the subnets within the MSK CF template, but for the ECS ALB you have them passed in - worth picking the same strategy for both? if so, any preferences?

jarohen (commented):
cheers @danmason :pray:

cloudformation/xtdb_ecs.yml
```diff
@@ -0,0 +1,239 @@
...
+        Version: "2012-10-17"
+        Statement:
+          - Sid: ''
+            Effect: Allow
+            Principal:
+              Service: ec2.amazonaws.com
+            Action: 'sts:AssumeRole'
+      ManagedPolicyArns:
+        - 'arn:aws:iam::aws:policy/service-role/AmazonEC2ContainerServiceforEC2Role'
+        - !Ref S3AccessPolicyArn
```
jarohen: can/should these go on the ECS service role rather than the machine itself?

cloudformation/xtdb_ecs.yml
```diff
@@ -0,0 +1,239 @@
...
+    Description: ARN of the managed policy offering access to all the MSK permissions
+    Type: String
+
+  ClusterName:
+    Description: Name of the desired ECS cluster.
+    Type: String
+    Default: xtdb-cluster
+
+  EC2InstanceType:
+    Description: EC2 Instance Type used for ECS Service (We recommend using one that is storage optimized)
```
jarohen, line 46: worth confirming that Docker can use the instance storage, rather than just the (much smaller) root partition of the machine

cloudformation/xtdb_msk.yml
danmason: Hmmm.. It's a fair question. On the one hand, I think I'd prefer to keep them closer with the stuff that's actually using them, so I'm happy to create them within the ALB  template, will need to also expose them from there, too.

On the other... It would really require a LOT of bits moving over from the vpc template, so maybe its better to keep all the VPC related setup together and expect bits to be passed to MSK (I'm okay with passing around lists of subnets and the like - wherever we create them we're going to have to pass them as a parameter to `xtdb-ecs` whatever happens so may as well make a pattern of it)

cloudformation/xtdb_msk.yml
danmason: I think since it's self referential (ie, security group pointing to itself) it _has_ to be done separately: https://stackoverflow.com/questions/46962263/cloudformation-securitygroup-circular-reference

cloudformation/xtdb_vpc.yml
danmason: :+1: Yep - that sounds much better, was being very permissive initially to get stuff working :)

cloudformation/xtdb_msk.yml
danmason: Decided to go with the approach of creating all the subnets in the VPC template and passing references around :)

cloudformation/xtdb_vpc.yml
danmason: WIll similarly have to do a self-referential rule :)

cloudformation/xtdb_alb.yml
danmason: As discussed - port 80 & HTTP for now, will add instructions for setting up HTTPs

Steps would be, in essence:
- Set up SSL certificate in ACM
- Update vpc security group to allow traffic on port 443
- Update ALB exposed port to 443, protocol to HTTPs, add certificate

cloudformation/xtdb_ecs.yml
danmason: Spent a fair bit of time on this! Pushed a commit that ensures the NVME drive gets mounted at `/mnt/nvme-storage` (this doesnt happen automatically, needed to update the launch configuration), and then used that as a volume mount on the task definition/container. 

Had a little play around to be able to SSH onto the node - can confirm the drive seems to get mounted and the volume seems to get made, so pretty happy it _should_ be using the correct storage now :)

cloudformation/xtdb_ecs.yml
danmason: I don't think they belong in the ECS service role - my understanding is that is more for the ECS service to register container instances. I did try placing them in the `TaskExecutionRole` - but that didn't work, it says the following when attempting to set up s3:

```
"User: arn:aws:sts::<accountid>:assumed-role/xtdb-stack-ecs-EC2InstanceRole-cq4KLVh142Ez/i-0221dfee3da46d752 is not authorized to perform: sqs:createqueue on resource: arn:aws:sqs:eu-west-1:<account-id>:xtdb-object-store-notifs-queue-7de3c104-82b7-4463-a224-d93f831776d7 because no identity-based policy allows the sqs:createqueue action (Service: Sqs, Status Code: 403, Request ID: fa4ced72-32cf-51b5-9c7f-b0923e8b76df, Extended Request ID: null)"}}
```

Seems to use the instance role for these things, so planning to leave as is?

danmason:
Further investigations into the problems with the HTTP client:
- Set up the ALB to log requests - outputting to a bucket.
- Setup the JVM to log requests from `hato`

# Shape of `hato` request

From the hato logs, we see the following format of request for a call to status (the response returns a 400):
```
; GET /status HTTP/1.1
; Connection: Upgrade, HTTP2-Settings
; Content-Length: 0
; Host: ecsalb-1380902940.eu-west-1.elb.amazonaws.com
; HTTP2-Settings: AAEAAEAAAAIAAAABAAMAAABkAAQBAAAAAAUAAEAA
; Upgrade: h2c
; User-Agent: Java-http-client/17.0.9
; accept: application/transit+json
; accept-encoding: gzip, deflate
```

We can contrast this to the debug request output from a `clj-http` call, which succeeds:
```
{:config nil,
 :method "GET",
 :requestLine
 #object[org.apache.http.message.BasicRequestLine 0x5c28c0e1 "GET http://ecsalb-1380902940.eu-west-1.elb.amazonaws.com/status HTTP/1.1"],
 :aborted false,
 :params
 #object[org.apache.http.params.BasicHttpParams 0x489ce5fd "[parameters={}]"],
 :protocolVersion
 #object[org.apache.http.HttpVersion 0x7146f22d "HTTP/1.1"],
 :URI
 #object[java.net.URI 0x98ca217 "http://ecsalb-1380902940.eu-west-1.elb.amazonaws.com/status"],
 :class org.apache.http.client.methods.HttpGet,
 :allHeaders
 [#object[org.apache.http.message.BasicHeader 0x24aad7ad "Connection: close"],
  #object[org.apache.http.message.BasicHeader 0x7861bc80 "accept-encoding: gzip, deflate"]]}
```

# Investigating the ALB log of a `400` call from Hato

> Elastic Load Balancing logs requests sent to the load balancer, including requests that never made it to the targets. For example, if a client sends a malformed request, or there are no healthy targets to respond to the request, the request is still logged. Elastic Load Balancing does not log health check requests.Each log entry contains the details of a single request (or connection in the case of WebSockets) made to the load balancer. For WebSockets, an entry is written only after the connection is closed. If the upgraded connection can't be established, the entry is the same as for an HTTP or HTTPS request.

## Anatomy of a failed response to `tx`:

```
http 2023-11-29T11:42:54.420508Z app/ECSALB/7aedb5977f9424f3 <my-ip>:53230 10.192.12.137:32773 0.001 0.001 0.000 400 400 347 231 "POST [http://ecsalb-1380902940.eu-west-1.elb.amazonaws.com:80/tx](http://ecsalb-1380902940.eu-west-1.elb.amazonaws.com/tx) HTTP/1.1" "Java-http-client/17.0.9" - - arn:aws:elasticloadbalancing:eu-west-1:<aws-account-id>:targetgroup/ECSTargetGroup/96e62e2d0f7ad026 "Root=1-656723be-7b45e32914ae34944583abd9" "-" "-" 0 2023-11-29T11:42:54.384000Z "forward" "-" "-" "10.192.12.137:32773" "400" "-" "-"
```

(mapping to syntax from <https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#access-log-entry-format>)

-   **type:** The type of request or connection - `http`
-   **time:** The time when the load balancer generated a response to the client - `2023-11-29T11:42:54.420508Z`
-   **elb:** The resource ID of the load balancer - `app/ECSALB/7aedb5977f9424f3`
-   **client:port:** The IP address and port of the requesting client - `<my-ip>:53230`
-   **target:port:** The IP address and port of the target that processed this request - `10.192.12.137:32773`
-   **request_processing_time:** The total time elapsed (in seconds, with millisecond precision) from the time the load balancer received the request until the time it sent the request to a target - `0.001`
-   **target_processing_time:** The total time elapsed (in seconds, with millisecond precision) from the time the load balancer sent the request to a target until the target started to send the response headers = `0.001`
-   **response_processing_time:** The total time elapsed (in seconds, with millisecond precision) from the time the load balancer received the response header from the target until it started to send the response to the client - `0.000`
-   **elb_status_code:** The status code of the response from the load balancer - `400`.
-   **target_status_code:** The status code of the response from the target. This value is recorded only if a connection was established to the target and the target sent a response - `400`
-   **received_bytes:** `347`
-   **sent_bytes:** `**231**`
-   **"request":** The request line from the client, enclosed in double quotes and logged using the following format: HTTP method + protocol://host:port/uri + HTTP version - `"POST` `[http://ecsalb-1380902940.eu-west-1.elb.amazonaws.com:80/tx](http://ecsalb-1380902940.eu-west-1.elb.amazonaws.com/tx)` `HTTP/1.1"`
-   **"user_agent":** A User-Agent string that identifies the client that originated the request - `"Java-http-client/17.0.9"`
-   **ssl_cipher:** `**-**`
-   **ssl_protocol**: `-`
-   **target_group_arn:** The Amazon Resource Name (ARN) of the target group - `arn:aws:elasticloadbalancing:eu-west-1:<aws-account-id>:targetgroup/ECSTargetGroup/96e62e2d0f7ad026`
-   **"trace_id":** The contents of the **X-Amzn-Trace-Id** header -  `"Root=1-656723be-7b45e32914ae34944583abd9"`
-   **"domain_name":** `-`
-   **"chosen_cert_arn":** `**-**`
-   **matched_rule_priority:** The priority value of the rule that matched the request - If no rule matched and the default action was taken, this value is set to 0 - `0`
-   **request_creation_time:** The time when the load balancer received the request from the client - `2023-11-29T11:42:54.384000Z`
-   **actions_executed:** The actions taken when processing the request, enclosed in double quotes - `forward`  (ie, send to target group)
-   **redirect_url:** `**-**`
-   **error_reason**: The error reason code, enclosed in double quotes. If the request failed, this is one of the error codes described in Error reason codes. If the actions taken do not include an authenticate action or the target is not a Lambda function, this value is set to -  `-`
-   **"target:port_list":** A space-delimited list of IP addresses and ports for the targets that processed this request, enclosed in double quotes - `"10.192.12.137:32773"`
-   **"target_status_code_list":** A space-delimited list of status codes from the responses of the targets, enclosed in double quote - `400`
-   **"classification":** The classification for desync mitigation, enclosed in double quotes. If the request does not comply with RFC 7230, the possible values are Acceptable, Ambiguous, and Severe. If the request complies with RFC 7230, this value is set to -, value was `-`
-   **"classification_reason":** `**-**`

## Observations

It seems as if a few interesting things happen here:
- The message **does** get forwarded to the target group, which returns a 400 from the server.
- The ELB returns a status of `400` - unclear if thats just forwarding the 400 from the server.
- We don't see any indication of the ALB returning an error/hitting desync mitigation.

This is all pretty interesting - current state of things:
- When we connect via hato directly to ec2, it works fine.
- Calls to the ALB endpoint from the browser, from CURL, and from clj-http - all work fine.
- When we connect via hato to the ALB endpoint:
  - It seems to get directed to the target group - ie, actually hits the server.
  - The server then seems to return a 400 - though never logs anything, indicating the server rejects the client request.
  - The loadbalancer then passes back out the 400 response.

danmason:
# Looking at Jetty logs

The following is outputted when using hato to connect to the ALB url for `status`:
```
2023-11-29T13:25:44.918+00:00	13:25:44 | DEBUG o.eclipse.jetty.server.HttpChannel | REQUEST for /status on HttpChannelOverHttp@155fadd7{s=HttpChannelState@74112d5{s=IDLE rs=BLOCKING os=OPEN is=IDLE awp=false se=false i=true al=0},r=2,c=false/false,a=IDLE,uri=http://ecsalb-1380902940.eu-west-1.elb.amazonaws.com/status,age=0}	Link 
2023-11-29T13:25:44.918+00:00	GET /status HTTP/1.1	Link 
2023-11-29T13:25:44.918+00:00	X-Forwarded-For: <my-ip> Link 
2023-11-29T13:25:44.918+00:00	X-Forwarded-Proto: http 	Link 
2023-11-29T13:25:44.918+00:00	X-Forwarded-Port: 80 	Link 
2023-11-29T13:25:44.918+00:00	Host: ecsalb-1380902940.eu-west-1.elb.amazonaws.com 	Link 
2023-11-29T13:25:44.918+00:00	X-Amzn-Trace-Id: Root=1-65673bd8-7aeea04e64c004a9632f9a70 	Link 
2023-11-29T13:25:44.918+00:00	Upgrade: h2c 	Link 
2023-11-29T13:25:44.918+00:00	Content-Length: 0 	Link 
2023-11-29T13:25:44.918+00:00	HTTP2-Settings: AAEAAEAAAAIAAAABAAMAAABkAAQBAAAAAAUAAEAA 	Link 
2023-11-29T13:25:44.918+00:00	User-Agent: Java-http-client/17.0.9 	Link 
2023-11-29T13:25:44.918+00:00	Accept: application/transit+json 	Link 
2023-11-29T13:25:44.918+00:00	Accept-Encoding: gzip, deflate 

13:29:54 | DEBUG org.eclipse.jetty.http.HttpParser | Parse exception: HttpParser{s=CONTENT,0 of 0} for HttpChannelOverHttp@21db7485{s=HttpChannelState@4309a6f{s=IDLE rs=BLOCKING os=OPEN is=IDLE awp=false se=false i=true al=0},r=1,c=false/false,a=IDLE,uri=http://ecsalb-1380902940.eu-west-1.elb.amazonaws.com/status,age=1}
2023-11-29T13:25:44.919+00:00	org.eclipse.jetty.http.BadMessageException: 400: null	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.server.HttpChannelOverHttp.upgrade(HttpChannelOverHttp.java:636)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.server.HttpChannelOverHttp.headerComplete(HttpChannelOverHttp.java:416)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.http.HttpParser.parseFields(HttpParser.java:1256)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.http.HttpParser.parseNext(HttpParser.java:1542)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.server.HttpConnection.parseRequestBuffer(HttpConnection.java:403)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:275)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:314)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:100)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.runTask(AdaptiveExecutionStrategy.java:421)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.consumeTask(AdaptiveExecutionStrategy.java:390)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.tryProduce(AdaptiveExecutionStrategy.java:277)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.run(AdaptiveExecutionStrategy.java:199)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread.run(ReservedThreadExecutor.java:411)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:969)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.doRunJob(QueuedThreadPool.java:1194)	Link 
2023-11-29T13:25:44.919+00:00	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1149)	Link 
2023-11-29T13:25:44.919+00:00	at java.base/java.lang.Thread.run(Thread.java:840)
``` 

In contrast, when we connect via hato DIRECTLY to the public ip of the EC2 instance hosting the node/task, we see the following (successful response):
```
2023-11-29T13:34:11.054+00:00	GET /status HTTP/1.1	Link 
2023-11-29T13:34:11.054+00:00	Content-Length: 0 	Link 
2023-11-29T13:34:11.054+00:00	Host: 34.244.60.251:32774 	Link 
2023-11-29T13:34:11.054+00:00	Upgrade: h2c 	Link 
2023-11-29T13:34:11.054+00:00	User-Agent: Java-http-client/17.0.9 	Link 
2023-11-29T13:34:11.054+00:00	Accept-Encoding: gzip, deflate 
13:34:11 | DEBUG o.eclipse.jetty.server.HttpChannel | onRequestComplete ServerHttpChannelOverHTTP2@49c42738{s=HttpChannelState@58580b0e{s=IDLE rs=BLOCKING os=OPEN is=IDLE awp=false se=false i=true al=0},r=1,c=false/false,a=IDLE,uri=http://34.244.60.251:327/}
``` 

Some notes:
- Seems like jetty throws an exception while trying to parse the HTTP fields.
- We dont pass HTTP2 settings or any of the ALB headers along with the non-alb connection request.
- Might be related to `HTTP2-Settings`?

cloudformation/xtdb_ecs.yml
jarohen: that's weird, I could have sworn AWS gave you a way to give permissions to the individual service/task - they're normally pretty fine grained when it comes to these things, I doubt they'd want to give any service running on that machine the same permissions, and it'd also mean that the permissions of the service depended on the machine it happens to run on?

jarohen (approved):

cloudformation/xtdb_ecs.yml
jarohen: just to clarify, IIUC there's two roles you can put on the task-def:
1. [execution role](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_execution_IAM_role.html), as you say - this is for setting up the container itself, pulling the image, writing logs, etc - i.e. I think this was the wrong one?
2. [task role](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html) - this is the one I thought was for giving permissions for the container to access other services (e.g. S3, SNS, SQS). `TaskRoleArn` on the CFN template.

That article does also [mention](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html#task-iam-role-considerations) to be careful when running containers on EC2 - that the containers inherit the permissions of the box itself.

danmason: Ahhhh - that was it! Got you - yeah, does sound like a **task role** is the way forward. I'll look about bringing that in and testing all reboots and works correctly :)

PR #2958, opened 2023-11-22
Setup and Build AWS Docker Image

### LINKED ISSUE #2895: Docker image for AWS deployments
related to #2888, but:
* includes Kafka module for use with MSK/Confluent Cloud
* includes S3 module for use as a remote obj-store
* ability to configure the above (see #2890)

danmason:
Resolves #2895 

- Sets up a `docker:aws` subproject with a selection of dependencies that it needs and the shadowJar task.
- Setting up an AWS based docker image with #env based configuration.
- Adding AWS docker image to our docker build tasks in CI/ built artifacts

### DIFF
--- a/docker/aws/Dockerfile
+++ b/docker/aws/Dockerfile
@@ -1,0 +1,32 @@
+FROM eclipse-temurin:17
+
+LABEL org.opencontainers.image.source=https://github.com/xtdb/xtdb
+LABEL org.opencontainers.image.description="XTDB 2.x"
+LABEL org.opencontainers.image.licenses="The MIT License"
+
+WORKDIR /usr/local/lib/xtdb
+
+ENTRYPOINT ["java", \
+    "-Dclojure.main.report=stderr", \
+    "-Dlogback.configurationFile=logback.xml", \
+    "--add-opens=java.base/java.nio=ALL-UNNAMED", \
+    "-Dio.netty.tryReflectionSetAccessible=true", \
+    "-cp","xtdb-aws.jar", \
+    "clojure.main", "-m", "xtdb.main"]
+
+HEALTHCHECK --start-period=15s --timeout=3s \
+    CMD curl -f http://localhost:3000/status || exit 1
+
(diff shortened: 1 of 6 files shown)


docker/bin/build-aws-image.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -e
+(
+    cd $(dirname $0)/..
+
+    if [ "$1" == "--clean" ] || ! [ -e build/libs/xtdb-standalone.jar ]; then
```
jarohen: aws?

docker/aws/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+ARG GIT_SHA
+ARG XTDB_VERSION
+ENV GIT_SHA=${GIT_SHA}
+ENV XTDB_VERSION=${XTDB_VERSION}
+
+LABEL org.opencontainers.image.version=${XTDB_VERSION}
+LABEL org.opencontainers.image.revision=${GIT_SHA}
+
+ADD aws_config.edn xtdb.edn
```
jarohen, line 31: I think you can now do:

```docker
COPY <<EOF xtdb.edn
{:xtdb.kafka/log {:topic-name #env XTDB_TOPIC_NAME
                  :bootstrap-servers #env KAFKA_BOOTSTRAP_SERVERS} 
 :xtdb.s3/bucket {:bucket #env XTDB_S3_BUCKET,
                  :prefix "xtdb-object-store"
                  :sns-topic-arn #env XTDB_SNS_TOPIC}
 :xtdb/server {:port 3000}}
EOF
```

if you'd prefer to inline this file?

jarohen (approved):

docker/aws/Dockerfile
danmason: As discussed - decided not to in-line it for now, but no strong preference - just a slight assumption that down the line if we set up configurable docker files that will probably be using external files

PR #2989, opened 2023-11-27
Resolve #1918 node out of sync

### LINKED ISSUE #1918: JDBC NodeOutOfSyncException should not happen after await-tx
There has been a [report](https://clojurians.slack.com/archives/CG3AM2F7V/p1678736012528299) of an Aurora Postgres user (JDBC tx-log and doc-store sharing a connection pool) experiencing the following:

> I just experienced some NodeOutOfSyncExceptions coming from tx-committed? - my code to submit a TX is basically this:
```clojure
(let [submitted-tx (xt/submit-tx node ops)]
  (xt/await-tx node submitted-tx)
  (if (xt/tx-committed? node submitted-tx)
    nil
    (throw (Error. "TX not committed"))))
```
> The message is Node out of sync - requested '{:xtdb.api/tx-id 3970, :xtdb.api/tx-time #inst "2023-03-13T19:15:45.803-00:00"}', available '{:xtdb.api/tx-time #inst "2023-03-13T19:15:45.806-00:00", :xtdb.api/tx-id 3969}'

I speculated that there may something happening in the connection pooling (i.e. using two different connections) that is causing the second API call to return slightly stale results.

Workaround:
> catching that particular error and retry tx-committed? after a short sleep.

Possibly related to https://github.com/xtdb/xtdb/issues/1917 (reported in the same thread)

UDPATE (20230921) Thomas Moerman on has experienced this also (running Postgres in Docker locally) https://clojurians.slack.com/archives/CG3AM2F7V/p1695280975177629?thread_ts=1678736012.528299&cid=CG3AM2F7V

wotbrew:
Solves a rare race condition that can cause out-of-order tx_time values for postgres tx logs.

Fixes #1918 

The postgres dialect is changed to use the STATEMENT_TIMESTAMP (for transactions only).

This is necessary because CURRENT_TIMESTAMP is decided at the start of the transaction before the tx_events table is locked.

History showing the problem:

- a TS0
- b TS1
- b lock wait
- a lock wait
- b lock acq
- b insert TS1
- b lock rel
- a lock acq
- a insert TS0
- a lock rel

I was only able to reproduce this issue for postgres at the current time, so the fix is pg specific.

### DIFF
--- a/modules/jdbc/src/xtdb/jdbc.clj
+++ b/modules/jdbc/src/xtdb/jdbc.clj
@@ -13,8 +13,7 @@
             [xtdb.io :as xio]
             [xtdb.system :as sys]
             [xtdb.tx.event :as txe]
-            [xtdb.tx.subscribe :as tx-sub]
-            [clojure.tools.logging :as log])
+            [xtdb.tx.subscribe :as tx-sub])
   (:import (clojure.lang MapEntry)
            (com.zaxxer.hikari HikariConfig HikariDataSource)
            (java.io Closeable)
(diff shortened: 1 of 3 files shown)


jarohen (approved):
cheers @wotbrew, lgtm :pray:

PR #2990, opened 2023-11-28
Query reference adjustments

Akeboshiwind:
The commits are all self contained so can be reverted/changed if there's anything you disagree with.

The biggest change I made was to have the operator types appear in the sidebar as I think they're super helpful.


jarohen (approved):
nice one @Akeboshiwind :pray: made a minor tweak to the 'operations' heading as the paragraph above is also all about operators - but otherwise all good :slightly_smiling_face:

PR #2991, opened 2023-11-28
Revisions to XTQL tutorial, made into a guide for docs + tests

Akeboshiwind:
- Renamed `dev/xtql/README.md` `XTQL Walkthrough` and made into a guide
- Minimally tweaked some wording
- Added tests
- Use [asciidoc include statements](https://docs.asciidoctor.org/asciidoc/latest/directives/include-tagged-regions/) to include the code directly from the tests
  - NOTE: paths are relative to the root of the docs subfolder, as that's where asciidoctor is run from

Caveats:
- Commented out some parts of the tests:
  - `pull` for JSON
  - `UPDATE` with subquery: https://github.com/xtdb/xtdb/issues/3050
  - `ASSERT`
- Uses yaml string to store json & sql
  - This is because of a few constraints:
        1. The format needs to support multiline strings (json doesn't)
        2. The examples need to be strings (json will later be passed to the http api as `xtql.json` will soon be removed)
        3. The strings themselves need to contain comments for asciidoc so as to not include the containing format
            (i.e. if you have the comments outside of the string it will include at least the surrounding quotes)

Closes xtdb/devrel#20


src/test/clojure/xtdb/tutorial_test.clj
```diff
@@ -14,69 +14,133 @@
...
+   [:put :orders {:xt/id 2, :customer-id :ivan, :currency :gbp, :order-value 150}]])
+
+(def promotions
+  [[:put :promotions {:xt/id 1, :promotion-type :christmas}]
+   [:put :promotions {:xt/id 2, :promotion-type :easter}]])
+
+(def documents
+  [[:put :documents {:xt/id "doc-id", :version 1}]
+   [:put :documents {:xt/id "doc-id2", :version 1}]])
+
```
FiV0: If you only use data once then inline it in the submit call directly. Otherwise this is good.

src/test/clojure/xtdb/tutorial_test.clj
```diff
@@ -14,69 +14,133 @@
...
 
 
 
   (testing "We can look up a single user-id by binding it in the from clause:"
 
-    (xt/q tu/*node*
-          '(from :users [{:xt/id "james"} first-name last-name]))
+    (t/is (= []
+             (xt/q tu/*node*
+                   '(from :users [{:xt/id "james"} first-name last-name]))))
```
FiV0: Can you replace this by something that returns data. (`:ivan` for example).

refset:
@Akeboshiwind I believe you can use `:key-fn :clojure` in the [query opts](https://docs.xtdb.com/reference/main/xtql/queries.html#:~:text=default%20false&#41;.-,%3Akey%2Dfn,-specifies%20how%20keys) for the sql-example tests so that they return the exact same results as the json-examples, and therefore you don't have to use two versions of the results

src/test/clojure/xtdb/tutorial_test.clj
```diff
@@ -0,0 +1,736 @@
...
+    (thrown-with-msg? java.lang.ClassCastException
+                      #"class clojure.lang.PersistentVector cannot be cast to class clojure.lang.MapEntry"
+      (x-json/parse-query
+        #ordered/map {"from" "users"
+                      "bind" ["age"]}))))
+
+(defn sql-example [name]
+  (get examples name))
+
+;; TODO: Inline some
```
FiV0: Still TODO?

src/test/clojure/xtdb/tutorial_test.clj
```diff
@@ -0,0 +1,736 @@
...
+  (let [node tu/*node*]
+    ;; tag::DML-Insert-xtql[]
+    (xt/submit-tx node
+      [(xt/insert-into :users
+         '(from :old-users [xt/id {:given-name first-name, :surname last-name}
+                            xt/valid-from xt/valid-to]))])
+    ;; end::DML-Insert-xtql[]
+    ,)
+
+  ;; TODO: Test valid time
```
FiV0: Still TODO?

src/test/clojure/xtdb/tutorial_test.clj
```diff
@@ -0,0 +1,736 @@
...
+  (xt/submit-tx tu/*node* old-users)
+
+  (t/is (= []
+           (xt/q tu/*node*
+                 '(from :users [first-name last-name]))))
+
+  (xt/submit-tx tu/*node*
+    [(xt/sql-op (sql-example "DML-Insert-sql"))])
+
+  ;; TODO: Test valid time
```
FiV0: Idem

FiV0:
@Akeboshiwind Nide work. Could you explain a bit more in the PR description (for our future selfs) how the (json + sql) from the yaml file as well as the tagged EDN from the test file actually make it into the ASCII doc. Just something that I can come back to, to understand it if the need arrises.

PR #2999, opened 2023-11-29
add `xt/template` macro

jarohen:
Allows `unquote` and `unquote-splicing` within a quoted form (inspiration from ['backtick'](https://github.com/brandonbloom/backtick/)).

e.g.

```clojure
(defn build-posts-query [{:keys [with-author?]}]
  (xt/template (from :posts [{:xt/id id} text
                             ~@(when with-author?
                                 '[author])])))"
```

* Need to requote once you're within an unquote(-splicing) form - same as Clojure macros. The user could re-call `xt/template` in there if it makes sense to.
* This pretty much obviates the need to go to 'mappy XTQL', I think?
* Naming RFC? bikeshedding welcome. I've taken this from 'backtick' itself.
  * `xt/quote`?
  * `xt/quq` (quote-unquote)?
  * I originally had `xt/ql` but, strictly speaking, it doesn't _just_ take a query - it takes any form - so I was wondering about a more general name.
  * :shrug:

### DIFF
--- a/api/src/main/clojure/xtdb/api.clj
+++ b/api/src/main/clojure/xtdb/api.clj
@@ -1,5 +1,6 @@
 (ns xtdb.api
   (:require [clojure.spec.alpha :as s]
+            [xtdb.backtick :as backtick]
             [xtdb.error :as err]
             [xtdb.protocols :as xtp]
             [xtdb.time :as time]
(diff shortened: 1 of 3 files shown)


refset:
> Naming RFC?

`mk`
`gen`
`fmt`
`splice`
`embed`
`inject`
`mix`
`blend`

(thank ChatGPT)

PR #3003, opened 2023-11-29
Fix concurrent schema migration errors for postgres/h2

### LINKED ISSUE #2776: JDBC store initialization from multiple nodes should not race and throw uncaught errors
Original report:

> there is a race condition during the SQL schema creation, present both when using PostgreSQL and embedded H2. In essence, multiple XTDB nodes starting up in parallel detect a missing table in the shared transaction store and try to initialize; as only one can succeed the others throw an exception. The situation likely occurs because checking for and creating the tx_events table and creating a sequence for the offset is not necessarily atomic.

Postgres:
> org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "pg_class_relname_nsp_index"
 Detail: Key (relname, relnamespace)=(tx_events_event_offset_seq, 2200) already exists.

H2:
> object already exists: TX_EVENTS_EVENT_KEY_IDX_2

This behaviour can be worked around but it unnecessarily clutters the logs.

Bonus idea:
> as a hint to possible solutions: database migration tools such as FlyWay also face such situations and they seem to use helper tables to ensure exclusive access

wotbrew:
Fixes #2776

Both postgres and H2 do not guarantee atomicity of CREATE * IF NOT EXISTS statements, even in a serializable transaction.

It is possible for two nodes to race on startup and cause an object already exists exception.

- Use an advisory lock + transactional DDL for postgres

- For H2 I struggled to get locking solutions to work at all, so for now we swallow those specific already exists errors.

### DIFF
--- a/modules/jdbc/src/xtdb/jdbc/h2.clj
+++ b/modules/jdbc/src/xtdb/jdbc/h2.clj
@@ -5,7 +5,8 @@
             [juxt.clojars-mirrors.nextjdbc.v1v2v674.next.jdbc :as jdbc]
             [juxt.clojars-mirrors.nextjdbc.v1v2v674.next.jdbc.result-set :as jdbcr])
   (:import [java.time LocalDate LocalTime OffsetDateTime ZoneOffset]
-           java.util.Date))
+           java.util.Date
+           (org.h2.jdbc JdbcSQLNonTransientException)))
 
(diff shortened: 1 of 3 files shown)


jarohen (approved):
cheers @wotbrew :pray:

PR #3008, opened 2023-11-29
Http response + more remote testing

### LINKED ISSUE #2963: Consistency bash on the Clojure HTTP API
* Check that it reasonably corresponds to the current `xt.api` API
* Ensure that it's documented in the OpenAPI spec (incl. types)

### LINKED ISSUE #2964: HTTP API returns JSON by default
* Transit users (including our Clojure HTTP client) will need to explicitly specify `Accept: application/transit+json`
* For queries, this is to return one result per line - i.e. [newline-delimited JSON](https://ndjson.org/)
* Added to the OpenAPI spec
* N.B. this is decoupled from the HTTP API accepting JSON requests

FiV0:
This PR adds:
- JSON support on the repsonse side.
  - `application/jsonl` for query results
  - `application/json` for anything else 
- Updates the openapi spec (still not fully flashed out as there still changes going on the tx-op side)
- Tests the remote http api a bit more basically.
- A custom jackson deserializer for JSON-LD. Json maps either get parsed to persistent maps or custom objects.

I am encoding non representable JSON values via JSON-LD `:bar -> {"@value" "bar", "@type" "xt:keyword"}`. 

Errors in the the return stream or in faulty return responses are encoded via `xt:error`. For example:
```clj
{"@type" "xt:error",
 "@value" {"xtdb.error/message" "data exception — division by zero",
           "xtdb.error/class" "xtdb.RuntimeException",
           "xtdb.error/data" {...}}}
```


http-server/src/main/clojure/xtdb/server.clj
```diff
@@ -183,7 +216,10 @@
                                                :muuntaja/decode handle-muuntaja-decode-error
                                                ::ri.exception/wrap (fn [handler e req]
                                                                      (log/warn e "response error")
-                                                                     (handler e req))})]
+                                                                     (let [response-format (:raw-format (:muuntaja/response req))]
+                                                                       (cond-> (handler e req)
+                                                                         (#{"appllication/json" "application/jsonl"} response-format)
```
jarohen: thai-po

http-server/src/main/clojure/xtdb/server.clj
```diff
@@ -104,6 +106,33 @@
...
+    (encode-to-output-stream [_ res _]
+      (fn [^OutputStream out]
+        (if-not (ex-data res)
+          (with-open [^IResultSet res res
+                      writer (io/writer out)]
+            (try
+              (doseq [el (iterator-seq res)]
+                (json/write el writer)
+                (.write writer "\n"))
+              ;; TODO how to to mark it as an error for the client
```
jarohen: the response status will likely do this.

in addition, I'd be tempted to return `{:message (.getMessage e), :data (ex-data e)}` here - think you might not even need a special case for `xtdb.RuntimeException` if `ex-data` handles non-IExceptionInfos gracefully

http-server/src/main/clojure/xtdb/server.clj
```diff
@@ -104,6 +106,33 @@
...
+
+    mf/EncodeToOutputStream
+    (encode-to-output-stream [_ res _]
+      (fn [^OutputStream out]
+        (if-not (ex-data res)
+          (with-open [^IResultSet res res
+                      writer (io/writer out)]
+            (try
+              (doseq [el (iterator-seq res)]
+                (json/write el writer)
```
jarohen: If we're expecting users to wrap input types in `{"@value": ..., "type": "..."}` (or similar), we should do the same on output so that the values round-trip correctly.

src/test/clojure/xtdb/remote_test.clj
```diff
@@ -13,3 +20,198 @@
...
+                                :form-params {:query '(from :docs [xt/id])
+                                              :basis {:tx tx}}
+                                :transit-opts xtc/transit-opts
+                                :url (http-url "query")})
+                 :body
+                 decode-transit*
+                 set))
+          "testing query")))
+
+(deftest json-repsonse-test
```
jarohen: tahi-po

src/test/clojure/xtdb/remote_test.clj
```diff
@@ -13,3 +20,198 @@
...
+                              ;; :transit-opts xtc/transit-opts
+                              :url (http-url "query")
+                              :throw-exceptions? false}
+                             #_#_identity identity)
+               (update :body decode-json)
+               (select-keys [:status :body])))
+        "illegal argument error")
+
+  ;; TODO how to mark this as an error
+  (t/is (= {:status 200,
```
jarohen: this should likely return a 400 status, if it doesn't already

http-server/src/main/clojure/xtdb/server.clj
FiV0: I agree. Can we do this via a postwalk before writing the value for now?

src/test/clojure/xtdb/remote_test.clj
FiV0: We are writing the result set to an output stream, when we are hitting this error we might have already written results to the output stream. I did this to replicate the exact behavior of an in memory node. So our client currently only throws when it actually consumes this error from the result set. So in transit you get something like 
```clj
{:foo "bar} ;; row 1
{:foo "bar"} ;; row 2
#xt/runtime-error {.....}
```
and only when one consumes the 3rd row the client throws. I don't know how to best represent this as an http result code.

http-server/src/main/clojure/xtdb/server.clj
FiV0: regarding repsonse status, see my comment below.

FiV0:
```clj
(try 
  (/ 1 0)
  (catch Throwable t
    {:message (.getMessage t) :data (ex-data t)}))
;; => {:message "Divide by zero", :data nil}

(try 
  (/ 1 0)
  (catch Throwable t
    (ex-data (throwable->ex-info t))))
;; => {:xtdb.error/error-type :unknown-runtime-error,
;;     :class "java.lang.ArithmeticException",
;;     :stringified "java.lang.ArithmeticException: Divide by zero"}
```
I'd prefer the latter.

src/test/clojure/xtdb/remote_test.clj
jarohen: ah, kk, I see. there'll likely be an answer for this in the HTTP standard - we shouldn't make one up :slightly_smiling_face:

http-server/src/main/clojure/xtdb/server.clj
FiV0: Jackson it is.

http-server/src/main/clojure/xtdb/server.clj
FiV0: We are going for one error type for all of JSON.

http-server/src/main/clojure/xtdb/jackson.clj
```diff
@@ -0,0 +1,72 @@
...
+     (encoder value gen)
+     (.writeEndObject gen))))
+
+(defn encode-throwable [^Throwable t ^JsonGenerator gen]
+  (let [mapper ^ObjectMapper (.getCodec gen)]
+    (.writeStartObject gen)
+    (.writeStringField gen "xtdb.error/message" (.getMessage t))
+    (.writeStringField gen "xtdb.error/class" (.getName (.getClass t)))
+    (.writeFieldName gen "xtdb.error/data")
+    (.writeRawValue gen (.writeValueAsString mapper (ex-data t)))
```
jarohen: do either `writeObject` or `writePOJO` work here, rather than going down to `writeRawValue`?

http-server/src/main/clojure/xtdb/jackson.clj
```diff
@@ -0,0 +1,72 @@
...
+  [{:keys [handlers]}]
+  (let [decoders (->> (for [[_ {:keys [tag decode]}] handlers] [tag decode]) (into {}))]
+    (reduce-kv
+     (fn [^SimpleModule module t {:keys [tag encode] :or {encode jt/encode-str}}]
+       (.addSerializer module t (serializer tag encode)))
+     (doto (SimpleModule. "JSON-LD")
+       (.addDeserializer Map (JsonLdValueOrPersistentHashMapDeserializer. decoders)))
+     handlers)))
+
+(def handlers {Keyword {:tag "xt:keyword"
```
jarohen, line 57: let's land these for now - in future we'll want to use standard JSON-LD types for these where they're available. e.g. `xsd:dateTime` etc (will continue to use `xt:` when a standard doesn't exist). 

Sets are handled separately, if memory serves - JSON-LD works with sets by default, with ordered Lists being the exception to the rule that you specify explicitly.

http-server/src/main/java/xtdb/jackson/JsonLdValueOrPersistentHashMapDeserializer.java
```diff
@@ -0,0 +1,72 @@
+package xtdb.jackson;
```
jarohen, line 1: Given this is significantly derived from Jsonista I'd add a notice to the top of this file with the Jsonista license - see `xtdb.backtick` for an example.

http-server/src/main/java/xtdb/jackson/JsonLdValueOrPersistentHashMapDeserializer.java
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    @SuppressWarnings("unchecked")
+    public Object deserialize(JsonParser jp, DeserializationContext ctxt) throws IOException, JsonProcessingException {
+        ObjectCodec mapper = jp.getCodec();
+        ObjectNode node = mapper.readTree(jp);
+
+        if (node.has("@type")) {
+            IFn decode = this.decoders.get(node.get("@type").asText());
+            if (decode != null) {
+                Object o = decode.invoke(mapper.readValue(node.get("@value").traverse(mapper), Object.class));
```
jarohen: nit: `return decode.invoke(..)`

http-server/src/main/clojure/xtdb/jackson.clj
```diff
@@ -0,0 +1,72 @@
...
+                        :encode jt/encode-keyword
+                        :decode keyword}
+               PersistentHashSet {:tag "xt:set"
```
jarohen: can we use `java.util.Set` here, just in case there are other concrete set types?

http-server/src/test/clojure/xtdb/remote_test.clj
```diff
@@ -0,0 +1,217 @@
...
+                           (if (instance? EOFException (.getCause e))
+                             ::finished
+                             (throw e))))]
+        (if (= ::finished value-read)
+          res
+          (recur (conj res value-read)))))))
+
+(defn- decode-transit [^String s] (first (decode-transit* s)))
+
+(def ^:private default-object-mapper
```
jarohen, line 43: worth us making the `xtdb.jackson` one available so that we're testing the real version?

http-server/src/test/java/xtdb/jackson/JsonLdValueOrPersistentHashMapDeserializerTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+class JsonLdValueOrPersistentHashMapDeserializerTest {
+
+    private final ObjectMapper objectMapper;
+    private IFn keywordFn = Clojure.var("clojure.core", "keyword");
+    private IFn setFn = Clojure.var("clojure.core", "set");
+
+    public JsonLdValueOrPersistentHashMapDeserializerTest() {
+        ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.configure(JsonParser.Feature.ALLOW_SINGLE_QUOTES, true);
```
jarohen: In the tests we have (at least) Java 17 available, which means we have text blocks:
```java
"""
{"@value": "...", "@type": "..."}
"""
```

Would prefer we used this rather than single quotes because it'll be more representative of the real world systems

http-server/src/main/resources/openapi.yaml
```diff
@@ -3,92 +3,189 @@ info:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Status'
   /tx:
     post:
       summary: Transaction
-      description: Submits a transaction to the cluster
+      description: > 
+        Submits a transaction to the cluster. 
+        See [XTQL Transactions](https://docs.xtdb.com/reference/main/xtql/txs) for details about different transaction operations.
```
jarohen: like this :ok_hand: same for queries please :)

these are also the endpoints for SQL txs/queries - could that link be included in the description too?

http-server/src/main/resources/openapi.yaml
```diff
@@ -3,92 +3,189 @@ info:
...
+  schemas:
+    Timeout:
+      description: timeout in milliseconds
+      type: integer
+    DefaultTz:
+      type: string
+      example: "America/Los_Angeles"
+    SystemTime:
+      type: string
+      example: "2024-01-01T00:00"
```
jarohen: examples for date-time here should end in `Z`. is there a `format` for 'date-time with TZ', or is this included in `date-time`?

http-server/src/test/java/xtdb/jackson/JsonLdValueOrPersistentHashMapDeserializerTest.java
```diff
@@ -0,0 +1,74 @@
...
+        SimpleModule module = new SimpleModule("JsonLd-test");
+
+        HashMap<String, IFn> decoders = new HashMap<String, IFn>();
```
jarohen, line 27: Given most (all?) of the encoders/decoders are Java methods, is it worth us having these written in Java, so that (again) you're testing real code here rather than having to replicate it in the tests?

jarohen (approved):
This is great @FiV0, cheers! :pray: some minor review comments below, but once they're addressed we're good to go :rocket:

http-server/src/main/clojure/xtdb/jackson.clj
FiV0: :+1:

http-server/src/test/java/xtdb/jackson/JsonLdValueOrPersistentHashMapDeserializerTest.java
FiV0: I'd prefer the IFn here as I would like to be able to easily extend this without writing any Java.

PR #3012, opened 2023-11-29
basic support for json roundtripping

FiV0:
This adds two little functions that help in roundtripping rows unlossy.


FiV0:
We are going to use Jackson Tagged values + out of band schemas for this.

PR #3030, opened 2023-12-07
XTDML builders -> macros so that `template` is done for you

jarohen:
I'm not 100% sure about this one - RFC.

> [They] were so preoccupied with whether or not they could, they didn’t stop to think if they should.

Essentially this makes `template` implicit in the XTDML builders. e.g.

```clojure
(-> (xt/update-table :users {:bind {:xt/id $uid, :version v}
                             :set {v (inc v)}})
    (xt/with-op-args {:uid :jms}))
```

* Any place that takes queries within these builders is auto-quoted, and unquote (`~`, `~@`) is fair game.
* Pro: users don't have to remember which parameters they need to quote these builders - they now quote nothing.
* Con: too much macro magic? doesn't look like a macro? users might just expect `v` to be a local variable pulled from the surrounding Clojure function?

Atm I'm probably a -0.3 - it's cool that we _can_ do this, and I thought it was worth seeing what it looked like, but ... :thinking:


jarohen:
I'll preemptively close this one on the assumption it's gone too far - if there's interest, happy to have a discussion.

PR #3033, opened 2023-12-11
Deserializers

### LINKED ISSUE #3044: Jackson Deserializers

FiV0:
This brings in a set of custom jackson deserializers for queries. We mostly check that required keys and optional keys parse correctly, but do not check for extra keys unrelated to a tx/query. 

- [x] ~custom decoding from CamelCase to kebab-case in documents~ snake_case everywhere 
- [x] ~pass to CamelCase for other constructs~ snake_case everywhere
- [x] ~`TransactionKey` is a java class in this PR. We need to separately remove `xtdb.api.TransactionKey` and intergrate it seperately.~ thanks @danmason 
- ~There are still some outstanding issues regarding encoding of ambiguous JSON types in Expr positions. (str literal vs symbol and exprCall vs Map literal).~ See  https://github.com/xtdb/xtdb/pull/3033#issuecomment-1858511341 below.

## Query
- [x] From 
- [x] Pipeline
- [x] Unify
- [x] Expr handling
- [x] BindDeserializer (outSpec)
- [x] ArgDeserialier (argSpec)
- [x] aggregate
- [x] limit
- [x] offset
- [x] join (join, left-join)
- [x] order-by
- [x] return
- [x] rel
- [x] unnest
- [x] where
- [x] with
- [x] without
- [x] subqueries
- [x] clauseDeserializer
- [x] QueryTailDeserializer
- [x] `xtdb/malformed ...` -> `xtql/malformed ...` everywhere
- [x] don't wrap `xtdb.IllegalArgumentException` in other `xtdb.IllegalArgumentException` throughout
- [x] deal with multiple elements in one ArgSpec/ColSpec/OutSpec map
- [ ] ~move deserializeBind to some abstract class?~
- [x] refactor Join/LeftJoin?
- [x] add currentTime to Basis
- [ ] from more complex `valid-time/system-time` form


api/src/main/java/xtdb/query/Query.java
```diff
@@ -311,7 +311,12 @@ static Call call(String ruleName, List<Expr> args) {
         return new Call(ruleName, args, null);
     }
 
-    final class Join implements UnifyClause {
+    abstract class AJoin implements UnifyClause {
```
jarohen: I'd go for `interface IJoin` here unless you specifically need an abstract class - there are fewer constraints on interface usage. I'd also not implement `UnifyClause` here - you can have both Join and LeftJoin implementing these.

More generally, this is what the principle of 'favouring composition over inheritance' actually looks like in practice :slightly_smiling_face:

api/src/main/java/xtdb/query/Query.java
```diff
@@ -360,13 +366,14 @@ private LeftJoin(Query query, List<ArgSpec> args, List<OutSpec> bindings) {
...
 
+        @Override
         public LeftJoin binding(List<OutSpec> bindings) {
             return new LeftJoin(query, args, bindings);
         }
 
         @Override
         public String toString() {
-            return stringifySeq("left-join", stringifyArgs(query, args), stringifyList(bindings));
+            return stringifySeq("join", stringifyArgs(query, args), stringifyList(bindings));
```
jarohen: left-join?

http-server/src/main/java/xtdb/jackson/CallDeserializer.java
```diff
@@ -0,0 +1,36 @@
...
+public class CallDeserializer extends StdDeserializer<Call> {
+
+    public CallDeserializer() {
+        super(Call.class);
+    }
+
+    @Override
+    public Call deserialize(JsonParser p, DeserializationContext ctxt) throws IllegalArgumentException, IOException {
+        ObjectMapper mapper = (ObjectMapper) p.getCodec();
+        BaseJsonNode node = mapper.readTree(p);
```
jarohen: it's not idiomatic to rely on `try`/`catch` in Java for casting exceptions (if that's what this and the other cases are trying to do) - best to `if (node instanceof ObjectNode) { ... } else { throw ...}`

http-server/src/main/java/xtdb/jackson/TxDeserializer.java
```diff
@@ -0,0 +1,56 @@
...
+    public Tx deserialize(com.fasterxml.jackson.core.JsonParser jp, DeserializationContext ctxt) throws IOException, JsonProcessingException {
+        ObjectMapper mapper = (ObjectMapper) jp.getCodec();
+        ObjectNode node = mapper.readTree(jp);
+        List<Ops> ops = new ArrayList<>();
+        LocalDateTime systemTime = null;
+        ZoneId defaultTz = null;
+
+        try {
+            if (node.has("tx_ops")) {
+                // The reason that we are fully walking the nodes is that Jsonista overwrites the standard `List` parser and
```
jarohen: are we getting any benefit from the Jsonista Clojure module now, or is it worth us going straight to Jackson? i.e. are we ever deserialising any Clojure data structures in this API?

http-server/src/main/java/xtdb/jackson/TxDeserializer.java
```diff
@@ -0,0 +1,56 @@
...
+            } else {
+                throw IllegalArgumentException.create(Keyword.intern("tx", "missing-tx-ops"), PersistentHashMap.EMPTY);
+            }
+            if (node.has("system_time")) {
+                systemTime = (LocalDateTime) mapper.readValue(node.get("system_time").traverse(mapper), Object.class);
+            }
+            if (node.has("default_tz")) {
+                defaultTz = (ZoneId)  mapper.readValue(node.get("default_tz").traverse(mapper), Object.class);
+            }
+        } catch (Exception e) {
```
jarohen: the exception handling is a little strange here - you're throwing IAE on L42 but then catching and ignoring it here - let's pair up and talk about the wider strategy?

http-server/src/main/java/xtdb/query/AJoinDeserializer.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public Query.AJoin deserialize(JsonParser p, DeserializationContext ctxt) throws IOException, JsonProcessingException {
+        ObjectMapper mapper = (ObjectMapper) p.getCodec();
+        ObjectNode node = mapper.readTree(p);
+
+        try {
+            Query.AJoin join = null;
+            if (node.has("join")) {
+                join = Query.join(mapper.treeToValue(node.get("join"), Query.class), deserializeArgs(mapper, (ArrayNode) node.get("args")));
+            } else {
```
jarohen: wouldn't necessarily assume `node.has("left_join")` here, I don't think? could we specify this in an `else if` and then throw an error in the case that it's neither?

http-server/src/main/java/xtdb/query/AggregateDeserializer.java
```diff
@@ -0,0 +1,44 @@
...
+            res.add(mapper.treeToValue(argSpecNode, ColSpec.class));
+        }
+        return res;
+    }
+    public Query.Aggregate deserialize(JsonParser p, DeserializationContext ctxt) throws IOException, JsonProcessingException {
+        ObjectMapper mapper = (ObjectMapper) p.getCodec();
+        ObjectNode node = mapper.readTree(p);
+
+        try {
+            return Query.aggregate(deserializeColSpec(mapper, (ArrayNode) node.get("aggregate")));
```
jarohen: rather than casting here, could we pass the raw node through to `deserializeColSpec` and have _that_ throw if it's the wrong type? seems like you're doing part of its job for it

http-server/src/main/java/xtdb/query/Basis.java
```diff
@@ -0,0 +1,6 @@
+package xtdb.query;
+
+import java.time.Instant;
+
+public record Basis(TransactionKey atTx, Instant currentTime){};
```
jarohen: this is great - I suspect this will become part of the Java API, so could it go in the `api` module instead? (and hence become a regular class, because Java 11)

http-server/src/main/java/xtdb/query/FromDeserializer.java
```diff
@@ -0,0 +1,45 @@
...
+
+            ArrayNode bindArray= (ArrayNode) node.get("bind");
+            for (JsonNode outSpecNode: bindArray) {
+                outSpecs.add(mapper.treeToValue(outSpecNode, OutSpec.class));
+            }
+            return query.binding(outSpecs);
+        } catch (IllegalArgumentException i) {
+            throw i;
+        } catch (Exception e) {
+            throw IllegalArgumentException.create(Keyword.intern("xtql", "malformed-from"), PersistentHashMap.create(Keyword.intern("json"), node.toPrettyString()), e);
```
jarohen: also, with `catch Exception`, rethrow IAE, you're assuming that any exception thrown in the child is a malformed-from - but you likely don't want to catch IOException and respond in this way, for example. do we need to catch and re-throw here?

http-server/src/main/java/xtdb/query/LimitDeserializer.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @Override
+    public Query.Limit deserialize (JsonParser p, DeserializationContext ctxt) throws IOException, JsonProcessingException {
+        ObjectMapper mapper = (ObjectMapper) p.getCodec();
+        JsonNode node = mapper.readTree(p);
+
+        try {
+            JsonNode limit = node.get("limit");
+            if (limit != null && limit.isNumber()) {
```
jarohen: does `isNumber` return true for floats too?

http-server/src/main/java/xtdb/query/QueryMap.java
```diff
@@ -0,0 +1,11 @@
+package xtdb.query;
+
+import clojure.lang.Keyword;
+
+import java.time.Duration;
+import java.time.ZoneId;
+import java.util.Map;
+
+public record QueryMap(Query query, Map<Keyword, Object> args, Basis basis, TransactionKey afterTx, Duration txTimeout,
```
jarohen: similarly to `Basis`, I suspect this is a candidate Java API class.

nit: naming: `QueryRequest`?

jarohen: also, `boolean` rather than `Boolean` if it can never be null

jarohen: and, if it is to become the Java API, we should expect `Map<String, Object> args` here, rather than making the Java folks write Clojure keywords

http-server/src/main/java/xtdb/query/TransactionKey.java
```diff
@@ -0,0 +1,6 @@
+package xtdb.query;
+
+import java.time.Instant;
+
+// TODO remove xtdb.api.TransactionKey and replace with this
+public record TransactionKey(Long txId, Instant systemTime) {}
```
jarohen: similarly to `Basis` and `QueryMap` - can this go in `api`?

jarohen: and also `long` rather than `Long`, don't think `txId` should ever be null?

http-server/src/main/java/xtdb/query/UnifyClauseDeserializer.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public Query.UnifyClause deserialize (JsonParser p, DeserializationContext ctxt) throws IOException, JsonProcessingException {
+        ObjectMapper mapper = (ObjectMapper) p.getCodec();
+        ObjectNode node = mapper.readTree(p);
+
+        try {
+            if (node.has("from")) {
+                return mapper.treeToValue(node, Query.From.class);
+            } if (node.has("where")) {
```
jarohen: heh, this works because your `if`s return, but I missed that these weren't `else if`s, given the indentation :sweat_smile:

http-server/src/test/java/xtdb/jackson/TxDeserializerTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class TxDeserializerTest {
+
+    private final ObjectMapper objectMapper;
+
+    public TxDeserializerTest() {
+        ObjectMapper objectMapper = new ObjectMapper();
+        SimpleModule module = new SimpleModule("JsonLd-test");
```
jarohen: I'm not a fan that these don't test the actual code - let's see what it takes to get the module itself into Java?

jarohen (changes requested):
hey folks - have had a run through, looking good so far :slightly_smiling_face: 

plenty of Java-related feedback here to run through, give me a shout if any of it is unclear.

cheers!

http-server/src/main/java/xtdb/jackson/TxDeserializer.java
FiV0: Yes, when we deserializing literals.

api/src/main/java/xtdb/query/Query.java
FiV0: If I don't implement `UnifyClause` at that level I need to write an extra deserializer for `join` and `left-join` which I tried to avoid in https://github.com/xtdb/xtdb/pull/3033/commits/bcd8375845d7ec19e9950bd4433e2d213dc44484. I can also just revert that.

FiV0: Addressed the interface bit.

http-server/src/main/java/xtdb/jackson/TxDeserializer.java
FiV0: yes, I want to keep the most inner IAE exception.

http-server/src/main/java/xtdb/query/AJoinDeserializer.java
FiV0: I mainly do it this way because the calling class assures there is one or the other. See `UnifyClauseDeser`.

http-server/src/main/java/xtdb/jackson/TxDeserializer.java
FiV0: My approach has been to write something like 
```java
  } catch (IllegalArgumentException i) {
           throw i;
  } catch (Exception e) {
           throw IllegalArgumentException.create(Keyword.intern("xtdb", "malformed-tx"), PersistentHashMap.create(Keyword.intern("json"), node.toPrettyString()));
  }
```
which I didn't do here.

http-server/src/main/java/xtdb/query/FromDeserializer.java
FiV0: See my comment above. We could catch something like `JsonProcessingException` in the second case?

http-server/src/test/java/xtdb/jackson/TxDeserializerTest.java
FiV0: Do you want me to move all mappers to Java?

FiV0: I would prefer the mappers to stay in `jackson.clj` to quickly add stuff as it comes up.

http-server/src/main/java/xtdb/query/QueryMap.java
danmason: Still yet to update the types of `args` here but I have moved it - shall update that on monday.

FiV0: Was just simple String replacement.

FiV0:
@jarohen Writing up the state of the Expr ambigiuty handling. Everything at the expression level is handled with normal non wrapped maps/normal json-ld types as in all other places. The exception are the following string json keys which are treated specially by the `ExprDeserializer`.
- `xt:exists`, `xt:q`, `xt:pull` and `xt:pull-many` which have a query as their value and the map also contains a `bind` key value pair
- `xt:call` which has a standard library fn as their value and an optional `args` key + vector pair.

The desision to parse these specially is based on the `xt:...` key and not on the map containing `bind` or `args`.
 
- all other values (including json-ld values) are treated as literals and parsed into expressions accordingly
- I have not taken a decision on literal strings. We could either create a special key for either string `{"xt:str" "foo"}` or a special key for symbols/vars/params `{"xt:var" "first_name"}`.


Some examples of the above ("simple" strings are treated as vars):
```clj
{"xt:call" "+"
 "args" ["foo" "bar"]}

;; =>
(Expr/call "+" [(Expr/lVar "foo") (Expr/lVar "bar")])

{"foo" 1}
;; =>
(Expr/map {"foo" (Expr/val 1)})

{"@type" "xt:set"
 "@value" [1 2]}

;; =>
(Expr/set #{(Expr/val 1) (Expr/val 2)})  
```

http-server/src/main/java/xtdb/jackson/TxDeserializer.java
jarohen: right - I wouldn't go with the `catch Exception` here - if it's an IAE, you want the inner one (as you say), maybe with some context at a push (but that's not what's happening here); if it's not an IAE (e.g. a general `IOException`) then it's likely not a malformed-tx

jarohen:
> @jarohen Writing up the state of the Expr ambiguity handling. Everything at the expression level is handled with normal non wrapped maps/normal json-ld types as in all other places. The exception are the following string json keys which are treated specially by the ExprDeserializer.

@FiV0 yep, aware this isn't what we started with but on balance I think this is a good call (pardon the pun) :clap: 

couple of clarifications:

* I'd double check exactly how JSON-LD handles discriminated unions (if indeed it does) and go with that - if it's distinguishing on compacted IRIs (essentially what this is doing) then that's fair game.
* Otherwise, I might be tempted to go with `@call` etc, which is how other built-ins are handled in JSON-LD.
* In long-form exprs, given we're moving to making literal maps the easy case, I'd also move to making literal strings the easy case - i.e. `xt:lvar`/`@lvar` being explicitly declared.
  * Separately, we could consider (given it's unambiguous) making string lvars available in short-forms - i.e. having `{"bind": ["foo", "bar", {"baz": {"@lvar": "renamedBaz"}}]}`

jarohen:
> Otherwise, I might be tempted to go with `@call` etc, which is how other built-ins are handled in JSON-LD.

Scratch this, we probably shouldn't - we should respect JSON-LD's space for reserved keywords. 

AFAICT `xt:call` _would_ be how JSON-LD represents discriminated unions - although I think it would then put the args in a sub-structure (rather than sibling keys). so `{"xt:call": ["+" & args]}`?

edit: although in JSON-LD, the vector syntax means that `xt:call` has many values - so map `{:f "+", :args [...]}` seems more idiomatic

http-server/src/test/java/xtdb/jackson/TxDeserializerTest.java
jarohen: I can see why this is preferable for development turnaround time in the short-term; long-term, we likely won't be adding things as frequently, and the value of testing the real code will overtake it. With that in mind, maybe once we've bottomed it out, we make the migration?

FiV0: All the "real" tests sit in Clojure anyway. Options are:
- remove all java tests, keep the rest as it is.
- move mappers to java to properly test the two Java deserializers tests I have written

http-server/src/main/java/xtdb/jackson/TxDeserializer.java
FiV0: As suggested below, we could use `JsonProcessingException` in place of Exception here. For instance, I don't do a `instanceof` checks everywhere. So there are definitely malformed tx/queries which won't get caught if I remove the second case.

PR #3047, opened 2023-12-14
Extract `TransactionKey` to java class, replace clojure record, update calls

danmason:
- Splitting out `TransactionKey` to separate Java class.
- Updating numerous usages of `TransactionKey` to use the new class, replacing various clojure record method calls.
- Updating the data readers/writers for `tx-key`

### DIFF
--- a/api/src/main/java/xtdb/api/TransactionKey.java
+++ b/api/src/main/java/xtdb/api/TransactionKey.java
@@ -1,0 +1,53 @@
+package xtdb.api;
+
+import java.time.Instant;
+import java.util.Objects;
+
+public class TransactionKey implements Comparable<TransactionKey> {
+    public long txId;
+    public Instant systemTime;
+
+    public TransactionKey(long txId, Instant systemTime) {
+        this.txId = txId;
+        this.systemTime = systemTime;
+    }
+
+    @Override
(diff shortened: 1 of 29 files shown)


api/src/main/java/xtdb/api/TransactionKey.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public long getTxId() {
+        return txId;
+    }
+
+    public Instant getSystemTime() {
+        return systemTime;
+    }
+
+    public TransactionKey setTxId(long txId) {
```
jarohen: no setters pls :cold_sweat:

modules/bench/src/main/clojure/xtdb/bench2/xtdb2.clj
```diff
@@ -92,20 +92,23 @@
...
              0)))
 
         compute-lag-abs
         (fn []
           (let [{:keys [latest-completed-tx] :as res} (xt/status node)]
             (or
              (when-some [[fut _] @last-submitted]
-               (let [tx-id (:tx-id @fut)]
-                 (when-some [{completed-tx-id :tx-id} latest-completed-tx ]
+               (let [tx-id (.txId @fut)
```
jarohen, line 109: reflection warning here?

PR #3051, opened 2023-12-14
Add Expr types `Set`, `Vec`, `Map` and `Null`

FiV0:


core/src/main/clojure/xtdb/xtql.clj
```diff
@@ -126,13 +132,20 @@
   (plan-expr [this] (param-sym (subs (.v this) 1)))
   (required-vars [_this] #{})
 
+  Expr$Vec
+  (plan-expr [this] (into [] (map plan-expr) (.elements this)))
+  (required-vars [this] (into #{} (mapv required-vars (.elements this))))
```
jarohen, line 137: needs to be mapcat in required vars as each child element will also return a collection

api/src/main/java/xtdb/query/Expr.java
```diff
@@ -3,16 +3,39 @@
...
+            return false;
+        }
+
+        @Override
+        public int hashCode() {
+            return 1;
+        }
+    }
+
+    static Null val() {
```
jarohen, line 35: this wants to be a static constant, like the Bool ones

api/src/main/java/xtdb/query/Expr.java
```diff
@@ -3,16 +3,39 @@
...
 import java.util.stream.Collectors;
 
 import static xtdb.query.QueryUtil.stringifyArgs;
 import static xtdb.query.QueryUtil.unmodifiableList;
 
 public interface Expr {
+
+    final class Null implements Expr {
+
+        public Null() {}
```
jarohen, line 16: private constructor so that we know that there's only one instance

api/src/main/java/xtdb/query/Expr.java
```diff
@@ -322,6 +345,81 @@ static Exists exists(Query query, List<ArgSpec> args) {
         return new Exists(query, args);
     }
 
+    final class Vec implements Expr {
```
jarohen, line 348: can this be List please, so that it's named for Java folks?

api/src/main/java/xtdb/query/Expr.java
```diff
@@ -322,6 +345,81 @@ static Exists exists(Query query, List<ArgSpec> args) {
...
+            return Objects.hash(elements);
+        }
+    }
+
+    static Vec vec(List<Expr> elements){
+       return new Vec(elements);
+    }
+
+    final class Set implements Expr {
+        public final java.util.Set<Expr> elements;
```
jarohen, line 374: at the risk of some inconsistency, I'd be tempted to call these three classes SetExpr so that you don't need the FQ java.util.Set

core/src/main/clojure/xtdb/xtql.clj
```diff
@@ -126,13 +132,20 @@
...
+
+  Expr$Set
+  (plan-expr [this] (into #{} (map plan-expr (.elements this)) ))
+  (required-vars [this] (into #{} (mapv required-vars (.elements this))))
+
+  Expr$Map
+  (plan-expr [this] (into {} (map (juxt (comp keyword util/str->normal-form-str key) (comp plan-expr val)) (.elements this))))
+  (required-vars [this] (into #{} (mapv required-vars (vals (.elements this)))))
```
jarohen, line 145: mapcat

jarohen (changes requested):

Issue #3043, opened 2023-12-14
Java API support


FiV0:
This ticket tracks work for moving the user specified data in the system from Clojure to Java.
- [x] parse clojure maps to corresponding JAVA api objects and update the places in `impl.node` and downwards accordingly. Probably some layer between node and xtdb.api. An example being `xtdb.query.Basis`.
- [x] instead of putting Clojure maps on the transaction log put Java maps with `Map<String, Object>` 
- [x] there are potentially also some changes to how we write documents

jarohen:
We're in a better place with this now - will close this card and create subsequent cards for future work.

PR #3060, opened 2023-12-19
Tweaks preparing node for a Java API

FiV0:
part of #3043

This PR adds a couple of changes mostly to the `node/impl.clj` to allow for the Java objects to be directly submitted through our submit and query protocols. This will make writing our JAVA API a lot simpler as well as opens the door for using the deserializers (`JSON` -> `Java Objects` -> `Node`) for the JSON request side.

- [x] ~The last commit is not really related to this PR, but I somehow had to change some transit files. Might be good to figure out why some files did not need to get updated when we brought in the `TransactionKey` class.~
- [x] Figure out how Put should be represented `Map<Keyword, Object>` vs `Map<String, Object>`.
  * Represent as `Map<?, ?>`.
  * Printing the result of `xt/put`, reading `#xt/put`: prints as the user entered.
  * Normalised as before, in tx-producer.
    * Accept the risk that if a user puts two attributes in that normalise to the same key (either in case-sensitivity, or keyword/string) then the behaviour is undefined.


FiV0:
@jarohen One thing I did not touch in this PR, but which is probably relevant for the Java API is that put operations still use a keyword for the table. I can also remove it in this one, but it's not important for JSON.

jarohen:
yep, we'll want tables to be strings eventually too

Issue #3083, opened 2024-01-08
Non-Clojure users configuring an XTDB node


jarohen:
Java users will be:
* configuring this in-process (likely for tests, but assume they may well want a full in-process node too)
* starting up remote clients

Users using the Docker image (outside of the AWS template) will also want to configure the node (e.g. Kafka, S3).

* Some analysis and prior art research required here - let's enumerate our options with benefits and tradeoffs.
* Implementation (obv)
* Documentation

---

* [x] #3096 
* [x] #2890 
* [x] #3112

jarohen:
See #2890 for an earlier idea (but don't feel beholden to it if there's a better approach).

Issue #2890, opened 2023-11-02
CLI/API to support YAML & file based configuration


jarohen:
Following the kotlin updates to the node configuration, we want to make numerous changes to the handling of file based config:
- [x] Want to ensure we support YAML file based configuration for the node - we prefer that it deserializes into the new `Config` class, so `kotlinx.serialization` seems a good bet. See be mostly default serdes, because of the way we've written the classes - see [**kaml**](https://github.com/charleskorn/kaml).
- [x] Update the Kotlin API to accept a `YAML` file and start a node. 
- [x] Updates to the Clojure API:
  - (Changes below apply to both `start-node` and `start-submit-client`.
  - Should accept a map, `Config`, `Path` or `File` here:
    - [x] Map should continue to do as it does (ie, use `apply-config` to build an `Config` object.
    - [x] If `Config` object passed in, simply pass through to Kotlin.
    - [x] If `File`/`Path` passed in, called `util/->path` to ensure we have a correctly formatted `Path`, then pass through to Kotlin.
    - [x] Update the docstring to reflect what we can pass in here, ensuring we mention that for `File`/`Path` we will only accept/use `YAML` files (as per the kotlin API) 
- [x] Updates to the CLI: 
  - [x] From the CLI: can probably remove the `--edn` option, expect stuff to be passed in as a file.
  - [x] At this point, can probably remove all JSON based node config.
  - [x] May want to decide on some changes in terms of how people use this.
  - [x] Make it use the updated kotlin/clojure API

PR #3089, opened 2024-01-11
Updates to CLI config/file based config.

danmason:
See #3083 & #2890

Makes a number of updates to the way that the CLI handles config/how file based config is handled:
- Moves the file based config out to a separate namespace (to be shared/reused later).
- Cleans up the CLI namespace of a number of things it doesn't particularly need (Ie, URL handling, never seemed to be possible to actually pass in a URL)
- Removes merging of config passed into CLI - with a preference for files passed in first.
  - Previously, if the user passed in both a file (using `-f`, or by having something like `xtdb.edn` present) and then passed in config string via `--edn` or `--json`, it would merge all of these configs. 
  - This behaviour struck me as odd, so I removed it.
  - The preference order is somewhat random - can change this if we think it's appropriate.
- Updates JSON config to support `ig/ref` using an object with `@ref`.
- Adds YAML based file config handling to the CLI: 
  - Currently only through passing in a `.yaml` file (with the `-f` option) or having `xtdb.yaml` present.
  - Could also check for `.yml` but decided against it - as `.yaml` seems to be the suggested standard file extension.
  - Similar to JSON, ALSO handles env vars / integrant refs (using a `!Env` and `!Ref` tag respectively) 
  - Doesn't handle `--yaml` (as we do for `--edn` and `--json` - seemed less sensible for YAML to be passed in as a string?)
- Add/update numerous tests for all of the above.

### DIFF
--- a/core/src/main/clojure/xtdb/cli.clj
+++ b/core/src/main/clojure/xtdb/cli.clj
@@ -1,82 +1,31 @@
 (ns ^:no-doc xtdb.cli
-  (:require [clojure.data.json :as json]
-            [clojure.edn :as edn]
-            [clojure.java.io :as io]
-            [clojure.string :as str]
+  (:require [clojure.java.io :as io]
             [clojure.tools.cli :as cli]
-            [clojure.tools.logging :as log]
-            [clojure.walk :as walk]
-            [xtdb.error :as err]
+            [clojure.tools.logging :as log] 
+            [xtdb.config :as config] 
             [xtdb.node :as xtn]
-            [xtdb.util :as util]
-            [juxt.clojars-mirrors.integrant.core :as ig])
-  (:import java.io.File
-           java.net.URL
-           java.util.Map))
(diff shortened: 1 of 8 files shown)


core/src/main/clojure/xtdb/cli.clj
```diff
@@ -1,83 +1,32 @@
...
-  (walk/postwalk
-   (fn [item]
-     (let [env-key (keyword "@env")]
-       (cond
-         (env-key item) (read-env-var (env-key item))
-         :else item)))
-   (json/read-str json-string :key-fn keyword)))
-
 (def cli-options
   [["-f" "--file CONFIG_FILE" "Config file to load XTDB options from - EDN, JSON"
```
refset: `(conj "YAML")`

danmason: Ah yes, thank you! Updated the validation but not the doc string.

PR #3101, opened 2024-01-17
Start node configuration via Kotlin API #3096

### LINKED ISSUE #3096: JVM users configuring an XTDB node
* Logs:
  * [x] Kafka log - a large list of parameters we'll need here but should be otherwise straightforward.
  * [x] Azure event hub log - we did implement this and we can add it as part of our work here - similarly, lots of parameters but otherwise straightforward (should be fairly straightforward)
* Buffer pool/Storage:
  * [x] In memory buffer pool - requires allocator as dependency, that seems to be all configuration wise.
  * [x] Local buffer pool - numerous parameters: data-dir, max-cache-bytes, max-cache-entries
  * [x] Remote buffer pool - depends on object stores.
    * object-store, disk-store (file path), max-cache-bytes and max-cache-entries
* Object stores:
  * [x] S3: requires params - bucket, sns-topic-arn, s3-configurator (defaults empty reify otherwise). Can provide prefix also.
  * [x] Azure: requires params - storage-account, container, servicebus-namespace and servicebus-topic-name Can provide prefix also
  * [x] Google Cloud: requires params - project-id, bucket, pubsub-topic
Can provide prefix also
  * :x: JDBC
    Quite a bit more complex - need to setup multiple other modules it relies on (ie, connection-pool, dialect)
    Questionable if we should do this? We've considered dropping JDBC for a while.
     * Think we shall leave this for now.
  * [x] Memory object store (in the tests only)
* (potentially third-party) Modules
  * [x] HTTP server
  * [x] Pgwire server
  * [x] FlightSQL server
  *  :x: (a hypothetical future) Metrics plugin
* [x] Log flusher - depends on a few things and has some parameters: dependent on log and indexer
  Configuration parameters - duration and on-heartbeat (this is a function it should run - might be awkward?)
* Numerous other parts we'll need to set up, but may be handled a bit differently given that they're not really "configurable":
  * Allocator - will be a shared part between numerous parts of the system, but doesnt look like we have any particular config for it so probably doesn't have a particular builder?
  * Log Watcher - again - nothing that seems to be configurable? Has dependencies on numerous other parts but that's all.
  * Indexer  - more system components, no user configurable parameters
  * Compactor - depends on allocator and buffer-pool but otherwise unconfigurable.
  * Metadata manager - dependency on buffer pool, but no configurable parameters.
  * Scan emitter - dependency on bufferpool and metadata manager, but no configurable parameters.
  * Ra-Query-Source - more of the same, depends on allocator, scan emitter, metadata manager, but not configurable
* [x] default-tz - global config param?
* [x] submit-client

danmason:
Resolves #3096

Covers most of the configurable XTDB modules, offering a number of Kotlin implemented factories, as well as converting to them from clojure. A lot of stuff in here, so a few things to mention:
- This doesn't cover any documentation updates - of which, there are **many**.
- `edn` configuration should work as expected, though we'll need to update to allow for YAML.
- As mentioned on the ticket, we haven't handled the `jdbc` module here, given that it is fairly outdated, complicated, and not much used in the codebase. Future of this TBD.
- While I have implemented the configuration of the **Azure Event Hub Log** - it seems to be currently broken (not discovered until now as we do not often run tests on it, considering the pricing of eventhubs we do not run it on the nightly). We can attempt to fix this - but not for this PR. 
- We do not particularly test a lot of this _from_ kotlin itself. I suspect we may want a bit of a run of that, at some point.
- There's probably a fair few lingering references to integrant from various build files - might want to clear these up. Not sure if we can yet clear up the `extra-config` stuff?


pgwire-server/src/main/kotlin/xtdb/api/PgWireServer.kt
```diff
@@ -0,0 +1,30 @@
+@file:JvmName("PgWireServer")
+
+package xtdb.api
+
+import xtdb.api.Xtdb.Module
+import xtdb.util.requiringResolve
+
+private val OPEN_SERVER = requiringResolve("xtdb.pgwire", "open-server")
+
+class PgwireServerModule(
```
danmason, line 10: Question: Naming? Do we have a preference?

src/test/clojure/xtdb/pgwire_test.clj
```diff
@@ -865,11 +865,11 @@
       (is (= [{:a {"b" 42}}] rs)))))
 
 (deftest start-stop-as-module-test
```
danmason, line 867: Only place we actually test pgwire inside the node - lots of other stuff in this module, not entirely related to configuration, but worth a think about.

PR #3109, opened 2024-01-22
Handling YAML file config from the Kotlin API

danmason:
See #2890

## Done in this PR

- Handling YAML decoding to Xdtb.Config
   - Uses `kotlinx.serialization` and `kaml` to decode YAML based config, along with all necessary changes to the various module configs to allow for this.
   - Adds tests for the YAMLDecoder to ensure can use all of our module config from it.
   - Modules specified with YAML tags, ie, `!S3` or `!Remote`.
   - Things that cannot be serialized/dont make sense from YAML marked as `Transient`
   - Using ServiceLoader to load polymorphic class deserializers from plugin modules. 
- Makes DefaultS3Configurator a data object (ie, saves us making a new instance each time, allows proper comparison)
- Update `InMemoryStorageFactory` to be a class such that we can specify/serialize it.
- Adds a Path/File arity to openNode()
  - Checks if `yaml` file extension - if not, throws. Only handles YAML. 
  - Checks if file exists, if not, throws. 
  - If yaml file and exists - passes to our YAML Decoding function then calls open() on the produced config.
  - Adds tests for all of the above behaviour.
- Adds very similar, openSubmitClient() with File/Path.

## TODO

Thinking about how we want to handle things Clojure & CLI side, ie how should the clojure API should work/interract with things. Currently:
- On node/start-node , we expect a config map, and convert it to `Xtdb$Config`
- Our CLI, when passed a file, grabs the contents, turns them to EDN config (does this for YAML, also), then passes the EDN config to node/start-node

I'm wondering about how/if we should handle file config on the clojure API - I'm thinking we have a few different options:
- We could make start-node allow `XTDB$Config` to be passed in directly, in which case it just passes through to Kotlin.
  - This might be useful - could make CLI fairly straight forward (it handles & convert file contents in clojure, then just passes the config it gets back out)
- We could extend `start-node` / add a new function specifically for handling a config file  (`start-node-from-config-file`, for example).
  - Certainly _can_ do this - though I'm not too sure if in-process clojure users will want to be using yaml/config files vs a config map that lives in their code.  
  - Could make it just a pass through to the Kotlin API - with all the behaviour that has (ie, checking file exists, only deals with YAML, accepts a File or Path) 
- Could also just leave Clojure `xtdb.node` as it is, and within the CLI we just build up `Xtdb$Config` objects based on what we receive, then call `.open` on the built Config Objects. The CLI doesn't necessary need to call to functions in `xtdb.node`, it's just been historically convenient for it to do so.

### DIFF
--- a/core/src/main/kotlin/xtdb/api/Xtdb.kt
+++ b/core/src/main/kotlin/xtdb/api/Xtdb.kt
@@ -10,9 +10,13 @@ import xtdb.ZoneIdSerde
 import xtdb.api.log.LogFactory
 import xtdb.api.storage.StorageFactory
 import xtdb.util.requiringResolve
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.Path
 import java.time.Duration
 import java.time.ZoneId
 import java.time.ZoneOffset
+import kotlin.io.path.extension
 
 object Xtdb {
     private val OPEN_NODE: IFn = requiringResolve("xtdb.node.impl", "open-node")
@@ -67,6 +71,21 @@ object Xtdb {
     @JvmOverloads
     fun openNode(config: Config = Config()) = config.open()
 
+    @JvmStatic
+    fun openNode(path: Path): IXtdb {
+        if (path.extension != "yaml") {
+            throw IllegalArgumentException("Invalid config file type - must be '.yaml'")
+        } else if (!path.toFile().exists()) {
+            throw IllegalArgumentException("Provided config file does not exist")
+        }
+
+        val yamlString = Files.readString(path)
+        val config = nodeConfig(yamlString)
+
+        return config.open()
+    }
+
     @JvmSynthetic
     fun openNode(build: Config.() -> Unit) = openNode(Config().also(build))
+
 }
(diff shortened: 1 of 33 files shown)


core/src/main/kotlin/xtdb/api/YamlConfigDecoder.kt
```diff
@@ -0,0 +1,77 @@
+@file:JvmName("YamlConfigDecoder")
```
jarohen: `YamlSerde` (and `YamlSerde.kt`) for consistency with `JsonSerde`?

core/src/main/kotlin/xtdb/api/Xtdb.kt
```diff
@@ -59,6 +71,26 @@ object Xtdb {
...
+            throw IllegalArgumentException("Provided config file does not exist")
+        }
+
+        val yamlString = Files.readString(path)
+        val config = nodeConfig(yamlString)
+
+        return config.open()
+    }
+
+    @JvmStatic
```
jarohen: I'd drop this, it's easy/well-known enough for Java/Kotlin folks to call `toPath()`

jarohen (approved):

PR #3114, opened 2024-01-23
Add handling of `!Env` tags in YAML Decoder, test.

### LINKED ISSUE #3112: Ensure YAML handles !Env tags
Following #2890 - want to ensure that the YAML has also got the ability to handle passed in environment variables, as we do in clojure.

Will want to handle !Env tags inside of our YAML decoding code.

danmason:
Resolves #3112 

Done outside of the usual kotlinx.serialization code since it's more of a pre-processing step - we'd otherwise have to ensure all decoders for all values handled `!Env` somehow. Looked into kaml's `YamlNode` but seemed like a bit of a dead end in terms of handling this specifically. As such, seemed simplest to do a bit of regex handling for it (similar, in a way, to how we use an #env reader for EDN). 

Tested using mockk to fake out setting the environment variable, seems to load correctly, and a null for a required parameter responds correctly..

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -47,6 +47,7 @@ dependencies {
     api(kotlin("stdlib-jdk8"))
     api("com.charleskorn.kaml","kaml","0.56.0")
 
+    testImplementation("io.mockk","mockk", "1.13.9")
     testImplementation(project(":http-server"))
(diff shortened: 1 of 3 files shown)


danmason:
Gonna go ahead and merge - if we want to revisit/make any changes we can, but at least this does what we want it to do.

PR #3115, opened 2024-01-24
Use custom string/path serde for handling !Env, remove string pre-processing

danmason:
Reworks the handling of `!Env` to make it part of the `kotlinx.serialization` handlign, using custom deserializers for particular classes to ensure we're only decoding Env for 'locations' of things, usually represented in remote classes as Strings/Paths. 

I think `handleEnvTag` is probably over complicated (I cannot imagine that `decoder.decodeString` has to do this much work to get at the scalar at the bottom), but that's mainly as a result to what we received in there (a YAMLInput with the whole Map surrounding the value we care to deserialize alongside a 'currentLocation'). Spent a fair while on this already, so thought I should make a PR.

### DIFF
--- a/core/src/main/kotlin/xtdb/api/YamlSerde.kt
+++ b/core/src/main/kotlin/xtdb/api/YamlSerde.kt
@@ -2,10 +2,16 @@
 
 package xtdb.api
 
-import com.charleskorn.kaml.Yaml
+import com.charleskorn.kaml.*
 import kotlinx.serialization.InternalSerializationApi
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.decodeFromString
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.encoding.decodeStructure
 import kotlinx.serialization.modules.SerializersModule
 import kotlinx.serialization.modules.polymorphic
 import kotlinx.serialization.modules.subclass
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/xtdb/api/YamlSerde.kt
```diff
@@ -14,10 +20,59 @@ import xtdb.api.log.InMemoryLogFactory
...
 import xtdb.api.log.LogFactory
 import xtdb.api.storage.ObjectStoreFactory
-import java.util.ServiceLoader
+import java.nio.file.Path
+import java.nio.file.Paths
+import java.util.*
 import java.util.ServiceLoader.Provider
 import kotlin.reflect.KClass
 
+object EnvironmentVariableProvider {
```
danmason, line 29: (made this into an object because `mockkstatic` was causing failures for SOME reason. Maybe was overriding some of the modules somehow, not sure)

core/src/main/kotlin/xtdb/api/YamlSerde.kt
```diff
@@ -14,10 +20,59 @@ import xtdb.api.log.InMemoryLogFactory
...
+fun envFromTaggedNode(taggedNode: YamlTaggedNode ): String {
+    if (taggedNode.tag == "!Env") {
+        val value = taggedNode.innerNode.yamlScalar.content
+        return EnvironmentVariableProvider.getEnvVariable(value) ?: throw IllegalArgumentException("Environment variable '$value' not found")
+    }
+    return taggedNode.innerNode.yamlScalar.content
+}
+fun handleEnvTag(input: YamlInput): String {
+    val currentLocation = input.getCurrentLocation()
+    val scalar = input.node
```
danmason: Seems to work at least 2 levels deep in the map, so presumably it only gives you the outer object surrounding the scalar. Not entirely sure.

PR #3128, opened 2024-01-30
Updates to `date_trunc` & supporting it from SQL

### LINKED ISSUE #3084: Add support for `date_trunc` function
https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-TRUNC - not in the SQL spec, but this is a de-facto standard.

* [x] `field` here is to be a compile-time constant. Can implement this in a similar manner to `get-field`, which should cut down on the (existing) code-gen for the function
* [x] to be added to the SQL grammar and tested
* [x] confirm the docs are up-to-date

danmason:
See #3084 

`DATE_TRUNC` from here https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-TRUNC
- Added DATE_TRUNC to `sql.ebnf`
- Added DATE_TRUNC to `xtdb.sql.plan`
- Added tests to `sql_test` - for planning and for behaviour.
- Included a direct cast to `long` on the forms inside of `zdt->ts` and `ldt->ts`. Saw some issues arise with `ArithmeticException: Integer Overflow` otherwise, related to how the reflector handles the forms.
- Updating the current implementation of `date_trunc`:
  - Code improved using the `DateTruncator` class and better sharing of field behaviour.
  - Supporting the (optional) third parameter for "time_zone"
    - Handled within the `sql.ebnf`, `plan`, added a new arity of codegen for date-trunc to handle this. 
    - Tests also added around this behaving itself. 
    - **Caveats**:
      - Also - If the user specifies an invalid timezone - we won't find out until the point of `expr/codegen-call`. Maybe we're okay with this? Error is fairly reasonable: `"java.time.zone.ZoneRulesException: Unknown time-zone ID: TZ/InvalidRegion (ZoneRulesProvider.java:882)"`. Added a test around this to sql_test.
      - Might be worth thinking about changes to the plan/analyze errors for certain cases. We should only really call `date_trunc` with three parameters IF the datetime we're passing in has a timezone. We don't allow this with `LocalDate` nor `LocalDateTime` (or interval, for that matter!) - user would get an error about missing codegen calls.   
  - Support for truncating intervals within `date_trunc`:
    - Handled within the `sql.ebnf`, `plan`, added a new arity of codegen for date-trunc to handle this. 
    - Tests also added around this behaving itself with different specified field values.
- Current docs updated - suspect we could do better than values in a table (formatting feels a bit hard to read), but all arities/supported values are at least mentioned in `temporal.adoc` now.

### DIFF
--- a/core/src/main/clojure/xtdb/expression/temporal.clj
+++ b/core/src/main/clojure/xtdb/expression/temporal.clj
@@ -8,6 +8,7 @@
            (java.time.temporal ChronoField ChronoUnit)
            [java.util Map]
            (org.apache.arrow.vector PeriodDuration)
+           [xtdb DateTruncator]
            [xtdb.vector IValueReader ValueBox]))
 
 (set! *unchecked-math* :warn-on-boxed)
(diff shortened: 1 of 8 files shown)


jarohen:
> We currently implement the field type a bit differently:
> We are not including WEEK, QUARTER, DECADE, CENTURY or MILLENNIUM

If these are easy enough (one liners, even?) let's include them

> We use MILLISECOND instead of MILLISECONDS and MICROSECOND instead of MICROSECONDS

Surprisingly, the SQL spec doesn't ever mention these - Postgres also works with `millisecond`, so my slight preference would be to standardise on singular throughout.

> We're not currently supporting this for intervals - only for types of datetime
> We're not supporting the (optional) third parameter for "time_zone"

If these are straightforward enough, particularly while you're in context, I'd say go for it.

core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -930,17 +941,79 @@
...
+                                                  "HOUR" `ChronoUnit/HOURS
+                                                  "MINUTE" `ChronoUnit/MINUTES
+                                                  "SECOND" `ChronoUnit/SECONDS
+                                                  "MILLISECOND" `ChronoUnit/MILLIS
+                                                  "MICROSECOND" `ChronoUnit/MICROS)))
+                             (.withZoneSameInstant ~trunc-timezone)) 
+                        (zdt->ts ts-unit)))}))
 
 (defmethod expr/codegen-call [:date_trunc :utf8 :date] [{[{field :literal} _] :args, [_ [_date _date-unit :as date-type]] :arg-types}]
   ;; FIXME this assumes epoch-day
```
refset, line 959: @danmason do you know if this FIXME still valid? and if so, whether it should/shouldn't be fixed as part of the PR?

danmason: Given that I haven't particularly changed the contents/moved it away from epoch-day, I'd assume the comment is still correct, though I can't say I'm sure exactly what it would necessitate changing. @jarohen  - perhaps? (git blame says this is from 18 months ago)

jarohen: yes, it's still valid - when the Arrow date type was first brought into the EE, it didn't properly support the two sub-types (day and milli). Everything within the EE therefore currently assumes that the intermediate representation used for dates is the date-day variant, as opposed to (e.g.) the support for multiple granularities in the timestamp-tz case.

so, tl;dr - it's still valid, and not one I'd expect fixed as part of this PR.

core/src/main/java/xtdb/util/MilleniumAdjuster.java
```diff
@@ -0,0 +1,14 @@
+package xtdb.util;
+
+import java.time.temporal.ChronoField;
+import java.time.temporal.Temporal;
+import java.time.temporal.TemporalAdjuster;
+
+public class MilleniumAdjuster implements TemporalAdjuster {
```
jarohen: thai-po

core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -930,17 +941,79 @@
...
+                                "MONTH" `(-> (.truncatedTo ChronoUnit/DAYS) (.withDayOfMonth 1))
+                                "WEEK" `(-> (.truncatedTo ChronoUnit/DAYS) (.with (WeekAdjuster.)))
+                                `(.truncatedTo ~(case field
+                                                  "DAY" `ChronoUnit/DAYS
+                                                  "HOUR" `ChronoUnit/HOURS
+                                                  "MINUTE" `ChronoUnit/MINUTES
+                                                  "SECOND" `ChronoUnit/SECONDS
+                                                  "MILLISECOND" `ChronoUnit/MILLIS
+                                                  "MICROSECOND" `ChronoUnit/MICROS)))
+                             (.withZoneSameInstant ~trunc-timezone))
```
jarohen: I suspect you'll want to do this TZ change before the truncation, so that it's truncated to (e.g.) the start of the year in the requested TZ?

core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -902,8 +903,13 @@
    :->call-code (fn [[_ x]]
                   (-> `(-> ~(ts->ldt x ts-unit)
                            ~(case field
+                              "MILLENNIUM" `(-> (.truncatedTo ChronoUnit/DAYS) (.with (MilleniumAdjuster.)) (.withDayOfYear 1))
```
jarohen: just a gut feel, but these seem relatively imperative, especially once there's no clean `truncatedTo` call. maybe more declarative to re-create the LDT? e.g. `(LocalDateTime/of (LocalDate/of (truncate-year year 1000) 1 1) LocalTime/MIDNIGHT)`?

perhaps let's pair, I might be talking BS

jarohen: separately, we can likely avoid the adjuster creation for every row by having those as static final fields instead - this could well be hot-loop code here

jarohen: or, if we did want to do it through the adjusters, we could do the whole adjustment there instead - which would cut down on the amount of generated code

jarohen (changes requested):
Hey @danmason, thanks!

SQL side all looks good, let's pair up on polishing the generated code

Separately, happy to run through what I meant by the first bullet point on #3084, good opportunity to fix that here.

core/src/main/clojure/xtdb/expression/temporal.clj
danmason: I would have thought the same, though looking at the behaviour from postgres their example seems to suggest the opposite:

```
SELECT date_trunc('day', TIMESTAMP WITH TIME ZONE '2001-02-16 20:38:40+00', 'Australia/Sydney');
Result: 2001-02-16 08:00:00-05
```

(Quite honestly I do question the usecase of this, with the optional timezone, in general. Just seems a bit... weird)

danmason: Of course, the example could also be wrong too, it's somewhat hard to really tell!

core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -103,15 +104,15 @@
   (if (= ts-unit :second)
     `(.toEpochSecond ~form)
     `(let [form# ~form]
-       (Math/addExact (Math/multiplyExact (.toEpochSecond form#) ~(types/ts-units-per-second ts-unit))
-                      (quot (.getNano form#) ~(quot (types/ts-units-per-second :nano) (types/ts-units-per-second ts-unit)))))))
+       (Math/addExact (Math/multiplyExact (long (.toEpochSecond form#)) ~(types/ts-units-per-second ts-unit))
```
mbutlerw, line 107: Do test it, but as per the reflection thread, I think you can get away with `~(with-tag expr 'long)` for these (adds a type hint), rather than introducing an explicit cast. 

Arguably a nitpick, but just FYI.

core/src/main/clojure/xtdb/expression/temporal.clj
danmason: Actually, scratch the above, think you're right

src/test/clojure/xtdb/expression_test.clj
```diff
@@ -137,15 +145,67 @@
...
-        (t/is (= (LocalDate/of 2022 3 29) (trunc "DAY")))
-        (t/is (= (LocalDate/of 2022 3 1) (trunc "MONTH")))
-        (t/is (= (LocalDate/of 2022 1 1) (trunc "YEAR")))
-        (t/is (= (LocalDate/of 2022 1 1) (project1 '(date-trunc "YEAR" (date-trunc "MONTH" date)) {:date ld})))))))
+    (t/is (= #time/zoned-date-time "2001-02-16T08:00-05:00[America/New_York]"
+             (project1 (list 'date-trunc "DAY" 'date "Australia/Sydney") test-doc)))))
+
+(t/deftest test-date-trunc-interval
+  (let [test-doc {:xt$id :foo,
+                  :interval (PeriodDuration. (Period/of 0 4 8) (Duration/parse "PT1H1M1.111111S"))}]
```
mbutlerw, line 198: Missing tests for anything larger than a quarter for intervals.

mbutlerw (approved):

PR #3140, opened 2024-02-02
Move query-ra to extra non test namespace

FiV0:
This move the `query-ra` fn to an extra `xtdb.query-ra` namespace.


core/src/main/clojure/xtdb/query_ra.clj
```diff
@@ -0,0 +1,67 @@
+(ns xtdb.query-ra
```
FiV0, line 1: @jarohen Not particular tied to `xtdb.query-ra`. Could also move it to `xtdb.operator` which sits currently empty.

PR #3160, opened 2024-02-08
Updates to `extract` function #3126

### LINKED ISSUE #3126: `extract` function wired up
extract from date-time already implemented in the EE.

* [x] extract from date, time, interval
* [x] temporal field to be compile-time literal - i.e. remove support for field to be dynamic
* [x] wire up in the SQL parser/analyser

danmason:
Resolves #3126

- Following the SQL spec here, mainly - read through and highlighted some relevant parts.
- Seems like the EBNF was pretty much done already so haven't changed it.
- Have made sure we support all potential values of the `extract_field` in the planner, added some tests for that.
- Updated the implementation:
  - Updates the structure of the codegen-call somewhat.
  - Ensures we support `SECOND` for all types that support second.
  - Splits the `timestamp-tz` implementation into `timestamp-tz` and `timestamp-local` - ensure we actually construct a ZonedDateTime with the specified tz from the args (so that we can use it to extract TIMEZONE offset related stuff).
  - Handling `TIMEZONE_HOUR` and `TIMEZONE_MINUTES` for all current implementations, according to the SQL Spec:
    - For a timestamp with a timezone (ie, timestamp-tz), we return values based on the offset.
    - Otherwise (for a local timestamp, date, or interval), we return an appropriate exception at compile time.
      - Did look into handling this within the planner, but it seems fairly hard to distinguish if a timestamp has a timezone, especially given that one can pass around any form of timestamp_expression.   
  - Ensure that we similarly throw exceptions when attempting to extract a datetime field that the type does not have (ie, extracting `HOUR` from a `DATE`.
  - Added a codegen calls and tests around extracting fields from both `interval` and `local-time`.
- Updated the table of miscellaneous functions in `temporal.adoc` docs to discuss all valid values when calling `EXTRACT`.

### DIFF
--- a/core/src/main/clojure/xtdb/expression/temporal.clj
+++ b/core/src/main/clojure/xtdb/expression/temporal.clj
@@ -921,6 +921,20 @@
                     "MINUTE" `(int 0)
                     "SECOND" `(int 0)))})
 
+(defmethod expr/codegen-call [:extract :utf8 :interval] [{[{field :literal} _] :args}]
+  {:return-type :i32
+   :->call-code (fn [[_ pd]]
+                  (let [period `(.getPeriod ^PeriodDuration ~pd)
+                        duration `(.getDuration ^PeriodDuration ~pd)]
+                    (case field
+                      "TIMEZONE_HOUR" (throw (UnsupportedOperationException. "Extract \"timezone_hour\" not supported for type interval"))
+                      "TIMEZONE_MINUTE" (throw (UnsupportedOperationException. "Extract \"timezone_minute\" not supported for type interval"))
+                      "YEAR" `(-> (.toTotalMonths ~period) (/ 12) (int))
+                      "MONTH" `(-> (.toTotalMonths ~period) (rem 12) (int))
+                      "DAY" `(.getDays ~period)
+                      "HOUR" `(-> (.toHours ~duration) (int))
+                      "MINUTE" `(-> (.toMinutes ~duration) (rem 60) (int))
+                      "SECOND" `(-> (.toSeconds ~duration) (rem 60) (int)))))})
 
 (defn field->truncate-fn
   [field]
--- a/core/src/main/clojure/xtdb/expression/temporal.clj
+++ b/core/src/main/clojure/xtdb/expression/temporal.clj
@@ -875,16 +875,36 @@
      :->call-code (fn [[s & _]]
                     `(parse-multi-field-interval (expr/resolve-string ~s) ~unit1 ~unit2))}))
 
-(defmethod expr/codegen-call [:extract :utf8 :timestamp-tz] [{[{field :literal} _] :args}]
-  {:return-type :i32
-   :->call-code (fn [[_ ts-code]]
-                  `(.get (.atOffset ^Instant (time/micros->instant ~ts-code) ZoneOffset/UTC)
+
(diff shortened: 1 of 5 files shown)


core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -875,28 +875,71 @@
...
-                    ;; we could inline the math here, but looking at sources, there is some nuance.
                     "YEAR" `(.getYear (LocalDate/ofEpochDay ~epoch-day-code))
                     "MONTH" `(.getMonthValue (LocalDate/ofEpochDay ~epoch-day-code))
                     "DAY" `(.getDayOfMonth (LocalDate/ofEpochDay ~epoch-day-code))
-                    "HOUR" `(int 0)
-                    "MINUTE" `(int 0)))})
+                    (throw (UnsupportedOperationException. (format "Extract \"%s\" not supported for type date" field)))))})
+
+(defmethod expr/codegen-call [:extract :utf8 :interval] [{[{field :literal} _] :args}]
+  ;; FIXME this assumes date-unit :day
```
mbutlerw: Had a quick look for an existing card, but couldn't find one.

No issue with fixing it a separate change but shall we card this up?

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -585,8 +585,15 @@
...
-     [:primary_datetime_field [:non_second_primary_datetime_field extract-field]] "FROM" ^:z es]
+    [:extract_expression "EXTRACT" [:primary_datetime_field [:non_second_primary_datetime_field extract-field]] "FROM" ^:z es]
+    ;;=>
+    (list 'extract extract-field (expr es))
+
+    [:extract_expression "EXTRACT" [:primary_datetime_field "SECOND"] "FROM" ^:z es]
+    ;;=>
+    (list 'extract "SECOND" (expr es))
+
+    [:extract_expression "EXTRACT" [:time_zone_field extract-field] "FROM" ^:z es]
```
mbutlerw, line 596: If you wanted to return an error in planning, you could choose to inspect the returned time object from `expr` and error based on that (Given an unambiguous mapping between SQL/XTQL/JAVA types).

mbutlerw (approved):
Nice test coverage :+1: 

LGTM, nothing that needs changing, just some comments.

core/src/main/clojure/xtdb/sql/plan.clj
danmason: Might be worthwhile doing at _some_ point - did seem like it would be fairly complex, though? Theres quite a lot of different structures of "datetime" we might received, as I understand it (ie, not sure how to really examine say, a structure representing a function call, and figure out the return types of it, for example)

danmason: Not sure if we have any existing examples of that in the analyzer?

core/src/main/clojure/xtdb/expression/temporal.clj
danmason: Sounds reasonable - it's also worth mentioning that I dont think this is relevant with `intervals`, may have copied over that FIXME accidentally, think it's specific to how we represent dates. Shall clean it up in that commit.

PR #3189, opened 2024-02-19
Adding `DATE_DIFF` - Subtracting datetime values & returning intervals in the EE/SQL.

### LINKED ISSUE #3124: `<datetime> - <datetime>` to return interval rather than duration.
SQL spec §6.33, general rule 7 (p316)

* returns year-month Arrow interval if years, months or year to month specified
* returns month-day-nano otherwise.
* interval qualifier to be a compile-time constant (see implementation note in #3084)

danmason:
See #3124 

## Assumptions

The following is based on my understanding of the issue & the SQL spec, namely:
- From the SQL EBNF, we also expect an interval qualifier to be passed through - ie, subtracting two datetime should look like this: `(DATETIME_1 - DATETIME_2) DAY`.
- We use the interval qualifier to:
  - Decide which interval type we want to return - as mentioned in the description of the issue, either `:year-month` or `month-day-nano`.
  - Decide what to calculate the difference between temporal values with.
  - Decide how to truncate/display the outputted interval.

## Implementation 

**Implemented in this PR:** 
- Codegen with variable number of arguments, function called `date-diff`. 
  - Takes two datetime values, and additionally either:
    - `start-field` - calculate difference in start-field and convert to truncated interval based on start-field  - ie, called for `(DATEDIFF dt1 dt2 "HOUR")`
    -  `start-field` & `precision` - calculate difference in start-field, with an extra precision provided for seconds - `ie, called for (DATEDIFF dt1 dt2 "SECOND (1, 3)")`. Convert to truncated interval based on start-field with appropriate second precision.
    -  `start-field` & `end-field` - calculate difference between the two in end-field and convert to truncated interval with least significant field being end-field and most significant field being start-field. ie, called for `(DATEDIFF dt1 dt2 "DAY TO SECOND")`
    -  `start-field`, `end-field` & `precision` - calculate difference between the two in end-field and convert to truncated interval, with appropriate provided precision for seconds ie, called for `(DATEDIFF dt1 dt2 "DAY TO SECOND (6)")`
  - Return type depends on the 'interval qualifier' - if qualifier starts with YEAR or MONTH, then return a `:year-month` interval, else `:month-day-nano`.
  - We do some additional validation in the codegen call:
    - If `precision` is provided, we check it has a valid value.
    - If using a multi field interval qualifier, we check the units are valud (ie, second field isnt larger than first field, only allowed multi field values used)
    - We keep the check from the previous implementation around infinite timestamps. 
  - Using `between` from `ChronoUnit` to calculate the difference between the two timestamps, and then return the result as a `PeriodDuration`.
      - Which ChronoUnit we use depends on the 'interval qualifier' - for example, if the qualifier is "YEAR", we use `ChronoUnit/YEARS` to calculate the difference in years.
      - In cases with multiple field interval qualifiers, we calculate period and duration accordingly - truncating the value based on the smallest time field value in the interval.
      - When precision is provided with `SECOND`, we also calculate in nanos and truncate to appropriate precision.
      - When only time based fields entered (ie, single field like `HOUR, SECOND` or multi field like `HOUR TO SECOND`), we return _everything_ in the Duration, calculated by our least significant field in the interval
  - We implement for other types `date`, `timestamp-tz` (and all other combinations of said types, ie for two dates, two timestamps with timezones, mixed types including timestamp-local, etc) using a set of codegen casts.
- SQL planning and parsing for `DATE_DIFF`, creating variable argument s-expressions as appropriate based on the provided interval qualifier structure.
- Numerous tests around behaviour:
  - Numerous tests around behaviour, various values, validations et al within `expression-test`. 
  - SQL planning and SQL query tests within `sql-test`.
- Fixing up previous tests that include `date`-`date` (in both XTQL and SQL)  

**Differences between DATE_DIFF and SQL spec**:
Some changes were also made between spec and the final result here, namely:
- Changing the syntax to a function call, `DATE_DIFF`. This was done for a few different reasons:
  - Ambiguity in the syntax - it was parsing as an `<interval primary>` with a `<numerical value function>` rather than the `<interval value expression>` we expected   
  - Better matching the experience of calling the underlying function from XTQL - when implementing the codegen we similarly implemented as a function called `date-diff`.  
- Slight differences in how we output/handle intervals:
  - `YEAR TO MONTH`, or `YEAR`, will be shown/handled in terms of months, due to how we display/pass around `interval/year-month` internally.  
  - As a result of using `Duration` - we will always display time based date differences in terms of `HOURS`, `MINUTES` and `SECONDS`, regardless of specified interval qualifier. E.G, if using `(DATE_DIFF dt1 dt2 "SECOND")`, and the result is MORE than 60 seconds, such as 66 seconds, will be outputted as `1 minute 6 seconds` in the PeriodDuration.
    
## TODO

Still have to write up docs around `DATE_DIFF` - likely it's own section in `temporal.adoc` given the general complexity/size of the options here (rather than part of one of the tables on there). Given the current size of PR and how long it has taken to get into a reviewable state, I'm inclined to open a ticket and do it as a followup push on `2.x`.

### DIFF
--- a/core/src/main/clojure/xtdb/expression/temporal.clj
+++ b/core/src/main/clojure/xtdb/expression/temporal.clj
@@ -374,33 +374,56 @@
   (defmethod expr/codegen-call [:+ t1 t2] [expr]
     (recall-with-flipped-args expr)))
 
-;;; subtract
+;; Subtraction
 
-(defmethod expr/codegen-call [:- :timestamp-tz :timestamp-tz] [{[[_ x-unit _], [_ y-unit _]] :arg-types}]
-  (-> (with-arg-unit-conversion x-unit y-unit
-        #(do [:duration %]) #(do `(Math/subtractExact ~@%)))
-      (update :->call-code wrap-throw-eot2 `Long/MAX_VALUE `Long/MAX_VALUE)))
+;; Calculate difference between two timestamps, returning the result as an Interval. 
+;; When subtracting two datetime values, we also provide an 'interval qualifier' (passed as a string). We use this to:
+;; - Determine the type of the resultant interval (either :year-month or :month-day-nano)
+;; - Decide the precision we truncate the result to (e.g. if the interval qualifier is "YEAR", we truncate the result to the nearest year)
+(defmethod expr/codegen-call [:date_diff :timestamp-local :timestamp-local :utf8] [{[_ _ {interval-qualifier :literal}] :args [[_ x-unit] [_ y-unit] _] :arg-types}]
+  (if (or (str/starts-with? interval-qualifier "YEAR")
+          (str/starts-with? interval-qualifier "MONTH"))
 
-(defmethod expr/codegen-call [:- :timestamp-local :timestamp-local] [{[[_ x-unit], [_ y-unit]] :arg-types}]
-  (-> (with-arg-unit-conversion x-unit y-unit
-        #(do [:duration %]) #(do `(Math/subtractExact ~@%)))
-      (update :->call-code wrap-throw-eot2 `Long/MAX_VALUE `Long/MAX_VALUE)))
+    ;; NOTE: As we are using `between` from ChronoUnit, we swap positions of x and y, as between caluclates (between start end) 
+    {:return-type [:interval :year-month]
+     :->call-code (fn [[x y]]
+                    `(PeriodDuration. ~(case interval-qualifier
+                                         "YEAR" `(Period/ofYears (.between ChronoUnit/YEARS ~(ts->ldt y y-unit) ~(ts->ldt x x-unit)))
+                                         "YEAR TO MONTH" `(Period/ofMonths (.between ChronoUnit/MONTHS ~(ts->ldt y y-unit) ~(ts->ldt x x-unit)))
+                                         "MONTH" `(Period/ofMonths (.between ChronoUnit/MONTHS ~(ts->ldt y y-unit) ~(ts->ldt x x-unit))))
+                                      Duration/ZERO))}
 
-(defmethod expr/codegen-call [:- :timestamp-local :timestamp-tz] [{[x [_ y-unit _]] :arg-types, :as expr}]
-  (-> expr (recall-with-cast x [:timestamp-local y-unit])))
+    {:return-type [:interval :month-day-nano]
+     :->call-code (fn [[x y]]
+                    `(PeriodDuration. (Period/ofDays (.between ChronoUnit/DAYS ~(ts->ldt y y-unit) ~(ts->ldt x x-unit)))
+                                      ~(case interval-qualifier
+                                         "DAY" `Duration/ZERO
+                                         "DAY TO HOUR" `(Duration/ofHours (rem (.between ChronoUnit/HOURS ~(ts->ldt y y-unit) ~(ts->ldt x x-unit)) 24))
+                                         "DAY TO MINUTE" `(Duration/ofMinutes (rem (.between ChronoUnit/MINUTES ~(ts->ldt y y-unit) ~(ts->ldt x x-unit)) 1440))
+                                         "DAY TO SECOND" `(Duration/ofSeconds (rem (.between ChronoUnit/SECONDS ~(ts->ldt y y-unit) ~(ts->ldt x x-unit)) 86400))
+                                         "HOUR" `(Duration/ofHours (rem (.between ChronoUnit/HOURS ~(ts->ldt y y-unit) ~(ts->ldt x x-unit)) 24))
+                                         "HOUR TO MINUTE" `(Duration/ofMinutes (rem (.between ChronoUnit/MINUTES ~(ts->ldt y y-unit) ~(ts->ldt x x-unit)) 1440))
+                                         "HOUR TO SECOND" `(Duration/ofSeconds (rem (.between ChronoUnit/SECONDS ~(ts->ldt y y-unit) ~(ts->ldt x x-unit)) 86400))
+                                         "MINUTE" `(Duration/ofMinutes (rem (.between ChronoUnit/MINUTES ~(ts->ldt y y-unit) ~(ts->ldt x x-unit)) 1440))
+                                         "MINUTE TO SECOND" `(Duration/ofSeconds (rem (.between ChronoUnit/SECONDS ~(ts->ldt y y-unit) ~(ts->ldt x x-unit)) 86400))
+                                         "SECOND" `(Duration/ofSeconds (rem (.between ChronoUnit/SECONDS ~(ts->ldt y y-unit) ~(ts->ldt x x-unit)) 86400)))))}))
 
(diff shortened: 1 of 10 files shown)


danmason:
Probably DOES make sense to treat this as a distinct function rather than `-` - can also aim to make SQL fairly similar to XTQL in how it calls it (ie, using a function like call syntax instead of `(DATETIME - DATETIME) INTERVALQUALIFIER`

Gonna take the following actions, for now:
- Rename `subtract_temporal` to `date_diff`
- Update the EBNF - interval value expression to use `DATEDIFF(DATETIME, DATETIME, INTERVALQUALIFIER)` instead of `(DATETIME - DATETIME) INTERVALQUALIFIER`
- May also switch this to use `primary_datetime_field` instead of `interval_qualifier` - given our lack of distinction when outputting the intervals, makes more sense.
- Implement the SQL planning for DATEDIFF and test.

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -138,6 +138,79 @@
...
 
+(defn ->date-diff-call [t1 t2 qualifier]
+  (r/zmatch qualifier
+            [:interval_qualifier
+             [:single_datetime_field [:non_second_primary_datetime_field start-field]]]
+            ;; =>
+            (list 'date_diff t1 t2 start-field)
+  
+            [:interval_qualifier
+             [:single_datetime_field [:non_second_primary_datetime_field start-field] [:unsigned_integer _leading-precision]]]
```
danmason, line 149: We're discarding "leading precision" here - seems we do not use this within our intervals either (presumably does not make so much sense given how we're storing them for Arrow)?

core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -374,33 +386,121 @@
...
+        nanos (.longValueExact (.setScale (.multiply (.subtract second-nanos seconds) 1e9M) (- (long precision) 9) BigDecimal/ROUND_DOWN))]
+    (Duration/ofSeconds seconds nanos)))
+
+(defn date-diff-start-end-fields [x-unit y-unit start-field end-field precision] 
+  {:return-type (if (= ["YEAR" "MONTH"] [start-field end-field])
+                  [:interval :year-month]
+                  [:interval :month-day-nano])
+   :->call-code (fn [[x y]]
+                  (case [start-field end-field]
+                    ["YEAR" "MONTH"] `(PeriodDuration. (Period/of (.between ChronoUnit/YEARS ~(ts->ldt y y-unit) ~(ts->ldt x x-unit))
```
danmason, line 438: Left out "MONTH" to "X" as it seemed a bit weird to represent (ie, how long is a MONTH in days?) - and we also don't support them in multi field intervals.This does mean that whenever we use the `MONTH-DAY-NANOS` type we're always using DAY in the `Period`.

src/test/clojure/xtdb/expression_test.clj
```diff
@@ -315,6 +315,232 @@
...
+        (t/is (= #xt/interval-ym "P14M" (date-diff "MONTH")))
+        (t/is (= #xt/interval-mdn ["P425D" "PT0S"] (date-diff "DAY")))
+        (t/is (= #xt/interval-mdn ["P0D" "PT10202H"] (date-diff "HOUR")))
+        (t/is (= #xt/interval-mdn ["P0D" "PT10202H4M"] (date-diff "MINUTE")))
+        (t/is (= #xt/interval-mdn ["P0D" "PT10202H4M26S"] (date-diff "SECOND")))))
+
+    (t/testing "with start-field and precision"
+      (letfn [(date-diff [start precision] (project1 (list 'date-diff 'zdt1 'zdt2 start precision) test-doc))]
+        ;; TODO - this fails because we're passing dates around with MICROSECOND precision
+        ;; (t/is (= #xt/interval-mdn ["P0D" "PT10921H14M30.111111111S"] (date-diff "SECOND" 9)))
```
danmason, line 372: Worth pointing out - not sure if theres a way to force timestamps that are passed into codegens/represented into nanos? Seems like typically we pass around all our timestamps in `microseconds`. I've similarly passed casted to timestamp-local with a `:micro` unit - if we'd prefer to change that to `nanos` we can do.

jarohen: this isn't quite true - we convert it into the unit of interest. see `with-arg-unit-conversion` and its usages

core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -374,33 +405,127 @@
...
```
jarohen, line 438: Could you let-bind the result of the `ts->ldt` codegen here for x and y? Not just for DRY - this will also mean it appears multiple times in the codegen (i.e. longer compile times) and will calculated multiple times at runtime (within a hot loop).

e.g. 

```clojure
`(let [~x-ldt-sym ~(ts->ldt x x-unit)]
   ~(case [start-field end-field]
      ["YEAR" "MONTH"] `(... (.between ... ~x-ldt-sym ~y-ldt-sym))))
```

jarohen: related: I might also be tempted to extract the calculation itself into a function that we don't have to compile every time, again in order to cut down the codegen but will likely also help convince the JiT compiler to AoT compile the calculations - it likely won't do this if it's technically a different function every time.

i.e.

```clojure
;; NB no quoting here, just a regular Clojure function 
;; also means it can be tested independently of all the codegen shenanigans if that makes life easier

(defn date-diff-year-month [^LocalDateTime y, ^LocalDateTime x])
  (PeriodDuration. (Period/of ...) ...))

;; call code
(case [start-field end-field]
  ["YEAR" "MONTH"] (date-diff-year-month y x)
  ...)
```

jarohen: ... having re-checked the SQL spec, I see it defines how to do this calculation without needing to translate into an LDT - §6.33, general rule 7 (p316). For us, this would likely mean converting the arguments into two `:timestamp-local`s with the same units (as we do in other temporal operations, IIRC), then converting to an interval using the start/end field. 

Let's discuss?

core/src/main/clojure/xtdb/sql/plan.clj
jarohen: You're probably right, but I'd prefer we spend some time to become more confident about it :slightly_smiling_face:

jarohen (changes requested):
Hey @danmason, cheers :pray: worth us having a chat about the implementation - I suspect we should stick closer to what it recommends in the SQL spec.

Come find me when it's a good time for a catchup, you know where I hang out :slightly_smiling_face:

danmason:
Decided against this as a

danmason:
Decided against this in the end - shall close the PR for now, will leave a comment on and close #3124 :)

PR #3192, opened 2024-02-20
Adding full expression support to inner XTQL args

### LINKED ISSUE #2996: Add support for expressions in all required places in XTQL
- [x] from-binding - expr
- [ ] from-binding - subq
- [x] rel-table - expr
- [ ] rel-table - subq
- [ ] rel-binding - expr
- [ ] rel-binding - subq
- [x] Aggregate - expr
- [x] Aggregate - subq
- [ ] order-by-spec - expr
- [ ] order-by-expr - subq
- [x] return - expr
- [ ] return - subq
- [x] where
- [x] with
- [x] unnest - expr
- [x] unnest - subq
- [x] subquery/join arguments

### LINKED ISSUE #3181: Problems with subquery parameters in XTQL
```clj
;; works
  (xt/q node '(-> (rel [{}] [])
                  (with {:foo $outer}))
        {:args {:outer 1}})
;; works
  (xt/q node '(-> (rel [{}] [])
                  (with {:foo (q (-> (rel [{}] [])
                                     (with {:foo 1})))})))
;; fails with an invalid logical plan
  (xt/q node '(-> (rel [{}] [])
                  (with {:foo (q (-> (rel [{}] [])
                                     (with {:foo $inner}))
                                 {:args [{:inner 1}]})}))) 
  
;; fails with missing column
  (xt/q node '(-> (rel [{}] [])
                  (with {:foo (q (-> (rel [{}] [])
                                     (with {:foo $inner}))
                                 {:args [{:inner $outer}]})}))
        {:args {:outer 1}})
```

FiV0:
This adds full expression in the `:arg` position of subqueries and explicit joins (`join`, `join-left`).
Examples:
```clj
;; expression in join args
(unify (from :a [{:xt/id aid} a b])
       (join (rel [{:b $a :c $c}] [b c])
             {:args [{:a (+ a 1) :c (+ 1 3)}]
              :bind [b c]}))
              
;; expression in subquery args
(-> (rel [{}] [])
    (with {:foo (q (rel [{:inner $inner}] [inner])
                   {:args [{:inner 1}]})}))
```

As the arguments for the subquery are (or can be dependent) on the accumulated plan, it is not possible to create a complete `join`/`subquery` plan that can just be joined with the accumulated plan. Instead the approach attaches three arguments to the planned `join`/`subquery`. 
- `arg-subqueries` - Any argument subqueries with which the accumulated plan needs to be wrapped.
- `tmp-expr-sym->apply-param-sym` - A mapping from temporary expr to an apply param symbol.
- `tmp-expr-sym->expr-vec` - A vector to be passed to be passed to the `:map` operator to create our temporary expressions.

This is then mostly used as follows:
```clj
[:apply :join tmp-expr-sym->apply-param
 [:map tmp-expr-sym->expr-vec
  (wrap-expr-subqueries acc-plan arg-subqueries)]
 actual-join-or-non-argument-subquery]
```
One has to assure that the `tmp-expr-sym`s are also projected away afterwards.

### DIFF
--- a/core/src/main/clojure/xtdb/xtql.clj
+++ b/core/src/main/clojure/xtdb/xtql.clj
@@ -72,13 +72,6 @@
        util/symbol->normal-form-symbol
        (with-meta {:correlated-column? true}))))
 
-(defn- args->apply-params [args]
-  (->> args
-       (mapv (fn [{:keys [l r _required-vars]}]
-               (MapEntry/create (apply-param-sym l) r)))
-       (into {})
-       (set/map-invert)))
-
 (defn- unifying-vars->apply-param-mapping [unifying-vars]
   ;; creates a param for each var that needs to unify, so that we can place the
   ;; equality predicate on the dependant side, within the left join.
@@ -95,9 +88,9 @@
   (col-sym
    (str "xt$_" (*gensym* "sqp"))))
 
-(defn gen-col []
-  (col-sym
-   (str "xt$_" (*gensym* "gc"))))
+(defn gen-col
+  ([] (col-sym (str "xt$_" (*gensym* "gc"))))
+  ([prefix] (col-sym prefix (str "xt$_" (*gensym* "gc")))))
 
 (declare plan-arg-spec)
 
@@ -115,6 +108,13 @@
              (keys)
              (set))
         expr/normalise-fn-name))
+
+(defn- plan-arg-bindings [args]
+  (let [arg-bindings (mapv plan-arg-spec args)
+        temporary-expr-symbols (map (comp #(gen-col %) :l) arg-bindings)]
+    {:arg-subqueries (mapcat (comp :subqueries :r) arg-bindings)
+     :tmp-expr-sym->apply-param-sym (zipmap temporary-expr-symbols (map (comp apply-param-sym :l) arg-bindings))
+     :tmp-expr-sym->expr-vec (mapv hash-map temporary-expr-symbols (map (comp :expr :r) arg-bindings))}))
 
 (extend-protocol ExprPlan
   Expr$Null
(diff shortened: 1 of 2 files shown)


core/src/main/clojure/xtdb/xtql.clj
```diff
@@ -685,47 +710,52 @@
...
+  (if (seq tmp-expr-sym->expr-vec)
     (let [{:keys [rels var->cols]} (with-unique-cols [acc-plan join-plan])
           [{acc-plan-with-unique-cols :ra-plan}
            {join-subquery-plan-with-unique-cols :ra-plan}] rels]
       (wrap-unify
-       {:ra-plan [:apply :cross-join args
-                  acc-plan-with-unique-cols
+       {:ra-plan [:apply :cross-join tmp-expr-sym->apply-param-sym
+                  [:map tmp-expr-sym->expr-vec
+                   (wrap-expr-subqueries acc-plan-with-unique-cols acc-provided-vars arg-subqueries)]
```
FiV0, line 722: I am implicitly making usage of the fact that `with-unique-cols` doesn't change the column names of the first plan given (`acc-plan` in this case). Hence I don't need to cleanup the `tmp-expr-sym`. I don't know if this approach is bad, as there are now implicit assumptions.

FiV0:
@mbutlerw If you think naming is bad, please let me know.

core/src/main/clojure/xtdb/xtql.clj
mbutlerw: IIRC this code was already making use of that assumption, so even if it is a mistake its not one introduced here.

When you say you don't need to clean up, you mean explicitly right? In that it will be projected away implicitly as a result of calling `wrap-unify` with the original col names.

core/src/main/clojure/xtdb/xtql.clj
```diff
@@ -116,6 +109,13 @@
              (set))
         expr/normalise-fn-name))
 
+(defn- plan-arg-bindings [args]
+  (let [arg-bindings (mapv plan-arg-spec args)
+        temporary-expr-symbols (map (comp #(*gensym* %) :l) arg-bindings)]
```
mbutlerw: I'd try to avoid using gensym directly, I think as a result you are missing a call to col-sym (and therefore metadata) here.

I'd use `gen-col` instead, feel free to extend to allow for a custom prefix string if you want the col to include the original col name.

mbutlerw (changes requested):
Only 1 tiny required change, overall looks really good, pretty much how I'd imagined it working.

Great work :+1:

core/src/main/clojure/xtdb/xtql.clj
FiV0: Yes exactly.

Issue #2424, opened 2022-09-12
`:table` round-tripping values via Java objects can cause units to be lost


jarohen:
e.g. `[:table [a] [{a (- #time/date "2021-12-24" #time/date "2021-12-23")}]]`

If the expression engine returns (e.g.) `[:duration :second]` for this, but the `#time/duration "PT24H"` object has a value->col-type of `[:duration :micro]`, then `:table` seems to try to insert a `[:duration :micro]` value into a `DurationVector` with `SECOND` units, yielding `PT24000000H`.

Only seems to affect types for which there is one Java type to many col-types.

jarohen:
got the params sent through as a relation now (a37fa61e61d51cef8931251fe6d55460366f566a) but preserving Arrow format here will involve refactoring the EE to better handle calculating single values. backlogging again for now.

Issue #435, opened 2019-11-25
Users can use all features of the old Java API from the new Java API


danmason:
Within the next vertical slice of the API, I wish to cover the contents of [**ICruxDatasource**](https://github.com/juxt/crux/blob/master/crux-core/src/crux/api/ICruxDatasource.java) and [**ICruxAPI**](https://github.com/juxt/crux/blob/master/crux-core/src/crux/api/ICruxAPI.java) in particular.

jarohen:
killing, we'll be migrating crux.api.alpha into crux.api

PR #3199, opened 2024-02-21
`date` - `date` returning an integer of days elapsed.

### LINKED ISSUE #3197: `date` - `date` to return an integer instead of a duration
Instead of returning a duration here, return an integer representing the difference in days between the two dates.

danmason:
Resolves #3197

### DIFF
--- a/src/test/clojure/xtdb/operator/table_test.clj
+++ b/src/test/clojure/xtdb/operator/table_test.clj
@@ -111,11 +111,6 @@
                            [:table [{x50 false}]]]]
                         {:params {'?x53 "RAIL"}}))))
 
-(t/deftest test-date-subtraction-bug-435
-  (t/is (= [{:a #time/duration "PT24H"}] ; it returned PT24000000H :/
-           (tu/query-ra
-            '[:table [a] [{a (- #time/date "2021-12-24" #time/date "2021-12-23")}]]))))
-
 (t/deftest test-incorrect-relation-params
   (t/is
    (thrown-with-msg? RuntimeException #"Table param must be of type struct list"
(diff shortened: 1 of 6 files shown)


src/test/clojure/xtdb/operator/table_test.clj
```diff
@@ -112,7 +112,7 @@
                         {:params {'?x53 "RAIL"}}))))
 
 (t/deftest test-date-subtraction-bug-435
```
danmason: Not sure on the exact issue/bug this is referencing (issue #435 is fairly old, might be from when this was on a separate board) - possible this test is redundant now given that we're not using durations? @jarohen

src/test/clojure/xtdb/operator/table_test.clj
```diff
@@ -112,7 +112,7 @@
...
-  (t/is (= [{:a #time/duration "PT24H"}] ; it returned PT24000000H :/
```
jarohen, line 115: ~mind duplicating this test rather than updating it? was a very specific bug that I wanted to keep a regression test for~

jarohen (approved):
cheers @danmason - just one request but good to merge without further review :pray:

src/test/clojure/xtdb/operator/table_test.clj
danmason: What do we want to duplicate here, exactly? Would it be using timestamps or something like that?

src/test/clojure/xtdb/operator/table_test.clj
jarohen: xtdb/core2#435, now #2424 - this was pre-migration :slightly_smiling_face:

src/test/clojure/xtdb/operator/table_test.clj
jarohen: hmm, maybe it is date-specific, so the issue goes away if `date - date` becomes integer... :thinking: 

alright, changed my mind, let's kill it - these comments should link from #2424 so should spot it when we come back to it

PR #3202, opened 2024-02-22
Add generic `length` function to EE 

### LINKED ISSUE #3125: `length` EE function
`length` expr for the below types should delegate to the functions below. 

* [x] alias for `character_length` on strings
* [x] alias for `octet_length` on varbinary
* [x] alias for `cardinality` on lists
* [x] collection size for sets 
* [x] non-absent field count for structs

**EDIT (Dan M - 22/02/2024):** 
- We did not previously implement `array_length` on lists - only `cardinality`. Have reflected note around aliasing to alias to that function.
- We do not implement `array_length` _or_ `cardinality` on sets - have updated TODO list to reflect that we need to implement for sets as well/

danmason:
See #3125 

Implements `length` within the expression engine:
- Implemented for a number of different types:
  - `utf8` string - calls through to codegen for `char-length`
  -  `varbinary` - calls through to codegen for `octet-length`
  -  `list` - calls through to codegen for `cardinality`
  - `set` - returns the number of elements in the set, using `count` (as it is passed a **Set**)
  -  `struct` returns the number of *non-absent* fields in the struct.
- Numerous tests in expression-test.
- Added to SQL parser/planning:
  - Handles a couple of literals/expressions in the grammar - particularly, `character_value_expression`, `string_value_expression` & `array_value_expression` 
  - Adds a number of tests to sql-test.
- Adds docs around this:
  - Created a new page, "Other Functions", under stdlib.
  - Goes through usages and what expressions are supported by `length`

### DIFF
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -12,7 +12,7 @@
            (java.nio ByteBuffer)
            (java.nio.charset StandardCharsets)
            (java.time Clock Duration Instant LocalDate LocalDateTime LocalTime OffsetDateTime ZoneOffset ZonedDateTime)
-           (java.util Arrays Date List Map UUID)
+           (java.util Arrays Date List Map Set UUID)
            (java.util.regex Pattern)
            (java.util.stream IntStream)
            (org.apache.arrow.vector PeriodDuration ValueVector)
@@ -1374,6 +1374,30 @@
 (defmethod codegen-call [:cardinality :list] [_]
   {:return-type :i32
    :->call-code #(do `(.size ~@%))})
+
+(defmethod codegen-call [:length :utf8] [expr]
+  (codegen-call (assoc expr :f :character_length)))
+
+(defmethod codegen-call [:length :varbinary] [expr]
+  (codegen-call (assoc expr :f :octet_length)))
+
+(defmethod codegen-call [:length :list] [expr]
+  (codegen-call (assoc expr :f :cardinality)))
+
+(defmethod codegen-call [:length :set] [_] 
+  {:return-type :i32
+   :->call-code #(do `(count ~@%))})
+
+(defn count-non-empty [m]
+  (reduce
+   (fn [^Integer non-empty [_x ^IValueReader y]]
+     (if (.isNull y) non-empty (Math/addExact non-empty 1)))
+   0
+   m))
+
+(defmethod codegen-call [:length :struct] [_] 
+  {:return-type :i32
+   :->call-code #(do `(count-non-empty ~@%))})
 
 (defn trim-array-view ^xtdb.vector.IListValueReader [^long trimmed-value-count ^IListValueReader lst]
   (reify IListValueReader
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -1388,12 +1388,13 @@
   {:return-type :i32
    :->call-code #(do `(count ~@%))})
 
-(defn count-non-empty [m]
-  (reduce
-   (fn [^Integer non-empty [_x ^IValueReader y]]
-     (if (.isNull y) non-empty (Math/addExact non-empty 1)))
-   0
-   m))
+(defn count-non-empty [m] 
+  (loop [n 0, xs (vals m)]
+    (if (seq xs)
+      (if (.isAbsent ^IValueReader (first xs))
+        (recur n (rest xs))
+        (recur (inc n) (rest xs)))
+      n)))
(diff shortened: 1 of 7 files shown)


docs/src/content/docs/reference/main/stdlib/other.adoc
```diff
@@ -0,0 +1,28 @@
...
+[source,sql]
+----
+LENGTH(<expr>)
+----
+
+Where `<expr>` is one of the following:
+
+* A **string** - returns the number of utf8 characters in the string
+** Acts as an alias for the `CHAR_LENGTH` function - see in link:string[String Functions]
+* A **varbinary** - returns the number of bytes in the varbinary
```
danmason, line 24: Usage of `varbinary` - seems to have leaked from arrow datatypes - (ie, https://docs.xtdb.com/reference/main/data-types.html#_scalar_types)

docs/src/content/docs/reference/main/stdlib/other.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+
+[source,sql]
+----
+LENGTH(<expr>)
+----
+
+Where `<expr>` is one of the following:
+
+* A **string** - returns the number of utf8 characters in the string
```
danmason, line 22: As a follow up/follow on work, probably want to line up with however we decide to consistently reference types/type names.

core/src/main/clojure/xtdb/expression.clj
```diff
@@ -1375,6 +1375,33 @@
...
+  (.size ^Set set))
+
+(defmethod codegen-call [:length :set] [_] 
+  {:return-type :i32
+   :->call-code #(do `(size-of-set ~@%))})
+
+(defn count-non-empty [m]
+  (reduce
+   (fn [^Integer non-empty [_x ^IValueReader y]]
+     (if (.isNull y) non-empty (Math/addExact non-empty 1)))
```
danmason: Originally done to cover "non empty field" requirement - nils are a value and as such are important to count - rather, we want to check here for **absent** (ie, there is no field).

docs/src/content/docs/reference/main/stdlib/other.adoc
```diff
@@ -0,0 +1,28 @@
...
+** Acts as an alias for the `CHAR_LENGTH` function - see in link:string[String Functions]
+* A **varbinary** - returns the number of bytes in the varbinary
+** Acts as an alias for `OCTET_LENGTH` function - see in link:string[String Functions]
+* A **list** - returns the number of elements in the list
+* A **set** - returns the number of elements in the set
+* A **struct** - returns the number of *non-empty* fields in the struct
```
danmason: **Non-absent** rather than **non-empty** (after change)

docs/src/content/docs/reference/main/stdlib/other.adoc
```diff
@@ -0,0 +1,28 @@
+---
+title: Other Functions
+---
+
+== Length
+
+We offer a generic function to calculate the length/cardinality of elements.
```
danmason: Remove "we offer a" - ie, avoid chattiness - also probably "length/cardinality"->"length of value"

core/src/main/clojure/xtdb/expression.clj
```diff
@@ -1375,6 +1375,33 @@
...
+(defmethod codegen-call [:length :varbinary] [expr]
+  (codegen-call (assoc expr :f :octet_length)))
+
+(defmethod codegen-call [:length :list] [expr]
+  (codegen-call (assoc expr :f :cardinality)))
+
+(defn size-of-set [^Set set] 
+  (.size ^Set set))
+
+(defmethod codegen-call [:length :set] [_]
```
danmason, line 1387: This does pass in a clojure persistent hashset, rather than a xtdb.vector.IListValueReader - can use `count`, though will probably want to investigate the hows/whys of us returning a Set here, rather than a reader.

(Can see the 'how' on our implementation of `SetVector` - we are actively creating a `PersistentHashSet`). Changing/updating this, if we need to, out of scope of here.

core/src/main/clojure/xtdb/expression.clj
danmason: Will need to implement `isAbsent` on IValueReader and it's implementers.

core/src/main/clojure/xtdb/expression.clj
danmason: Have moved this over to `count` - leaving comment open RE: types and the like, in case we wish to do more in the future here.

mbutlerw (approved):
LGTM :shipit:

PR #3205, opened 2024-02-26
Add more caching to the MetadataManager

### LINKED ISSUE #3210: Improving performance of loading metadata files into VSRs
On OLTP queries, loading metadata files into VSRs is taking up a significant proportion of the query - this is amortised on longer running OLAP queries but on otherwise low-latency queries this is quite an overhead.

In a quick and dirty test we were caching the result of `open-meta-file` and this was already bringing down the biggest chunk of metadata related calculation from ~20% to almost 0. How I propose to incorporate this fix into the metadata manager. Bring a configurable cache (Caffeine) into the MetadataManager. This cache will keep a set of open items for a given meta data file around. These include:
- an open VSR pointing to the metadata relation
- an open buffer (underlying the above VSR)
- pre-calculated page-idxs (already done now and kept forever)

I propose to change 
```clj
(tableMetadata [metadata-mgr meta-rel-rdr file-path])
=>
(tableMetadata [metadata-mgr file-path])
```
and reopening the meta-relation only when the "MetaFile" has been pushed out of the cache.

The reified `ITableMetadata` will then become a `deftype TableMetadata` which additionally implements `AutoCloseable` .

On a second pass, we could also cache page-idx-preds for a particular column set. For this (I think) `*column->pushdown-bloom*` has to go though.

FiV0:
`open-meta-files` was loading buffers and VSR's over and over again for the metadata. This change adds a cache to the MetadataManager that keeps some of these longer around. The cache is currently defaulting to 128 in size, which (assuming Metadata files are roughly the same size) accounts for roughly ~1G of data. 

We add a `TableMetadata` defrecord which implements `ITableMetadata` and also contains the trie, arrow buffer and metadata relation reader on top of previously calculated metadata. This gets cached. The last one, either MetadataManager evicting the particular `TableMetadata` or a query using the `TableMetadata`, closes the referenced objects. One thing to note is that previously we were caching `table-metadata-idxs` indefinitely and without bound, they now also get removed one things get evicted from the `MetadataManager`.

### DIFF
--- a/core/src/main/clojure/xtdb/compactor.clj
+++ b/core/src/main/clojure/xtdb/compactor.clj
@@ -18,7 +18,8 @@
            xtdb.vector.IRelationWriter
            xtdb.vector.IRowCopier
            xtdb.vector.IVectorWriter
-           xtdb.vector.RelationReader))
+           xtdb.vector.RelationReader
+           (xtdb.metadata IMetadataManager)))
 
 #_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
 (definterface ICompactor
@@ -102,19 +103,19 @@
   (vw/->vec-writer allocator "xt$recency"
                    (FieldType/notNullable #xt.arrow/type [:timestamp-tz :micro "UTC"])))
 
-(defn exec-compaction-job! [^BufferAllocator allocator, ^IBufferPool buffer-pool, {:keys [page-size]}
-                            {:keys [^Path table-path path trie-keys out-trie-key]}]
+(defn exec-compaction-job! [^BufferAllocator allocator, ^IBufferPool buffer-pool, ^IMetadataManager metadata-mgr
+                            {:keys [page-size]} {:keys [^Path table-path path trie-keys out-trie-key]}]
   (try
     (log/infof "compacting '%s' '%s' -> '%s'..." table-path trie-keys out-trie-key)
 
-    (util/with-open [meta-files (LinkedList.)
+    (util/with-open [table-metadatas (LinkedList.)
                      data-rels (trie/open-data-rels buffer-pool table-path trie-keys nil)]
       (doseq [trie-key trie-keys]
-        (.add meta-files (trie/open-meta-file buffer-pool (trie/->table-meta-file-path table-path trie-key))))
+        (.add table-metadatas (.openTableMetadata metadata-mgr (trie/->table-meta-file-path table-path trie-key))))
 
-      (let [segments (mapv (fn [{:keys [trie] :as meta-file} data-rel]
-                             {:trie trie, :meta-file meta-file, :data-rel data-rel})
-                           meta-files
+      (let [segments (mapv (fn [{:keys [trie] :as _table-metadata} data-rel]
+                             {:trie trie, :data-rel data-rel})
+                           table-metadatas
                            data-rels)
             schema (->log-data-rel-schema (map :data-rel segments))]
 
@@ -219,13 +220,14 @@
 
 (defmethod ig/prep-key :xtdb/compactor [_ opts]
   (into {:allocator (ig/ref :xtdb/allocator)
-         :buffer-pool (ig/ref :xtdb/buffer-pool)}
+         :buffer-pool (ig/ref :xtdb/buffer-pool)
+         :metadata-mgr (ig/ref :xtdb.metadata/metadata-manager)}
         opts))
 
 (def ^:dynamic *page-size* 1024)
 (def ^:dynamic *l1-file-size-rows* (bit-shift-left 1 18))
 
-(defmethod ig/init-key :xtdb/compactor [_ {:keys [allocator ^IBufferPool buffer-pool]}]
+(defmethod ig/init-key :xtdb/compactor [_ {:keys [allocator ^IBufferPool buffer-pool metadata-mgr]}]
   (let [page-size *page-size*
         l1-file-size-rows *l1-file-size-rows*]
     (util/with-close-on-catch [allocator (util/->child-allocator allocator "compactor")]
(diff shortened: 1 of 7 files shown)


core/src/main/clojure/xtdb/metadata.clj
```diff
@@ -311,23 +315,47 @@
...
+  (columnNames [_] col-names)
+  (rowIndex [_ col-name page-idx] (.get page-idx-cache [col-name page-idx]))
+  (iidBloomBitmap [_ page-idx]
+    (let [bloom-rdr (-> (.structKeyReader metadata-rdr "columns")
+                        (.listElementReader)
+                        (.structKeyReader "bloom"))]
+
+      (when-let [bloom-vec-idx (.get page-idx-cache ["xt$iid" page-idx])]
+        (when (.getObject bloom-rdr bloom-vec-idx)
+          (bloom/bloom->bitmap bloom-rdr bloom-vec-idx)))))
```
FiV0, line 336: @jarohen I kept the `ITableMetadata` around but things can also be grabbed directly from the record (see trie). I could either slim down the interface (just `rowIndex` and `iidBloomBitmap`) or move everything to the protocol and make the `defrecord` a `deftype`.

core/src/main/clojure/xtdb/compactor.clj
```diff
@@ -102,19 +103,19 @@
...
 
-    (util/with-open [meta-files (LinkedList.)
+    (util/with-open [table-metadatas (LinkedList.)
                      data-rels (trie/open-data-rels buffer-pool table-path trie-keys nil)]
       (doseq [trie-key trie-keys]
-        (.add meta-files (trie/open-meta-file buffer-pool (trie/->table-meta-file-path table-path trie-key))))
+        (.add table-metadatas (.tableMetadata metadata-mgr (trie/->table-meta-file-path table-path trie-key))))
 
-      (let [segments (mapv (fn [{:keys [trie] :as meta-file} data-rel]
-                             {:trie trie, :meta-file meta-file, :data-rel data-rel})
```
FiV0, line 116: In scan as well as here we were attaching this `meta-file`, I don't know if there was reason to that or if it is some leftover from earlier implementations.

core/src/main/clojure/xtdb/metadata.clj
```diff
@@ -59,7 +63,7 @@
 (definterface IMetadataManager
   (^void finishChunk [^long chunkIdx, newChunkMetadata])
   (^java.util.NavigableMap chunksMetadata [])
-  (^xtdb.metadata.ITableMetadata tableMetadata [^xtdb.vector.RelationReader metaRelReader, ^java.nio.file.Path metaFilePath])
+  (^xtdb.metadata.ITableMetadata tableMetadata [^java.nio.file.Path metaFilePath])
```
jarohen: if the caller needs to close this, can we rename it to `openTableMetadata`?

core/src/main/clojure/xtdb/metadata.clj
```diff
@@ -311,23 +315,47 @@
...
-  (reify ITableMetadata
-    (metadataReader [_] metadata-reader)
-    (columnNames [_] col-names)
-    (rowIndex [_ col-name page-idx] (.get page-idx-cache [col-name page-idx]))
-    (iidBloomBitmap [_ page-idx]
-      (let [bloom-rdr (-> (.structKeyReader metadata-reader "columns")
-                          (.listElementReader)
-                          (.structKeyReader "bloom"))]
+(defrecord TableMetadata [^HashTrie trie
+                          ^RelationReader meta-rel-reader
```
jarohen, line 319: possibly a bit out of context, but it's a little confusing that we have `meta-rel-reader` and `metadata-rdr` - maybe one/both of these can have a more specific name now that they're closer together?

core/src/main/clojure/xtdb/metadata.clj
jarohen: aware it's not massively Clojure-y, but I don't mind this - the strongly-typed interface means the caller doesn't need to worry as much about type hinting.

that said, column-names is just a set, so maybe not as required there. :shrug:

core/src/main/clojure/xtdb/metadata.clj
jarohen: latter is reading the leaves, IIUC?

core/src/main/clojure/xtdb/metadata.clj
```diff
@@ -353,7 +379,8 @@
 
   AutoCloseable
   (close [_]
-    (.clear chunks-metadata)))
+    (.clear chunks-metadata)
+    (util/close (vals (.asMap table-metadata-cache)))))
```
jarohen: nit: `util/close` takes maps, no need for `vals`

jarohen (approved):
nice one @FiV0, good unification here :ok_hand: 

minor nits/suggestions, but good to go without further review :pray:

core/src/main/clojure/xtdb/metadata.clj
FiV0: I essentially keep `meta-rel-reader` around for closing and `metadata-rdr` for reading the leafs. Wanted to avoid grabbing the `leaf-reader` every time, but agree it's a bit redundant.

jarohen: not worried about it being redundant, we do that in plenty of places - more concerned about the naming :slightly_smiling_face: i.e. shall we call the metadata-rdr some variant on 'metadata-leaf-rdr' instead?

core/src/main/clojure/xtdb/compactor.clj
jarohen: this was to do with lifecycle management, IIRC - previously (when it was based on postwalk) we had to keep the meta-file obj around. I suspect this is significantly simpler now, so yep, not entirely surprised we can remove this :slightly_smiling_face:

PR #3206, opened 2024-02-26
Remove unused jdbc/postgres test and clean up docker instructions

mbutlerw:


danmason (approved):
Given lack of support for JDBC object store of recent object store protocol changes (meaning that these tests would fail even if they were in the docker compose) and the prior plan to remove JDBC as an object store, happy to approve this

Issue #3196, opened 2024-02-21
Add cast between `duration` and `interval` in the EE


danmason:
Want to add the ability to `cast` between durations and intervals: 
- `duration`-> `interval` casting: 
  - Can be cast to `month-day-nano` - months == 0 within a `month-day-nano`, (value should be all in days).
  - Cannot be cast to `year-month` interval, and decide not to cast `day-time`.
- `interval`-> `duration` casing:
  - a `month-day-nano` can be cast to a `duration` if and only if months==0.
  - a `year-month` doesn't cast to a duration, and decided to not cast `day-time`.
In both cases, we assume a day to == 24h, as standard.

See descriptions of interval types in arrow [**here**](https://github.com/apache/arrow/blob/main/format/Schema.fbs#L240-L435).

Issue #2252, opened 2022-09-03
Error handling for inserting strings into time cols


refset:
```
jdt=> start transaction read write;                                                                                                                                                INSERT INTO Prop_Owner (id, customer_number, property_number, application_time_start, application_time_end)
VALUES (1, 145, 7799997, '1998-01-03', '1999-01-01');
commit;
BEGIN
INSERT 0 0
ERROR:  internal error on commit (report as a bug)
```

```
5:36:52 | ERROR core2.util | Uncaught exception:
java.lang.AbstractMethodError: Receiver class core2.vector$eval4732$fn$reify__4734 does not define or inherit an implementation of the resolved method 'abstract long readLong(int)' of interface core2.vector.IMonoVectorReader.
	at core2.indexer$$reify__17557$fn$reify__17564.accept(indexer.clj:473)
	at core2.operator.rename.RenameCursor$reify__12434.accept(rename.clj:41)
	at core2.operator.table.TableCursor$fn__12741.invoke(table.clj:33)
	at core2.operator.table.TableCursor.tryAdvance(table.clj:31)
	at core2.operator.rename.RenameCursor.tryAdvance(rename.clj:31)
	at core2.operator.ResultCursor.tryAdvance(operator.clj:38)
	at java.base/java.util.Spliterator.forEachRemaining(Spliterator.java:332)
	at core2.indexer$$reify__17557$fn__17563.invoke(indexer.clj:451)
	at core2.indexer$$reify__17557.indexOp(indexer.clj:450)
	at core2.indexer$$reify__17796.indexOp(indexer.clj:613)
	at core2.indexer.Indexer.indexTx(indexer.clj:686)
	at core2.ingester$eval17984$fn$reify__17988.acceptRecord(ingester.clj:76)
	at core2.log$tx_handler$fn__17917.invoke(log.clj:34)
	at clojure.core.protocols$iter_reduce.invokeStatic(protocols.clj:49)
	at clojure.core.protocols$fn__8230.invokeStatic(protocols.clj:75)
	at clojure.core.protocols$fn__8230.invoke(protocols.clj:75)
	at clojure.core.protocols$fn__8178$G__8173__8191.invoke(protocols.clj:13)
	at clojure.core$reduce.invokeStatic(core.clj:6886)
	at clojure.core$reduce.invoke(core.clj:6868)
	at core2.log$handle_polling_subscription$fn__17923.invoke(log.clj:48)
	at clojure.lang.AFn.run(AFn.java:22)
[...]

PR #3217, opened 2024-02-28
Casting between string and temporals - #3123

### LINKED ISSUE #3123: Casting strings <--> temporal values
Expecting/producing ISO8601 formats

See `CAST` in SQL Spec (Section 6.13 in SQL 2011).

* [x] string <--> date
* [x] string <--> time
* [x] string <--> timestamp
* [x] string <--> timestamp with timezone

## EDIT 1

Given that we don't typically surface Duration as a type within SQL, have left it out of the initial PR - can discuss & decide to do it as a follow up: 
* string <--> duration (months/years must be zero - e.g. `P5DT1H`)

## EDIT 2 (2024/03/01)

A couple of current points of consideration:
- Do we handle casting to VARCHAR(n)?
  - In the initial version of the PR, we did handle `n` and truncated our strings accordingly.
  - We could well leave this out/unsupported - it certainly does raise a number of questions re: should we retain truncation if the string is used elsewhere, should it be part of the type system, etc?
  - Might make sense to go with (initially) JUST unlimited varchars and to comment out anything to do with `(n)` - unless we particularly see a need arise to support varchar(n).
    - Would be treated as an additiion to the spec.
  - To that end - could comment out everything in `character_string_type` aside from just `varchar`, such that these errors get surfaced at the point of parsing. 
- What to do with Durations?
  - We certainly do expose Durations throughout various places (`datetime - datetime`), so its not like we dont have any concept them at the query level.
  - There's a fair argument about Durations being more generally useful as a measure of time elapsed.
  - Given our usage of them, might make sense to expose them a bit more generally as one of our data types?
  - Should probably also allow specifying precision here?

Current plan:
- VARCHAR:
  - Remove handling of VARCHAR(n) & truncating on length
  - Comment out anything unused in `character_string_type` such that we get parsing errors.
- Handle casting to Duration:
  - Add as a type to the SQL, plan for it.
  - Add to expression engine.
  - Support precision (as we do for timestamp/time) 
  - List as one of our data types in docs.

danmason:
**Issue:** #3123 
**Github Action Runs**: See [**here**](https://github.com/danmason/xtdb/actions?query=branch%3Acasting-string-to-temporals-3123++)

- Within the expression engine, implements casting between **utf8 strings** and **temporal** values. Temporal values supported:
  - Time without timezone
  - Timestamp with timezone
  - Timestamp without timezone
  - Date
- When casting to a temporal type  - ensure we handle `precision` being provided, changing type accordingly (and defaulting to micros otherwise)
- When casting to a VARCHAR, can optionally pass `(n)` to truncate the output string.
- Using the inbuilt Java temporal classes - ensures we both parse and output in terms of ISO specified strings. 
- Has some custom exception handling to present a decently formatted message when failing to parse a temporal string.
- Add necessary parts to the EBNF and SQL Planner to be able to convert between these types.
- Adds a number of tests around the casting behavior to both `temporal_test` and `sql_test`.

For reviewers - see the comments on the PR for a few caveats/questions on behavior.

## TODOs

Summarized from discussions on Slack, stuff to do:
- [x] Updates to units when casting to temporal types.
  - Default to `micro` as the unit when casting to temporal types.
  - Allow specification of `fractional-precision` , handle within the expression engine.
    - More than 6 we use nano as the unit
    - 6 or less we use micro as the unit
    - Truncate accordingly based on precision specified.
- [x] Updates to casting to VARCHAR:
  - Allow `n` to be provided, truncate accordingly.
  - Allow VARCHAR without `n` specified - do not truncate in that case.
- [x] Can probably clear up handling of and mentions of `Duration` here  (as both something to cast from and something to cast to), and drop from the issue.
- [x] Returning of `runtime exceptions` when failing to parse a temporal string - stops indexer from halting

### DIFF
--- a/core/src/main/clojure/xtdb/sql/plan.clj
+++ b/core/src/main/clojure/xtdb/sql/plan.clj
@@ -138,6 +138,10 @@
                             {::err/message (str "Cannot build interval for: "  (pr-str qualifier))
                              :qualifier qualifier}))))
 
+(defn cast-temporal-with-precision [e type fractional-precision]
+  (let [fp (parse-long fractional-precision)]
+    (list 'cast e [type (if (<= fp 6) :micro :nano)] {:precision fp})))
+
 (defn cast-expr [e cast-spec]
   (r/zmatch cast-spec
     [:exact_numeric_type "INTEGER"]
@@ -155,9 +159,39 @@
     (list 'cast e :f32)
     [:approximate_numeric_type "DOUBLE" "PRECISION"]
     (list 'cast e :f64)
+            
+    [:datetime_type "DATE"]
+    (list 'cast e [:date :day])
+            
+    [:datetime_type "TIME"]
+    (list 'cast e [:time-local :micro])
+    [:datetime_type "TIME" [:with_or_without_time_zone "WITHOUT" "TIME" "ZONE"]]
+    (list 'cast e [:time-local :micro])
+    [:datetime_type "TIME" [:unsigned_integer fractional-precision]]
+    (cast-temporal-with-precision e :time-local fractional-precision)
+    [:datetime_type "TIME" [:unsigned_integer fractional-precision] [:with_or_without_time_zone "WITHOUT" "TIME" "ZONE"]]
+    (cast-temporal-with-precision e :time-local fractional-precision)
 
+    [:datetime_type "TIMESTAMP"]
+    (list 'cast e [:timestamp-local :micro])
+    [:datetime_type "TIMESTAMP" [:with_or_without_time_zone "WITHOUT" "TIME" "ZONE"]]
+    (list 'cast e [:timestamp-local :micro])
+    [:datetime_type "TIMESTAMP" [:unsigned_integer fractional-precision]]
+    (cast-temporal-with-precision e :timestamp-local fractional-precision)
+    [:datetime_type "TIMESTAMP" [:unsigned_integer fractional-precision] [:with_or_without_time_zone "WITHOUT" "TIME" "ZONE"]]
+    (cast-temporal-with-precision e :timestamp-local fractional-precision)
+    
     [:datetime_type "TIMESTAMP" [:with_or_without_time_zone "WITH" "TIME" "ZONE"]]
     (list 'cast-tstz e)
+    [:datetime_type "TIMESTAMP" [:unsigned_integer fractional-precision] [:with_or_without_time_zone "WITH" "TIME" "ZONE"]]
+     (let [fp (parse-long fractional-precision)]
+       (list 'cast-tstz e {:precision fp :unit (if (<= fp 6) :micro :nano)}))
+
+    [:character_string_type "VARCHAR"]
+    (list 'cast e :utf8)
+
+    [:character_string_type "VARCHAR" [:character_length [:unsigned_integer length]]]
+    (list 'cast e :utf8 {:length (parse-long length)})
 
     (throw (err/illegal-arg :xtdb.sql/parse-error
                             {::err/message (str "Cannot build cast for: " (pr-str cast-spec))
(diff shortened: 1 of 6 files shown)


core/src/main/resources/xtdb/sql/parser/sql.ebnf
```diff
@@ -384,6 +384,12 @@ correlation_name
     | boolean_type
     | datetime_type
     | interval_type
+    | duration_type
+    | varchar_type
+    ;
+
+varchar_type
```
danmason: Separate VARCHAR type - open to think a bit more about this, but `character_string_type`/`character_large_object_type` currently isnt being used anywhere, and it made sense to create a separate "VARCHAR" type than use `character_string_type`.

In any case - I dont think we want to use the current VARCHAR type in `character_string_type` as a cast target, otherwise casting to a string/varchar would look like this:
> `CAST(DATE '2011-01-01' AS VARCHAR(n))`

(Which would require a user to specify a character length)

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -155,10 +155,26 @@
     (list 'cast e :f32)
     [:approximate_numeric_type "DOUBLE" "PRECISION"]
     (list 'cast e :f64)
-
+            
+    [:datetime_type "DATE"]
+    (list 'cast e [:date :day])
+    [:datetime_type "TIME"]
+    (list 'cast e [:time-local :micro])
```
danmason, line 167: When casting to temporal, need to specify the specific type of temporal we want. Went with `:<temporal> :micro` in most cases, given that seems to be most fairly typically used (it is _also_ what is used when calling `cast-tstz`, so there is some prior art of doing that).

Happy to change this - we could use `:nano` here so we don't lose nanos while casting from a string with nanos specified (would need to change the `with time zone` cast to not use `cast-tstz`)

core/src/main/resources/xtdb/sql/parser/sql.ebnf
```diff
@@ -495,6 +501,10 @@ interval_type
     : 'INTERVAL' interval_qualifier
     ;
 
+duration_type 
+    : 'DURATION'
```
danmason: Not sure if we really expose `DURATION` much to SQL - if we want to cast to duration we will need some concept of a duration type, so added one here as a possible cast target (will need similar for #3196)

src/test/clojure/xtdb/node_test.clj
```diff
@@ -259,8 +259,8 @@ WHERE foo.xt$id = 1"]])]
       (t/is (= {{:xt/id 1, :v 1} 2, {:xt/id 1, :v 2} 1}
                (q2 {:basis {:at-tx tx2}, :default-all-valid-time? true}))))))
 
-(t/deftest test-error-handling-inserting-strings-into-app-time-cols-397
```
danmason: (See #2252)

By my understanding, this test relied on inserting a string to an app time column failing mid transaction. Post our changes - this now fails at the point of running `submit-tx` due to attempting and failing to cast the string to a zoned date time. Updated the test to instead insert an integer - seems to have previous behavior, testing transaction rollback (I believe).

core/src/main/clojure/xtdb/sql/plan.clj
mbutlerw: This seems reasonable for now.

core/src/main/resources/xtdb/sql/parser/sql.ebnf
mbutlerw: I think `VARCHAR(n)` is the correct way to go here, its both spec compliant and standard behaviour in other databases. 

If we wish to add a no-limit version as well (e.g. `VARCHAR`) I'm not against that.

mbutlerw (changes requested):

mbutlerw (approved):

PR #3219, opened 2024-03-01
Updates to string<-->temporal casting: Handling durations, changes to Varchar

### LINKED ISSUE #3123: Casting strings <--> temporal values

danmason:
**Github Action Runs**: [**Link**](https://github.com/danmason/xtdb/actions/workflows/build.yml?query=branch%3Atemporal-casting-changes++)
Resolving some of the extra actions from #3123 

- VARCHAR changes:
  - Only handling `varchar` (without `(n)` specified).
  - Commenting out unused parts of `character string type`
- `String <--> Duration`
  - Added `duration_type` as a cast target in the SQL grammar, add plan for it. 
  - Handled in the EE.
  - Handling `precision` being specified as well (ie, `DURATION (9)`).
  - Adds `Duration` as a data type in the data type table.

### DIFF
--- a/core/src/main/clojure/xtdb/expression/temporal.clj
+++ b/core/src/main/clojure/xtdb/expression/temporal.clj
@@ -339,6 +339,26 @@
                             (local-time->nano))
                       (with-conversion :nano tgt-tsunit)))})
 
+(defn alter-duration-precision [^long precision ^Duration duration]
+  (if (= precision 0)
+    (.withNanos duration 0)
+    (.withNanos duration (let [nanos (.getNano duration)
+                               factor (Math/pow 10 (- 9 precision))]
+                           (* (Math/floor (/ nanos factor)) factor)))))
+
+(defn duration->nano ^long [^Duration d]
+  (.toNanos d))
+
+(defmethod expr/codegen-cast [:utf8 :duration] [{[_ tgt-tsunit :as target-type] :target-type  {:keys [precision]} :cast-opts}]
+  (when precision (ensure-fractional-precision-valid precision))
+  {:return-type target-type
+   :->call-code (fn [[s]]
+                  (-> `(->> (expr/resolve-string ~s)
+                            (parse-with-error-handling "duration" #(Duration/parse %))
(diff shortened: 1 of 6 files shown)


jarohen (approved):
grand, cheers @danmason :pray:

PR #3224, opened 2024-03-04
Updates to interval comparison/equality #2307

### LINKED ISSUE #2307: Interval comparison / equality (§8.2)
p439, §8.2 general rule 7.

* Can only compare within year-month and within month-day-nanos.
* Roughly: convert to the lowest common interval unit, convert to integer and compare.
* Additional fun for us is within MDN intervals (for non-zero months, which don't exist in SQL) - we have a few options:
  1. prevent it, statically - type error
  2. dynamically allow it, runtime error for non-zero months
  3. same, but say that intervals with same months are comparable, different months compare to null
  4. magically convert 30d=1m (like Postgres)

  If we go for 3 or 4, we could extend the comparisons to between YM/DT/MDN, but we lose the invariant that 'X + A = R and X + B = R iff A = B'.

danmason:
**Issue:** #2307
**Github Action Runs**: See [**here**](https://github.com/danmason/xtdb/actions/workflows/build.yml?query=branch%3Acompare-intervals+)

Within this PR, there's a number of updates & additions to comparison/equality between intervals. In particular:
- Can compare only within same typed interval - if differing intervals are compared, we throw an unsupported operation exception.
- If attempting to compare a `:day-time` interval - we throw an unsupported operation exception.
- For `year-month` intervals:
  - Convert to total months & compare that.
- For `month-day-nano` intervals:
  - Will throw a runtime error if either has `months > 0` (we have a few other places where in general we handle MDNs that way, so shall stick with it).
  - Otherwise, convert to a total of the MDN in `nanosseconds` and compare that.

Some notes:
- Currently, `single_field_interval` does return `day-time` when the field is not `YEAR`/`MONTH` so we will have the case that one cannot compare between those.

### DIFF
--- a/core/src/main/clojure/xtdb/expression/temporal.clj
+++ b/core/src/main/clojure/xtdb/expression/temporal.clj
@@ -656,6 +656,31 @@
   (with-arg-unit-conversion x-unit y-unit
     (constantly :i32) #(do `(Long/compare ~@%))))
 
+(defn compare-ym-intervals ^long [^PeriodDuration x ^PeriodDuration y]
+  (Long/compare (.toTotalMonths (.getPeriod x)) (.toTotalMonths (.getPeriod y))))
+
+(defn compare-mdn-intervals ^long [^PeriodDuration x ^PeriodDuration y]
+  (let [^Period x-period (.getPeriod x)
+        ^Period y-period (.getPeriod y)]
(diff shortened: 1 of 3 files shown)


mbutlerw (approved):
LGTM :shipit:

PR #3228, opened 2024-03-05
Casting between durations <--> MDN intervals #3196

### LINKED ISSUE #3196: Add cast between `duration` and `interval` in the EE

danmason:
**Issue:** #3196
**Github Action Runs**: See [**here**](
https://github.com/danmason/xtdb/actions?query=branch%3Acast-intervals-and-durations-3196+)

Handles casting between Intervals and Duration:
- For both - we implicitly assume 1 day == 24 hours
- For `interval-->duration`:
  - Based on the ticket, only a `month-day-nano` interval can be converted to a duration - attempting to cast from a `day-time` or `year-month` interval throws an unsupported operation exception. 
  - We ONLY convert a `month-day-nano` **if** total months == 0.
  - When converting to a duration we also consider `precision` (if passed in). We use an appropriate return type (defaults to `micro`), and truncate duration accordingly. 
- For `duration-->interval`:
  - We **always** return a `month-day-nano` interval.
  - This interval will only ever have `days` - `month` will always = 0.
  - We add appropriate SQL planning/parsing for this.
  - We assume no interval qualifier is specified - casting to an interval looks like `CAST(foo.duration AS INTERVAL)`
- Add a number of appropriate tests to `temporal_test` and `sql_test`.

We also make a small change to the `with-conversion` function in `temporal.clj`:
- Separating out certain calls into their own function so we can type hint parameters and the result.
- This prevents certain unexpected `arithmetic exception: integer overflow` being thrown due to unclear types/reflection.

### DIFF
--- a/core/src/main/resources/xtdb/sql/parser/sql.ebnf
+++ b/core/src/main/resources/xtdb/sql/parser/sql.ebnf
@@ -501,7 +501,7 @@ with_or_without_time_zone
     ;
 
 interval_type
-    : 'INTERVAL' interval_qualifier
+    : 'INTERVAL' [ interval_qualifier ]
     ;
 
 row_type
--- a/core/src/main/clojure/xtdb/expression/temporal.clj
+++ b/core/src/main/clojure/xtdb/expression/temporal.clj
@@ -447,6 +447,19 @@
                             (duration->nano))
                       (with-conversion :nano tgt-tsunit)))})
 
+
+(defn duration->mdn-interval [^Duration d]
+  (let [days-in-duration (.toDays d)]
+    (PeriodDuration. (Period/ofDays days-in-duration)
+                     (.minusDays d days-in-duration))))
+
+(defmethod expr/codegen-cast [:duration :interval] [{[_ d-unit] :source-type}]
+  {:return-type [:interval :month-day-nano]
+   :->call-code (fn [[d]]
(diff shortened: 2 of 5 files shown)


core/src/main/resources/xtdb/sql/parser/sql.ebnf
```diff
@@ -501,7 +501,7 @@ with_or_without_time_zone
     ;
 
 interval_type
-    : 'INTERVAL' interval_qualifier
+    : 'INTERVAL' [ interval_qualifier ]
```
danmason, line 504: Making `interval_qualifier` optional here, such that we can use `CAST(... AS INTERVAL)`

PR #3233, opened 2024-03-11
Casting to and from intervals

### LINKED ISSUE #2308: Casting to/from intervals
See §6.13, general rule 19, p247

* [x] Interval literals
  * IQ becomes optional. if present, parse as SQL (i.e. disallow ISO); if absent, parse as ISO.
* [x] Duration literals
  * SQL extension
  * parsed as ISO8601

* [x] Interval --> String
* [x] String --> Interval
  * IQ becomes optional; if not provided then it returns a full MDN interval (cf. postgres) in ISO8601 format
* [x] String --> Duration
  * expecting ISO8601 only. can't auto-infer units here (because value is dynamic)
* [x] Duration --> String
* [x] Duration --> Interval
  * If IQ present, normalise according to SQL spec
  * If IQ absent, MDN interval with just time (i.e. no 24H->1D cast)
* [x] Interval --> Interval
  * If IQ is absent, this is a no-op
  * If IQ is present, truncate and normalise as per SQL spec.
    e.g. if `DAY TO HOUR` specified, normalise with 24H -> 1D.
* [x] Int -> Single Field Interval
* [x] Docs for all of the above

### Notes

- Removed  'day-time intervals to/from Arrow durations' from this ticket  - we handle `MDN interval`<-->`Duration` in #3228
- Remvoe 'Single field intervals to ints' casting from this ticket - we lose that information within the expression engine, so it's fairly vague/undefined how exactly that should work. Could always follow up with extra functions/operations on durations/intervals at a later point.

danmason:
Initial cut of #2308 
**Test Runs:** [**Here**](https://github.com/danmason/xtdb/actions?query=branch%3Acasting-to-and-from-intervals++)

Implements the following: 
- Handling interval qualifier in the plan and returning a map of relevant info - useful as cast opts/literal args. 
- Adding cast to Interval with interval qualifier to SQL plan.
- Adding int -> interval with single field interval qualifier casting.
- Adding string -> interval with interval qualifier casting - calling down to single field interval / multi field interval respectively.
- Adding interval -> ISO 8601 string casting
- Adds numerous tests around all of the above.

Above will likely change within our follow up work on intervals - leaving out casting between interval types here.

### DIFF
--- a/core/src/main/clojure/xtdb/expression/temporal.clj
+++ b/core/src/main/clojure/xtdb/expression/temporal.clj
@@ -498,6 +498,17 @@
     (->multi-field-interval-call expr)
     (->single-field-interval-call expr)))
 
+(defn interval->iso-string [^PeriodDuration x]
+  (let [period-str (.toString (.getPeriod x))
+        duration-str (.toString (.getDuration x))]
+    (str period-str (str/replace duration-str #"^PT" "T"))))
+
+(defmethod expr/codegen-cast [:interval :utf8] [_]
+  {:return-type :utf8
+   :->call-code (fn [[pd]]
+                  `(-> (interval->iso-string ~pd)
+                       (string->byte-buffer)))})
+
 ;;;; SQL:2011 Operations involving datetimes and intervals
 (defn- recall-with-cast
   ([expr cast1 cast2] (recall-with-cast expr cast1 cast2 expr/codegen-call))
--- a/core/src/main/clojure/xtdb/expression/temporal.clj
+++ b/core/src/main/clojure/xtdb/expression/temporal.clj
@@ -477,6 +477,27 @@
                     "MINUTE" `(PeriodDuration. Period/ZERO (Duration/ofMinutes ~x))
                     "SECOND" `(PeriodDuration. Period/ZERO (Duration/ofSeconds ~x))))})
 
+(defn ->single-field-interval-call [{{:keys [start-field leading-precision fractional-precision]} :cast-opts}]
+  (let [expr {:op :call
+              :f :single_field_interval
+              :args [{} {:literal start-field} {:literal leading-precision} {:literal fractional-precision}]
+              :arg-types [:utf8 :utf8 :int :int]}]
+    (expr/codegen-call expr)))
+
+(defn ->multi-field-interval-call [{{:keys [start-field end-field leading-precision fractional-precision]} :cast-opts}]
+  (let [expr {:op :call
+              :f :multi_field_interval
+              :args [{} {:literal start-field} {:literal leading-precision} {:literal end-field} {:literal fractional-precision}]
+              :arg-types [:utf8 :utf8 :int :utf8 :int]}]
+    (expr/codegen-call expr)))
+
+(defmethod expr/codegen-cast [:utf8 :interval] [{{:keys [start-field end-field]} :cast-opts :as expr}]
+  ;; Calls down to multi-field-interval if both start-field and end-field present, 
+  ;; else calls to single-field-interval if only start-field is present
+  (if (and start-field end-field)
+    (->multi-field-interval-call expr)
+    (->single-field-interval-call expr)))
+
 ;;;; SQL:2011 Operations involving datetimes and intervals
 (defn- recall-with-cast
   ([expr cast1 cast2] (recall-with-cast expr cast1 cast2 expr/codegen-call))
--- a/core/src/main/clojure/xtdb/expression/temporal.clj
+++ b/core/src/main/clojure/xtdb/expression/temporal.clj
@@ -460,6 +460,23 @@
                        (Duration/ofNanos)
                        (duration->mdn-interval)))})
 
+(defmethod expr/codegen-cast [:int :interval] [{{:keys [start-field end-field]} :cast-opts}]
+  (when end-field (throw (err/illegal-arg :xtdb.expression/attempting-to-cast-int-to-multi-field-interval
+                                          {::err/message "Cannot cast integer to a multi field interval"
+                                           :start-field start-field
+                                           :end-field end-field})))
+  {:return-type (if (#{"YEAR" "MONTH"} start-field)
+                  [:interval :year-month]
+                  [:interval :month-day-nano])
+   :->call-code (fn [[x]]
+                  (case start-field
+                    "YEAR" `(PeriodDuration. (Period/ofYears ~x) Duration/ZERO)
+                    "MONTH" `(PeriodDuration. (Period/ofMonths ~x) Duration/ZERO)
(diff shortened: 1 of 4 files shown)


core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -460,6 +460,107 @@
...
 
+(defmethod expr/codegen-cast [:int :interval] [{{:keys [start-field end-field]} :cast-opts}]
+  (when end-field (throw (err/illegal-arg :xtdb.expression/attempting-to-cast-int-to-multi-field-interval
+                                          {::err/message "Cannot cast integer to a multi field interval"
+                                           :start-field start-field
+                                           :end-field end-field})))
+  {:return-type (if (#{"YEAR" "MONTH"} start-field)
+                  [:interval :year-month]
+                  [:interval :month-day-nano])
+   :->call-code (fn [[x]]
```
danmason, line 471: Could probably call `single-field-interval` here, though I re-implemented such that we'd return month-day-nano. Likely one to update when we take a look at how we should handle single field interval and the like within plan.

core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -460,6 +460,107 @@
...
+                  (case start-field
+                    "YEAR" `(PeriodDuration. (Period/ofYears ~x) Duration/ZERO)
+                    "MONTH" `(PeriodDuration. (Period/ofMonths ~x) Duration/ZERO)
+                    "DAY" `(PeriodDuration. (Period/ofDays ~x) Duration/ZERO)
+                    "HOUR" `(PeriodDuration. Period/ZERO (Duration/ofHours ~x))
+                    "MINUTE" `(PeriodDuration. Period/ZERO (Duration/ofMinutes ~x))
+                    "SECOND" `(PeriodDuration. Period/ZERO (Duration/ofSeconds ~x))))})
+
+(defn ->single-field-interval-call [{{:keys [start-field leading-precision fractional-precision]} :cast-opts}]
```
jarohen, line 480: as discussed, let's go the other way here longer term - getting single-field/multi-field interval to accept these as static arguments

core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -460,6 +460,107 @@
...
+    (not= minutes 0) (conj {:field "MINUTE" :rank 3})
+    (not= seconds nanos 0) (conj {:field "SECOND" :rank 4})))
+
+(defn- format-end-field [start-field end-field nanos?]
+  (cond
+    (and (= start-field end-field) nanos?) "SECOND(1,9)"
+    nanos? "SECOND(9)"
+    :else end-field))
+
+(defn- ->interval-str [days hours minutes seconds nanos start-rank end-rank]
```
jarohen: this is likely where you're getting your boxed-math warnings - all of the numeric parameters get boxed up here to call this function. 

unfortunately we can't put primitive type-hints on functions with more than 4 parameters in Clojure - possible to take the period and duration in here instead, perhaps?

core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -460,6 +460,107 @@
...
+        abs-total-months (abs (.toTotalMonths period))
+        year (quot abs-total-months 12)
+        month (mod abs-total-months 12)
+        interval-string (cond-> ""
+                          (not= year 0) (str year (when (not= month 0) "-"))
+                          (not= month 0) (str month))
+        interval-qualifier (cond-> ""
+                             (not= year 0) (str "YEAR" (when (not= month 0) " TO "))
+                             (not= month 0) (str "MONTH"))]
+    (format "INTERVAL '%s%s' %s" sign interval-string interval-qualifier)))
```
jarohen: I don't think we're to include `INTERVAL` nor the IQ in the string rendering, according to the spec?

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -138,6 +138,112 @@
                             {::err/message (str "Cannot build interval for: "  (pr-str qualifier))
                              :qualifier qualifier}))))
 
+(defn interval-qualifier->map [qualifier]
+  (r/zmatch qualifier
+            [:interval_qualifier
```
jarohen: wonder whether we can avoid the cross-product of different possibilities here with a couple of extracted fns for start/end?

core/src/main/clojure/xtdb/expression/temporal.clj
jarohen: (i.e. no action required in this PR)

core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -460,6 +460,107 @@
...
+            nanos (.toNanosPart abs-duration)
+            present-fields (get-present-fields days hours minutes seconds nanos)
+            {start-field :field start-rank :rank} (first present-fields)
+            {end-field :field end-rank :rank} (last present-fields)
+            formatted-end-field (format-end-field start-field end-field (> nanos 0))
+            interval-str (->interval-str days hours minutes seconds nanos start-rank end-rank)
+            interval-qualifier (if (= start-field end-field) formatted-end-field (format "%s TO %s" start-field formatted-end-field))]
+        (format "INTERVAL '%s%s' %s" sign interval-str interval-qualifier)))))
+
+(defmethod expr/codegen-cast [:interval :utf8] [{[_ iunit] :source-type}]
```
jarohen: given that the new proposal is to return ISO8601 here, I might be tempted to migrate this pre-PR to cut down on the churn? (not a strongly-held opinion)

core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -460,6 +460,107 @@
...
+(defmethod expr/codegen-cast [:utf8 :interval] [{{:keys [start-field end-field]} :cast-opts :as expr}]
+  ;; Calls down to multi-field-interval if both start-field and end-field present, 
+  ;; else calls to single-field-interval if only start-field is present
+  (if (and start-field end-field)
+    (->multi-field-interval-call expr)
+    (->single-field-interval-call expr)))
+
+(defn ym-interval->string [^PeriodDuration x]
+  (let [period (.getPeriod x)
+        sign (if (.isNegative period) "-" "")
```
jarohen: nit: I think `quot` might already handle putting the `-` sign in for you, save you taking the abs and then re-adding the sign

jarohen (changes requested):
Thanks @danmason - few minor tweaks here pre-merge, but otherwise let's get this landed and then pair up :slightly_smiling_face:

PR #3236, opened 2024-03-13
Copy row removal

### LINKED ISSUE #2739: Remove copying in scan
Currently, scan copies all of its input vectors - this was to get scan merging the tries without having an indirect vector that can range over multiple input vectors.

Needing to create copiers and then actually copy the rows introduces a significant performance overhead for small and large queries respectively.

FiV0:
Previously we were copying data in scan to a new relation that was then passed out of scan. We now create views over the existing content columns. For temporal columns we still copy to a new relation because of the resolution of the polygons.
The main new class is a `IndirectMultiVectorReader` which is a view over multiple `IVectorReader`s. It should behave like a `DuvReader` in case the underlying vectors don't have a monomorphic type and otherwise like a reader as returned by `getField`. 

Todo:
- [x] check performance
    - [x] TPC-H (SF 0.4) - hot and cold run
         - ~75s on main (uncompacted - with background tasks)
         - ~66.5s with copy row removal (uncompacted - with background tasks)
    - [x] Auctionmark 30sec run - number of get-item calls
         - ~72k calls on main (with one thread failure)
         - ~67k calls with copy row removal (with two thread failures)
         - failures are related to #3280
- [x] Review

### DIFF
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -22,6 +22,7 @@
            (java.io Closeable)
            java.nio.ByteBuffer
            (java.nio.file Path)
+           (com.carrotsearch.hppc IntArrayList)
            (java.util ArrayList Comparator HashMap Iterator LinkedList Map PriorityQueue)
            (java.util.function IntPredicate Predicate)
            (java.util.stream IntStream)
@@ -38,7 +39,8 @@
            xtdb.operator.IRelationSelector
            (xtdb.trie ArrowHashTrie$Leaf EventRowPointer HashTrie HashTrieKt LiveHashTrie$Leaf MergePlanNode MergePlanTask)
            (xtdb.util TemporalBounds TemporalBounds$TemporalColumn)
-           (xtdb.vector IRelationWriter IRowCopier IVectorReader IVectorWriter RelationReader RelationWriter)
+           (xtdb.vector IRelationWriter IRowCopier IVectorReader IVectorWriter RelationReader RelationWriter
+                        IMultiVectorRelationFactory IVectorIndirection$Selection IndirectMultiVectorReader)
            (xtdb.watermark ILiveTableWatermark IWatermarkSource Watermark)))
 
 (s/def ::table symbol?)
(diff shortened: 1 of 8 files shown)


core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -117,30 +119,43 @@
...
+                         [(.rowCount rel) (-> (.readerForName rel "op") (.legReader :put))])
+                       leaf-rels)
+        reader-indirection (IntArrayList.)
+        vector-indirection (IntArrayList.)]
+    (letfn [(->indirect-multi-vec [col-name reader-selection vector-selection]
+              (let [normalized-name (util/str->normal-form-str col-name)
+                    readers (ArrayList.)]
+                (if (= normalized-name "xt$iid")
+                  (doseq [^RelationReader leaf-rel leaf-rels]
+                    (if-let [rdr (some-> (.readerForName leaf-rel "xt$iid")
```
jarohen: nit: shouldn't need absence handling here - `xt$iid` is guaranteed present

core/src/main/kotlin/xtdb/vector/IVectorIndirection.kt
```diff
@@ -0,0 +1,37 @@
+package xtdb.vector
+
+interface IVectorIndirection {
+    fun valueCount(): Int
+
+    fun getIndex(idx: Int): Int
+
+    operator fun get(idx: Int): Int {
+        return getIndex(idx)
+    }
```
jarohen: in Kotlin, this is usually one-lined: 
```suggestion
    operator fun get(idx: Int) = getIndex(idx)
```

core/src/main/kotlin/xtdb/vector/IndirectMultiVectorReader.kt
```diff
@@ -0,0 +1,303 @@
...
+    private fun safeReader(idx: Int): IVectorReader {
+        return readers[readerIndirection[idx]] ?: throw unsupported()
+    }
+
+    override fun valueCount(): Int {
+        return readerIndirection.valueCount()
+    }
+
+    override fun getName(): String {
+        return readers.filterNotNull().first().name
```
jarohen: nit: worth having `name` as a field on the vector?

core/src/main/kotlin/xtdb/vector/IndirectMultiVectorReader.kt
```diff
@@ -0,0 +1,303 @@
...
+
+    private val fields: List<Field?> = readers.map { it?.field }
+    private val legReaders = ConcurrentHashMap<Keyword, IVectorReader>()
+    private val vectorField by lazy (LazyThreadSafetyMode.PUBLICATION){
+        MERGE_FIELDS.applyTo(RT.seq(fields.filterNotNull())) as Field
+    }
+
+    companion object {
+        private val MERGE_FIELDS: IFn = requiringResolve("xtdb.types/merge-fields")
+        private val VEC_TO_WRITER: IFn = requiringResolve("xtdb.vector.writer/->writer")
```
jarohen: this is now `FieldVectorWriters.writerFor` - the Clojure one is just a call to that method (for historical reasons).

core/src/main/kotlin/xtdb/vector/IndirectMultiVectorReader.kt
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    override fun isNull(idx: Int): Boolean {
+        val readerIdx = readerIndirection[idx]
+        return readerIdx < 0 || readers[readerIdx] == null || readers[readerIdx]!!.isNull(vectorIndirections[idx])
+    }
+
+    override fun getBoolean(idx: Int): Boolean {
+        return safeReader(idx).getBoolean(vectorIndirections[idx])
+    }
```
jarohen: when it's a longer line, can still write it as an expression:
(IntelliJ will offer this as a refactoring)

```suggestion
    override fun getBoolean(idx: Int) =
        safeReader(idx).getBoolean(vectorIndirections[idx])
```

core/src/main/kotlin/xtdb/vector/IndirectMultiVectorReader.kt
```diff
@@ -0,0 +1,303 @@
...
+    override fun getObject(idx: Int): Any? {
+        return safeReader(idx).getObject(vectorIndirections[idx])
+    }
+
+    override fun getObject(idx: Int, keyFn: IKeyFn<*>?): Any? {
+        return safeReader(idx).getObject(vectorIndirections[idx], keyFn)
+    }
+
+    override fun structKeyReader(colName: String): IVectorReader {
+        // TODO what happens if underlying readers have a different type?
```
jarohen: this is one we'll need to handle

core/src/test/kotlin/xtdb/vector/IndirectMultiVectorReaderTest.kt
```diff
@@ -0,0 +1,307 @@
...
+    private fun readMaps(valueReader: IValueReader): Any? {
+        return when (val o = valueReader.readObject()) {
+            is Map<*, *> -> o.mapValues { readMaps(it.value as IValueReader) }
+            else -> o
+        }
+    }
+
+    @Test
+    fun testMonomorphicStructVectors() {
+        val fooField = Field("foo", FieldType(false, BOOL_TYPE, null, null), null)
```
jarohen: nit: `FieldType.notNullable`

core/src/main/kotlin/xtdb/vector/IVectorIndirection.kt
```diff
@@ -0,0 +1,37 @@
+package xtdb.vector
```
jarohen: meta: if you're doing straight Java -> Kotlin, these are best committed separately straight to `main` :pray:

core/src/main/kotlin/xtdb/vector/IVectorIndirection.kt
```diff
@@ -0,0 +1,37 @@
...
+    fun valueCount(): Int
+
+    fun getIndex(idx: Int): Int
+
+    operator fun get(idx: Int): Int {
+        return getIndex(idx)
+    }
+
+    @Suppress("DEPRECATION")
+    @JvmRecord
```
jarohen: nit: probably doesn't need to be a record now that it's a Kotlin data class?

core/src/main/kotlin/xtdb/vector/IVectorIndirection.kt
```diff
@@ -0,0 +1,37 @@
...
+interface IVectorIndirection {
+    fun valueCount(): Int
+
+    fun getIndex(idx: Int): Int
+
+    operator fun get(idx: Int): Int {
+        return getIndex(idx)
+    }
+
+    @Suppress("DEPRECATION")
```
jarohen: what's the deprecation here, OOI? can we avoid it?

jarohen (changes requested):
cheers @FiV0 :pray: :clap:

all minor changes apart from the handling child polymorphism - happy to chat about that if it'd be helpful :slightly_smiling_face:

core/src/main/kotlin/xtdb/vector/IndirectMultiVectorReader.kt
FiV0: I think I put the `TODO` in when I started off. This should be handled fine by the recursive usage of IMVR.

core/src/main/kotlin/xtdb/vector/IVectorIndirection.kt
FiV0: It's on the `formatted` fn below.

jarohen: ah, I see - Kotlin string interpolation instead?

jarohen (approved):

PR #3244, opened 2024-03-14
Add AGE(dt, dt) function to EE and SQL.

### LINKED ISSUE #3198: Add support for `age(dt, dt)` function
See functions here https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-TABLE

- To return an interval - exact format & granularity to be decided.
- MDN with populated full months?

danmason:
**Related Issue:** #3198 
**GH Action Runs:** [Here](https://github.com/danmason/xtdb/actions?query=branch%3Aadd-age-fn-3198++)

Adds the `Age(dt, dt)` function, returning a **month-day-nano** interval representing the time between the two dates:
- Adds to the expression engine.
- Adds into SQL parsing, planning.
- Adds it under `temporal.adoc` docs.
- Adds numerous tests to both temporal-test and sql-test to check behaviour.

### DIFF
--- a/core/src/main/clojure/xtdb/expression/temporal.clj
+++ b/core/src/main/clojure/xtdb/expression/temporal.clj
@@ -1440,6 +1440,49 @@
                       "MILLISECOND" `(->period-duration ~period (.truncatedTo ~duration ChronoUnit/MILLIS))
                       "MICROSECOND" `(->period-duration ~period (.truncatedTo ~duration ChronoUnit/MICROS)))))})
 
+(defn ->mdn-interval-between [^LocalDateTime end-dt ^LocalDateTime start-dt]
+  (let [period-between (Period/between (.toLocalDate start-dt) (.toLocalDate end-dt))
+        period-days (.getDays period-between)
+        duration-between (Duration/between start-dt end-dt)
+        start-lt (.toLocalTime start-dt)
+        end-lt (.toLocalTime end-dt)
+        period-day-adjustment (cond
+                                (and (.isNegative period-between)
+                                     (.isBefore start-lt end-lt))
+                                1
+
+                                (and (not (.isNegative period-between))
+                                     (not= period-days 0)
+                                     (.isAfter start-lt end-lt))
+                                -1
+
+                                :else 0)
+        adjusted-period (.plusDays period-between period-day-adjustment)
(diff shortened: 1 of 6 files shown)


core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -1440,6 +1440,49 @@
                       "MILLISECOND" `(->period-duration ~period (.truncatedTo ~duration ChronoUnit/MILLIS))
                       "MICROSECOND" `(->period-duration ~period (.truncatedTo ~duration ChronoUnit/MICROS)))))})
 
+(defn ->mdn-interval-between [^LocalDateTime end-dt ^LocalDateTime start-dt]
+  (let [period-between (Period/between (.toLocalDate start-dt) (.toLocalDate end-dt))
+        period-days (.getDays period-between)
+        duration-between (Duration/between start-dt end-dt)
+        start-lt (.toLocalTime start-dt)
+        end-lt (.toLocalTime end-dt)
+        period-day-adjustment (cond
```
jarohen, line 1449: mind walking me through the logic here when you get a mo?

(I'm sure it's fine, it's just not intuitive to me!)

jarohen (approved):

danmason:
Some notes on some of the logic in here-

### Why we do period-day-adjustment:
An example - End DT: 2020-01-01 00:00:00, Start DT: 2019-12-31 12:00:00  
- Actual difference is 12 hours, but the period difference will be calculated as 1 day, duration as 36 hours (we remove any days from the duration prior to output) 
- The period differs as the dates differ - so we will need to correct the period to 0 days as the overall difference isn't 1 whole day.

We use local time to calculate if we need to adjust the period (below is in the positive difference case): 
  - If the times were the same, there actually WOULD be a period difference of 1 day 
     - (ie, 2020-01-01 00:00:00 and 2019-12-31 00:00:00)  .
  - If the start time was AFTER the end time, there would be a difference of greater than 1 day, so again, there would be a full day difference between the two.
     - (ie, 2020-01-01 12:00:01 and 2019-12-31 12:00:00)  
  - As start time is BEFORE the end time, there is < day difference, so we need to adjust the period accordingly (minus 1 day)  

Same applies with negative differences, though in reverse.

PR #3254, opened 2024-03-21
Make data types a subsection & add temporal types doc.

danmason:
Preview docs here: https://temporal-types-docs-2308.d2zcybuz6k9g4m.amplifyapp.com/

Breaking `data types` out into an (uncollapsed) subsection, with a separate document for "temporal types":
- Lists and gives examples of creating the various types.
  - Clojure types split out under their own section in the `drivers` docs.  
- Talks about details around intervals.
- Talks about casting behaviour

We also update various usages/mentions of temporal types and their constructors to point appropriately towards the relevant places.

### DIFF
--- a/docs/src/content/docs/reference/main/data-types/temporal-types.adoc
+++ b/docs/src/content/docs/reference/main/data-types/temporal-types.adoc
@@ -1,0 +1,143 @@
+---
+title: XTDB Temporal Types
+---
+
+NOTE: The contents of this page focus mainly on construction and casting of Temporal Types. For information on functions which apply to temporal types, see the link:../stdlib/temporal[Temporal Functions] page.
+
+This document outlines the supported types of XTDB, alongside their usage, construction and casting behaviors.
+
+== Supported Temporal Types
+
+[cols="3,5",options="header"]
+|===
+|XTDB Type |Description
+
+|`TIMESTAMP [WITHOUT TIMEZONE]`
+|Represents both date and time *without* a time zone offset.
+
+|`TIMESTAMP WITH TIMEZONE`
+|Represents both date and time *with* a time zone offset.
+
+|`DATE`
+|Represents a date without a time component.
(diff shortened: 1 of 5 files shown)


docs/src/content/docs/reference/main/data-types/temporal-types.adoc
```diff
@@ -0,0 +1,143 @@
...
+TIME '14:30:00' -- Time without time zone.
+INTERVAL '1-2' YEAR TO MONTH -- A Year-Month Interval specified with an interval qualifier.
+INTERVAL 'P1Y2M' -- The Same Interval specified in a ISO8601 format.
+INTERVAL '1 2:30:10.111111' DAY TO SECOND(6) -- A Day-Time Interval specified with an interval qualifier.
+INTERVAL 'P1DT2H30M10.111111S' -- The Same Interval specified in a ISO8601 format.
+DURATION 'P20H3M11.1S' -- A Duration specified in ISO8601 format.
+PERIOD(DATE '1998-01-05', DATE '1998-01-12') -- A Period specified with two dates.
+----
+
+=== XTQL/Clojure Examples
```
refset: We might be better to have XTQL/JSON examples here instead, or as well, but ahead of the XTQL/Clojure. Or perhaps both should live under the various "Drivers" sections (might be a good opporunity to use asciidoc Includes?). I'm aware the overview data types page currently lacks JSON and shows Clojure, but that might be better to rethink also rather than follow.

danmason: Current thoughts/changes might make here:
- Move Clojure examples out to a separate document "temporal types".
- Create an issue around missing XTQL JSON types.
- Keep SQL examples on the page - and maybe add JSON as/when they're ready to go.

Can merge what we have afterwards as it should be an improvement on what we already have :)

PR #3264, opened 2024-03-27
Add table/schema privilege function parsing, always returning true.

### LINKED ISSUE #3216: Rudimentary support for Postgres table access control functions
Metabase (see #3093) makes a [query](https://github.com/metabase/metabase/blob/1ea996899716252675644f8f1e1bb0c09399e115/src/metabase/driver/postgres.clj#L863-L866) at startup that calls two table access control functions: `pg_catalog.has_table_privileges` and `pg_catalog.has_schema_privileges`.

* Before we have finer-grained access control, we should return that the current user has full read/write access to all schemas/tables/views
* This may involve changes to the SQL grammar/parser. TBC: 
  * what is this syntax implying? i.e. can functions be scoped to a schema, or is `pg_catalog` a receiver?
  * is it SQL-spec compliant?

danmason:
**Github Action Runs**: [Here](https://github.com/danmason/xtdb/actions?query=branch%3Apostgres-table-access-functions-3216++)
**Linked Issue:** Resolves #3216 

Adds initial support for some of the table access control/"Access Privilege Inquiry Functions" to our SQL:
- Works with or without `pg_catalog.` specified the function call.
  - If it is, we essentially handle it in the parser and discard it prior to planning.
- Currently implementing:
  - All the arities for `has_table_privilege`
  - All the arities for `has_schema_privilege` 
- Always returning true from the planner.

### DIFF
--- a/src/test/clojure/xtdb/sql_test.clj
+++ b/src/test/clojure/xtdb/sql_test.clj
@@ -1896,35 +1896,19 @@
 
 (t/deftest test-postgres-access-control-functions
   ;; These current functions should always should return true
+  (t/are [sql expected] (= expected (plan-expr sql))
+    "has_table_privilege('xtdb','docs', 'select')" true
+    "has_table_privilege('docs', 'select')" true
+    "pg_catalog.has_table_privilege('docs', 'select')" true
 
-  (t/testing "has table privilege"
-    (t/is (= [{:xt/column-1 true}]
-             (xt/q tu/*node* "SELECT has_table_privilege('xtdb','docs', 'select') FROM (VALUES 1) AS x")))
+    "has_schema_privilege('xtdb', 'public', 'select')" true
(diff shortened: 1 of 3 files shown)


src/test/clojure/xtdb/sql_test.clj
```diff
@@ -1893,3 +1893,39 @@
   
   (t/is (= [{:xt/column-1 "public"}]
            (xt/q tu/*node* "SELECT current_schema FROM (VALUES 1) AS x"))))
+
+(t/deftest test-postgres-access-control-functions
+  ;; These current functions should always should return true
```
mbutlerw, line 1898: Could dry up these test examples, so that the `expr` is templated into the query for each case. See `plan-expr` for inspiration, or change these to plan tests and use it directly.

mbutlerw (approved):

PR #3267, opened 2024-03-28
Handle `routine_invocation` - use for `has_table_privilege` and `has_schema_privilege`.

### LINKED ISSUE #3216: Rudimentary support for Postgres table access control functions

danmason:
**Github Action Runs**: [Here](https://github.com/danmason/xtdb/actions?query=branch%3Aroutine-invocation-3216++)
**Linked Issue:** Resolves #3216 

Adds handling for `routine_invocation` and updates `has_table_privilege` and `has_schema_privilege` to use it.

### DIFF
--- a/core/src/main/clojure/xtdb/sql/plan.clj
+++ b/core/src/main/clojure/xtdb/sql/plan.clj
@@ -393,6 +393,14 @@
        ;; should never be reached
        (throw (err/illegal-arg :xtdb.sql/parse-error {::err/message "Invalid Period Predicand"}))))))
 
+(defn routine-invocation [_schema routine-name _fn-opts]
+  (case (str/lower-case routine-name)
+    "has_table_privilege" true
+    "has_schema_privilege" true
+    (throw (err/illegal-arg :xtdb.sql/parse-error
+                            {::err/message (str "Attempting to invoke unknown routine - " routine-name)
+                             :routine-name routine-name}))))
+
 (defn expr [z]
   (r/zmatch z
     [:column_reference _]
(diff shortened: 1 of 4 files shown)


core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -393,6 +393,14 @@
        ;; should never be reached
        (throw (err/illegal-arg :xtdb.sql/parse-error {::err/message "Invalid Period Predicand"}))))))
 
+(defn routine-invocation [_schema routine-name _fn-opts]
```
danmason, line 396: Essentially ignoring both schema and shape of SQL args here - _could_ potentially check schema (assume PG_CATALOG) if present and then point to a function under that namespace, but not I'm sure that really adds anything for us or is really necessary?

core/src/main/resources/xtdb/sql/parser/sql.ebnf
```diff
@@ -380,6 +380,10 @@ schema_name
     : [ schema_name <period> ] !#'(?i)(\b(SYSTEM_TIME|VALID_TIME)\b)' identifier
     ;
 
+(* Following the lead of regular_identifier here, filtering known function calls *)
+routine_identifier
```
danmason, line 384: Essentially looking at EVERY other potential `fn_name()` type invocation in the grammar and explicitly checking that the routine name is NOT one of those - in a similar way to how regular_identifier disambiguates things that look like column names.

danmason:
Decided against doing this for now due to ambiguity of grammar and need for workarounds.

PR #3272, opened 2024-04-02
[WIP] Schema flattening

FiV0:
Placeholder PR for schema flattening.

Still lots of stuff todo:
- [ ] Fix issues with Sparse Union vectors upstream (see https://github.com/apache/arrow-java/issues/108)
- [ ] Copiers SUV -> DUV, DUV -> SUV


FiV0:
Can come back to this if we are really seeing the indirection of the toplevel DUV's being a problem.

PR #3285, opened 2024-04-08
Add test framework

l4e21:


lang/python/README.md
```diff
@@ -11,6 +11,19 @@ To install the dependencies, run:
 poetry install
 ```
 
+## Testing
+Through localhost connection
+```shell
+poetry run pytest
+```
+
+Through Docker
```
Akeboshiwind: Took a stab at changing the docker-compose to make this one command:
```yaml
version: '3'
services:
  db:
    image: ghcr.io/xtdb/xtdb-standalone-ea
    ports:
      - "6543:3000" 
    stdin_open: true 
    tty: true
    healthcheck:
      test: ["CMD", "curl", "-f", "http://localhost:3000/status"]
      interval: 1s
      timeout: 1s
      retries: 10
      start_period: 5s
  app:
    environment:
        DATABASE_HOST: db
        DATABASE_PORT: 3000
    build:
        context: .
        dockerfile: Dockerfile
    depends_on:
      db:
        condition: service_healthy
```

With the final command being: `docker-compose up --build --abort-on-container-exit`

Not sure if it's better though as I'd never heard of `healthcheck` or `--abort-on-container-exit` before 🤷

lang/python/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN pip install poetry
+
+
+ENV PATH="/root/.poetry/bin:${PATH}"
+
+COPY ./xtdb_juxt ./xtdb_juxt
+COPY ./pyproject.toml .
+COPY ./README.md .
+
+RUN poetry install
```
Akeboshiwind, line 11: Small note, if you make this:
```Dockerfile
COPY ./pyproject.toml .
RUN poetry install

COPY ./xtdb_juxt ./xtdb_juxt
```

Then docker will only re-run `poetry install` (which can take a while) when the pyproject.toml changes.
Not super important because I think github actions doesn't cache image layers, but nice for local development :)

lang/python/README.md
l4e21: Does seem to depend on the start period but works with a start period of 30s without variation.

PR #3288, opened 2024-04-08
Add infix regex operators to SQL #3213

### LINKED ISSUE #3213: Infix SQL regex operators
Addition to the SQL spec - see Postgres ['POSIX regex functions'](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-POSIX-REGEXP)

* `~`, `~*`, `!~`, `!~*`
* Regex language to be Java's - i.e. don't try to be Postgres/SQL regex compatible
* Expect regex to be a compile-time literal - i.e. to be extracted in `parse-list-form` and pre-compiled.
* If the [SQL-spec compliant functions](https://www.postgresql.org/docs/current/functions-matching.html#POSIX-VS-XQUERY) don't already exist, create those and make the Postgres ones aliases (same as #3211)

danmason:
**Github Action Runs**: [Here](https://github.com/danmason/xtdb/actions?query=branch%3Ainfix-regex-operators-3213++)
**Linked Issue:** Resolves #3213 

- Adds the infix regex operators from postgres - see [here](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-POSIX-REGEXP).
  - Essentially - these are parsed in the grammar and plan to use the existing `like_regex` function (which already can handle being non case sensitive.
  - Adds some tests under `sql_test` for the planning and usage of these. 
- Fixes a small issue within the implementation to `like_regex`:
  - Discovered while in `sql_test` - Seemed to essentially treat the compiled pattern as if it was a clojure string regex, could see that from the codegen, so it was disregarding the tags (such as saying the pattern should be case insensitive) 
  - Seemed to come as a result from compiling the pattern ahead of time - so fixed this by constructing/compiling the pattern as part of the executed code.
  - Added a test around `like_regex` using case insensitive tag at the `sql_test` level to verify this as fixed.

### DIFF
--- a/src/test/clojure/xtdb/sql_test.clj
+++ b/src/test/clojure/xtdb/sql_test.clj
@@ -537,6 +537,53 @@
   (t/is (= [{:match true}]
            (xt/q tu/*node* "SELECT ('ABC' LIKE_REGEX 'a' FLAG 'i') as match FROM (VALUES 1) AS x"))))
 
+(t/deftest test-postgres-regex-expr
+  (t/are [sql expected] (= expected (plan-expr sql))
+
+    "foo.a ~ foo.b" '(like-regex x1 x2 "")
+    "foo.a ~* foo.b" '(like-regex x1 x2 "i")
+
+    "foo.a !~ foo.b" '(not (like-regex x1 x2 ""))
+    "foo.a !~* foo.b" '(not (like-regex x1 x2 "i"))))
+
+(defn pg-regex-query [val op pattern]
+  (let [query (format "SELECT ('%s' %s '%s') as match FROM (VALUES 1) AS x" val op pattern)
+        [{:keys [match]}] (xt/q tu/*node* query)]
+    match))
+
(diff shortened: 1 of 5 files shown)


src/test/clojure/xtdb/sql_test.clj
```diff
@@ -530,6 +530,60 @@
     "foo.a NOT LIKE_REGEX foo.b" '(not (like-regex x1 x2 ""))
     "foo.a NOT LIKE_REGEX foo.b FLAG 'i'" '(not (like-regex x1 x2 "i"))))
 
+(t/deftest test-like-regex-query-case-insensitive
```
danmason, line 533: Was not previously working until the change to `expression.clj` - looking at the prior codegen, I believe it was essentially treating the compiled pattern (`(Pattern/compile re-literal flag-int)`) as a regular regex without any flags set. Compiling the pattern at the point of code execution seemed to fix these (and also some of the tests within `pg-regex-query`, that were failing for the same reason)

PR #3289, opened 2024-04-09
Refactor node to expose query phases

### LINKED ISSUE #3255: Pgwire returns fields in order of the SQL statement
Relies on work outlined in https://github.com/xtdb/xtdb/issues/3266

Which parts of the pgwire server protocol card end up being implemented as part of this feature can be outlined in the PR.

mbutlerw:
Operations such as parsing, preparing, binding and executing a query feature in many SQL protocols and message flows. This commit takes the opportunity to expose these as first class operations on an XTDB node and various query objects.

The first new operation is prepareQuery, which is called on an XTDB node, with an unparsed query and a set of query opts and returns and after analysing and planning the query returns a PreparedQuery object. It is at this point the structure of the database is captured (base tables, the columns on those tables, and their respective types where relevant). The owner of a PreparedQuery is assumed to be the caller, with it being their responsibility to hold reference to and reuse as they see fit.

This PreparedQuery object exposes bind as a method, which when provided a further set of query options (including params/args) will return a BoundQuery object. It is at this point that the input data for a query is captured (values of arguments/params to the query, which rows in the base tables can be seen). The eventual result of executing a BoundQuery should be deterministic. To allow for greater flexibility, if the caller provides params as a query option, it is assumed they are taking responsibility for opening and lifecycle of the param relation. If instead the args option is provided, the BoundQuery takes responsibility for opening and the param relation on the callers behalf and will close said relation when the BoundQuery itself is closed. A BoundQuery object exposes 2 main methods. The former is an optional columnFields method, which returns an ordered list of the columnFields which will be returned by the query. This can be thought of as describing the result set of a given BoundQuery.

The latter is openCursor, which as the name suggests opens a cursor for that query, which when stepped through returns the results of executing the BoundQuery.

As of this commit prepareQuery, and therefore access to the subsequent operations, is only available on the IXtdbInternal interface. These are used by, and in turn re-exposed by, various internal clients to an XTDB node such as pgwire and flightsql. Currently the only public query operation remains openQueryAsync, which returns a stream of results, however this has been reimplemented in terms of these lower level operations.

Commit also renames IRaQuerySource to IQuerySource to better represent it's new functionality as well as removes the caching as part of qs.prepareRaQuery because as mentioned above, reuse of PreparedQueries is the responsibility of the caller. However we may want to reinstate a cache on the users behalf as part of openQueryAsync if there is a sufficient performance benefit. Instead we now opt to cache planning of queries as part of qs.planQuery.

### DIFF
--- a/api/src/main/clojure/xtdb/api.clj
+++ b/api/src/main/clojure/xtdb/api.clj
@@ -162,22 +162,25 @@
    (-> @(q& node q+args opts)
        (rethrowing-cause))))
 
+(defn ->QueryOptions [{:keys [args after-tx basis tx-timeout default-tz default-all-valid-time? explain? key-fn], :or {key-fn :kebab-case-keyword}}]
+  (-> (QueryOptions/queryOpts)
+      (cond-> (instance? Map args) (.args ^Map args)
+              (sequential? args) (.args ^List args)
+              after-tx (.afterTx after-tx)
+              basis (.basis (Basis. (:at-tx basis) (:current-time basis)))
+              default-tz (.defaultTz default-tz)
+              tx-timeout (.txTimeout tx-timeout)
+              (some? default-all-valid-time?) (.defaultAllValidTime default-all-valid-time?)
+              (some? explain?) (.explain explain?))
+
+      (.keyFn (serde/read-key-fn key-fn))
+
+      (.build)))
+
 (extend-protocol xtp/PNode
   IXtdb
-  (open-query& [this query {:keys [args after-tx basis tx-timeout default-tz default-all-valid-time? explain? key-fn], :or {key-fn :kebab-case-keyword}}]
-    (let [query-opts (-> (QueryOptions/queryOpts)
-                         (cond-> (instance? Map args) (.args ^Map args)
-                                 (sequential? args) (.args ^List args)
-                                 after-tx (.afterTx after-tx)
-                                 basis (.basis (Basis. (:at-tx basis) (:current-time basis)))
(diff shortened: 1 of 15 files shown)


api/src/main/clojure/xtdb/api.clj
```diff
@@ -162,22 +162,25 @@
    (-> @(q& node q+args opts)
        (rethrowing-cause))))
 
+(defn ->QueryOptions [{:keys [args after-tx basis tx-timeout default-tz default-all-valid-time? explain? key-fn], :or {key-fn :kebab-case-keyword}}]
```
jarohen, line 165: this is in the `xtdb.api` namespace - can you confirm you want it to become part of the public API?

core/src/main/clojure/xtdb/node/impl.clj
```diff
@@ -102,6 +108,39 @@
     {:latest-completed-tx (.latestCompletedTx indexer)
      :latest-submitted-tx (xtp/latest-submitted-tx this)})
 
+  IXtdbInternal
+  (^CompletableFuture compileQuery [_ ^String query, ^QueryOptions query-opts]
```
jarohen: @mbutlerw: just a feeling from looking at the PR (might be because I'm not 100% in context) - it feels like we've got two layers of this prepare/bind/execute now, but that they don't quite tie up?

maybe it's also because XTQL doesn't go this route - could it?

worth a chat/demo :slightly_smiling_face:

jarohen (approved):
nice one @mbutlerw, thanks mate :pray:

PR #3303, opened 2024-04-11
Implement SELECT without FROM in SQL

### LINKED ISSUE #2225: Support SQL queries without `FROM`
If a SQL statement doesn't contain a `FROM`, you can think of it as selecting from `[:table [{}]]` (i.e. 1 row, no cols).

danmason:
Resolves #2225 

- Updates the grammar to make table_specification optional.
- Update the planner to support `query_specification` **without** a FROM clause:
  - Calls `build-query-specification` with a nil table expression.
  - If `build-query-specification` is passed a  nil table expression, we use `[:table [] [{}]]` when building the `:project`   
- Fix to `analyze` to handle zipper missing a table expression - using `find-first` to find the select_list instead of znth. 
- Add a specific test to check `SELECT` without from.
- Update multiple other tests using unnecessary `FROM (VALUES 1) as x` to select without from.

### DIFF
--- a/core/src/main/clojure/xtdb/sql/analyze.clj
+++ b/core/src/main/clojure/xtdb/sql/analyze.clj
@@ -718,7 +718,7 @@
                      {:ref ag})])
 
                 nil))]
-      (let [sl (r/$ ag -2)
+      (let [sl (r/find-first (partial r/ctor? :select_list) ag)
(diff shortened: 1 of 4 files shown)


jarohen (approved):
sweet, cheers @danmason :pray:

PR #3339, opened 2024-04-26
add gradle wrapper symlinks

l4e21:
Gradle wrapper should probably be present in submodules (fixes jacking in when running clojureRepl on submodules which currently yields an error if the IDE cannot find the gradle wrapper)

I've looked at automatically generating wrappers for subprojects when running `wrapper` but I don't think that's worth it.


jarohen:
@l4e21 which IDE is this, btw?

We tend to run a mega-REPL from the root of the project - if that's not covering what you're looking to do, chances are there's something else up.

Issue #3338, opened 2024-04-25
Memory leak in query with subquery


FiV0:
Running into some memory leak when querying data from the bench products dataset
```clj
  (in-ns 'xtdb.bench.products)

  (defn local-node
    ([]
     (local-node "dev/products/products"))
    ([path-prefix]
     (xtn/start-node {:log [:local {:path (str path-prefix "-log")}]
                      :storage [:local {:path (str path-prefix "-storage")}]})))
  
  (with-open [node (local-node)
              is (-> products-file io/input-stream (GZIPInputStream.))]
    (store-documents! node (transit-seq (transit/reader is :msgpack))))

  (with-open [node (local-node)]
    (Thread/sleep 1000)
    (xt/q node '(-> (from :products [{:xt/id id} {:gtin "06420618721502"}])
                    (with {:more-data (q (from :products [{:xt/id $id} gs1-data])
                                         {:args [id]})}))))
```

FiV0:
Narrowed it down to this RA plan
```clj
[:apply
 :single-join
 {xt$id ?id}
 [:table [{:xt$id "62e2e8cd715db4a7db2cfd5e"}]] 
 [:scan {:table products}
  [{xt$id (= xt$id ?id)} gs1_data]]]
```

FiV0:
Cross chunk boundary issue
```clj
(t/deftest test-chunk-boundary-different-struct-types
  (with-open [node (xtn/start-node {:indexer {:rows-per-chunk 20}})]
    (xt/submit-tx node (for [i (range 20)]
                         [:put-docs :xt_docs {:xt/id i :foo {:bar 42}}]))

    (xt/submit-tx node (for [i (range 20 40)]
                         [:put-docs :xt_docs {:xt/id i :foo {:bar "42"}}]))

    (t/is (= []
             (tu/query-ra
              '[:apply :cross-join {}
                [:table [{}]]
                [:scan {:table xt_docs} [foo]]]
              {:node node})))))
```

PR #3343, opened 2024-04-29
Enable `getLeg` on base ValueVectorReader

FiV0:
So the problem is that IMVR might have monomorphic struct vectors in the underlying vectors. For example `[:struct {"bar" :i64}]` and `[:stuct {"bar" :utf8}]`. From the point of view of IMVR the "bar" attribute type is a union. The expression engine than might call `getLeg` on the "bar" `IValueReader`'s, but at that point the value readers are primitive `IValueReader`'s.

Options were:
1. Just return the primitive `getLeg` name in every base vector. Implemented here.
2. Do something more complicated in the `IValueReader` of IMVR. The problem is that an `getObject` call needs to be modified quite heavily in the case of struct readers. So I opted for 1.

Happy to consider other options, as `getLeg` calls should "normally" only be called on unions.

### DIFF
--- a/core/src/main/java/xtdb/vector/ValueVectorReader.java
+++ b/core/src/main/java/xtdb/vector/ValueVectorReader.java
@@ -9,6 +9,7 @@ import org.apache.arrow.vector.holders.NullableIntervalDayHolder;
 import org.apache.arrow.vector.holders.NullableIntervalMonthDayNanoHolder;
 import org.apache.arrow.vector.types.pojo.ArrowType;
 import org.apache.arrow.vector.types.pojo.Field;
+import xtdb.Types;
 import xtdb.api.query.IKeyFn;
 import xtdb.types.IntervalDayTime;
 import xtdb.types.IntervalMonthDayNano;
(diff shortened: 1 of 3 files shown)


jarohen (commented):
thanks @FiV0 - couple of points to consider:

* It used to be the case that the presence of non-null `legs` made the metadata treat vectors as unions - I'm assuming this is no longer the case, given this works?
* Good that you've fixed this issue, obv, but it sounds like the memory leak would still be there if something error'd in a similar place - could we use this opportunity to also tighten up the cleanup now that we know where the error's coming from?

FiV0:
> It used to be the case that the presence of non-null legs made the metadata treat vectors as unions - I'm assuming this is no longer the case, given this works?

Ok, have to check that explicitly.
 
> Good that you've fixed this issue, obv, but it sounds like the memory leak would still be there if something error'd in a similar place - could we use this opportunity to also tighten up the cleanup now that we know where the error's coming from?

Yes, good point. I think this error got likely swallowed somewhere. I'll try to find the place for that.

FiV0:
This actually doesn't seem to fix the memory leak, so might actually just be something that showed up while investigating #3338 or there are two issues.

jarohen (approved):

PR #3346, opened 2024-05-02
Add UI inputs to docs Fiddle

Akeboshiwind:
See the [fiddle-demo](https://pr-3346.d2zcybuz6k9g4m.amplifyapp.com/fiddle-demo#live-queries) page for more on how it works


refset:
The built version (via the preview URL you shared) doesn't work right on Chromium 124.0 - producing several JS errors:
image

Meaning components aren't initialized properly as a result:
image

Akeboshiwind:
Thanks! I *think* this will fix all those bugs.
Looks like I'll need to do more tests with the production build 😅

refset:
Cool, all seems to work okay now :+1: 

However there are still some errors showing (just in case these weren't visible to you already)

image

Akeboshiwind:
Yup, thankfully both of those are expected 👍 
The first one is for showing off the errors for the Fiddle component, and the second one is for showing off errors for the LiveQuery component

PR #3349, opened 2024-05-02
No longer using `compile-sql` to encode-sql-args lists.

danmason:
Similar to a137d41

### DIFF
--- a/core/src/main/clojure/xtdb/log.clj
+++ b/core/src/main/clojure/xtdb/log.clj
@@ -196,25 +196,26 @@
       (.writeObject xtql-op-writer (ClojureForm. op))
       (.endStruct xtql-writer))))
 
-(defn encode-sql-args [^BufferAllocator allocator, query, arg-rows]
-  (let [plan (sql/compile-query query)
-        {:keys [^long param-count]} (meta plan)
+(defn encode-sql-args [^BufferAllocator allocator, arg-rows]
+  (if (apply not= (map count arg-rows))
+    (throw (err/illegal-arg :sql/arg-rows-different-lengths
+                            {::err/message "All SQL arg-rows must have the same number of columns"
+                             :arg-rows arg-rows}))
+    (let [param-count (count (first arg-rows))
+          vecs (ArrayList. param-count)]
+      (try
+        (dotimes [col-idx param-count]
+          (.add vecs
+                (vw/open-vec allocator (symbol (str "?_" col-idx))
+                             (mapv #(nth % col-idx nil) arg-rows))))
 
-        vecs (ArrayList. param-count)]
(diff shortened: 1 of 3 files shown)


core/src/main/clojure/xtdb/log.clj
```diff
@@ -196,17 +196,14 @@
...
-(defn encode-sql-args [^BufferAllocator allocator, query, arg-rows]
-  (let [plan (sql/compile-query query)
-        {:keys [^long param-count]} (meta plan)
-
-        vecs (ArrayList. param-count)]
+(defn encode-sql-args [^BufferAllocator allocator, arg-rows]
+  (let [max-param-count (apply max (map count arg-rows))
+        vecs (ArrayList. max-param-count)]
     (try
-      ;; TODO check arg count in each row, handle error
```
danmason, line 205: If indeed we do see have a mismatch here expecting to throw an error at a later point when compiling the query.

jarohen: @danmason could you pop a test in to confirm this?

core/src/main/clojure/xtdb/log.clj
```diff
@@ -196,17 +196,14 @@
       (.writeObject xtql-op-writer (ClojureForm. op))
       (.endStruct xtql-writer))))
 
-(defn encode-sql-args [^BufferAllocator allocator, query, arg-rows]
-  (let [plan (sql/compile-query query)
-        {:keys [^long param-count]} (meta plan)
-
-        vecs (ArrayList. param-count)]
+(defn encode-sql-args [^BufferAllocator allocator, arg-rows]
+  (let [max-param-count (apply max (map count arg-rows))
```
jarohen: `apply max` will fail here if `arg-rows` is empty - I assume that's handled elsewhere?

core/src/main/clojure/xtdb/log.clj
```diff
@@ -226,7 +223,7 @@
         (.writeObject query-writer sql)
 
         (when-let [arg-rows (.argRows op)]
```
jarohen: `not-empty` call here around `argRows` would alleviate the `apply max` concern above

jarohen (approved):
couple of things to address but otherwise good to go, cheers @danmason :pray:

core/src/main/java/xtdb/vector/RelationReader.java
```diff
@@ -33,6 +33,10 @@ public int rowCount() {
         return rowCount;
     }
 
+    public int argCount() {
```
jarohen: I wouldn't add this here - it's not something that makes sense for all RelationReaders

PR #3356, opened 2024-05-07
Provide Python DB API (PEP 249) class

l4e21:


FiV0:
@l4e21 Can we maybe create a custom error client side and wrap with that the value of `obj['@value']`? Similar to what we have in Clojure, a RuntimeException and an IllegalArgumentException. Something along the lines of
```py
class MyCustomError(RuntimeError):
    def __init__(self, message):
        # Call the base class constructor with the error message
        super().__init__(message)
```
If there is a RuntimeError (haven't checked) in Python use that, etc..

FiV0:
An example is here: https://github.com/xtdb/xtdb/blob/5c45f16d6c0ccdc30b285813a2ee9eade6e07915/http-server/src/test/clojure/xtdb/remote_test.clj#L180-L211

lang/python/xtdb/test/test_encoding.py
```diff
@@ -107,3 +107,29 @@ def test_timestamps(client):
     client.submit_tx([SqlTx("INSERT INTO trades (xt$id, price) VALUES (1, 100)")])
     query_result = client.query(Sql("SELECT trades.xt$id, trades.price, trades.xt$system_from, trades.xt$system_to FROM trades FOR ALL SYSTEM_TIME"))
     assert isinstance(query_result[0]["xt$system_from"], datetime)
+
+def test_basis_change(client):
+    client.set_basis("2020-01-01T12:34:56Z")
```
refset: basis only relates to querying, these should be `set_tx_time` instead - IIUC? i.e. I don't think they will be having any affect

l4e21: The basis for the client changes both the basis for querying and the system time for the transaction. The tests pass so it definitely works. But it's true that it can be made clearer and separated out. The reason I did it like this is because in my mind if I changed something like the 'basis' expecting both to change and only one changed it can maybe cause confusion to the user

PR #3363, opened 2024-05-13
Fixes for the nightly test failures #3362

### LINKED ISSUE #3362: Nightly CI failing
Nightly tests have been failing for a while & have not been ran against main for quite a while since we moved `2.x` over to `main`. As such, need to go through the list of failures and make any appropriate fixes to get the tests passing again. See recent test failures [**here**](https://github.com/xtdb/xtdb/actions/runs/9061295271/job/24892775420)

danmason:
Resolves #3362

**Successful 'Nightly on Demand' Run**: https://github.com/xtdb/xtdb/actions/runs/9066674997

Fixes numerous issues with the remote object stores to ensure they work properly and that we can pass the nightly tests:
- A lot of issues were fixed by version bumps for S3 and Google Cloud.
- Some test updates for kafka test and numerous test updates for the remote object stores.
- Most of the remaining remote object store failures were caused by behaviour in RemoteBufferPool that would cause errors on `finish-chunk!`:
  - Essentially - within RemoteBufferPool we were resolving numerous inputted paths against "storage-root"
    - Storage-root essentially being used to prefix files with a version number `v02` that we can update when storage format changes in a backwards-incompatible way.
  - Because we were handling this within RemoteBufferPool, internally object store filenames would include this prefix, and in places we were passing back the filenames with the `v02` prefix - this lead to failures/errors within compacting/chunk finishing as we would
    - Call to list files.
    - Get back filenames WITH their `v02` version prefix.
    - Pass said listed filenames back to caller, who would then attempt to call `GET` on the filenames.
    - RemoteBufferPool would then attempt to prefix them _again_, so would be attempting to get files like `v02/v02/tables/...`, which would not be present/found and would throw an error.
  - Elsewhere - ie, in the LocalBufferPool - we just treat the storage version prefix as part of the full directory name - none of the filenames will ever particularly see/care about it.
  - As such - made a change here to follow a similar pattern in RemoteBufferPool/Remote Object Stores:
    - RemoteBufferPool will just pass object keys/filenames that it gets - it doesn't do any resolving or the like on the Paths.
    - When creating the remote object store, we add the `v02` prefix to the existing object store `prefix` value set on configuration, so it becomes part of the folder structure without being passed around in filenames in the system.
    - This fixed the majority of the `finish-chunk` calls that were previously throwing errors.
- Also fixed an issue within the S3 file watcher where it would see slightly different filenames passed back from the SQS subscriptions:
  - Essentially, URL encoding certain values - ie, `table/xt$txs` became `table/xt%24txs`.
  - Now ensure that we decode the strings received from SQS properly.

### DIFF
--- a/modules/s3/src/main/clojure/xtdb/s3/file_list.clj
+++ b/modules/s3/src/main/clojure/xtdb/s3/file_list.clj
@@ -4,6 +4,7 @@
             [clojure.tools.logging :as log] 
             [xtdb.util :as util])
   (:import [java.lang AutoCloseable]
+           [java.net URLDecoder]
            [java.nio.file Path]
            [java.util UUID NavigableSet]
            [software.amazon.awssdk.core.exception AbortedException]
@@ -121,7 +122,10 @@
                                                 event-type (cond
                                                              (string/starts-with? event-name "ObjectCreated") :create
                                                              (string/starts-with? event-name "ObjectRemoved") :delete)
-                                                filename (util/->path (get-in s3-event [:s3 :object :key]))
+                                                filename (-> s3-event
+                                                             (get-in [:s3 :object :key])
+                                                             (URLDecoder/decode "utf8")
+                                                             (util/->path))
                                                 ;; If prefix present - need to filter for objects that start with prefix
                                                 ;; And then relativize the path - otherwise, can leave it unaltered and unfiltered
                                                 file (if prefix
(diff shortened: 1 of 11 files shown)


modules/s3/src/main/clojure/xtdb/s3/file_list.clj
```diff
@@ -86,6 +86,9 @@
       (:Records)
       (first)))
 
+(defn url-decode-filename [filename]
+  (string/replace filename "%24" "$"))
+
```
FiV0: Can me make sure this is the only problem. I think there could be other funky table names etc..

mbutlerw (approved):

modules/s3/src/main/clojure/xtdb/s3/file_list.clj
danmason: Used `URLDecoder` to cover our bases here.

PR #3365, opened 2024-05-15
`maybePromote` for creating row copiers

FiV0:
This adds `maybePromote` to the `IVectorWriter` interface. `maybePromote` returns a (potentially) new `IVectorWriter` that can accommodate the field passed in. The intended usage is 
```clj
(.maybePromote column-wtr (.getField column-rdr))
(let [copier (.rowCopier column-wtr column-rdr)]
  (.copyRow copier 0))
```
Issues that should probably have a clear answer:
 - One can call `maybePromote` on a union leg writer. Normally one should not need to call this as a user on a legWriter that changes the promotes the leg itself, but there is nothing preventing it currently and we use it for promoting potential children of a leg.
 - In a couple of places we are now calling `maybePromote` and assume that the existing writer only gets mangled in place (for example the `put` leg is a struct vector to which a new child might be added, but the struct vector itself doesn't get promoted). 

TODO:
- [ ] performance check


core/src/main/clojure/xtdb/compactor.clj
```diff
@@ -40,7 +40,10 @@
             sf-copier (-> (.readerForName data-rdr "xt$system_from") (.rowCopier sf-wtr))
             vf-copier (-> (.readerForName data-rdr "xt$valid_from") (.rowCopier vf-wtr))
             vt-copier (-> (.readerForName data-rdr "xt$valid_to") (.rowCopier vt-wtr))
-            op-copier (-> (.readerForName data-rdr "op") (.rowCopier op-wtr))]
+            op-rdr (.readerForName data-rdr "op")
+            ;; makes the assumption that the op writer only gets mangled in place
```
jarohen, line 44: try without this - we should already have the full field specified ahead-of-time (because we write the schema out earlier than this)

FiV0:
@jarohen In case you looking at this, please let me merge as I might still want to do some tweaks.

FiV0:
Closing in Favor of #3374.

PR #3370, opened 2024-05-21
Fix escape sequence `SyntaxWarning`

LtTempletonPeck:
Replace invalid escape sequences in regular expression string literals which was [upgraded](https://docs.python.org/3/whatsnew/3.12.html#other-language-changes) from `DeprecationWarning `to a `SyntaxWarning` in  Python 3.12

When starting the script using Python +3.12 the following output is displayed:
```shell
/Users/LtTempletonPeck/xtsql.py:463: SyntaxWarning: invalid escape sequence '\s'
  self.accept = re.sub('=\s*', '', arg)
/Users/LtTempletonPeck/xtsql.py:469: SyntaxWarning: invalid escape sequence '\s'
  self.url = re.sub('=\s*', '', arg)
```

Alternatively raw string literals could be used:
```python
            self.accept = re.sub(r'=\s*', '', arg)
```

### DIFF
--- a/docs/public/xtsql.py
+++ b/docs/public/xtsql.py
@@ -460,13 +460,13 @@ class XtdbConsole(cmd.Cmd):
     def do_accept(self, arg):
         'Sets or shows the accepted mime type.'
         if arg:
-            self.accept = re.sub('=\s*', '', arg)
+            self.accept = re.sub('=\\s*', '', arg)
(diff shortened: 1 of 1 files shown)


refset:
Thanks for spotting & fixing :pray:  also TIL about Python raw string literals!

PR #3371, opened 2024-05-21
docs: fix gradle xtdb-core dependency

LtTempletonPeck:
Fix `xtdb-core` dependency in Gradle `build.gradle.kts` example.

The correct artifactid is in the Maven `pom.xml` [example](https://github.com/xtdb/xtdb/blob/d1d6d457fcb17205707a03ac61934060f0bbafa5/docs/src/content/docs/drivers/java/getting-started.adoc?plain=1#L49)


jarohen:
Thanks @LtTempletonPeck! :pray:

PR #3374, opened 2024-05-22
promoteChildren + fix #3338

### LINKED ISSUE #3338: Memory leak in query with subquery

FiV0:
This adds a method `promoteChildren(field : Field)` to the `IVectorWriter` interface. The composite writers take care of dealing with internal promotion themselves when writing values which they can't yet hold.

 The new `promoteChildren` method should be used when creating multiple row copiers on the same writer. In this case the writer needs to know upfront the potential child changes for every reader it will receive data from.
```clj
(let [struct-int-field (types/->field "foo" #xt.arrow/type :struct false (types/col-type->field "bar" :i64))
      struct-str-field (types/->field "foo" #xt.arrow/type :struct false (types/col-type->field "bar" :utf8))]
  (with-open [struct-int-vec (.createVector struct-int-field tu/*allocator*)
              struct-str-vec (.createVector struct-str-field tu/*allocator*)]

    ;; write something to the two vectors

    (with-open [wtr (vw/->vec-writer tu/*allocator* "my-new-struct" (FieldType/notNullable #xt.arrow/type :struct))]
      (.promoteChildren wtr struct-int-field)
      (.promoteChildren wtr struct-str-field)
      (let [int-copier (.rowCopier wtr struct-int-vec)
            ;; creating this copier previously invalidated the `int-copier` just above
            str-copier (.rowCopier wtr struct-str-vec)]
        (.copyRow int-copier 0)
        (t/is (= [{:bar 42}]
                 (tu/vec->vals (vw/vec-wtr->rdr wtr))))
        (.copyRow str-copier 0)
        (t/is (= [{:bar 42} {:bar "forty-two"}]
                 (tu/vec->vals (vw/vec-wtr->rdr wtr))))))))
```

### DIFF
--- a/src/test/clojure/xtdb/operator/scan_test.clj
+++ b/src/test/clojure/xtdb/operator/scan_test.clj
@@ -72,6 +72,22 @@
                      [:scan {:table xt_docs} [foo]]]
                    {:node node}))))))
 
+(t/deftest test-row-copying-different-struct-types-between-chunk-boundaries-3338
+  (with-open [node (xtn/start-node {:indexer {:rows-per-chunk 20}})]
+    (xt/submit-tx node (for [i (range 20)]
+                         [:put-docs :xt_docs {:xt/id i :foo {:bar 42}}]))
+
+    (xt/submit-tx node (for [i (range 20 40)]
+                         [:put-docs :xt_docs {:xt/id i :foo {:bar "forty-two"}}]))
+
+    (t/is (= #{{:foo {:bar 42}} {:foo {:bar "forty-two"}}}
+             (set (tu/query-ra
+                   ;; the cross-join copies data from the underlying IndirectMultiVectorReader
+                   '[:apply :cross-join {}
+                     [:table [{}]]
+                     [:scan {:table xt_docs} [foo]]]
+                   {:node node}))))))
+
 (t/deftest test-smaller-page-limit
   (with-open [node (xtn/start-node {:indexer {:page-limit 16}})]
     (xt/submit-tx node (for [i (range 20)] [:put-docs :xt_docs {:xt/id i}]))
--- a/core/src/main/kotlin/xtdb/vector/DenseUnionVectorWriter.kt
+++ b/core/src/main/kotlin/xtdb/vector/DenseUnionVectorWriter.kt
@@ -4,6 +4,7 @@ import clojure.lang.Keyword
 import org.apache.arrow.vector.ValueVector
 import org.apache.arrow.vector.complex.DenseUnionVector
 import org.apache.arrow.vector.complex.replaceChild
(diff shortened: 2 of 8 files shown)


src/test/clojure/xtdb/operator/scan_test.clj
```diff
@@ -72,6 +72,21 @@
...
+  (with-open [node (xtn/start-node {:indexer {:rows-per-chunk 20}})]
+    (xt/submit-tx node (for [i (range 20)]
+                         [:put-docs :xt_docs {:xt/id i :foo {:bar 42}}]))
+
+    (xt/submit-tx node (for [i (range 20 40)]
+                         [:put-docs :xt_docs {:xt/id i :foo {:bar "forty-two"}}]))
+
+    (t/is (= #{{:foo {:bar 42}} {:foo {:bar "forty-two"}}}
+             (set (tu/query-ra
+                   '[:apply :cross-join {}
```
jarohen, line 86: I'm missing the nuance of this being apply cross-join rather than a straight scan?

src/test/clojure/xtdb/operator/scan_test.clj
```diff
@@ -72,6 +72,21 @@
                      [:scan {:table xt_docs} [foo]]]
                    {:node node}))))))
 
+(t/deftest test-row-copying-different-struct-types-bewteen-chunk-boundaries-3338
```
jarohen: nit: seplling

jarohen (approved):
nice one @FiV0, significantly simpler - this PR definitely doesn't tell the story of the journey but it's in a much better place now as a result :sweat_smile:

src/test/clojure/xtdb/operator/scan_test.clj
FiV0: A `scan` doesn't copy. The cross join is using the `MVR` row copier. If you have a simpler example let me know. I can also add a comment if you think that would be helpful.

PR #3392, opened 2024-06-04
Use NotifyingSubscriberHandler for local-directory-log

### LINKED ISSUE #3382: local-directory-log using notifying subscriber handler
Currently we're polling, but the sleep is showing up in the AuctionMark profile - we should be able to use the notifying subscriber handler here in a similar vein to the in-memory log.

FiV0:
This uses the notifying subscriber handler for the local directory log. 

As the local log might need to reread previously written transactions (called records in the log), we get the `latest-submitted-tx` from the log file (this is just an offset). We also do some simple checks (record-seperator in place, offset + record size line up with total log size) to assure we are not reading from a corrupted log. 

Closes #3382.

This doesn't deal with the azure log for which the issue of #3382 also exists.

### DIFF
--- a/core/src/main/clojure/xtdb/log.clj
+++ b/core/src/main/clojure/xtdb/log.clj
@@ -5,7 +5,6 @@
             [xtdb.error :as err]
             [xtdb.node :as xtn]
             xtdb.protocols
-            [xtdb.sql :as sql]
             [xtdb.trie :as trie]
             [xtdb.types :as types]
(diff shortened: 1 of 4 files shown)


FiV0:
@jarohen Spinning without sleeping still seems to give more throughput in case of Auctionmark (Left this PR, right spinning, middle old).
Selection_031

jarohen (approved):
cheers @FiV0! :pray:

Issue #3396, opened 2024-06-07
AuctionMark running in AWS


jarohen:
* [x] AM running in a Docker container
  * SF 0.1 
* [x] CloudFormation templates for the infrastructure:
  * MSK
  * ECS on EC2, using i3.* machine type for its instance storage.
    * 3 concurrent worker tasks
    * Local disk cache being stored on the underlying machine's instance storage
  * S3/SNS object-store

PR #3404, opened 2024-06-10
Add `cloud-benchmark` and files for basic setup for running Auctionmark on AWS 

danmason:
See #3396

Add `cloud-benchmark` project:
- Has a main file/main project that takes some environment var configured options & node config set inside of a config yaml file and then runs auctionmark with said options against the configured node.
  - Node config expected under `cloud-config.yaml` - we can add this within Docker files depending on which cloud needs which node config.
  - Expects various Environment vars for some of the key auctionmark configs, ie:
    - `AUCTIONMARK_DURATION`
    - `AUCTIONMARK_SCALE_FACTOR`
    - `AUCTIONMARK_LOAD_PHASE`    
  - Can set `SLACK_WEBHOOK_URL` to post messages to a webhook - have set this up under #xtdb-run-stability-notifications for now

Adds the `cloud-benchmark:aws` subproject:
- Includes the above `cloud-benchmark` project and AWS related dependencies, and has a task to build a Shadow JAR
- Docker, cloudformation, scripts and README included for setting up an AWS infra, creating/pushing an AWS benchmark docker image, and running a cluster of ECS tasks.
- Reuses/references some of the old work/templates for running an S3/MSK Kafka backed node on AWS.

Also adds the `cloud-benchmark:local` subproject:
- Not necessarily the quickest way of running local auctionmark, but can build/tag/push the 'local' dockerfile to cloud if we want to easily/quickly run without cloud infra, and also can setup various constraints and the like fairly easily using the docker file.
- Has a few commented parts in the Dockerfile for running/exposing ports for Yourkit Profiling.

### To note:
- Number of errors being thrown when using AWS infra backed node.
- We do not currently surface errors thrown in auctionmark threads, so if an error is thrown inside of the `concurrently` block we will not see it in the slack message.


jarohen (approved):
sweet, nice one @danmason - appreciate the thoroughness of the docs too :pray:

modules/bench/src/main/clojure/xtdb/bench/auctionmark.clj
```diff
@@ -666,4 +665,4 @@
                              :duration duration
                              :freq (Duration/ofMillis (* 0.2 (.toMillis duration)))
                              :job-task {:t :call, :transaction :index-item-status-groups, :f (b/wrap-in-catch index-item-status-groups)}}]}
-            (when sync {:t :call, :f #(tu/then-await-tx (:sut %))})])}))
+            (when sync {:t :call, :f #(catchup (:sut %))})])}))
```
danmason: Primarily updated to remove the dependency on `test-util` (caused some issues when building the shadow JAR, since test-util doesn't live under any of the modules, as far as I can tell). File was recently updated to add/use `catchup` - though there IS a difference here:

```clojure
(defn then-await-tx
  (^TransactionKey [node]
   (then-await-tx (latest-submitted-tx node) node nil))

  (^TransactionKey [tx node]
   (then-await-tx tx node nil))

  (^TransactionKey [tx node timeout]
   (idx/await-tx tx node timeout)))
```

danmason: Rather than catchup:

```clojure
(defn catchup [node]
  (let [{:keys [latest-completed-tx]} (xt/status node)]
    (loop [last-tx latest-completed-tx]
      (Thread/sleep 100)
      (let [{:keys [latest-completed-tx]} (xt/status node)]
        (when-not (= last-tx latest-completed-tx)
          (recur latest-completed-tx))))))
```

Might explain our query 'memory leaks' - might not be letting things finish properly.

PR #3414, opened 2024-06-13
Return postgres types over pgwire for monomorphic types, update pg_catalog

### LINKED ISSUE #3232: Pgwire to return primitive types for monomorphic columns

mbutlerw:
PR creates an explicit type mapping between arrow/xtdb
types and postgres types, including how to read/write those
types to and from pgwire.

This enables us to accept (as params) and return (from queries)
postgres types via pgwire, both in binary and text format.

PR also updates pg_catalog to include more information, including postgres types.

### DIFF
--- a/core/src/main/clojure/xtdb/types.clj
+++ b/core/src/main/clojure/xtdb/types.clj
@@ -6,9 +6,14 @@
             [xtdb.util :as util])
   (:import (clojure.lang MapEntry)
            [java.io Writer]
+           [java.nio.charset StandardCharsets]
+           [java.nio ByteBuffer]
+           [java.util UUID]
            (org.apache.arrow.vector.types DateUnit FloatingPointPrecision IntervalUnit TimeUnit Types$MinorType UnionMode)
            (org.apache.arrow.vector.types.pojo ArrowType ArrowType$Binary ArrowType$Bool ArrowType$Date ArrowType$Decimal ArrowType$Duration ArrowType$FixedSizeBinary ArrowType$FixedSizeList ArrowType$FloatingPoint ArrowType$Int ArrowType$Interval ArrowType$List ArrowType$Map ArrowType$Null ArrowType$Struct ArrowType$Time ArrowType$Time ArrowType$Timestamp ArrowType$Union ArrowType$Utf8 Field FieldType)
+           xtdb.api.query.IKeyFn
            xtdb.Types
+           [xtdb.vector IVectorReader]
            (xtdb.vector.extensions TransitType KeywordType SetType UriType UuidType)))
 
 (set! *unchecked-math* :warn-on-boxed)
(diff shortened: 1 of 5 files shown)


src/test/clojure/xtdb/pgwire_test.clj
```diff
@@ -118,7 +117,7 @@
     "require" :unsupported))
 
 (defn- jdbc-conn ^Connection [& params]
-  (jdbc/get-connection (apply jdbc-url params)))
+  (jdbc/get-connection (doto (apply jdbc-url params) println)))
```
jarohen: cheeky println 🙊

src/test/clojure/xtdb/information_schema_test.clj
```diff
@@ -115,6 +197,134 @@
                                [schemaname tablename tableowner tablespace]]
                              {:node tu/*node*})))))
 
+(deftest test-pg-class
```
jarohen, line 199: here comes a git conflict with #3412

pgwire-server/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1455,19 +1410,19 @@
...
 
-    (when (= :prepared-statements describe-type)
-      (throw (UnsupportedOperationException.
-              ;;TODO consider if we can send some kind of describe statement for prepared statments
-              ;;Not possible to return accurate param and return types but maybe we can return something
-              ;;generic enough to not upset clients
-              "XTDB does not support describing prepared statements,
-               please bind the statement and describe the portal")))
+   (if (= :prepared-stmt describe-type)
+     nil #_(throw (UnsupportedOperationException.
```
jarohen: prefer not to leave these in - if we don't need it, can it be removed?

pgwire-server/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -433,10 +383,9 @@
     :cols [{:column-name "transaction_isolation" :column-oid oid-varchar}]
     :rows [["read committed"]]}
 
-   ;; ODBC issues this query by default, you may be able to disable with an option
-   {:q "select oid, typbasetype from pg_type where typname = 'lo'"
-    :cols [{:column-name "oid", :column-oid oid-varchar} {:column-name "typebasetype", :column-oid oid-varchar}]
-    :rows []}
+   {:q "show timezone"
```
jarohen, line 322: this can probably go into the grammar now, and be supported as a proper query rather than a hard coded string

core/src/main/clojure/xtdb/types.clj
```diff
@@ -704,3 +708,167 @@
...
        (into {} (map (juxt key (comp #(merge-fields % null-field) val))))))
+
+(defn- read-utf8 [^bytes barr] (String. barr StandardCharsets/UTF_8))
+
+(defn utf8
+  "Returns the utf8 byte-array for the given string"
+  ^bytes [s]
+  (.getBytes (str s) StandardCharsets/UTF_8))
+
+(def pg-types
```
jarohen, line 720: worth creating a xtdb.pgwire.types ns for these? 

if nothing else, core doesn't (yet) depends on pgwire, so this creates a (conceptual, not code) cycle

jarohen: although info-schema potentially does already... let's leave these for now, reckon we'll merge pgwire into core soon enough

jarohen (changes requested):
cheers @mbutlerw - happy to land this ahead of #3412, and we'll deal with the conflict tomorrow?

pgwire-server/src/main/clojure/xtdb/pgwire.clj
mbutlerw: Yes, although didn't want to burden this PR with said changes. Will spin out these plus other tooling/postgres dialect related features into their own cards.

PR #3439, opened 2024-06-24
Trying to remove 'appendices' in docs

jarohen:
The existence of 'appendices' is arguably a bit of a smell of documentation that doesn't really belong anywhere.

* Moving community/roadmap out of 'appendices' into the intro section
* Inline 'mission' into the otherwise-anaemic welcome page

RFC: 
1. the two pages still in appendices - these seem to be covered elsewhere? i.e. in the various tutorials/getting started. Shall we remove them?
2. there seems to be quite a lot of overlap between 'getting started' and the 'immutability walkthrough' - do we need both?


jarohen:
stale

PR #3466, opened 2024-07-04
Size limiting local-disk-cache using Caffeine based evictor

### LINKED ISSUE #3424: Prevent local disk caches from exceeding configured limits
To avoid hitting any "No space left on device" errors (or causing other processes in the system to hit such errors), XTDB's local disk caches should support configurable limits (e.g. fixed absolute size, or % of total volume size)

danmason:
Associated issue: #3424
Github Action Runs: [Here](https://github.com/danmason/xtdb/tree/limit-disk-cache-size)

## Description

Allows the setting of a "max size in bytes" for the local disk cache.
- You can configure a "max size" in bytes for the local disk cache to have.
  - Defaulting to 10GB - really just a random number - I have been testing with 1GB, 5GB, etc for example.
  - It's not entirely a hard limit but we aim to keep under it generally.
- Utilizes a caffeine cache to save buffer paths and model their relative weightings/sizes - also where we set our limit:
  - The concept essentially being that we can use Caffeine and it's semantics to evict files once it's beyond a certain max size.
  - Uses custom weightings to achieve this - essentially weighing function run whenever a file is inserted/updated - and is based on it's file size on disk.
  - We prevent files that are currently being used from being deleted by using caffeine's [pinning mechanism](https://github.com/ben-manes/caffeine/wiki/Faq#pinning-entries) - in the case of a 'maxweight' based eviction policy:
    - If we want a file to be preserved we can return a `weight` of `0`.
    - I've accomplished this via the value stored in Caffeine, which has a `pinned?` key.
  - Load up any files previously on the disk under the `local-disk-cache` path when starting the cache.
 - Pinning behaviour:
   - Files are pinned when they are first fetched to be returned (ie, from `getBuffer`), pinning just prior to them being memory mapped.
   - Files are unpinned when the ArrowBuf that memory maps them is released from memory - passing a custom `release-fn` into the allocation-manager they are using.     
   - At the point of files being pinned, we adjust/lower the max weight of the cache to account for them towards the overall cache weight.
   - When they are unpinned, we re-add their size on disk back to the max weight.  
- Added configuration of the above into the relevant docs.
- Using AsyncCache in Caffeine to achieve all of this.

## Thoughts
- The `release-fn` inside of `release0` feels a bit hacky - perhaps not in the right place, but seems to work - our intention here is very much to remove files as and when we consider them "unused" by anything.

## Observations
- When ran with a very low disk cache size limit (ie, 1GB), we can very well go over that depending on how much is currently being held in memory. Running at _slightly_ higher limits though, (ie, 5GB on Auctionmark), we seem to keep under the max fairly well, and actively remove files. Expecting at more proper size limits (ie, 100GB), should keep at that limit fairly well. 
- Most of our previous observed issues here (ie, with heap size OoMs, with file not found), seem to be fixed (OoM's still caused with lower amounts of memory), seemed to run okay on a 1hour long run (even with a pretty small disk cache limit of 5GB)

## Update

In summary:
- I have fixed the issues that seem to cause race conditions.
- It seems to no longer hang on google cloud after moving over to use the async caffeine cache.
- I've fixed a bug causing it to not respect the size limit.

Ran it against google cloud auctionmark for an hour, with a limited size disk cache of 5GB, and it seemed to respect that number just fine. At this point, opening it up as ready for review.

### DIFF
--- a/core/src/main/clojure/xtdb/buffer_pool.clj
+++ b/core/src/main/clojure/xtdb/buffer_pool.clj
@@ -1,15 +1,16 @@
 (ns xtdb.buffer-pool
-  (:require [juxt.clojars-mirrors.integrant.core :as ig]
+  (:require [clojure.tools.logging :as log]
+            [juxt.clojars-mirrors.integrant.core :as ig]
             [xtdb.node :as xtn]
             [xtdb.object-store :as os]
             [xtdb.util :as util])
   (:import (clojure.lang PersistentQueue)
-           (com.github.benmanes.caffeine.cache Cache Caffeine RemovalListener Weigher)
-           [java.io ByteArrayOutputStream Closeable]
+           (com.github.benmanes.caffeine.cache AsyncCache Cache Caffeine RemovalCause RemovalListener Weigher Policy$Eviction)
+           [java.io ByteArrayOutputStream Closeable File]
            (java.nio ByteBuffer)
            (java.nio.channels Channels ClosedByInterruptException)
            [java.nio.file FileVisitOption Files LinkOption Path]
-           [java.nio.file.attribute FileAttribute]
+           [java.nio.file.attribute FileAttribute BasicFileAttributes]
            [java.util NavigableMap TreeMap]
            (java.util NavigableMap)
            [java.util.concurrent CompletableFuture ForkJoinPool TimeUnit]
@@ -329,10 +330,21 @@
         (upload-multipart-buffers object-store k (arrow-buf->parts arrow-buf))
         nil))))
 
+(defn update-evictor-key [^AsyncCache local-disk-cache-evictor ^Path k update-fn]
+  (.. local-disk-cache-evictor asMap (compute k (reify BiFunction
+                                                  (apply [_ _k fut] (update-fn fut)))))) 
+
+(defn set-cache-max-weight [^AsyncCache cache ^long new-size] 
+  (let [sync-cache (.synchronous cache)
+        ^Policy$Eviction eviction-policy (.get (.eviction (.policy sync-cache)))]
+    (.setMaximum eviction-policy (max 0 new-size))))
+
 (defrecord RemoteBufferPool [allocator
                              ^Cache memory-store
                              ^Path local-disk-cache
-                             ^ObjectStore object-store]
+                             ^AsyncCache local-disk-cache-evictor
+                             ^ObjectStore object-store
+                             !evictor-max-weight]
   IBufferPool
   (getBuffer [_ k]
     (let [cached-buffer (cache-get memory-store k)]
@@ -345,16 +357,38 @@
 
         :else
         (let [buffer-cache-path (.resolve local-disk-cache k)]
-          (-> (if (util/path-exists buffer-cache-path)
-                ;; todo could this not race with eviction? e.g exists for this cond, but is evicted before we can map the file into the cache?
-                (CompletableFuture/completedFuture buffer-cache-path)
-                (do (util/create-parents buffer-cache-path)
-                    (.getObject object-store k buffer-cache-path)))
-              (util/then-apply
-                (fn [path]
-                  (let [nio-buffer (util/->mmap-path path)
-                        create-arrow-buf #(util/->arrow-buf-view allocator nio-buffer)]
-                    (cache-compute memory-store k create-arrow-buf)))))))))
+          (->
+           (update-evictor-key local-disk-cache-evictor k
+                               (fn [^CompletableFuture fut]
+                                 (-> (if (util/path-exists buffer-cache-path)
+                                       (-> fut
+                                           (util/then-apply 
+                                            (fn [{:keys [pinned?]}]
+                                              (log/tracef "Key %s found in local-disk-cache - returning buffer" k)
+                                              {:path buffer-cache-path :previously-pinned? pinned?})))
+                                       (do (util/create-parents buffer-cache-path)
+                                           (log/debugf "Key %s not found in local-disk-cache, loading from object store" k)
+                                           (-> (.getObject object-store k buffer-cache-path)
+                                               (util/then-apply 
+                                                (fn [path]
+                                                  {:path path :previously-pinned? false})))))
+                                     (util/then-apply
+                                      (fn [{:keys [path previously-pinned?]}]
+                                        (let [file-size (util/size-on-disk path)
+                                              nio-buffer (util/->mmap-path path)
+                                              buffer-release-fn (fn []
+                                                                  (set-cache-max-weight local-disk-cache-evictor (swap! !evictor-max-weight + file-size))
+                                                                  (update-evictor-key local-disk-cache-evictor k
+                                                                                      (fn [_] (CompletableFuture/completedFuture {:pinned? false :file-size file-size}))))
+                                              create-arrow-buf #(util/->arrow-buf-view allocator nio-buffer buffer-release-fn)
+                                              buf (cache-compute memory-store k create-arrow-buf)]
+                                          {:pinned? true :file-size file-size :ctx {:buf buf :previously-pinned? previously-pinned?}}))))))
+           (util/then-apply
+            (fn [{:keys [file-size ctx]}] 
+              (let [{:keys [previously-pinned? buf]} ctx] 
+                (when-not previously-pinned? 
+                  (set-cache-max-weight local-disk-cache-evictor (swap! !evictor-max-weight - file-size)))
+                buf))))))))
 
   (listAllObjects [_] (.listAllObjects object-store))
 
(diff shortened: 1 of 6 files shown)


src/test/clojure/xtdb/buffer_pool_test.clj
```diff
@@ -254,3 +255,45 @@
...
+        (insert-utf8-to-local-cache bp (util/->path "a") 4)
+        (insert-utf8-to-local-cache bp (util/->path "b") 4)
+        (t/is (= 2 (file-count local-disk-cache)))) 
+      
+      (t/testing "going above max size - all entries pinned (ie, in memory cache) - should return all elements"
+        (insert-utf8-to-local-cache bp (util/->path "c") 4)
+        (t/is (= 3 (file-count local-disk-cache)))) 
+      
+      (t/testing "entries unpinned (cleared from memory cache by new entries) - should evict entries"
+        (insert-utf8-to-local-cache bp (util/->path "d") 4)
```
danmason: Need to insert a number of entries to start shifting things off of the memory cache/"release" them in order for them to be considered evictable, and similarly we need to insert multiple since all pinned entries have a weight of `0` (ie, need to insert 'enough' for the weight to go over our configured max size). 

This did also initially look at what filenames were in the cache but due to the semantics of Caffeine (ie, Window TinyLfu caching rather than LRU) it's a bit unpredictable as to what exactly gets removed (I'm sure this test could probably be a bit clearer, sort of shifted into actually running nodes against this to see behaviour on a larger scale)

core/src/main/clojure/xtdb/buffer_pool.clj
```diff
@@ -436,12 +445,42 @@
 
 (set! *unchecked-math* :warn-on-boxed)
 
+(defn ->caffeine-cache ^com.github.benmanes.caffeine.cache.Cache [size ^Path local-disk-cache]
+  (log/infof "Creating local-disk-cache with size limit of %s bytes" size)
+  (let [cache (-> (Caffeine/newBuilder)
+                  (.maximumWeight size)
+                  (.weigher (reify Weigher
+                              (weigh [_ k v]
```
danmason: Our `pinning` mechanism in action - setting anything with a `:pinned?` value to weight 0 (so it will not be considered for weight connected eviction)

src/test/clojure/xtdb/buffer_pool_test.clj
ben-manes: We actually inject a bit of randomness into TinyLFU to ensure it is unpredictable in order to protect against HashDoS attacks. That's because we don't want an artificially inflated popularity score for the LRU's victim (least recent) to block the addition of a very popular new arrival (candidate). Nor do we want to always admit or else we lose scan resistance (e.g. a multi-pass analysis loop). So if both options are popular, we accept the candidate 1% of the time to break free to not suffer a degraded state. This makes testing the policy harder, so in our own tests we do an ugly hack by setting the thread-local random seed it a known value. I wouldn't go that far in user code since we might adjust the policy, e.g. we added hill climbing to adapt it, so tests should be behavioral making your approach nicer.

ben-manes:
> This pinning/unpinning approach does have the effect of us going over our configured 'max size' - at least by the amount of ArrowBufs floating around unclosed/unreleased, since our pinning method does reduce the overall current weight of the caffeine cache.

If useful, you could also adjust the cache's maximum size after pinning / unpinning in order to try to respect the limit. That can be done using the [cache.policy()](https://www.javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/latest/com.github.benmanes.caffeine/com/github/benmanes/caffeine/cache/Policy.html) api. It will penalize the caller by requiring that it takes the eviction lock and performs a maintenance cycle (evict if over capacity) which is typically fast.

Pinning is kind of a weird feature so this approach of adjusting to a zero weight was the best compromise that I could think of, though I'm sure others might really dislike it. Some caches allow for skipping the entry from consideration, e.g. Coherence's [EvictionApprover](https://docs.oracle.com/en/middleware/fusion-middleware/coherence/12.2.1.4/java-reference/com/tangosol/net/cache/ConfigurableCacheMap.EvictionApprover.html), but what if nothing can be evicted? Generally that would be an overflow that regularly triggers O(n) scans, though the pinning could be moved off of the main LRU list to avoid that walk. Ehcache tried to resolve that with its [EvictionAdvisor](https://www.ehcache.org/documentation/3.0/eviction-advisor.html) by simply evicting the entry anyway after a threshold, but that completely breaks the stateful needs for this capability and serves only to let users try to correct a dumb eviction policy (at the time they accidentally implemented LRU as MRU so the hit rates were very poor and I think they tried to compensate rather than investigate to find the underlying error). Simply opt'ing out of eviction and adjusting the capacity if needed to compensate felt like the best approach.

src/test/clojure/xtdb/buffer_pool_test.clj
danmason: @ben-manes Thank you for the responses on here! Some very useful suggestions and background for us.

Will probably take a similar approach here as you do in your tests :)

danmason:
> If useful, you could also adjust the cache's maximum size after pinning / unpinning in order to try to respect the limit. That can be done using the [cache.policy()](https://www.javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/latest/com.github.benmanes.caffeine/com/github/benmanes/caffeine/cache/Policy.html) api. It will penalize the caller by requiring that it takes the eviction lock and performs a maintenance cycle (evict if over capacity) which is typically fast.

We had a few chats and came to a similar conclusion/solution - I think for most of our intents/purposes here it makes plenty of sense for it to work like that :)

core/src/main/clojure/xtdb/buffer_pool.clj
```diff
@@ -329,18 +346,20 @@
 
         :else
         (let [buffer-cache-path (.resolve local-disk-cache k)]
-          (-> (if (util/path-exists buffer-cache-path)
-                ;; todo could this not race with eviction? e.g exists for this cond, but is evicted before we can map the file into the cache?
-                (CompletableFuture/completedFuture buffer-cache-path)
+          (-> (if (.getIfPresent local-disk-cache-evictor buffer-cache-path)
```
FiV0: Regarding race conditions and missing files: I suspect you run into the race where `getIfPresent` returns true (thread 1), some other thread removes the file on disc (thread 2) and then you are trying to mmap it again. If you want to make this thread safe you can look at 
```clj
(.. local-disk-cache asMap compute/computeIfPresent (reify BiFunction ....))
```
We did something similar in my PR: https://github.com/FiV0/xtdb/blob/c21398a6dd51f009845de4c5625e7a2c631c8b29/core/src/main/clojure/xtdb/buffer_pool.clj#L47

core/src/main/clojure/xtdb/util.clj
```diff
@@ -447,26 +451,30 @@
...
+                          (.put (.duplicate nio-buffer))
+                          (.clear)))
+         address (MemoryUtil/getByteBufferAddress nio-buffer)
+         size (.remaining nio-buffer)
+         allocation-manager (proxy [AllocationManager] [allocator]
+                              (getSize [] size)
+                              (memoryAddress [] address)
+                              (release0 []
+                                (try-free-direct-buffer nio-buffer)
+                                (when release-fn (release-fn))))
```
FiV0, line 470: Would of course really like to avoid this hack, but will not block it on that.

FiV0 (changes requested):

core/src/main/clojure/xtdb/buffer_pool.clj
danmason: Yeah, this was absolutely a problem - saw this a lot more frequently locally when fixing up the tests. Changed this now, shall mark as resolved - thanks @FiV0 :)

danmason:
The approach where we lower the max weight of the cache when files are pinned seemed to work fairly well - setting the cache size _very_ low (1GB), we certainly do see ourselves rise above that a bit based on what is currently pinned in memory (seen it go as high as 2.5GB), but we do then fairly proactively clear files up as we go. I imagine this approach should work fairly well with reasonably set size limits on the local disk cache. 

Going to observe how it behaves with a slightly higher size cap, 5GB against auctionmark.

core/src/main/clojure/xtdb/buffer_pool.clj
```diff
@@ -345,16 +357,42 @@
...
+                                      (fn [{:keys [path previously-pinned?]}]
+                                        (let [file-size (util/size-on-disk path)
+                                              nio-buffer (util/->mmap-path path)
+                                              buffer-release-fn (fn []
+                                                                  (update-evictor-key local-disk-cache-evictor k
+                                                                                      (fn [^CompletableFuture fut]
+                                                                                        (some-> fut
+                                                                                                (util/then-apply (fn [{:keys [pinned? file-size]}]
+                                                                                                                   (when pinned?
+                                                                                                                     (set-cache-max-weight local-disk-cache-evictor (swap! !evictor-max-weight + file-size)))
```
jarohen: as discussed - atom watcher

core/src/main/kotlin/xtdb/api/storage/Storage.kt
```diff
@@ -109,10 +110,13 @@ object Storage {
         val localDiskCache: Path,
         var maxCacheEntries: Long = 1024,
         var maxCacheBytes: Long = 536870912,
+        // 10GB default
+        var maxLocalDiskCacheBytes: Long = 10737418240,
```
jarohen: REPL hacking:

```clojure
;; fixed size first, no default
;; then percentage of total space, defaulted to 75%

(.getTotalSpace (Files/getFileStore (util/->path "/home/james/src/xtdb/xtdb2/src/test/clojure/xtdb/test_json.clj")))
```

jarohen (approved):

jarohen:
@ben-manes thanks for all your insight on this one :relaxed: :pray:

PR #3476, opened 2024-07-08
Add `plan-q` for lazy querying to the Clojure API.

### LINKED ISSUE #3468: Adding a lazier Clojure query API
`xt/q` currently brings the whole result-set into memory, but a user may want to stream the result out to (e.g.) an HTTP response, or a file.

Options:

1. XT1-style `open-q`
2. next.jdbc-style `plan-q` (name tbc)
3. something else. RFC?

FiV0:
Adds a `plan-q` function to the Clojure API which returns an `IReduceInit` for controlling the realized result set.

Second phrase of the doc string was stolen from `next.jdbc`.

Todo:
- [x] ~I haven't checked the lower level bits of the jetty server in combination with our clojure http client and how the http response is handled. Is the `InputStream` hato (our http client) actually buffered to disk when received or will this also fill up memory?~ We are assuming the client deals fast enough with the incoming data stream or the client deals somehow with the back pressure on the result stream if rows get consumed slower than the response comes in.
- [x] If this PR goes in, we could potentially implement `xt/q` based on this.

### DIFF
--- a/api/src/main/clojure/xtdb/api.clj
+++ b/api/src/main/clojure/xtdb/api.clj
@@ -13,14 +13,14 @@
             [xtdb.error :as err]
             [xtdb.protocols :as xtp]
             [xtdb.serde :as serde]
-            [xtdb.time :as time]
-            [xtdb.tx-ops :as tx-ops])
-  (:import (java.io Writer)
-           (java.util List Map)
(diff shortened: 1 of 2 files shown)


jarohen:
Looking good so far, agree with both of your todos :)

PR #3489, opened 2024-07-11
Add support for datetimes in pgwire

### LINKED ISSUE #3458: Timestamp types in pgwire

mbutlerw:
Commit also includes various refactors to the pgwire server and tests

Didn't end up using the env for type encoding, but having done the refactor, inclined to leave it in place for future types/changes in behaviour. Can eventually refactor it away if it proves unnecessary.


src/test/clojure/xtdb/pgwire_test.clj
```diff
@@ -1380,3 +1404,115 @@
...
+
+               (.next rs)
+               (t/is (= val (.getObject rs 1 type)))))))))
+
+(deftest test-timestamptz
+  (let [type OffsetDateTime
+        pg-type "timestamptz"
+        val #xt.time/offset-date-time "2024-07-03T19:01:34-07:00"
+        val2 #xt.time/offset-date-time "2024-07-03T19:01:34.695959+03:00"]
+    (doseq [binary? [true false]]
```
jarohen, line 1439: nice - I'd be tempted to add something like `(t/testing (if binary? "binary" "text") ...)` round here so that test fails tell you which mode you're in

core/src/main/clojure/xtdb/types.clj
```diff
@@ -717,55 +721,76 @@
   ^bytes [s]
   (.getBytes (str s) StandardCharsets/UTF_8))
 
+(def iso-local-date-time-formatter-with-space
```
jarohen, line 724: now that pgwire's in core I'd be tempted to shift the pg-specific stuff out to xtdb.pgwire.types or similar?

jarohen (approved):
Nice one @mbutlerw :clap:

Think you're aware, but generally speaking, looking to move away from the canned queries where possible, now that it's relatively straightforward for us to add proper parsing. dw about it for this change but maybe a separate refactoring we can do in a quiet moment.

PR #3502, opened 2024-07-17
Add first-row to L0 and L1filenames - ignore superseded L1 files

### LINKED ISSUE #3498: Superseded L1 files get returned by `current-trie-files`
I ran into this while investigating the slow down in Auctionmark (https://github.com/xtdb/xtdb/issues/3452).

Currently the compactor creates a new L1 file from the most recent L0 file and a previous L1 file (accumulation of previous L0 files) if the accumulation of all L0 files doesn't go over the chunk boundary (~ 100k rows). Meaning if many chunks have finished but the accumulation of all L0 files for a particular table doesn't add up to more than 100k rows, there should be only one L1 file for that table. This is currently not the case for the `item-table` in Auctionmark. The `current-trie-files` fn returns all 10 L1 files even though only the newest is relevant. The data returned is still correct as there is no extra data or incorrect temporal resolution happening, we just might have the same event from multiple L1 files.

@jarohen already explained part of a solution to me. L1 files currently contain level (1), next row and row count as part of the their filename. To determine if a file has been superseded or not, we need to add the starting row as well to the filename, as it is otherwise not possible to determine from the filename if it has been superseded or not.

FiV0:
This is a fix for https://github.com/xtdb/xtdb/issues/3498. We are adding a first row to L0 and L1 file names. The PR mainly checks that we are only picking one L1 file per first-row (if not superseded by an L2+) and that we are picking the L1 file with the largest next-row.


core/src/main/clojure/xtdb/trie.clj
```diff
@@ -310,6 +312,9 @@
 
           {l0-trie-keys 0, l1-trie-keys 1, :as level-grouped-file-names} (file-names->level-groups file-names)
 
+          ;; filtering superseded L1 files
+          l1-trie-keys (->> l1-trie-keys (sort-by (juxt :first-row :next-row) #(compare %2 %1)) (util/distinct-by :first-row) reverse)
```
jarohen: We can cut down on some of the more expensive operations and intermediate data structures here because the file names are already (and still, after this PR :clap:) lexicographically sorted - the files with the same first row should (IIRC) already be next to each other in `l1-trie-keys`.

i.e. you should then be able to `(partition-by :first-row)` rather than sorting and distincting, `(map last)` and possibly save yourself the `reverse` too - `(into [] (comp (partition-by :first-row) (map last)) l1-trie-keys)`?

jarohen (approved):
Nice one @FiV0, good implementation and good set of tests - well done mate :clap: :bow: 

Just the one suggestion in the comments, and the version bump as discussed, then this should be good to go :slightly_smiling_face: 

I'd be interested to see how much of a performance gain this is, too - in theory, this would have been us scanning through more data as time goes on, which you've now removed :thinking:

core/src/main/clojure/xtdb/trie.clj
FiV0: Hadn't realized that we have this guarantee on the buffer pool.

PR #3512, opened 2024-07-19
Scan recency filter for older recency pages

FiV0:
This adds the recency filter to scan for older pages than the query range of scan. 

The formula is essentially `while recency of page < min(validTo.lower, systemTo.lower)` we are ignoring those pages. 

Todo:
- [x] Walk through scan recency test.

### DIFF
--- a/core/src/main/kotlin/xtdb/trie/HashTrie.kt
+++ b/core/src/main/kotlin/xtdb/trie/HashTrie.kt
@@ -5,9 +5,12 @@ import org.apache.arrow.memory.util.ArrowBufPointer
 import xtdb.trie.ArrowHashTrie.IidBranch
 import xtdb.trie.ArrowHashTrie.RecencyBranch
 import xtdb.trie.HashTrie.Node
+import xtdb.util.TemporalBounds
+import java.time.Instant
 import java.util.*
 import java.util.function.Predicate
 import java.util.stream.Stream
+import kotlin.math.min
 
 internal typealias RecencyArray = LongArray
 
@@ -82,9 +85,10 @@ data class MergePlanNode(val segment: ISegment, val node: Node<*>)
 class MergePlanTask(val mpNodes: List<MergePlanNode>, val path: ByteArray)
 
 @Suppress("UNUSED_EXPRESSION")
-fun toMergePlan(segments: List<ISegment>, pathPred: Predicate<ByteArray>?): List<MergePlanTask> {
+fun toMergePlan(segments: List<ISegment>, pathPred: Predicate<ByteArray>?, temporalBounds: TemporalBounds = TemporalBounds()): List<MergePlanTask> {
     val result = mutableListOf<MergePlanTask>()
     val stack = ObjectStack<MergePlanTask>()
+    val minRecency = min(temporalBounds.validTo.lower, temporalBounds.systemTo.lower)
 
     val initialMpNodes = segments.mapNotNull { seg -> seg.trie.rootNode?.let { MergePlanNode(seg, it) } }
     if (initialMpNodes.isNotEmpty()) stack.push(MergePlanTask(initialMpNodes, ByteArray(0)))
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/xtdb/trie/HashTrie.kt
```diff
@@ -97,11 +101,13 @@ fun toMergePlan(segments: List<ISegment>, pathPred: Predicate<ByteArray>?): List
...
                 for (mpNode in mergePlanTask.mpNodes) {
-                    // TODO filter by recencies #3166
                     val recencies = mpNode.node.recencies
                     if (recencies != null) {
-                        for (i in recencies.indices) {
+                        var i = 0
+                        while (i < recencies.size && recencies[i] <= minRecency) i++
+                        while (i < recencies.size) {
                             newMpNodes += MergePlanNode(mpNode.segment, mpNode.node.recencyNode(i))
+                            i++
```
FiV0: This 5 lines are actually the logic, so probably good to have a second :eyes: over this.

src/test/clojure/xtdb/operator/scan_test.clj
```diff
@@ -676,3 +679,56 @@
...
+                       (count (tu/query-ra '[:scan {:table docs :for-valid-time [:between #inst "2025" nil]} [xt$id]]
+                                           {:node node}))))
+              ;; 2024 and inf
+              (t/is (= 2 @merge-tsk-cnt))
+
+
+              ;; everything
+              (t/is (= 5
+                       (count (tu/query-ra '[:scan {:table docs :for-valid-time :all-time} [xt$id]] {:node node}))
+                       @merge-tsk-cnt)))))))))
```
FiV0: This test is a bit of a mouthful (or however you would say that for a test). So maybe we can find something simpler.

core/src/main/kotlin/xtdb/util/TemporalBounds.kt
```diff
@@ -20,11 +21,19 @@ class TemporalBounds {
         fun gte(operand: Long) = apply { lower = max(operand, lower) }
 
         fun inRange(operand: Long) = operand in lower..upper
+
+        override fun toString(): String {
```
jarohen: these might be better achieved by making TemporalBounds and TemporalColumn data classes, if that works?

FiV0: Can make them data classes. The override for `TemporalColumn` still needs to happen if I want a human friendlier output or is there a way to annotate data fields with special printers?

src/test/clojure/xtdb/operator/scan_test.clj
```diff
@@ -676,3 +679,57 @@
...
+          (tu/finish-chunk! node)
+          ;; compaction happens in 2025
+          (c/compact-all! node)
+
+          (let [->old-merge-tasks scan/->merge-tasks
+                merge-tsk-cnt (atom 0)
+                query-opts {:node node
+                            :basis {:at-tx tx-key
+                                    :current-time (:system-time tx-key)}}]
+            (with-redefs [scan/->merge-tasks (fn [& args]
```
jarohen, line 705: as discussed, investigate whether we can call `->merge-tasks` as a unit test here instead

src/test/clojure/xtdb/trie_test.clj
```diff
@@ -29,39 +29,78 @@
...
+                                           (-> (trie/->Segment (->arrow-hash-trie log-root))
+                                               (assoc :seg :log))
+                                           (-> (trie/->Segment (->arrow-hash-trie log2-root))
+                                               (assoc :seg :log2))]
+                                          nil
+                                          (TemporalBounds.))
+                  (merge-plan-nodes->path+pages))))))
+
+(deftest test-merge-plan-recency-filtering
+  (with-open [al (RootAllocator.)
```
jarohen, line 73: todos:

- [x] comment up exactly what the test case is stressing - what's the case, what are you hoping the algo will do?
- [x] are there other cases to consider here which might catch the algo out?

src/test/clojure/xtdb/operator/scan_test.clj
jarohen: as discussed, let's remove the with-redefs here, just test the final answer, and rely on the unit tests below to confirm that it got there in the correct manner.

value of these tests is in showing that, even for cases which might cause the algo trouble, we still return the correct answers - i.e. we haven't missed a page

jarohen (approved):
nice one @FiV0 :clap:

PR #3523, opened 2024-07-24
generate_series

### LINKED ISSUE #3212: `generate_series` function
See Postgres ['set-returning functions'](https://www.postgresql.org/docs/current/functions-srf.html).

This may need to be an operator (similar to `rel`) rather than an EE function.

egg-juxt:


jarohen:
hey @egg-juxt - fixed in 6ee9e48bb31bb37f4cfeda6d49e39a648737eed5. :pray:

PR #3540, opened 2024-07-30
Filter newer pages that don't overlap in system-time and valid-time

### LINKED ISSUE #3522: Ruling out earlier pages/tries for valid-time / system-time range queries
The proposal outlines how to skip newer/later pages and (sub) tries when doing valid-time/system-time range queries.

- The PR #3512 skips older pages than the query and should be merged before this work  goes in.
- The issue #3518 is an extension to this work and would (likely) make the algorithm below quicker.

Currently we have temporal metadata per page. #3518 would add this metadata per trie node and per trie itself.
I am describing the algorithm for valid-time, but the same applies to system-time.

The actual algorithm based on the (new) metadata. Assuming we get a query for (q_valid_from, q_valid_to) (would be an epsilon for a point query). 

* We first find the relevant data based on recency buckets. Earlier recency buckets can definitely be filtered out (The existing PR https://github.com/xtdb/xtdb/pull/3512 covers this). 
* We then move FORWARD in system time (across tries, nodes and pages, if the former two have the appropriate metadata). If a page overlaps the q_valid_bounds with min valid_from and max valid_to we take that page also into account when doing temporal resolution. 
* This information should be stored as metadata, so that the pages/files might not need to be fetched. In case a trie or node overlaps, we need to look at finer grained metadata (subtries or pages) to decide which pages to actually take into account. If the metadata does not overlap at trie or node level the whole trie and subtrie can be skipped respectively.
 
Examples (assuming we have pages for 2021, 2022 and 2023 and current):
- i.e. if interested in 2021 (query valid time range is in 2021), and 2023 overlaps but nothing else we take pages 2021 and 2023
- i.e if interested in 2022 but 2023 and current don’t overlap we only need to look at 2022

It is counter-intuitive that in the first example we don't need to take page 2022 into account when doing temporal resolution even though we have taken 2023. The metadata assures us that nothing in the 2022 page can effect anything in the projected data.    

For completeness different query scenarios:
- sys_now, vt_now - not much changes. look at all current pages and do temporal resolution.
- sys_all, vt_all - not much changes, we need to do full bitemp resolution anyway
- sys_past_point, vt_now - skip later pages
- sys_now, vt_range - where we benefit the most of this extra metadata
- sys_range, vt_range - The most generalized approach of the above. The temporal bounds are passed to the merge plan calculation and the metadata for system-time as well as valid-time get taken into account. All the above scenarios are covered by this one and use the same code path.

Just to be clear. There won’t be a big case statement on the above, this is just a summary of what would actually happen in these scenarios.

FiV0:
This uses the temporal metadata information to filter pages in scan.

The two main advantages are:
- historical system time queries filter out newer pages completely
- valid time queries only include pages that can effect the requested valid time range

Todo:
- [x] rebase
- [x] index bump
- [x] refactor of `->merge-tasks` with data structures that are easier to setup
- [x] move comments in tests into `t/is / t/testing` strings
- [x] What cases (in terms of valid-time pruning) does this NOT cover?

### DIFF
--- a/core/src/main/clojure/xtdb/metadata.clj
+++ b/core/src/main/clojure/xtdb/metadata.clj
@@ -30,6 +30,7 @@
            xtdb.IBufferPool
            (xtdb.metadata ITableMetadata PageIndexKey)
            (xtdb.trie ArrowHashTrie HashTrie)
+           (xtdb.util TemporalBounds TemporalDimension)
            (xtdb.vector IVectorReader IVectorWriter)
            (xtdb.vector.extensions KeywordType SetType TransitType UriType UuidType)))
 
@@ -320,13 +321,16 @@
     {:col-names (into #{} col-names)
      :page-idx-cache page-idx-cache}))
 
+
 (defrecord TableMetadata [^HashTrie trie
                           ^Relation meta-rel
                           ^ArrowBuf buf
                           ^IVectorReader metadata-leaf-rdr
                           col-names
                           ^Map page-idx-cache
-                          ^AtomicInteger ref-count]
+                          ^AtomicInteger ref-count
+                          ^IVectorReader min-rdr
+                          ^IVectorReader max-rdr]
   ITableMetadata
   (metadataReader [_] metadata-leaf-rdr)
   (columnNames [_] col-names)
@@ -341,11 +345,20 @@
         (when (.getObject bloom-rdr bloom-vec-idx)
           (bloom/bloom->bitmap bloom-rdr bloom-vec-idx)))))
 
+  (temporalBounds[_ page-idx]
+    (let [^long system-from-idx (.get page-idx-cache (PageIndexKey. "xt$system_from" page-idx))
+          ^long valid-from-idx (.get page-idx-cache (PageIndexKey. "xt$valid_from" page-idx))
+          ^long valid-to-idx (.get page-idx-cache (PageIndexKey. "xt$valid_to" page-idx))]
+      (TemporalBounds. (TemporalDimension. (.getLong min-rdr valid-from-idx) (.getLong max-rdr valid-to-idx))
+                       (TemporalDimension. (.getLong min-rdr system-from-idx) Long/MAX_VALUE))))
+
   AutoCloseable
   (close [_]
     (when (zero? (.decrementAndGet ref-count))
       (util/close meta-rel)
       (util/close buf))))
+
+(def ^:private temporal-col-type-leg-name (name (types/arrow-type->leg (types/->arrow-type [:timestamp-tz :micro "UTC"]))))
 
 (defn ->table-metadata ^xtdb.metadata.ITableMetadata [^IBufferPool buffer-pool ^Path file-path]
   (util/with-close-on-catch [buf (.getBuffer buffer-pool file-path)]
@@ -357,8 +370,16 @@
           (let [rdr (.getOldRelReader rel)
                 ^IVectorReader metadata-reader (-> (.readerForName rdr "nodes")
                                                    (.legReader "leaf"))
-                {:keys [col-names page-idx-cache]} (->table-metadata-idxs metadata-reader)]
-            (->TableMetadata (ArrowHashTrie. nodes-vec) rel buf metadata-reader col-names page-idx-cache (AtomicInteger. 1))))))))
+                {:keys [col-names page-idx-cache]} (->table-metadata-idxs metadata-reader)
+                temporal-col-types-rdr (some-> (.structKeyReader metadata-reader "columns")
+                                               (.listElementReader)
+                                               (.structKeyReader "types")
+                                               (.structKeyReader temporal-col-type-leg-name))
+
+                min-rdr (some-> temporal-col-types-rdr (.structKeyReader "min"))
+                max-rdr (some-> temporal-col-types-rdr (.structKeyReader "max"))]
+            (->TableMetadata (ArrowHashTrie. nodes-vec) rel buf metadata-reader col-names page-idx-cache (AtomicInteger. 1)
+                             min-rdr max-rdr)))))))
 
 (deftype MetadataManager [^IBufferPool buffer-pool
                           ^Cache table-metadata-cache
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -13,7 +13,7 @@
             [xtdb.metadata :as meta]
             xtdb.object-store
             [xtdb.time :as time]
-            [xtdb.trie :as trie]
+            [xtdb.trie :as trie :refer [MergePlanPage]]
             [xtdb.types :as types]
             [xtdb.util :as util]
             [xtdb.vector.reader :as vr]
@@ -231,10 +231,6 @@
 
 (defrecord LeafPointer [ev-ptr rel-idx])
 
-(defprotocol MergePlanPage
-  (load-page [mpg buffer-pool vsr-cache])
-  (test-metadata [msg]))
-
 (deftype TrieCursor [^BufferAllocator allocator, ^Iterator merge-tasks, ^IRelationWriter out-rel
                      col-names, ^Map col-preds,
                      ^TemporalBounds temporal-bounds
@@ -253,7 +249,7 @@
                   calculate-polygon (bitemp/polygon-calculator temporal-bounds)
                   bitemp-consumer (->bitemporal-consumer out-rel col-names)
                   leaf-rdrs (for [leaf leaves
-                                  :let [^RelationReader data-rdr (load-page leaf buffer-pool vsr-cache)]]
+                                  :let [^RelationReader data-rdr (trie/load-page leaf buffer-pool vsr-cache)]]
                               (cond-> data-rdr
                                 iid-pred (.select (.select iid-pred allocator data-rdr params))))
                   [temporal-cols content-cols] ((juxt filter remove) temporal-column? col-names)
@@ -349,7 +345,7 @@
         (test [_ path]
           (zero? (HashTrie/compareToPath iid-ptr path)))))))
 
-(defrecord ArrowMergePlanPage [data-file-path ^IntPredicate page-idx-pred ^long page-idx]
+(defrecord ArrowMergePlanPage [data-file-path ^IntPredicate page-idx-pred ^long page-idx ^ITableMetadata table-metadata]
   MergePlanPage
   (load-page [_mpg buffer-pool vsr-cache]
     (util/with-open [rb (bp/open-record-batch buffer-pool data-file-path page-idx)]
@@ -359,29 +355,20 @@
         (vr/<-root vsr))))
 
   (test-metadata [_mpg]
-    (.test page-idx-pred page-idx)))
+    (.test page-idx-pred page-idx))
+
+  (temporal-bounds [_mpg] (.temporalBounds table-metadata (int page-idx))))
+
+(def ^:private non-constraint-bounds (TemporalBounds.))
 
 (defrecord LiveMergePlanPage [^RelationReader live-rel trie ^LiveHashTrie$Leaf leaf]
   MergePlanPage
   (load-page [_mpg _buffer-pool _vsr-cache]
     (.select live-rel (.mergeSort leaf trie)))
 
-  (test-metadata [_mpg] true))
+  (test-metadata [_mpg] true)
 
-(defn ->merge-task [mp-pages]
-  (let [leaves (ArrayList.)]
-    (loop [[mp-page & more-mp-pages] mp-pages
-           node-taken? false]
-      (if mp-page
-        (let [take-node? (test-metadata mp-page)]
-
-          (when (or take-node? node-taken?)
-            (.add leaves mp-page))
-
-          (recur more-mp-pages (or node-taken? take-node?)))
-
-        (when node-taken?
-          (vec leaves))))))
+  (temporal-bounds [_msg] non-constraint-bounds))
 
 (defmethod ig/prep-key ::scan-emitter [_ opts]
   (merge opts
@@ -449,7 +436,7 @@
                                                 :param-types (update-vals param-fields types/field->col-type)}]
                                (MapEntry/create col-name
                                                 (expr/->expression-selection-spec (expr/form->expr select-form input-types)
-                                                                                     input-types))))
+                                                                                  input-types))))
                            (into {}))
 
             metadata-args (vec (for [[col-name select] selects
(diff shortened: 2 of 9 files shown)


FiV0:
Struggling to find some good and not hairy abstraction for testing, probably good to walk/chat this through with someone.

core/src/main/clojure/xtdb/metadata.clj
```diff
@@ -351,6 +354,27 @@
         (when (.getObject bloom-rdr bloom-vec-idx)
           (bloom/bloom->bitmap bloom-rdr bloom-vec-idx)))))
 
+  (temporalBounds[_ page-idx]
+    ;; TODO when badly distributed there are pages with no data
```
FiV0: At some point I was running into the case where the tries were pointing to empty pages with badly distributed data. I don't know if we should consider this a bug or if such cases should be filtered elsewhere in the trie merging logic.

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -361,29 +362,68 @@
...
+                    (cond-> non-taken-pages
+                      (and node-taken?
+                           (.intersectsSystemTime page-bounds query-bounds)
+                           (.canBoundValidTime query-bounds page-bounds))
+                      (conj mp-page)))))
+
+         (when node-taken?
+           (loop [[page & more-pages] non-taken-pages]
+             (when page
+               (let [smallest-valid-from (min-valid-from (temporal-bounds page))]
```
FiV0: Getting `temperol-bounds` a second time here.

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -235,7 +235,8 @@
 
 (defprotocol MergePlanPage
   (load-page [mpg buffer-pool vsr-cache])
-  (test-metadata [msg]))
+  (test-metadata [msg])
+  (temporal-bounds [msg]))
```
FiV0: I added a quite generic `temporal-bounds` method. We could add more methods that are finer (`min-sys-from`, `max-sys-to` ....), but the temporal-bounds seemed the best suited.

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -361,29 +362,68 @@
...
+               take-node? (and (.intersects page-bounds query-bounds)
+                               (test-metadata mp-page))]
+
+           (if take-node?
+             (do
+               (.add leaves mp-page)
+               (recur more-mp-pages
+                      true
+                      (cond-> largest-valid-to
+                        (not (instance? LiveMergePlanPage mp-page))
```
FiV0: The instance check was something we wanted to avoid with the `MergePlanNode` protocol. The problem is that on the one hand we don't have metadata on the live-trie and hence are just returning temporal-bounds that are unconstrained (i.e. always take the live trie), on the other hand we don't want these temporal bounds to be included into the max-valid-to calculation.

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -361,29 +362,68 @@
...
+                    node-taken?
+                    largest-valid-to
+                    (cond-> non-taken-pages
+                      (and node-taken?
+                           (.intersectsSystemTime page-bounds query-bounds)
+                           (.canBoundValidTime query-bounds page-bounds))
+                      (conj mp-page)))))
+
+         (when node-taken?
+           (loop [[page & more-pages] non-taken-pages]
```
FiV0: This loop could be made conditional on `valid-to` projections under the assumption we allow for version coalescing.

core/src/main/kotlin/xtdb/util/TemporalBounds.kt
```diff
@@ -4,34 +4,66 @@ import java.time.Instant
...
+
+    fun canBoundValidTime(other: TemporalBounds) =  overlapsValidTime(other) && validTo.lower <= other.validTo.upper
+    fun canBoundSytemTime(other: TemporalBounds) =  overlapsSystemTime(other) && systemTo.lower <= other.systemTo.upper
+    fun canBound(other: TemporalBounds) = canBoundValidTime(other) && canBoundSytemTime(other)
+
+    fun union(other: TemporalBounds) = TemporalBounds(
+        validFrom = validFrom.union(other.validFrom).union(other.validTo),
+        validTo = validTo.union(other.validFrom).union(other.validTo),
+        systemFrom = systemFrom.union(other.systemFrom),
+        systemTo = systemTo.union(other.systemTo))
```
FiV0: There is a lot of stuff here which is currently unused. I was just switching often between these and also wanted to include the symmetric operations. I can also slim this done if wished for.

src/test/clojure/xtdb/trie_test.clj
```diff
@@ -136,12 +137,286 @@
...
+                                               (load-page [_ _ _] {:seg seg :page-idx (.getDataPageIndex ^ArrowHashTrie$Leaf node)})
+                                               (test-metadata [_] true)
+                                               (temporal-bounds [_] (page-bounds-fn (.getDataPageIndex ^ArrowHashTrie$Leaf node)))))
+                                           query-bounds)]
+                          {:path (vec (.getPath mpt))
+                           :pages (mapv #(scan/load-page % nil nil) leaves)}))))
+       (sort-by :path)))
+
+(deftest test-to-merge-tasks-with-temporal-filters
+  (let [
```
FiV0: This is a test I had before the refactoring. It essentially combines `toMergePlan` with `->merge-task`. If we consider the above test (`test-->merge-task`) sufficient I can also delete the code related to this.

FiV0:
@jarohen I added a bunch of comments inline where there is room for discussion + issues.

core/src/main/clojure/xtdb/metadata.clj
```diff
@@ -351,6 +354,27 @@
...
+    (if-let [temporal-col-types-rdr (-> (.structKeyReader metadata-leaf-rdr "columns")
+                                        (.listElementReader)
+                                        (.structKeyReader "types")
+                                        (.structKeyReader temporal-col-type-leg-name))]
+      (let [min-rdr (.structKeyReader temporal-col-types-rdr "min")
+            max-rdr (.structKeyReader temporal-col-types-rdr "max")
+            ^long system-from-idx (.get page-idx-cache (PageIndexKey. "xt$system_from" page-idx))
+            bounds (TemporalBounds.)]
+        ;; assumes sys-from, valid-from, valid-to always appear in that order
```
jarohen: this is what the page-idx-cache is for - unless/until it shows up in the profiling I'd err on the side of being safe

core/src/main/clojure/xtdb/metadata.clj
jarohen: yes, ideally - when the page is created. there's some logic here already, IIRC, but clearly it isn't catching a certain case...

core/src/main/clojure/xtdb/operator/scan.clj
jarohen: would be good to understand this one a bit better :sweat_smile:

edit: maybe this is a sign that re-using the temporal bounds isn't quite sound?

core/src/main/clojure/xtdb/operator/scan.clj
jarohen: I'm not 100% on this one - it seems like we're re-using TemporalBounds for something it wasn't intended for. That is, originally this was intended for the temporal constraints of a query (and the methods in there assume that); now, we might still have lower and upper on valid-from/sys-from etc but they've got subtly different meanings, so the assumptions of the original methods may be invalid.

For the cost of a new (albeit similar) class, I think separating these out might make it less error-prone?

src/test/clojure/xtdb/operator/scan_test.clj
```diff
@@ -694,42 +694,95 @@
...
-    (binding [c/*page-size* 1
-              c/*l1-file-size-rows* 256
-              c/*ignore-signal-block?* true]
-      (with-open [node (xtn/start-node {:log [:in-memory {:instant-src (tu/->mock-clock (tu/->instants :year))}]
-                                        :storage [:local {:path (io/file (.toFile node-dir) "objects")}]
-                                        :indexer {:rows-per-chunk 1}})]
-        ;; 2020 - 2025
-        (let [tx-key (last (for [i (range 6)]
-                             (xt/execute-tx node [[:put-docs :docs {:xt/id 1 :col i}]])))]
+  (binding [c/*page-size* 1
```
jarohen, line 705: This is a good high-level test, definitely worth keeping - but part of the aim of our refactoring earlier in the week was that we can then easier unit-test the lower-level function. With the new tests as they stand, we don't necessarily know that we've filtered the data out at the page level?

jarohen: aha - they're in trie-test. worth bringing those tests into this namespace, given they're testing scan functions?

src/test/clojure/xtdb/trie_test.clj
```diff
@@ -136,24 +137,301 @@
...
+                                                [2 micros-2020 micros-2021]
+                                                [3 micros-2021 micros-2022]
+                                                [4 micros-2022 Long/MAX_VALUE]]]
+                      (->mock-merge-plan-page page min-vf max-vt))
+                    query-bounds)
+                   (mapv :page)))]
+
+
+      (let [query-bounds (TemporalBounds.)]
+        ;; stricly inside 2020
```
jarohen: I'd be tempted by a letfn/helpers here to create the query-bounds?

core/src/main/kotlin/xtdb/util/TemporalBounds.kt
jarohen: this suggests that we might have mis-modelled the valid-time and system-time parts of the TemporalBounds?

e.g. maybe rather than TB (vf,vt,sf,st) -> TC(lower, upper) we change it up slightly to TB(valid, sys) -> TC(min-from, max-from, min-from, min-to)?

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -360,29 +361,67 @@
...
+                        (max (max-valid-to page-bounds)))
+                      non-taken-pages))
+
+             (recur more-mp-pages
+                    node-taken?
+                    largest-valid-to
+                    (cond-> non-taken-pages
+                      (and node-taken?
+                           (.intersectsSystemTime page-bounds query-bounds)
+                           (.canBoundValidTime query-bounds page-bounds))
```
jarohen: (this stuff's hard to visualise, so entirely possible I've not got this one right/missed a trick)

IIUC, the criterion here is that we want to include a later page if (the user has projected vt, and) it can bound the valid-times of the earlier _page_, not the query as a whole?

:exploding_head: :thinking:

jarohen (commented):
Hey mate, well done for getting the PR raised here, not an easy one in the slightest :sweat_smile:

Few questions and thoughts below, happy to have a chat if it'd be easier

core/src/main/clojure/xtdb/metadata.clj
FiV0: See https://github.com/xtdb/xtdb/issues/3580

core/src/main/clojure/xtdb/operator/scan.clj
FiV0: Should mostly be resolved by https://github.com/xtdb/xtdb/pull/3583.

core/src/main/kotlin/xtdb/util/TemporalBounds.kt
FiV0: See https://github.com/xtdb/xtdb/pull/3583

core/src/main/clojure/xtdb/operator/scan.clj
FiV0: Let's come back to this after the rebase and let's talk it through.

src/test/clojure/xtdb/operator/scan_test.clj
FiV0: Move `->merge-task` to `trie.clj`

FiV0:
I did some rudimentary testing via https://github.com/xtdb/xtdb/pull/3579. With a lot of history the bottleneck becomes the `temporalBounds` calculation. Currently there is still some overhead per page and as the history grows that overhead becomes more and more apparent.
 
For example a simple aggregate over the last week: 
- 150 ms with two years of data
- 2sec + with ten years of data

Potential options:
- Add metadata at higher leves in the tries, so that the whole trie or partial subtrees can be quickly pruned. See https://github.com/xtdb/xtdb/issues/3518
- Add metadata for files/tries. Then one wouldn't even need to look at the tries.

core/src/main/clojure/xtdb/metadata.clj
```diff
@@ -351,6 +354,21 @@
         (when (.getObject bloom-rdr bloom-vec-idx)
           (bloom/bloom->bitmap bloom-rdr bloom-vec-idx)))))
 
+  (temporalBounds[_ page-idx]
+    ;; TODO when badly distributed there are pages with no data
+    (if-let [temporal-col-types-rdr (-> (.structKeyReader metadata-leaf-rdr "columns")
```
FiV0: Try to push reader creation to constructor.

FiV0:
The numbers in the above benchmark need to be taken with a grain of salt. As I am backfilling more eagerly (system time is squashed with respect to valid time), the recency filtering can likely not unfold its full power.
 
In theory we would see the following scenarios for a period of 10 years (2014 - 2024):
- Aggregate 2024: 2014-2023 should mostly be filtered by pruning older recency pages hence quite high up in the tries
- Aggregate 2023: 2014-2022 (see above), for 2024 we would still need to look at all the temporal metadata of the pages
- Aggregate 2014: Worst-case scenario (in terms of filtering trie pages), no recency filter can be applied in the tries so we (currently) need to look at the temporal bounds per page for 2015-2024

L-shaped world:
- old pages still get filtered by recency
- newer pages can always bound the pages in the VT query range. In this case we would really benefit by only including them if `xt/valid-to` gets projected.

In general if people don't explicitly bound their valid-to times this will preform poorly (in it's current form) when going back in history without being interested in the recent history (ie. query for data in 2014).

core/src/main/clojure/xtdb/metadata.clj
FiV0: Tried to push this into the constructor, but running into issues with a nil `temporal-col-types-rdr`. Some issue when data is badly distributed. I thought this is fixed, but there still seem to be cases when the metadata doesn't exist.

FiV0: If a trie is empty then these readers don't exist.

jarohen (approved):

PR #3560, opened 2024-08-05
Renamed aggregation functions ANY/SOME to BOOL_OR. Added BOOL_AND synonym for EVERY

### LINKED ISSUE #3554: `ANY`/`SOME`/`EVERY` aggregate expressions renamed to `BOOL_OR`, `BOOL_AND`
See [this note](https://www.postgresql.org/docs/9.5/functions-aggregate.html#:~:text=As%20for%20any%20and%20some%2C%20it%20seems%20that%20there%20is%20an%20ambiguity%20built%20into%20the%20standard%20syntax) from Postgres regarding the ambiguity in the SQL spec, and the approach they took to work around it.

egg-juxt:


core/src/main/clojure/xtdb/operator/group_by.clj
```diff
@@ -564,10 +564,9 @@
                                            :args [{:op :local, :local acc-local}
                                                   {:op :local, :local val-local}]}})))
 
-(defmethod ->aggregate-factory :all [agg-opts] (bool-agg-factory :and agg-opts))
-(defmethod ->aggregate-factory :every [agg-opts] (->aggregate-factory (assoc agg-opts :f :all)))
-(defmethod ->aggregate-factory :any [agg-opts] (bool-agg-factory :or agg-opts))
-(defmethod ->aggregate-factory :some [agg-opts] (->aggregate-factory (assoc agg-opts :f :any)))
+(defmethod ->aggregate-factory :bool_and [agg-opts] (bool-agg-factory :and agg-opts))
+(defmethod ->aggregate-factory :every [agg-opts] (->aggregate-factory (assoc agg-opts :f :bool_and)))
```
jarohen, line 568: let's drop `EVERY` here too, seems inconsistent to leave that on its own just because the SQL spec doesn't include it in `ANY`/`ALL`/`SOME`

jarohen (approved):
cheers @egg-juxt - one minor change and good to go :)

core/src/main/clojure/xtdb/operator/group_by.clj
egg-juxt: I thought about it, but I left it because PostgreSQL does have it: https://www.postgresql.org/docs/9.5/functions-aggregate.html

PR #3561, opened 2024-08-05
Add `StringMapWithEnvVarsSerde`, use within Kafka. Resolves #3557

### LINKED ISSUE #3557: `!Env` YAML tag to work in nested maps
We're adding the ability to source Kafka config directly from our YAML map. We'd still like secrets to be submitted through env vars, but these only apply at the top-level atm - let's make them available in nested maps too.

danmason:
Adding a new deserializer class such that we can handle environment variables inside of nested maps - and then use within kafka. This would allow you to set things within the `propertiesMap` in Kafka and pass them in via environment variables (ie, can pass in things like `sasl.jaas.config` from a secret)

### DIFF
--- a/core/src/main/kotlin/xtdb/api/YamlSerde.kt
+++ b/core/src/main/kotlin/xtdb/api/YamlSerde.kt
@@ -8,6 +8,8 @@ import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.descriptors.PrimitiveKind
 import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
 import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.descriptors.buildSerialDescriptor
+import kotlinx.serialization.descriptors.StructureKind
 import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/xtdb/api/YamlSerde.kt
```diff
@@ -78,6 +80,34 @@ object StringWithEnvVarSerde : KSerializer<String> {
     }
 }
 
+object StringMapWithEnvVarsSerde : KSerializer<Map<String, String>> {
+    @OptIn(kotlinx.serialization.ExperimentalSerializationApi::class, kotlinx.serialization.InternalSerializationApi::class)
+    override val descriptor: SerialDescriptor = buildSerialDescriptor("StringMapWithEnvVarsSerde", StructureKind.MAP)
+
+    override fun serialize(encoder: Encoder, value: Map<String, String>) {
+        encoder.encodeString(value.toString())
```
jarohen: I doubt this'll serialize in valid YAML? if it's not used, I'd throw UOE instead - make it really clear :slightly_smiling_face:

danmason: Yeah, that's a fair point - I'll do that, I think they do the same in the `kaml` tests as well :)

PR #3573, opened 2024-08-07
Default unspecified param types to utf8/text

mbutlerw:
Changes prepareQuery to assume the type of any param without an explicit type to be utf8. It does this by comparing the param count of the query to the number param-types its provided, defaulting the difference.

Commit also exposes the calculated paramFields as part of PreparedQuery. This allows the pgwire server to correctly describe the expected type of any such defaulted params.

This is not yet true for DML statements as they do not use the preparedQuery lifecycle, and so require explicit types.

Commit also brings in pg2 as a test dep to allow easier testing of default param types and postgres native params ($1, $2 ..)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -255,6 +255,7 @@ dependencies {
     testImplementation("org.clojure", "test.check", "1.1.1")
     testImplementation("clj-kondo", "clj-kondo", "2023.12.15")
     testImplementation("com.github.seancorfield", "next.jdbc", "1.3.939")
+    testImplementation("com.github.igrishaev", "pg2-core", "0.1.17")
 
     // For generating clojure docs
     testImplementation("codox", "codox", "0.10.8")
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -1094,9 +1094,14 @@
           (read-binary session param)
           (read-text session param))))))
 
-(defn- cmd-exec-dml [{:keys [conn-state] :as conn} {:keys [dml-type query transformed-query params] :as stmt}]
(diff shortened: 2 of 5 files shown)


egg-juxt:
Hi Matt, I have just tested your branch. Unfortunately, defaulting to text doesn't work for all cases. TypeORM will for example build:
```
SELECT "user"."_id" AS "user__id", "user"."firstName" AS "user_firstName", "user"."lastName" AS "user_lastName", "user"."age" AS "user_age" 
FROM "user" "user" 
WHERE "user"."age" >= $1
```

The parameter is expected to be handled as a number, but errors on the xt side:

```
xtdb.expression$emit_projection.invokeStatic(expression.clj:1493) (prepare-expr expr) => 
  {:args [{:args [{:op :variable, :var-type :i64, :variable age}
                  {:op :param, :param ?_0, :param-type :utf8}],
           :f :>=,
           :op :call}],
   :f :boolean,
   :op :call}
10:38:40.738 [pgwire-connection--pool-4-thread-5] ERROR xtdb.pgwire - #error {
 :cause class java.nio.DirectByteBuffer cannot be cast to class java.lang.Number (java.nio.DirectByteBuffer and java.lang.Number are in module java.base of loader 'bootstrap')
 :via
 [{:type java.lang.ClassCastException
   :message class java.nio.DirectByteBuffer cannot be cast to class java.lang.Number (java.nio.DirectByteBuffer and java.lang.Number are in module java.base of loader 'bootstrap')
   :at [clojure.lang.Util compare Util.java 152]}]
 :trace
 [[clojure.lang.Util compare Util.java 152]
  [clojure.core$eval49410$fn__49411 invoke NO_SOURCE_FILE 0]
  [xtdb.expression$$reify__28299 project expression.clj 1555]
  [xtdb.expression$$reify__28308 select expression.clj 1563]
  [xtdb.operator.scan.TrieCursor$fn__33873$fn__33905 invoke scan.clj 305]
 
```

PR #3576, opened 2024-08-08
Refactor ->merge-tasks, introduce protocol MergePlanNode

FiV0:
This PR tries to pull apart the logic of merging tries (meta files) from IO related operations (getting metadata) by making the trie merging logic more pure and testable in isolation. The main motivator for this is https://github.com/xtdb/xtdb/pull/3540, but as the temporal filtering becomes more complicated, having separated phases will help with easier test setups.

We introduce a new protocol `MergePlanNode` that currently only contains `load-page` and `test-metadata` but which can be extended in the future with other methods required for page filtering (for example temporal metadata). This protocol can then be mocked for trie merging and the filters be relaxed and tightened respectively.  

Potentially more refactring work:
- [ ] Check if the compactor merge logic can somehow be unified with the scan merge logic

### DIFF
--- a/src/test/clojure/xtdb/trie_test.clj
+++ b/src/test/clojure/xtdb/trie_test.clj
@@ -1,7 +1,7 @@
 (ns xtdb.trie-test
   (:require [clojure.java.io :as io]
             [clojure.test :as t :refer [deftest]]
-            [xtdb.operator.scan :as scan]
+            [xtdb.operator.scan :as scan :refer [MergePlanPage]]
             [xtdb.test-json :as tj]
             [xtdb.test-util :as tu]
             [xtdb.time :as time]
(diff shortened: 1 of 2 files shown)


src/test/clojure/xtdb/trie_test.clj
```diff
@@ -138,6 +138,22 @@
                                             temporal-bounds)
                     (merge-plan-nodes->path+pages))))))))
 
+(defn ->test-merge-plan-node [page metadata-matches?]
+  (reify MergePlanPage
+    (load-page [_ _ _] page)
```
jarohen: might be worth throwing UOE here, make it clear that this isn't supported for this implementation and confirm that we're not accidentally using it in tests that try to load the page

jarohen (approved):
one minor change, otherwise all good.

good to see the simplicity of the test too - job done :ok_hand:

src/test/clojure/xtdb/trie_test.clj
FiV0: But I am using it in the test just below. Essentially identifying the page (page + segment in the tests to come).

PR #3583, opened 2024-08-09
Simplify TemporalBounds from hypercubes to rectangles

### LINKED ISSUE #3581: Simplify temporal bounds
Temporal bounds are currently 4 temporal columns which are `valid-from`, `valid-to`, `system-from` and `system-to`. Each temporal column is made up of a lower bound and upper bound. This let's us create quite complex hypercubes in the 4 dimensional query space. In practice though we are currently just upper bounding `valid-from` and `system-from` and lower bounding `valid-to` and `system-to`. The ticket proposes to slim down the temporal bounds to just two dimensions `valid-time` and `system-time` with an lower and upper bound respectively. This means that certain queries can't be expressed with temporal bounds anymore (but not all queries could be expressed with the current version). This means some changes to scan and temporal resolution to use the new temporal bounds.

If we decide to push more complex scan predicates on the temporal columns further down, we can always run multiple queries with the slimmed down temporal bounds.

Currently:
TemporalBounds (vf,vt,sf,st) -> TemporalColumn(lower, upper)

Proposal:
TemporalBounds(valid-time, system-time) -> TemporalDimension(lower, upper)

FiV0:
As per https://github.com/xtdb/xtdb/issues/3581 move from:

TemporalBounds (vf,vt,sf,st) -> TemporalColumn(lower, upper)

to

TemporalBounds(valid-time, system-time) -> TemporalDimension(lower, upper)

~For completeness, I also restate the semantics at the boundaries for the temporal dimensions and events. An event is made up of vf, vt, sf and st. vt and st are non inclusive, meaning the event ceases to exists at vt and st respectively. In particular if vf = vt or sf = st then event is non existent. The lower and upper bound of a temporal dimension are inclusive. This means an event with vf = upper and vf < vt gets included and an event with vt = lower does NOT get included. Similarly for sf and st.~

We decided to make the `upper` of `TemporalDimension(lower, upper)` non inclusive. This also means that an event intersection becomes easier as we don't have to take special care of events ending at `lower`. Both events and temporal bounds describe a rectangle where the upper bound of the two time dimensions is non inclusive. 

Todo:
- [x] ~move from `lower/upper` to `min-from, max-to`~ moved back to `lower`/`upper`

### DIFF
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -38,8 +38,9 @@
            xtdb.ICursor
            (xtdb.metadata IMetadataManager ITableMetadata)
            xtdb.operator.SelectionSpec
-           (xtdb.trie ArrowHashTrie$Leaf EventRowPointer EventRowPointer$Arrow HashTrie HashTrieKt LiveHashTrie$Leaf MergePlanNode MergePlanTask)
-           (xtdb.util TemporalBounds TemporalBounds$TemporalColumn)
+           (xtdb.trie ArrowHashTrie$Leaf EventRowPointer EventRowPointer$Arrow HashTrie
+                      HashTrieKt LiveHashTrie$Leaf MergePlanNode MergePlanTask)
+           (xtdb.util TemporalBounds TemporalDimension)
            (xtdb.vector IMultiVectorRelationFactory IRelationWriter IVectorReader IVectorWriter IndirectMultiVectorReader RelationReader RelationWriter)
            (xtdb.watermark ILiveTableWatermark IWatermarkSource Watermark)))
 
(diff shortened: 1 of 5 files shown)


core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -82,34 +82,34 @@
...
                   (case tag
                     :at (let [[at] args
                               at-μs (->time-μs at)]
-                          (.lte start-col at-μs)
-                          (.gt end-col at-μs))
+                          (.lte temporal-dimension at-μs)
+                          (.gte temporal-dimension at-μs))
 
                     ;; overlaps [time-from time-to]
                     :in (let [[from to] args]
```
FiV0: Would be good to clarify what overlap exactly means here. I understand it as the intervals need to have a `valid-from` strictly less than the `to` from `[:in from to]`.

core/src/main/kotlin/xtdb/util/TemporalBounds.kt
```diff
@@ -4,34 +4,30 @@ import java.time.Instant
...
-            && this.validTo.inRange(validTo)
-            && this.systemFrom.inRange(systemFrom)
-            && this.systemTo.inRange(systemTo))
+data class TemporalBounds(
+    val validTime: TemporalDimension = TemporalDimension(),
+    val systemTime: TemporalDimension = TemporalDimension()
+){
+    fun intersects(other: TemporalBounds) = this.validTime.intersects(other.validTime) && this.systemTime.intersects(other.systemTime)
+    fun intersects(validFrom: Long, validTo: Long, systemFrom: Long, systemTo: Long) =
+        (this.validTime.intersects(validFrom, validTo-1) && this.systemTime.intersects(systemFrom, systemTo-1))
```
FiV0: These `-1` are currently here but could also go somewhere else as they are not really `TemporalBounds` related.

Let's say you have the valid-time interval [2 2], which is a point. Intersection would mean we also include a valid-time range of the form [0 2], but we only want to include [0 2+epsilon]. Previously this could easily be modeled by requiring vf [2 inf] and  vt [-inf 2+epsilon] in the temporal bounds.

FiV0: As we moved to non inclusive `maxTo` this conversation is out of data.

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -71,47 +71,44 @@
...
-
-                    :between (let [[from to] args]
-                               (.gt end-col (->time-μs (or from [:now])))
-                               (when-let [to-μs (some-> to ->time-μs)]
-                                 (.lte start-col to-μs)))
+    (let [^TemporalDimension sys-dim (apply-constraint for-system-time)
+          bounds (TemporalBounds. (apply-constraint for-valid-time) sys-dim)]
+      ;; we further constrain bases on tx
+      (when-let [system-time (some-> at-tx (.getSystemTime) time/instant->micros)]
+        (.setMaxTo sys-dim (min (inc system-time) (.getMaxTo sys-dim)))
```
FiV0: This is now the only location where a bit more fiddly logic is required. We first create TemporalBounds based on `for-system-time` (if existant) and then further constrain based on `at-tx`.

PR #3590, opened 2024-08-13
Allow specification of full endpoints for storageaccount/service bus namespace

### LINKED ISSUE #3586: Ensure can pass full storageaccount/servicebus endpoint into Azure Blob Storage
We currently take in the storage account/service bus namespace **names** and construct the URLs/endpoints within the Azure module, ie, 
```
(cond-> (-> (BlobServiceClientBuilder.)
            (.endpoint (str "https://" storage-account ".blob.core.windows.net"))
            (.credential credential)
            (.buildClient)))
```

This causes issues when used inside of private networks, where you need some ability to pass in the full private endpoint, so we should offer a mechanism to configure/provide this.

danmason:
Resolves #3586

Makes `storageAccount` and `serviceBusNamespace` optional (defaulted to `null`) and allows passing in of `storageAccountEndpoint` and `serviceBusNamespaceFQDN`. At least one of each type of parameter needs to be specified, otherwise we throw an IAE. 

Allows users to explicitly provide the relevant endpoints which will be used directly by the APIs instead of constructing endpoints/FQDNs from the passed in names.

### DIFF
--- a/docs/src/content/docs/config/storage/azure.adoc
+++ b/docs/src/content/docs/config/storage/azure.adoc
@@ -84,17 +84,29 @@ storage: !Remote
   objectStore: !Azure
     # -- required
 
-    # The storage account which has the storage container
+    # --- At least one of storageAccount or storageAccountEndpoint is required
+
+    # The name of the storage account which has the storage container
     # (Can be set as an !Env value)
     storageAccount: storage-account
+
+    # The full endpoint of the storage account which has the storage container
+    # (Can be set as an !Env value)
+    # storageAccountEndpoint: https://storage-account.privatelink.blob.core.windows.net
     
     # The name of the blob storage container to be used as the object store
     # (Can be set as an !Env value)
     container: xtdb-container
 
-    # The service bus namespace which contains the serviceBusTopicName
+    # --- At least one of serviceBusNamespace or serviceBusNamespaceFQDN is required
+
+    # The name of the service bus namespace which contains the serviceBusTopicName
     # (Can be set as an !Env value)
     serviceBusNamespace: xtdb-service-bus
+
+    # The fully qualified domain name of the service bus namespace which contains the serviceBusTopicName
+    # (Can be set as an !Env value)
+    # serviceBusNamespaceFQDN: xtdb-service-bus.privatelink.servicebus.windows.net
 
     # The name of the service bus topic which is collecting notifications from the container
     # (Can be set as an !Env value)
--- a/modules/azure/src/main/clojure/xtdb/azure.clj
+++ b/modules/azure/src/main/clojure/xtdb/azure.clj
@@ -4,6 +4,7 @@
             [xtdb.azure.log :as tx-log]
             [xtdb.azure.object-store :as os]
             [xtdb.buffer-pool :as bp]
+            [xtdb.error :as err]
             [xtdb.log :as xtdb-log]
             [xtdb.node :as xtn]
             [xtdb.time :as time]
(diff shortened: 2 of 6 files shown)


modules/azure/src/main/kotlin/xtdb/api/storage/AzureBlobStorage.kt
```diff
@@ -88,18 +90,22 @@ object AzureBlobStorage {
...
      * @property prefix A file path to prefix all of your files with - for example, if "foo" is provided all xtdb files will be located under a "foo" directory.
+     * @property userManagedIdentityClientId The client ID of the user managed identity to use for authentication, if applicable
+     * @property storageAccountEndpoint The full endpoint of the [storage account](https://learn.microsoft.com/en-us/azure/storage/common/storage-account-overview) which has the [container] to be used as an object store
+     * @property serviceBusNamespaceFQDN The fully qualified domain name of the [Service Bus namespace](https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-messaging-overview#namespaces) which contains the [serviceBusTopicName] collecting notifications from the [container]
      */
     @Serializable
     @SerialName("!Azure")
     data class Factory(
-        val storageAccount: String,
+        val storageAccount: String? = null,
```
danmason, line 100: Feel like the kotlin here is a bit confusing, since we're making a required parameter NOT required - so comments appreciated here/if theres a better way to go about this.

docs/src/content/docs/config/storage/azure.adoc
```diff
@@ -84,18 +84,30 @@ storage: !Remote
...
+    # storageAccountEndpoint: https://storage-account.privatelink.blob.core.windows.net
     
     # The name of the blob storage container to be used as the object store
     # (Can be set as an !Env value)
     container: xtdb-container
 
-    # The service bus namespace which contains the serviceBusTopicName
+    # --- At least one of serviceBusNamespace or serviceBusNamespaceFQDN is required
+
+    # The name of the service bus namespace which contains the serviceBusTopicName
```
danmason, line 103: To note, the fact one is a full URL/endpoint and the other is a fully-qualified domain name is more of a decision imposed by Azure and it's APIs

modules/azure/src/main/clojure/xtdb/azure.clj
```diff
@@ -21,34 +22,43 @@
...
+    storage-account-endpoint (.storageAccountEndpoint storage-account-endpoint)
+    servicebus-namespace-fqdn (.serviceBusNamespaceFQDN servicebus-namespace-fqdn)))
 
 ;; No minimum block size in azure
 (def minimum-part-size 0)
 
-(defn open-object-store [^AzureBlobStorage$Factory factory]
-  (let [user-managed-identity-id (.getUserManagedIdentityClientId factory)
+(defn open-object-store [^AzureBlobStorage$Factory factory] 
+  (let [storage-account-endpoint (cond
```
jarohen, line 38: would be good to refactor this into the Kotlin so that we verify config ASAP (i.e. less time with potentially incorrect config), but this one can wait til post-merge.

modules/azure/src/main/kotlin/xtdb/api/storage/AzureBlobStorage.kt
jarohen: not that I can think of :slightly_smiling_face:

modules/azure/src/main/clojure/xtdb/azure.clj
danmason: Yeah - there's a fair few things I'd LIKE to do on top of the kotlin config, but I'm not sure how easy the `kaml` serde stuff makes our life here..

PR #3594, opened 2024-08-14
Switch to use AKS for single node benchmark

### LINKED ISSUE #3592: Switch Azure benchmark over to using AKS for single node run
In order to work around limitations with container app volumes we're switching to AKS. We have a working setup for a single node run so should replace the existing single node setup.

As a starting point:
- Remove eventhub kafka setup.
- Terraform infra to setup the kubernetes cluster.
- Kubernetes config to run a single node job using local txLog and azure blob storage.
- Mounting persistent volume claims for the local disk cache.

tggreene:
Use AKS instead of container apps for the single node benchmark run. One caveat is that you may need to comment out the kubernetes resources when destroying the cluster.

Closes #3592

### DIFF
--- a/cloud-benchmark/azure/terraform/main.tf
+++ b/cloud-benchmark/azure/terraform/main.tf
@@ -143,7 +143,6 @@ resource "azurerm_role_assignment" "cloud_benchmark_eventhub_receive" {
   scope                = azurerm_eventhub.cloud_benchmark.id
 }
 
-# Container App Config
 resource "azurerm_container_registry" "acr" {
   name                = "cloudbenchmarkregistry"
   resource_group_name = azurerm_resource_group.cloud_benchmark.name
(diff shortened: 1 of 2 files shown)


cloud-benchmark/azure/terraform/main.tf
```diff
@@ -183,108 +182,153 @@ resource "azurerm_container_app" "cloud_benchmark_single_node" {
...
+    name      = local.service_account_name
+    namespace = kubernetes_namespace.cloud_benchmark[0].metadata[0].name
+  }
+}
 
-      env {
-        name  = "XTDB_AZURE_EVENTHUB_NAMESPACE"
-        value = azurerm_eventhub_namespace.cloud_benchmark.name
-      }
+resource "kubernetes_deployment" "cloud_benchmark" {
```
danmason, line 254: All looking pretty sensible to me so far :)

I was keen for us to try setting up a persistent volume (might be worth a look at some of the google-cloud kubernetes files) given that was one of our big reasons to switching over to AKS.

danmason:
Closing this - opened a new one against my fork, #3602

PR #3604, opened 2024-08-16
Add metadata to trie recency branches

### LINKED ISSUE #3518: Coarser temporal metadata on tries
Currently we have temporal metadata information (min valid-time, min sys-time etc..) only on per page level. 
We can add metadata information per trie as well as for the inner nodes of the trie to rule out whole tries early or shortcircuit when calculating merge tasks.

FiV0:
This is a simple attempt to add metadata to the recency branches. Nothing fancy, just saving the same meta data that exists for pages but for a whole subtree. There are no functional changes, only the format on disk changes as I am not yet making use of this additional meta data.

As the writing for tries happens top to bottom, I currently recalculate metadata on every node (ie slower compaction). An optimization would be to calculate the meta data per page and then do a bottom up merging, but that would have required a lot more changes to how we currently write tries.

This would also require an index bump if it goes in.


core/src/main/kotlin/xtdb/trie/ArrowHashTrie.kt
```diff
@@ -54,6 +57,9 @@ class ArrowHashTrie(private val nodesVec: Vector) :
         override fun recencyNode(idx: Int) =
             checkNotNull(forIndex(path, recencyIdxVec.getInt(startIdx + idx)))
 
+        fun temporalBounds(): TemporalBounds {
+            TODO("Add temporal bounds for earlier pruning!")
```
FiV0, line 61: This is where the new functionality would come in.

FiV0:
We could also potentially slim down the metadata we save on the nodes. I assume the largest increase in size comes from blooms (which are not the first use case I have in mind). Downside is that if we ever want more metadata we need to do an index bump.

FiV0:
Just some first suggestions from @jarohen if I continue this:
- add the metadata at every level, use the page-idx from one level below to merge the current level
- add explicit fields for `xt$system_from`, `xt$valid_from` and `xt$valid_to`.

jarohen:
closing this one, feel free to poke me and re-open it if you come back to it :slightly_smiling_face:

PR #3630, opened 2024-08-23
Move schema to watermark

### LINKED ISSUE #3603: Refactor: move schema to watermark

mbutlerw:
Begins the process of moving the schema object onto the watermark, by first moving the result of  `allTableColNames` from scanEmitter onto the watermark.

PR also removes the need for a zero dependency prepare-ra, by supplying minimal deps in the query-ra test util.

### DIFF
--- a/core/src/main/clojure/xtdb/query.clj
+++ b/core/src/main/clojure/xtdb/query.clj
@@ -200,7 +200,7 @@
                      (let [table-info-at-execution-time (with-open [wm (.openWatermark wm-src)]
                                                           (.scanFields scan-emitter wm
                                                                        (mapcat #(map (partial vector (key %)) (val %))
-                                                                               (scan/tables-with-cols wm-src scan-emitter))))]
+                                                                               (scan/tables-with-cols wm-src))))]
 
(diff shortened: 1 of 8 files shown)


core/src/main/clojure/xtdb/query.clj
```diff
@@ -136,104 +136,101 @@
...
+                  clock (Clock/fixed current-time default-tz)]
+
+              (reify
+                BoundQuery
+                (columnFields [_]
+                  (->column-fields ordered-outer-projection fields))
+                (openCursor [_]
+                  (when relevant-schema-at-prepare-time
+                    (let [table-info-at-execution-time (with-open [wm (.openWatermark wm-src)]
+                                                         (.scanFields scan-emitter wm
```
jarohen, line 199: I'm assuming this is next to go onto the wm?

jarohen (approved):

core/src/main/clojure/xtdb/query.clj
mbutlerw: Yes, I think it should be possible to retire scanFields entirely, replacing it with a schema-map with Fields as leaf entries.

PR #3632, opened 2024-08-23
Add DurationAdd transform

Akeboshiwind:
This will remain undocumented for now.

In a future PR we should add:
- `$Value` and `$Key` child classes like the default kafka connectors
- The ability to work on schema'd data & Structs


Akeboshiwind:
Not going to be used so no point merging atm

PR #3633, opened 2024-08-23
[WIP] Use the V2 field syntax for DurationAdd

Akeboshiwind:
Adds support for the new [V2 field syntax](https://cwiki.apache.org/confluence/display/KAFKA/KIP-821%3A+Connect+Transforms+support+for+nested+structures#KIP821:ConnectTransformssupportfornestedstructures-ReplaceField).

Not sure if this should be merged yet as it doesn't work for `output.field` because the [methods required are private](https://github.com/apache/kafka/blob/771b9576b00ecf5b64ab6e8bedf04156fbdb5cd6/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/field/SingleFieldPath.java).

(To be merged after #3632)

### DIFF
--- a/modules/kafka-connect/build.gradle.kts
+++ b/modules/kafka-connect/build.gradle.kts
@@ -21,6 +21,7 @@ dependencies {
 
     api(kotlin("stdlib-jdk8"))
     implementation("org.apache.kafka:connect-api:3.8.0")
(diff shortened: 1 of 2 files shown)


Akeboshiwind:
Not going to be used so no point merging atm

PR #3635, opened 2024-08-23
Update XTQL links

federico-razzoli:
XTQL links in the README were broken. Replaced them with the most appropriate URL I've found.


refset:
Thank you @federico-razzoli :pray:

PR #3644, opened 2024-08-28
Supporting window functions with `PARTITION BY` and `ORDER BY`

### LINKED ISSUE #3465: Support for ROW_NUMBER window function
Several tools (Metabase, DBeaver) immediately make a metadata query involving a ROW_NUMBER window function to generate a column index within a table.

Options:
1. Implement enough to make the `ROW_NUMBER` window function work
2. Spot the queries that common tools send, return some canned results

Notes - what might (1) involve?
* Window functions can be thought of in two parts: the window definition and the projection function.
  * Window definitions consist of a partition and an ordering - e.g. `PARTITION BY a, b ORDER BY c DESC, d`. These definitions can be specified ad-hoc, or as a named window to be reused.
  * Projection functions are then run on the rows within a window - e.g. `ROW_NUMBER`, `LEAD(x)`.
  * IIUC (just intuition through usage, haven't definitively analysed the spec), a standard SQL 'group-by' can potentially be thought of as a special case of a window function - the window definition only uses the partition side (i.e. no ordering); the projection is one row for the whole partition rather than one output row for every input.
    * I suspect this subtlety (collapsing rows rather than 1:1) might be enough to justify a slightly different operator for window functions? Alternatively, we could consider including it in `:group-by`.

In terms of changes required, (1) might involve:
* [ ] creating a `:window` operator - materializes the input relation, partitions it, sorts the partitions, runs the projection. We'd want the ability to construct multiple window orderings in one operator given we're materializing the input, so I'm imagining a selection vector for each window definition?
* [ ] updating the SQL grammar to parse out window function projections, constructing a `:window` operator

Edit (by @FiV0): 

For a first pass we decided to put out of scope:
- framing
- aggregations

The first minimal thing to run should be something like:
- `ROW_NUMBER() OVER (PARTITION BY X ORDER BY Y)`

FiV0:
This adds window function support for `ROW_NUMBER() OVER (PARTITION BY x ORDER BY z) AS rn`. 

The operator currently looks like
```clj
[:window {:windows {window1 {:partition-cols [a]
                             :order-specs [[b]]
                             :frame {}}}
          :projections [{col-name {:window-name window1
                                   :window-agg (row-number)}}]}
 plan]
```

The algorithm currently works on the whole input relation. This could be parallelized later or the grouping step goes to disk (as `GROUP BY` also works completely in memory this is something for another ticket). We use the same grouping logic from `GROUP BY` to partition the relation. We then sort the whole relation by prepending the grouping vector and then pass `grouping vector`, `sorted indirection vector` and the input relation to the WindowFunctionAggregate to generate some specific aggregate. I have not touched frames in this PR, although I make the assumption of the default behaviour of `BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW`

Things that need further work / thought:
- [x] I currently copy the whole relation (sort does a similar thing), we could try to use some retaining logic.
- [x] The current version supports one window function (although multiple aggregates). My current approach would be to treat window functions separately and only merge them later (in some optimization step). Window functions with different partitions need to run anyway after one another (except if we plan to parallelize). 
  - Let's throw with multiple window functions
- [x] I added a `min`/`max` aggregate (without going through the EE) to test if one can add some more complicated window aggregates easily.
  - Comment it out or put it on a branch.
- [x] Some `order-by` planning logic in the SQL planner is reused. As the planner contains some state, we need to make sure that is correctly used. 
  - Should be ok from @jarohen and @FiV0 understanding.

### DIFF
--- a/core/src/main/clojure/xtdb/operator/window.clj
+++ b/core/src/main/clojure/xtdb/operator/window.clj
@@ -1,0 +1,193 @@
+(ns xtdb.operator.window
+  (:require [clojure.spec.alpha :as s]
+            [xtdb.expression :as expr]
+            [xtdb.logical-plan :as lp]
+            [xtdb.operator.group-by :as group-by]
+            [xtdb.operator.order-by :as order-by]
+            [xtdb.rewrite :refer [zmatch]]
+            [xtdb.types :as types]
+            [xtdb.util :as util]
+            [xtdb.vector.reader :as vr]
+            [xtdb.vector.writer :as vw])
+  (:import (clojure.lang IPersistentMap)
+           (com.carrotsearch.hppc LongLongMap LongLongHashMap)
+           (java.io Closeable)
+           (java.util LinkedList List Spliterator)
+           (org.apache.arrow.memory BufferAllocator)
+           (org.apache.arrow.vector BigIntVector)
+           (xtdb ICursor)
+           (org.apache.arrow.vector.types.pojo Field)
+           (xtdb.operator.group_by IGroupMapper)
+           (xtdb.vector RelationWriter IVectorWriter)))
+
+(s/def ::window-name symbol?)
+
+;; TODO
+(s/def ::frame any?)
+;; TODO assert at least one is present
+(s/def ::partition-cols (s/coll-of ::lp/column :min-count 1))
+(s/def ::window-spec (s/keys :opt-un [::partition-cols ::order-by/order-specs ::frame]))
+(s/def ::windows (s/map-of ::window-name ::window-spec))
+
+
+(s/def ::window-agg-expr
+  (s/or :nullary (s/cat :f simple-symbol?)
+        :unary (s/cat :f simple-symbol?
+                      :from-column ::lp/column)
+        :binary (s/cat :f simple-symbol?
+                       :left-column ::lp/column
+                       :right-column ::lp/column)))
+
+(s/def ::window-agg ::window-agg-expr)
+(s/def ::window-projection (s/map-of ::lp/column (s/keys :req-un [::window-name ::window-agg])))
+(s/def ::projections (s/coll-of ::window-projection :min-count 1))
+
+(defmethod lp/ra-expr :window [_]
+  (s/cat :op #{:window}
+         :specs (s/keys :req-un [::windows ::projections])
+         :relation ::lp/ra-expression))
+
+(comment
+  (s/valid? ::lp/logical-plan
+            '[:window {:windows {window1 {:partition-cols [a]
+                                          :order-specs [[b]]
+                                          :frame {}}}
+                       :projections [{col-name {:window-name window1
+                                                :window-agg (row-number)}}]}
+              [:table [{}]]]))
+
+
+#_{:clj-kondo/ignore [:unused-binding :clojure-lsp/unused-public-var]}
+(definterface IWindowFnSpec
+  (^xtdb.vector.IVectorReader aggregate [^org.apache.arrow.vector.IntVector groupMapping
+                                         ^ints sortMapping
+                                         ^xtdb.vector.RelationReader in-rel]))
+
+#_{:clj-kondo/ignore [:unused-binding :clojure-lsp/unused-public-var]}
+(definterface IWindowFnSpecFactory
+  (^clojure.lang.Symbol getToColumnName [])
+  (getToColumnField [])
+  (^xtdb.operator.window.IWindowFnSpec build [^org.apache.arrow.memory.BufferAllocator allocator]))
+
+#_{:clj-kondo/ignore [:unused-binding]}
+(defmulti ^xtdb.operator.window.IWindowFnSpecFactory ->window-fn-factory
+  (fn [{:keys [f from-name from-type to-name zero-row?]}]
+    (expr/normalise-fn-name f)))
+
+(defmethod ->window-fn-factory :row_number [{:keys [to-name]}]
+  (reify IWindowFnSpecFactory
+    (getToColumnName [_] to-name)
+    (getToColumnField [_] (types/col-type->field :i64))
+
+    (build [_ al]
+      (let [^BigIntVector out-vec (-> (types/col-type->field to-name :i64)
+                                      (.createVector al))
+            ^LongLongMap group-to-cnt (LongLongHashMap.)]
+        (reify
+          IWindowFnSpec
+          (aggregate [_ group-mapping sortMapping]
+            (let [offset (.getValueCount out-vec)
+                  row-count (.getValueCount group-mapping)]
+              (.setValueCount out-vec (+ offset row-count))
+              (dotimes [idx row-count]
+                (.set out-vec (+ offset idx) (.putOrAdd group-to-cnt (.get group-mapping (aget sortMapping idx)) 0 1)))
+              (vr/vec->reader out-vec)))
+
+          Closeable
+          (close [_] (.close out-vec)))))))
+
+
+(deftype WindowFnCursor [^BufferAllocator allocator
+                         ^ICursor in-cursor
+                         ^IPersistentMap static-fields
+                         ^IGroupMapper group-mapper
+                         order-specs
+                         ^List window-specs
+                         ^:unsynchronized-mutable ^boolean done?]
+  ICursor
+  (tryAdvance [this c]
+    (boolean
+     (when-not done?
+       (set! (.done? this) true)
+
+       (try
+         (let [grouping-vec-name "grouping-vec"]
+           ;; TODO we likely want to do some retaining here instead of copying
+           (util/with-open [rel-wtr (RelationWriter. allocator (for [^Field field static-fields]
+                                                                 (vw/->writer (.createVector field allocator))))
+                            ;; TODO check why grouping can be null
+                            ^IVectorWriter grouping-wtr (-> #xt.arrow/field ["grouping-vec" #xt.arrow/field-type [#xt.arrow/type :i32 true]]
+                                                            (.createVector allocator)
+                                                            vw/->writer)]
+
+             (.forEachRemaining in-cursor (fn [in-rel]
+                                            (vw/append-rel rel-wtr in-rel)
+                                            (with-open [group-mapping (.groupMapping group-mapper in-rel)]
+                                              (vw/append-vec grouping-wtr (vr/vec->reader group-mapping)))))
+
+             (let [rel-rdr (vw/rel-wtr->rdr rel-wtr)
+                   sort-mapping (order-by/sorted-idxs (vr/rel-reader (conj (seq rel-rdr) (vw/vec-wtr->rdr grouping-wtr)))
+                                                      (into [[(symbol grouping-vec-name)]] order-specs))]
+               (doseq [^IWindowFnSpec window-spec window-specs]
+                 (.aggregate window-spec (.getVector grouping-wtr) sort-mapping))
+
+               (util/with-open [window-cols (->> window-specs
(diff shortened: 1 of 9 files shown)


core/src/main/clojure/xtdb/operator/window.clj
```diff
@@ -0,0 +1,237 @@
...
+                     (do
+                       (.accept c out-rel)
+                       true)
+                     false))))))
+         (finally
+           (util/try-close group-mapper)
+           (run! util/try-close window-specs))))))
+
+  (characteristics [_]
+    (bit-or Spliterator/DISTINCT Spliterator/IMMUTABLE))
```
jarohen: these aren't necessarily distinct, because you're no longer returning one row per group

core/src/main/clojure/xtdb/operator/window.clj
```diff
@@ -0,0 +1,237 @@
...
+                           (doseq [^IWindowFnSpecFactory factory window-fn-factories]
+                             (.add window-fn-specs (.build factory allocator)))
+
+                           (WindowFnCursor. allocator in-cursor (order-by/rename-fields fields)
+                                            (group-by/->group-mapper allocator (select-keys fields partition-cols))
+                                            order-specs
+                                            (vec window-fn-specs)
+                                            false)
+
+                           (catch Exception e
```
jarohen: `Throwable`, if you're using it for cleanup, because this ensures you clean up your resources no matter what gets thrown. `Exception` if you're catching it, because you usually don't want to catch `Error`s

(Throwables = Exceptions + Errors)

core/src/main/clojure/xtdb/sql.clj
```diff
@@ -12,4 +12,4 @@
                (plan/->logical-plan)
                (vary-meta assoc :param-count (:param-count (meta plan)))
                #_ (doto clojure.pprint/pprint)))))
-      #_util/lru-memoize))
+      util/lru-memoize))
```
jarohen: this probably needs a `;; <<no-commit>>` :wink:

src/test/clojure/xtdb/operator/window_test.clj
```diff
@@ -0,0 +1,122 @@
...
+                                 [::tu/blocks '{a :i64, b :i64} blocks]]
+                                #_{:with-col-types? true})
+                   sort-fn))))]
+
+    (t/is (= [] (run-test '{:partition-cols [a]
+                            :order-specs [[b]]}
+                          '{rn (row-number)}
+                          [[] []])))
+
+    (t/is (= [{:a 1, :b 20, :rn 0}
```
jarohen: I'd probably compare sets here so that the reader can better see/verify what's going on

(or `frequencies` if it's a bag)

src/test/clojure/xtdb/sql_test.clj
```diff
@@ -15,7 +15,7 @@
   ([sql opts] (sql/compile-query sql opts))
   ([sql] (plan-sql sql {})))
 
-(def regen-expected-files? false) ;; <<no-commit>>
+(def regen-expected-files? true) ;; <<no-commit>>
```
jarohen: are your git hooks running? :sweat_smile:

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -1569,6 +1584,50 @@
 
       agg-sym))
 
+  (visitWindowFunctionExpr [{:keys [^Map !windows] :as this} ctx]
```
jarohen, line 1586: for now, can we throw an error here if the user specifies a frame, so that it doesn't fail silently?

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -1569,6 +1584,50 @@
...
+            [window-name window-spec] (-> (.windowNameOrSpecification ctx) (.accept this))]
+        (.put !windows window-sym {:windows {window-name window-spec}
+                                   :projections [{window-sym (-> window-projection
+                                                                 (assoc :window-name window-name))}]})
+
+        window-sym)))
+
+  (visitWindowNameOrSpecification [this ctx]
+    (if-let [_window-name-ctx (.windowName ctx)]
+      (throw (UnsupportedOperationException.))
```
jarohen: can this one have a "TODO: ..." message so that the user knows what isn't supported?

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -1927,20 +2007,22 @@
...
           plan [:intersect l-plan (rename-col-syms r-plan)]]
       (->QueryExpr (if-not (.ALL ctx)
                      [:distinct plan]
                      plan)
                    l-col-syms)))
 
   (visitQuerySpecification [{:keys [out-col-syms order-by-ctx]} ctx]
-    (let [qs-scope (->QuerySpecificationScope scope
+    (let [
+          ;; group-invar-col-tracker (->GroupInvariantColsTracker env qs-scope (HashSet.))
```
jarohen: if this isn't required, can it be removed?

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -1569,6 +1584,50 @@
...
+        window-sym)))
+
+  (visitWindowNameOrSpecification [this ctx]
+    (if-let [_window-name-ctx (.windowName ctx)]
+      (throw (UnsupportedOperationException.))
+      (-> (.windowSpecification ctx) (.accept this))))
+
+  (visitRowNumberWindowFunction [{{:keys [!id-count]} :env} _]
+    (let [window-sym (-> (->col-sym (str "xt$row_number_" (swap! !id-count inc)))
+                         (vary-meta assoc :window-out-sym? true))]
```
jarohen, line 1604: reckon we need to include `:window-out-sym?` in the `->projected-col-expr` exclusion list?

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -1747,6 +1808,25 @@
         [:project (mapv :col-sym projected-cols) plan]
         plan))))
 
+(defn- wrap-windows [plan windows]
+  (let [{:keys [windows] :as window-opts} (apply merge-with into (vals windows))
```
jarohen, line 1815: nit: `(into {} cat (vals windows))`, or similar?

i.e. something that's more obviously getting all the vals into a single map

core/src/main/clojure/xtdb/operator/window.clj
```diff
@@ -0,0 +1,237 @@
...
+                                      (.get group-mapping (aget sortMapping idx))
+                                      (fn [_group-idx cnt] (or (some-> cnt inc) 0)))))))
+
+          (finish [_]
+            (vr/vec->reader out-vec))
+
+          Closeable
+          (close [_] (.close out-vec)))))))
+
+(doseq [op [:min :max]]
```
jarohen: worth chucking this in a `comment` block if it's currently just at 'hack' stage so the next person understands the situation :slightly_smiling_face: 

(I haven't reviewed the impl)

core/src/main/clojure/xtdb/operator/window.clj
```diff
@@ -0,0 +1,237 @@
...
+(defmethod ->window-fn-factory :row_number [{:keys [to-name]}]
+  (reify IWindowFnSpecFactory
+    (getToColumnName [_] to-name)
+    (getToColumnField [_] (types/col-type->field :i64))
+
+    (build [_ al]
+      (let [^BigIntVector out-vec (-> (types/col-type->field to-name :i64)
+                                      (.createVector al))
+            ;; TODO this can become large if groups are small
+            ^HashMap group-to-cnt (HashMap.)]
```
jarohen: if this is a primitive map, let's use something from `hppc` here

jarohen: you might even want an `IntArrayList`, where the position in the list is the group idx

core/src/main/clojure/xtdb/operator/window.clj
```diff
@@ -0,0 +1,237 @@
...
+  (fn [{:keys [f from-name from-type to-name zero-row?]}]
+    (expr/normalise-fn-name f)))
+
+(defmethod ->window-fn-factory :row_number [{:keys [to-name]}]
+  (reify IWindowFnSpecFactory
+    (getToColumnName [_] to-name)
+    (getToColumnField [_] (types/col-type->field :i64))
+
+    (build [_ al]
+      (let [^BigIntVector out-vec (-> (types/col-type->field to-name :i64)
```
jarohen, line 84: I'd now use an `IVectorWriter` here, save you faffing about with `setValueCount` etc below

core/src/main/clojure/xtdb/operator/window.clj
```diff
@@ -0,0 +1,237 @@
...
+
+             (.forEachRemaining in-cursor (fn [in-rel]
+                                            (vw/append-rel rel-wtr in-rel)
+                                            (with-open [group-mapping (.groupMapping group-mapper in-rel)]
+                                              (vw/append-vec grouping-wtr (vr/vec->reader group-mapping)))))
+
+             (let [rel-rdr (vw/rel-wtr->rdr rel-wtr)
+                   sort-mapping (order-by/sorted-idxs (vr/rel-reader (conj (seq rel-rdr) (vw/vec-wtr->rdr grouping-wtr)))
+                                                      (into [[(symbol grouping-vec-name)]] order-specs))]
+               (doseq [^IWindowFnSpec window-spec window-specs]
```
jarohen, line 164: given we're doing this all in one pass (cf. group-by where we call aggregate once for each input block) it might be simpler to combine aggregate and finish in the window operator?

jarohen (changes requested):
Well done @FiV0 :bow: minor changes below, but really glad to see this working, awesome work :relaxed:

Could you also add this to the railroad diagrams in the reference docs as part of this PR?

core/src/main/clojure/xtdb/sql.clj
FiV0: This was something that sneaked in with some previous commits. Updated `main` with `no-commit`.

core/src/main/clojure/xtdb/operator/window.clj
FiV0: Will fix, but this was straight copied from `group-by.clj`.

core/src/main/clojure/xtdb/sql/plan.clj
FiV0: Happy to use something else, but this does something different
```clj
(into {} cat [{:foo [1]} {:foo [2]}])
;; => {:foo [2]}

(apply merge-with into [{:foo [1]} {:foo [2]}])
;; => {:foo [1 2]}
```

core/src/main/clojure/xtdb/operator/window.clj
FiV0: A bit annoying that the group mapper doesn't give you the number of groups.

core/src/main/clojure/xtdb/operator/window.clj
FiV0: But I need the vector later. Are you saying I should then convert to old style?

core/src/main/clojure/xtdb/operator/window.clj
FiV0: Used `LongLongHashMap` from hppc, could use `IntArrayList` if I change the group mapper to return the number of groups.

jarohen:
Merged, c2ef052cec15017236ab6b253624e42c39290403 - thanks @FiV0 🙏

PR #3645, opened 2024-08-28
AKS Monitoring

tggreene:
This is quite a small change, requires the user add instrumentation key manually as this is somewhat a secret.


cloud-benchmark/azure/README.adoc
```diff
@@ -160,3 +160,25 @@ kubectl delete pvc xtdb-pvc-local-caches --namespace cloud-benchmark
 ``` 
 
 NOTE: You do not necessarily _need_ to delete all of the above between runs - you can also change the kubernetes config map to use slightly different directory names (ie, changing bucket prefix, new local-disk-cache directory, etc) to avoid conflicts between runs. 
+
+== Monitoring
+
+To see the benchmark metrics you need to supply an application insights instrumentation key via an environment variable in the `single-node-auctionmark.yaml` or `multi-node-auctionmark.yaml` file as below:
```
danmason, line 221: I think in our case it's probably fine that we manually add the instrumentation key in (not too dissimilar to the SLACK_WEBHOOK_URL) but it might be worth looking into how we might be able to pass/use secrets from AKS (could also allow us to pass in the user managed identity client id) - could be useful as and when we update the templates

PR #3646, opened 2024-08-28
Add support for pg ::regclass type and cast

### LINKED ISSUE #3416: Support Postgres OID alias types
See the [Postgres docs](https://www.postgresql.org/docs/current/datatype-oid.html)

e.g.

```sql
SELECT * FROM pg_attribute WHERE attrelid = 'mytable'::regclass;

-- effectively expands to

SELECT * FROM pg_attribute
  WHERE attrelid = (SELECT oid FROM pg_class WHERE relname = 'mytable');
```

Note: 

> While that doesn't look all that bad by itself, it's still oversimplified. A far more complicated sub-select would be needed to select the right OID if there are multiple tables named mytable in different schemas. The regclass input converter handles the table lookup according to the schema path setting, and so it does the “right thing” automatically. Similarly, casting a table's OID to regclass is handy for symbolic display of a numeric OID.

mbutlerw:
Commit uses the newly available schema value available on watermark to enable casting to and from an approximation of the postgres regclass type. This type represents the OID of a pg class (which is currently synonymous with a table in XTDB), and uniquely identifies that relation via a hash of its namespaced table name (schema-name + table-name). This makes it possible to refer to tables by their name within SQL queries.

Commit does this by exposing the schema as an implicitly available variable to the expression engine.

### DIFF
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -7,7 +7,7 @@
             [xtdb.util :as util]
             [xtdb.vector.reader :as vr]
             [xtdb.vector.writer :as vw])
-  (:import (clojure.lang Keyword MapEntry)
+  (:import (clojure.lang IPersistentMap Keyword MapEntry)
            (java.nio ByteBuffer)
            (java.nio.charset StandardCharsets)
            (java.time Clock Duration Instant LocalDate LocalDateTime LocalTime OffsetDateTime ZoneOffset ZonedDateTime)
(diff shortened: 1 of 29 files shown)


jarohen (approved):

core/src/main/clojure/xtdb/expression.clj
```diff
@@ -1500,6 +1504,7 @@
           (assert return-type (pr-str expr))
           {:!projection-fn (delay
                              (-> `(fn [~(-> rel-sym (with-tag RelationReader))
+                                       ~(-> schema-sym (with-tag IPersistentMap))
```
mbutlerw, line 1507: Having the Kotlin interfaces define this as a Java `Map` but tag it as a Clojure `IPersistentMap` felt correct here. But wanted to flag.

PR #3651, opened 2024-08-28
Moving file listing out to the Log Interface/Kafka.

### LINKED ISSUE #3631: Using another Kafka topic to replace the cloud-specific obj-store file watchers
We currently expect users to set up obj-store notifications on each of the three clouds via SNS/SQS and their equivalents. We already have Kafka in those environments, so let's use a new Kafka file-notification topic to broadcast these updates instead, and simplify the infrastructure requirements.

danmason:
Resolves #3631 

Switches over to using the log interface & kafka for messaging between a cluster:
- Remove previous object store messaging.
- Update documentation around setting all of the object stores up.
- Update tests, sample config and benchmarks.
- Update documentation around the node - want to clear you HAVE to use Kafka for messaging across a cluster of nodes when using RemoteBufferPool.

Have updated the various tests and cloud benchmark tests, and have also ran this against the AWS benchmark to verify it behaves as it should.

What follows is a set of instructions on how to migrate from the current world of file notifications to the new one:

# Migrating File List Watching in an XTDB Cluster to Kafka

Within XTDB, we have a need to track file changes across nodes while using remote storage implementations. As of the most recent update, our mechanism for handling this has changed - as such, what will follow is an explanation of what is currently there, and a guide on migrating this across.

## Context

As we make great use of the list/subsets of the list of files within XTDB (ie, in scanning, indexing), we need access to the full file list across the entire cluster (ie, what is actually on the remote object store, rather than simply what files has the current node written). Without some mechanism of communicating this between nodes in place - we would have to do this by calling out to **list objects** on the object store - which would quickly become computationally (and physically) expensive.  

## Previous Implementation 

When running multiple XTDB nodes in a cluster, pointed at the same remote storage/object store (ie, using our cloud Object Storage for **S3**, **Azure Blobs** and **Google Cloud Storage**), we previously required the user to configure the following:
- Some form of blob notifications to be sent on the cloud when changes were made to files (ie, they were added, deleted, etc).
- Some method of gathering these blob notifications
  - e.g, in AWS, we would send blob messages from the bucket to an SNS topic 
- Some method of reading these blob notifications and keeping a list of active files across the cluster within an XTDB node
  - e.g, in AWS, we would create an SQS queue on starting a node that would receive messages from the SNS topic and adjust a list accordingly.

Setting this up would typically involve:
- Creating any necessary infrastructure on the cloud to support this behaviour.
- Passing in various IDs/endpoints to these services on our node configuration.
- Ensuring the XTDB node had all of the relevant permissions for all of the relevant services.

Generally speaking, this approach does indeed prevent the need for constant "list objects" calls - only requiring a single call to list out objects when initializing the XTDB node (such that it has some idea of what objects have been written previously). This does have the downside of both making XTDB far more complex to deploy, infrastructure wise, and also creates a potentially costly dependency on other services within the cloud.

## New Approach

We've moved to a new approach for handling this - making use of our **Transaction Log** implementation for handling the communication between XTDB nodes. Within an XTDB cluster on the cloud, we currently expect the user to make use of a shared, multi node implementation of the Transaction Log (which we currently provide using **Apache Kafka**). 

As such - we now create two topics when using Kafka as the XTDB transaction log - the `txTopic` for the transaction log itself, and the `filesTopic` for sending file notifications. Each node will send and subscribe to these file notifications, and this forms the basis of our file list.

## Migrating your Cluster

Actionable steps to switch over to the new approach - as it *is* a breaking change:
- (If using a manually provisioned Kafka Topic) - create a new topic for the file notifications. 
  - Can see required & recommended config for this topic on the [Kafka configuration docs.](https://docs.xtdb.com/config/tx-log/kafka.html) 
  - Ensure XTDB has the correct permissions in place to READ, WRITE, and DESCRIBE this topic.
- With the Kafka topic in place (or if the XTDB node has permissions to automatically create Kafka topics), update your XTDB configuration:
  - Remove any usages of the external services from the configuration of your remote Object stores:
    - For **S3**: remove `snsTopicArn`
    - For **Google Cloud Storage**: remove `pubSubTopic`. 
    - For **Azure Blob Storage**: remove `serviceBusNamespace` & `serviceBusTopicName`.
  - Update your configuration for the Kafka log:
    - `topicName` -> `txTopic`.
    - `pollDuration` -> `txPollDuration` (if present)
    - Add the following:
      - `filesTopic` - to point at your file notification topic.
    - Remove the following, if present:
      - `topicConfiguration` - previously used to pass in a map of topic configuration options to use when creating the transaction log topic. If XTDB is used to automatically create the topics, we will configure them with our required settings. If you wish to have a different configuration for your topic, we would recommend creating it manually (as above).
-  Once the XTDB nodes are up and running again, using Kafka for handling file notifications, you can clear up any extraneous infrastructure on your cloud of choice:
  - On AWS, you can remove: 
    - The `SNS` subscription from your S3 bucket
    - The `SNS` topic for file notifications.
    - Any `SQS` queues previously created by XTDB
    - Any permissions for the above.
  - On Google Cloud, you can remove:
    - The subscription on Pub/Sub to any storage notifications from the bucket.
    - The Pub/Sub topic itself.
    - Any permissions for the above.
  - On Azure, you can remove:
    - The `Service Bus` topic.
    - The `Service Bus` namespace. 
    - The `EventGrid` subscription on `Azure Blob Storage` notifications.
    - The `EventGrid` topic.
    - Any permissions for using the above.

If using the sample templates for any of the cloud providers, we have updated these accordingly.


danmason:
We might want to consider the naming of "Transaction Log" as a module, given that in certain cases it's more like "Transaction Log++"

jarohen (approved):

PR #3698, opened 2024-09-14
Improved error messages.

bombaywalla:
Added ::err/message with the expected type where it was missing.

Currently:
```
user> (xt/submit-tx node [:put-docs :my-table {:xt/id :foo}])
Execution error (IllegalArgumentException) at xtdb.serde/iae-reader (serde.clj:254).
Illegal argument: 'xtql/malformed-tx-op'
```


jarohen:
thanks :relaxed:

PR #3704, opened 2024-09-16
Handling, logging and rethrowing Azure SDK errors - updating logback level of `reactor`

### LINKED ISSUE #3703: Handle, log and rethrow all Azure SDK errors within the XTDB module
We want to ensure that at _all_ points where we call the Azure SDK we catch any errors appropriately. For any errors/exceptions that we dont have separate handling code for, we want to catch them, log them out within the xtdb.azure logs and re-throw them. This also allows us to bump the logging on the classes that Azure SDK uses (ie, the azure sdk itself, reactor, http client issues etc) up to `ERROR` such that we have less chatty logs and we're responsible to log any errors/exceptions that are relevant to us. 

- [x] Handle, log and rethrow misc throwables/errors from our Azure SDK calls.
- [x] Update the logback/logging levels of the Azure Benchmark + Azure Docker Image
  - More specifically, want to bump reactor to `ERROR`. 
- [x] Test against nightly azure tests. 
- [ ] Test against a 1Hour benchmark.

danmason:
Resolves #3703 

Should make a lot of the underlying HTTP connection/azure API calls a lot less chatty, and ensures we properly handle/log out any errors that actually matter to us.

Going to run against the Azure Benchmark, ensure it behaves itself.

### DIFF
--- a/cloud-benchmark/src/main/resources/logback.xml
+++ b/cloud-benchmark/src/main/resources/logback.xml
@@ -9,4 +9,5 @@
   </root>
   <logger name="xtdb" level="INFO" />
   <logger name="com.azure" level="ERROR" />
+  <logger name="reactor" level="ERROR" />
 </configuration>
--- a/modules/azure/src/main/clojure/xtdb/azure/object_store.clj
+++ b/modules/azure/src/main/clojure/xtdb/azure/object_store.clj
@@ -15,20 +15,36 @@
            xtdb.api.storage.ObjectStore
            [xtdb.multipart IMultipartUpload SupportsMultipart]))
 
+(defmacro rethrowing-reactor-cause [form]
+  `(try
+     ~form
+     (catch Throwable e#
+       (throw (Exceptions/unwrap e#)))))
+
 (defn- get-blob 
   ([^BlobContainerClient blob-container-client blob-name]
    (try
-     (-> (.getBlobClient blob-container-client blob-name)
-         (.downloadContent)
-         (.toByteBuffer))
+     (rethrowing-reactor-cause
+      (-> (.getBlobClient blob-container-client blob-name)
+          (.downloadContent)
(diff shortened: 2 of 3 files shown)


cloud-benchmark/src/main/resources/logback.xml
```diff
@@ -9,4 +9,6 @@
   </root>
   <logger name="xtdb" level="INFO" />
   <logger name="com.azure" level="ERROR" />
+  <logger name="reactor" level="ERROR" />
+  <logger name="reactor.netty" level="ERROR" />
```
jarohen: shouldn't need `reactor.netty` here - Java logging forms a hierarchy, so if `reactor.netty` isn't specified it defaults to `reactor`

modules/azure/src/main/clojure/xtdb/azure/object_store.clj
```diff
@@ -24,7 +24,15 @@
      (catch BlobStorageException e
        (if (= 404 (.getStatusCode e))
          (throw (os/obj-missing-exception blob-name))
-         (throw e)))))
+         (throw e)))
+     (catch Throwable e
+       (let [unwrapped-e (Exceptions/unwrap e)]
```
jarohen: we have a macro, `util/rethrowing-cause` which does this unwrapping - if you wrap the code on L21-23 with that, you can then use a standard `catch` clause to check for `InterruptedException`

jarohen (approved):
couple of small recommendations; otherwise, looks fine :relaxed:

modules/azure/src/main/clojure/xtdb/azure/object_store.clj
danmason: Wrote up a new one in `azure.clj` - `rethrowing-reactor-cause` :)

PR #3707, opened 2024-09-17
Tag metrics with node id, add config for monitoring across nodes on Azure

### LINKED ISSUE #3701: Ensure all prometheus metrics are available from Azure AKS benchmark
Want to add whatever config/setup we need such that we can log out all of our relevant prometheus metrics from the Azure AKS benchmark.
- [x] Should write up any relevant instructions necessary for setting it up.
- [x] Want to ensure we can distinguish between nodes - probably want to look at "dimensions" for this.
- [x] Auctionmark benchmark gauges available.
- [x] JVM gauges available.
- [x] Setup a basic dashboard of things we care about in debugging the benchmark.

danmason:
Resolves #3701 

- Adds a common tag to all reported metrics, `node-id`. 
  - This is currently either grabbed from an environment variable or generated with a short uuid suffix.
  - We can use this `node-id` to filter metrics within azure.
  -  We use the env var to easily identify bench nodes 
- Setup a dashboard on azure application insights with relevant information for debugging auctionmark.
  - We split the above across nodes using the new node tag
  - Committed the current template within the repo - we could likely make a general purpose one for people to use too.
- Update instructions for setup.  

Preview of the dashboard on application insights:
image

### DIFF
--- a/cloud-benchmark/azure/README.adoc
+++ b/cloud-benchmark/azure/README.adoc
@@ -218,6 +218,12 @@ There is a helper script to do _all_ of the above, which can be run by:
 
 == Monitoring
 
+=== Setup
+
+To monitor the benchmark, we can use Azure Application Insights. We set up an applications insights resource in the terraform configuration, and output an instrumentation key. 
+
+=== Collecting metrics from node
+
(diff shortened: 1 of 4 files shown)


tggreene (approved):
Looks good to me

PR #3717, opened 2024-09-19
Add metric gauges for child-allocators in XTDB.

### LINKED ISSUE #3702: Add gauges for monitoring memory usage of XTs child allocators
Within arrow we use a number of "child allocators" for direct memory operations - different components in XT have their own child allocators associated with it which we can use to keep track of what is allocating which memory.

Want to create a gauge for each of these child allocators to monitor node direct memory and where it is being allocated:
- [x] Buffer Pool 
- [x] Compactor
- [x] Indexer
- [x] Log Watcher
- [x] Live Index

danmason:
Resolves #3702 

Allows us to keep track of allocated direct memory on a per child allocator basis:
- Needed to pass `metrics/registry` down to a lot of the XTDB modules, so makes a dependency between the two (might allow for further monitoring gauges, if appropriate).
  - More complex in buffer pool but not dissimilar to us passing down the "file cache"/log.
- Passes tests and have tested against `azure_test`, seems to work well.
- Gauges getting added as expected, have seen these looking at `localhost:8080` on the dev node.

### DIFF
--- a/core/src/main/clojure/xtdb/metrics.clj
+++ b/core/src/main/clojure/xtdb/metrics.clj
@@ -11,6 +11,7 @@
            (java.util.function Supplier)
            java.util.List
            (java.util.stream Stream)
+           (org.apache.arrow.memory BufferAllocator)
            (xtdb.api Xtdb$Config)
            (xtdb.api.metrics Metrics Metrics$Factory PrometheusMetrics$Factory)))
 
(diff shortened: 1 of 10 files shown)


core/src/main/clojure/xtdb/metrics.clj
```diff
@@ -42,6 +43,9 @@
        (cond-> (:unit opts) (.baseUnit (str (:unit opts))))
        (.register reg))))
 
+(defn add-allocator-gauge [reg meter-name ^BufferAllocator allocator]
+  (add-gauge reg meter-name (fn [] (.getAllocatedMemory allocator)) {:unit "bytes"}))
```
danmason, line 47: If there were other metrics we were interested in here could fairly easily replace "meter-name" with "allocator-name" or something similar and register multiple different gauges here. `AllocatedMemory` seemed the most obviously useful to us as we move forward with stability work.

jarohen (approved):
sweet :ok_hand:

PR #3718, opened 2024-09-19
Add support for `rank`, `dense-rank` and multiple window functions to window operator

FiV0:
This adds:
- use `hccp.LongArrayList` instead of `LongLongHashMap` in `row-number` and all subsequent functions
- support for the `rank` and `dense-rank` functions
- multiple window functions

Previously the window columns would get filled per the order the rows were coming into the operator and we would sort the incoming relation based on the window sorting. With multiple window functions this approach became a mess (I tried). Instead we now fill the window column according to the sorting and then all window columns can just be appended to the relation coming into the window operator.

### DIFF
--- a/core/src/main/clojure/xtdb/operator/order_by.clj
+++ b/core/src/main/clojure/xtdb/operator/order_by.clj
@@ -54,22 +54,25 @@
       (doto (ArrowFileWriter. root nil (Channels/newChannel os))
         (.start) (.writeBatch) (.end)))))
 
+(defn order-specs->comp ^Comparator [^RelationReader read-rel, order-specs]
+  (reduce (fn [^Comparator acc, [column {:keys [direction null-ordering]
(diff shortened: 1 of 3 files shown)


jarohen (approved):
Thanks @FiV0, this looks great :ok_hand: With some additional E2E (SQL) tests I'd be happy to merge as the operator tests seem sound, but it'd be good to get a walk-through of how it works - Monday during/after the showcase?

FiV0:
> I'd be happy to merge as the operator tests seem sound, but it'd be good to get a walk-through of how it works - Monday during/after the showcase?

Sound good to me

jarohen:
(discussed) not going to wait on the SQL planning side, will submit a separate card

jarohen:
backed out the third commit here, to be re-submitted

FiV0:
Test

PR #3719, opened 2024-09-19
Add Bitwise Operators

### LINKED ISSUE #3714: Support bitwise operators
(edited) Query from metabase:
```sql
SELECT 
  tmp.TABLE_CAT, 
  tmp.TABLE_SCHEM, 
  tmp.TABLE_NAME, 
  tmp.NON_UNIQUE, 
  tmp.INDEX_QUALIFIER, 
  tmp.INDEX_NAME, 
  tmp.TYPE, 
  tmp.ORDINAL_POSITION, 
  /*trim(
    both '"' 
    from 
      pg_catalog.pg_get_indexdef(
        tmp.CI_OID, tmp.ORDINAL_POSITION, 
        false
      )
  ) AS COLUMN_NAME, */
  CASE tmp.AM_NAME WHEN 'btree' THEN CASE tmp.I_INDOPTION[tmp.ORDINAL_POSITION - 1] & 1 :: smallint WHEN 1 THEN 'D' ELSE 'A' END ELSE NULL END AS ASC_OR_DESC, 
  tmp.xCARDINALITY, 
  tmp.PAGES, 
  tmp.FILTER_CONDITION 
FROM 
  (
    SELECT 
      NULL AS TABLE_CAT, 
      n.nspname AS TABLE_SCHEM, 
      ct.relname AS TABLE_NAME, 
      NOT i.indisunique AS NON_UNIQUE, 
      NULL AS INDEX_QUALIFIER, 
      ci.relname AS INDEX_NAME, 
      CASE i.indisclustered WHEN true THEN 1 ELSE CASE am.amname WHEN 'hash' THEN 2 ELSE 3 END END AS TYPE, 
      (
        information_schema._pg_expandarray(i.indkey)
      ).n AS ORDINAL_POSITION, 
      ci.reltuples AS xCARDINALITY, 
      ci.relpages AS PAGES, 
      pg_catalog.pg_get_expr(i.indpred, i.indrelid) AS FILTER_CONDITION, 
      ci.oid AS CI_OID, 
      i.indoption AS I_INDOPTION, 
      am.amname AS AM_NAME 
    FROM 
      pg_catalog.pg_class ct 
      JOIN pg_catalog.pg_namespace n ON (ct.relnamespace = n.oid) 
      JOIN pg_catalog.pg_index i ON (ct.oid = i.indrelid) 
      JOIN pg_catalog.pg_class ci ON (ci.oid = i.indexrelid) 
      JOIN pg_catalog.pg_am am ON (ci.relam = am.oid) 
    WHERE 
      true 
      --AND n.nspname = E 'public' 
      --AND ct.relname = E 'users'
  ) AS tmp 
ORDER BY 
  NON_UNIQUE, 
  TYPE, 
  INDEX_NAME, 
  ORDINAL_POSITION
```

Throws this error:
```
Errors parsing SQL statement:
  - line 18:84 token recognition error at: '&'
```

Akeboshiwind:
Fixes #3714

### DIFF
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -184,6 +184,19 @@
          :decimal 'bigdec}
         types/col-type-head))
 
+(def type->unchecked-cast
+  (comp {:bool 'boolean
+         :i8 'unchecked-byte
+         :i16 'unchecked-short
+         :i32 'unchecked-int
+         :i64 'unchecked-long
+         :f32 'unchecked-float
+         :f64 'unchecked-double
+         :timestamp-tz 'unchecked-long
+         :duration 'unchecked-long
+         :decimal 'bigdec}
+        types/col-type-head))
+
 (def idx-sym (gensym 'idx))
 (def rel-sym (gensym 'rel))
 (def schema-sym (gensym 'schema))
@@ -755,12 +768,9 @@
   {:return-type (types/least-upper-bound arg-types)
    :->call-code #(do `(* ~@%))})
 
-(defmethod codegen-call [:bit_not :int] [{:keys [arg-types]}]
-  (let [return-type (types/least-upper-bound arg-types)]
-    {:return-type return-type
-     :->call-code (fn [emitted-args]
-                    (list (type->cast return-type)
-                          `(bit-not ~@(with-math-integer-cast return-type emitted-args))))}))
+(defmethod codegen-call [:bit_not :int] [{[x-type] :arg-types}]
+  {:return-type x-type
+   :->call-code #(do `(bit-not ~@%))})
 
 (defmethod codegen-call [:bit_and :int :int] [{:keys [arg-types]}]
   (let [return-type (types/least-upper-bound arg-types)]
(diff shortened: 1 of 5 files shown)


core/src/main/clojure/xtdb/expression.clj
```diff
@@ -755,6 +755,48 @@
   {:return-type (types/least-upper-bound arg-types)
    :->call-code #(do `(* ~@%))})
 
+(defmethod codegen-call [:bit_not :int] [{:keys [arg-types]}]
+  (let [return-type (types/least-upper-bound arg-types)]
```
jarohen: with bit-not, you don't need the LUB of the arg-types - there's only one.

core/src/main/clojure/xtdb/expression.clj
```diff
@@ -755,6 +755,48 @@
...
+
+(defmethod codegen-call [:bit_xor :int :int] [{:keys [arg-types]}]
+  (let [return-type (types/least-upper-bound arg-types)]
+    {:return-type return-type
+     :->call-code (fn [emitted-args]
+                    (list (type->cast return-type)
+                          `(bit-xor ~@(with-math-integer-cast return-type emitted-args))))}))
+
+(defmethod codegen-call [:bit_shift_left :int :int] [{:keys [arg-types]}]
+  (let [return-type (types/least-upper-bound arg-types)]
```
jarohen: I'd be tempted (but check Postgres) to just use the type of the left-hand argument here - the difference it makes is in overflow handling.

edit: checked Postgres: `SELECT 1::smallint << 16::int;` => 0; I'd guess it gives 2^16 for us as it stands?

jarohen (changes requested):
Nice one @Akeboshiwind :clap: just some pedantry around the type widening, otherwise good to go

Akeboshiwind:
As discussed on Slack it turns out Postgres does weird things with the shift operators:

```
SELECT 3::smallint << 16; => 0
SELECT 3::smallint << 32; => 3
SELECT 3::smallint << 31; => 0

SELECT 3::int << 32; => 3
SELECT 3::int << 31; => -2147483648

SELECT 3::int << 38; => 48
SELECT 3::int << 12 << 12 << 12; => 0
```

Basically it looks like Postgres does `mod(n, 32)` on the rhs.

We decided to leave this as the JVM default for now (which means always truncating) until some tool eventually requires it 😅

PR #3720, opened 2024-09-19
Consolidating on pgwire for non-Clojure users

jarohen:
Now that our Postgres wire-compatibility is at a decent level with moderate tooling support, we should consolidate on pgwire at this stage of the beta, to avoid maintaining multiple APIs going into GA.

This PR:

* removes JSON support on the HTTP API - the HTTP API now only supports Transit/JSON, for the Clojure remote client. This meant a significant code reduction in the serdes for HTTP requests/responses.
* removes the HTTP-based JS/Python SDKs - these users can now interact with XTDB through standard PostgreSQL drivers.

Next up, we can consider deprecating/removing the Java/Kotlin APIs too (again, in favour of PostgreSQL tooling).

We'd like to preserve the ability for these users to easily and quickly create transient in-memory nodes. Rather than direct API access, though, we can start up the PG server and give users a temporary connection string - either isolated per-connection on a single port, or a new port for each node.


jarohen:
Closing this for now, deferring the decision to evaluate demand.

If you are intending to use the HTTP JSON API for non-Clojure use-cases (as opposed to PostgreSQL compatible tooling), please do get in touch through our usual channels. Thanks!

PR #3728, opened 2024-09-24
Update docker tag in  installation-via-docker.adoc

aangiel:
Adds 'nightly' tag to docker image, because it couldn't find manifest.


jarohen:
Thanks @aangiel :relaxed: 

I'll also add a note about the beta1 tag :ok_hand:

jarohen:
Ah, or rather, I've pushed the `latest` tag s.t. the original now works :laughing: 

In any event, will still add the notes about the different tags.

Issue #3411, opened 2024-06-11
Allowing users to specify schema type constraints on a per-keypath basis


jarohen:
XTDB has always been a schema-less database, and we intend it to remain as such. Particularly, the value we see from this is two-fold:

* It's not always possible to fully understand your schema ahead-of-time - in non-trivial projects, you learn more about it as you go, and we value flexibility here.
* Specifying schema is tedious and onerous.

Secondary to that is the nature of the schema that one must specify in a traditional SQL database. Thinking of a product catalogue use case, for example, product data is very heterogeneous - all products will likely have a title and description, some will have dimensions/colours, some will have data like energy efficiency ratings, clothing might be XS/XL/age 4-5/size 10 (US 8)/34R etc. Real-world data (much as some data architects might try!) rarely fits into a single-size box - we value the DX of working with flexible documents rather than strict, pre-modelled rows.

That said, on the other end of the spectrum is a fully-dynamic system where anything goes. This has its own tradeoffs, especially in a database (and, within that, especially in a bitemporal database) - you only need one erroneous string in an otherwise int column, for example, to cause pain on the read-side.

there-must-be-a-better-way.png

We propose, in a similar manner to Clojure's [spec](https://clojure.org/about/spec), that:

* It's common for composites (e.g. structs, documents, 'products') to have varying key-sets.
* At the same time, it's rarely intentional for an individual key to either change type, or have multiple types.

As such, we will associate a type (well, two - see below) with each key-path - either a top-level column, a nested key within a struct, or the type of a list's elements - while leaving the top-level documents and nested structs to contain any number of those columns/keys.

Spec:

* Each key-path will have a 'write-type' and a 'read-type'.
  * The 'read-type' is the data actually present in the column, over all time - e.g. if a column is mainly floats but has some strings for historical reasons, this type will be `Union<Float, Str>`. This will be used to determine the result types of queries.
    * Internally, we essentially maintain this already.
  * The 'write-type' of a key-path is the current type-constraint on that key-path - if a transaction attempts to insert a different type, it will be aborted.
* The 'write-type' is set on the first insertion to each key-path.
  * We could have chosen (indeed, still could choose) to start flexible and allow the user to narrow it later - we decided on balance that it's more likely for a column to be monomorphic, and that users can explicitly opt-in to polymorphism if required.
* Users should be able to change the write-type using SQL-standard `ALTER TABLE` syntax.
[...]

PR #3747, opened 2024-09-26
Try to infer pg-wire read functions from DML statements type hints

### LINKED ISSUE #3622: Add support for pg param type hints `$1::type`
Postgres allows the users to type hint parameters within SQL strings, which effectively declares the incoming type of that param. This is especially useful for cases/clients without an additional way to specify param type information.

`INSERT INTO my_table (name, created_at) VALUES ('John Doe', $1::timestamp);`

Note that while it shares its syntax with a runtime type cast, my understanding is that this instead assigns the type of the param as outlined in Type Casts https://www.postgresql.org/docs/8.3/sql-expressions.html
> A cast applied to an unadorned string literal represents the initial assignment of a type to a literal constant value

The param types should be discovered as a result of the parse command, and be available from that point onwards. This means for DML we will likely need to parse/prepare the query string, for purposes of returning parameter descriptions and decoding incoming params, even though we will then place the original SQL string back on the wire. 

Further Reading:
https://www.postgresql.org/docs/8.3/sql-syntax-lexical.html#SQL-SYNTAX-CONSTANTS-GENERIC

TODO:

- [ ] Uncomment test case written as part of #3572

FiV0:
This is an attempt (rather hacky) to infer read functions for parameters if the client doesn't send us parameter type information. An example of such a client is `node-postgres`. For clients that send us type information nothing should change as that type information (except for an undefined type) takes priority. So the order is:
- type information send by the client (other than undefined)
- type information inferred from `::INT`, `::FLOAT` ... etc..
- undefined type

 Just be aware that something like `INSERT INTO docs(_id, foo) VALUES ($1::INT, $2::FLOAT)` will now try to read an int and a float for parameters when using `node-postgres`. If the users genuinely wanted to send us strings that should be cast to int and float respectively, there might be subtle differences (I don't think for these two types) when going through the EE instead of reading params. This last issue definitely is a problem when using binary encoding (which `node-postgres` doesn't support).
 
 We are still rejecting DML if no parameter information can be inferred.

TODO:
- [ ] test other DML statements
- [ ] make it work for queries?
- [ ] fix reflection warnings (there are a lot because of `proxy`)

### DIFF
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -11,7 +11,7 @@
             [xtdb.node.impl]
             [xtdb.query]
             [xtdb.serde :as serde]
-            [xtdb.sql.plan :as plan]
+            [xtdb.sql.plan :as plan :refer [->CastArgsVisitor]]
             [xtdb.time :as time]
             [xtdb.types :as types]
             [xtdb.util :as util])
@@ -30,7 +30,7 @@
            (org.antlr.v4.runtime ParserRuleContext)
            [org.apache.arrow.vector PeriodDuration]
            org.postgresql.util.PGobject
-           (xtdb.antlr SqlVisitor Sql$DirectlyExecutableStatementContext)
+           (xtdb.antlr SqlVisitor SqlBaseVisitor Sql$DirectlyExecutableStatementContext Sql$ParamExprContext)
            [xtdb.api ServerConfig Xtdb$Config]
            xtdb.api.module.XtdbModule
            xtdb.IResultCursor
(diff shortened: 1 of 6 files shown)


core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -292,46 +299,69 @@
...
           (.accept (plan/parse-statement sql-trimmed)
-                   (reify SqlVisitor
-                     (visitDirectSqlStatement [this ctx] (.accept (.directlyExecutableStatement ctx) this))
-
-                     (visitSetSessionVariableStatement [_ ctx]
+                   (proxy
+                       [SqlBaseVisitor]
+                       []
+                     (defaultResult [] {})
+                     (aggregateResult [aggregate nextResult] (merge aggregate nextResult))
```
FiV0, line 372: It is potentially expensive to walk the tree twice.

FiV0:
Let's try to infer the types for queries via standard planning.

FiV0:
We decided against this approach in favor of something like #3411.

PR #3750, opened 2024-09-26
Minor fix in the SQL Quickstart docs

njprov:
Hey! Hope you don't mind the drive-by PR for a minor fix in the docs. I'm happy send this some other way if you'd prefer. I'm working through the SQL Quickstart and noticed that the psql version of the "When did you know it?" section is missing the `INSERT` command. PS I like the SQL interface in the v2 changes. I saw yall at HYTRADBOI a few years ago but didn't get very far into testing it out at the time. Starting with SQL is easier for me.

### DIFF
--- a/docs/src/content/docs/quickstart/sql-overview.mdx
+++ b/docs/src/content/docs/quickstart/sql-overview.mdx
@@ -532,6 +532,9 @@ INSERT INTO people (_id, name, favorite_color, _valid_from)
   </Fiddle>
   <Card slot="2">
 ```sql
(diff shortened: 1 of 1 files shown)


njprov:
Also, the next section, yesterday's report with today's data, has different queries in the XT Play and psql tabs. I'm not sure which one is intended.

refset:
Hey @njprov much appreciated, thank you for taking such a thorough look! :pray: 

> I'm not sure which one is intended.

Definitely the XT Play tab :sweat_smile: I'll push a fix for that now also

PR #3767, opened 2024-10-08
SET ROLE statement accepted with no effect

### LINKED ISSUE #3757: Have `SET ROLE ...` as a separate grammer statement
As part of https://github.com/xtdb/xtdb/issues/3745, I relaxed the grammar to allow `SET parameter value`. That should probably only be available to `SET ROLE`.

https://www.postgresql.org/docs/current/sql-set-role.html
https://www.postgresql.org/docs/current/sql-set.html

Related: https://github.com/xtdb/xtdb/issues/3626

egg-juxt:


core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1554,6 +1556,7 @@
       :canned-response (cmd-write-canned-response conn canned-response)
       :set-session-parameter (cmd-set-session-parameter conn parameter value)
       :set-session-characteristics (cmd-set-session-characteristics conn session-characteristics)
+      :set-role nil
```
egg-juxt, line 1559: @mbutlerw Is this the right way to add a no-op statement, for which we don't have an implementation yet?

mbutlerw: Not sure there's an established pattern yet, this seems like a reasonable approach for now.

mbutlerw (approved):

FiV0 (approved):
LGTM. We should probably document this at some point, but that can probably be done in https://github.com/xtdb/xtdb/issues/3626

PR #3776, opened 2024-10-10
Lower level pg-wire tests

### LINKED ISSUE #3653: Add error handling around reading of params in pgwire
We should catch and return a pg error in the case we are unable to parse/read params in `xtify-param`

FiV0:
This mainly adds a namespace `pgwire-protocol-test` that lets you more easily test protocol violations and lower level features of our pgwire server. For example, the last commit tests if a driver sends us a parameter of claimed type x, but the given text is not parsaeble for that type.

### DIFF
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -14,7 +14,7 @@
             [xtdb.time :as time]
             [xtdb.types :as types]
             [xtdb.util :as util])
-  (:import [clojure.lang PersistentQueue]
+  (:import [clojure.lang PersistentQueue ExceptionInfo MapEntry]
            [java.io ByteArrayInputStream ByteArrayOutputStream Closeable DataInputStream DataOutputStream EOFException IOException InputStream OutputStream PushbackInputStream]
            [java.lang Thread$State]
            [java.net ServerSocket Socket SocketException]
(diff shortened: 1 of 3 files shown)


src/test/clojure/xtdb/pgwire_protocol_test.clj
```diff
@@ -0,0 +1,345 @@
...
+    ;; Reserve 4 bytes for the message length (to be set later)
+    (.writeInt startup-message 0)
+    ;; this is the normal 'hey I'm a postgres client' if ssl is not requested
+    (.writeInt startup-message 196608)
+    (write-string-parameter startup-message "user" user)
+    (write-string-parameter startup-message "database" database)
+    (.write startup-message 0)
+    (let [^bytes message-bytes (.toByteArray startup-message-ba)
+          length (alength message-bytes)]
+      (System/arraycopy (int-to-bytes length) 0 message-bytes 0 4)
```
FiV0, line 103: I tried to rewrite this function in the style of all the others, but something goes awefully wrong in `reac-c-string` on the server side. I haven't fully figured out what is wrong. It is probably good to debug it, but didn't want to blog this PR on it.

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1630,6 +1664,12 @@
             #'msg-password nil
 
             (cmd-send-error conn (err-protocol-violation "unknown client message"))))))
+    (catch ExceptionInfo e
```
FiV0: Not sure this is actually the right place to catch things. Probably good to talk through a general error handling approach for these kind of errors.

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -650,17 +670,24 @@
...
+   :message \M
+   :detail \D
+   :position \P
+   :where \W})
+
 (def ^:private io-error-notice-field
   "An io-data type that writes a (vector/map-entry pair) k and v as an error field."
-  {:read no-read
+  {:read (fn read-error-or-notice-field [^DataInputStream in]
+           (let [field-key ((clojure.set/map-invert error-or-notice-type->char) (char (.readByte in)))]
```
mbutlerw: Alias cleanup.

FiV0:
The last commit actually has the same issue as the other `sql.Timestamp` test in that it depends on the the region it's being run.

PR #3778, opened 2024-10-10
rewrite != to <> in lexer #3773

### LINKED ISSUE #3773: != does not work where <> does - they should be equivalent
works OK:
```
xtdb=> select 1 from pg_type where typrelid <> 0;
 _column_1
-----------
(0 rows)
```
fails:
```
xtdb=> select 1 from pg_type where typrelid != 0;
ERROR:  != not applicable to types i32 and i64
```

This is from a query that postgrex uses (the elixir pgwire client)

Per https://www.postgresql.org/docs/current/functions-comparison.html:
> <> is the standard SQL notation for “not equal”. != is an alias, which is converted to <> at a very early stage of parsing. Hence, it is not possible to implement != and <> operators that do different things.

mpisanko:


mpisanko:
would u pls have a look @mbutlerw

mbutlerw (dismissed):

mbutlerw (changes requested):
Change looks good overall, but could we add a test for this if there isn't already coverage.

mbutlerw (approved):

jarohen (changes requested):
@mpisanko I know this is a three-liner, but I'd probably fix this in plan.clj (something like `((some-fn {'!= '<>} identity) (symbol (.getText (.compOp ctx))))` in `visitComparisonPredicate{,Part2}`?), or maybe as an alias in the expression engine - reason being I think this _may_ alter any error messages, changing the user's original query...?

jarohen (approved):

PR #3791, opened 2024-10-15
Documenting how to customize logging for the Docker image

egg-juxt:
I didn't think this was stuff for the official docs, but still handy to document, so I've put it inside the docker image.


refset:
Thanks @egg-juxt we'll add a `Troubleshooting` page to go into more depth also

refset:
Thanks @egg-juxt we'll add a `Troubleshooting` page to go into more depth also

PR #3800, opened 2024-10-17
Pgwire cleanup

### LINKED ISSUE #3768: Missing `_id` in pgwire INSERT never returns
`INSERT INTO foo (notid) VALUES (1);`

### LINKED ISSUE #3780: Dealing with closed stream or other bad/malicious data in pgwire
For example `read-c-string` in pgwire.clj will currently blow up if the stream closes unexpectedly
```clj
(defn- read-c-string
  "Postgres strings are null terminated, reads a null terminated utf-8 string from in."
  ^String [^InputStream in]
  (loop [baos (ByteArrayOutputStream.)
         x (.read in)]
    (if (zero? x) 
      (String. (.toByteArray baos) StandardCharsets/UTF_8)
      (recur (doto baos
               (.write x))
             (.read in)))))
```

I have not gotten a simple test for this, but have definitely run into an issue where this function blows up the Heap space.

### LINKED ISSUE #3784: Better error handling approach in pgwire server
Currently the error handling in the pgwire server is a bit all over the place. Sometimes we throw and rethrow. Sometimes we send a message. I think it would be good to find some unified approach especially as the `cmd-fns` are used from the `simple query` as well as the `extended query` workflow. 

By unified approach we mean:
- One place where we catch errors (for example `handle-msg`)
- If it is an error that should be dealt with via the protocol, deal with it immediately, dont rethrow. Try to skip subsequent commands until we go back to square one (`sync` or `read-for-query`). 
- Other bits?

FiV0:
This PR adds a couple of changes to the pgwire namespace that should help with testing and organizing the code better:
- Testing a `Connection` (and the protocol) without any need for a real (socket) connection 
- simplify a couple of `cmd-...` calls
- deal with error propagation more consistently

### DIFF
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -1079,6 +1079,9 @@
         (do (.unread in (byte x))
             (recur in (assoc acc (read-c-string in) (read-c-string in))))))))
 
+(defn skip-until-sync? [{:keys [conn-state] :as _conn}]
+  (:skip-until-sync @conn-state))
+
 (defn cmd-startup [conn]
   (loop [{{:keys [in]} :frontend, :keys [server], :as conn} conn]
     (let [{:keys [version msg-in]} (read-version in)]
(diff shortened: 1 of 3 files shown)


jarohen (approved):
:ok_hand:

PR #3806, opened 2024-10-18
Elixir support - add missing bits to pgwire (goes alongside #3773)

### LINKED ISSUE #3737: Elixir hello-world
See: [ecto](https://github.com/elixir-ecto/ecto) and [Postgrex](https://github.com/elixir-ecto/postgrex)

* [ ] Add a minimal test-suite under the lang harness
* [ ] Fix initial metadata query (info-schema-test, search 3737)

mpisanko:


core/src/main/clojure/xtdb/types.clj
```diff
@@ -1062,36 +1150,149 @@
...
+                               (.setCharAt sb (dec (.length sb)) \}))
+                             (.append sb "}"))
+                           (utf8 (.toString sb))))
+           :write-binary (fn [_env ^IVectorReader list-rdr idx]
+                           (let [^IVectorReader el-rdr (.listElementReader list-rdr)
+                                 sb (StringBuilder. "{")]
+                             (if (zero? (.getListCount list-rdr idx))
+                               (.append sb "}")
+                               (do
+                                 (loop [lidx (.getListStartIndex list-rdr idx)]
```
jarohen: nit: `dotimes`?

core/src/main/clojure/xtdb/types.clj
```diff
@@ -1112,13 +1313,19 @@
    [:timestamp-tz :micro] :timestamptz
    :tstz-range :tstz-range
    [:interval :month-day-nano] :interval
+   [:list :i32] :_int4
+   [:list :i64] :_int8
 
-   #_#_ ; FIXME not supported by pgjdbc until we sort #3683 and #3212
-   :transit :transit})
+   #_#_; FIXME not supported by pgjdbc until we sort #3683 and #3212
```
jarohen: nit: try to minimise your diff, where possible, to make it easier for the PR reader to figure out what's changed :pray: 

(this one's obviously very minor, but a good habit to get into)

core/src/main/clojure/xtdb/types.clj
```diff
@@ -300,11 +301,19 @@
 (def num-types (descendants col-type-hierarchy :num))
 (def date-time-types (descendants col-type-hierarchy :date-time))
 
+(defn is-list-type [col-type]
```
jarohen: nit: naming - `list-type?`?

core/src/main/clojure/xtdb/types.clj
```diff
@@ -1130,21 +1337,48 @@
...
                        (->> (into #{} (map (partial col-type->pg-type fallback-pg-type)))))]
+     (log/tracef "field->pg-type %s, col-type %s, field-name %s" (pr-str col-types) col-type field-name)
      (-> (if-let [col-type (->unified-col-type col-types)]
            col-type
            (col-type->pg-type fallback-pg-type col-type))
          (pg-types)
          (set/rename-keys {:oid :column-oid})
          (assoc :field-name field-name
                 :column-name field-name)))))
+
```
jarohen: we don't currently tend to check these in - they tend to become tests in the test suite (if appropriate)

build.gradle.kts
```diff
@@ -104,7 +104,7 @@ allprojects {
                 implementation(deps.clojure.lang)
 
                 testRuntimeOnly("dev.clojurephant", "jovial", "0.4.1")
-                nrepl("cider", "cider-nrepl", "0.47.1")
+                nrepl("cider", "cider-nrepl", "0.50.1")
```
jarohen: this commit can go direct to `main`, dw about including these in a PR (see git.adoc)

src/test/clojure/xtdb/pgwire_test.clj
```diff
@@ -1982,8 +1985,36 @@ ORDER BY t.oid DESC LIMIT 1"
...
+          (t/is (= [{"ids" "_int8"}]
+                   (result-metadata stmt)
+                   (result-metadata rs)))
+
+          (t/is (= [{"ids" [1 2]}] (rs->maps rs))))))))
+
+(deftest test-complex-array
+  (t/testing "complex array from query used by elixir client"
+    (with-open [conn (jdbc-conn)]
+      (with-open [stmt (.prepareStatement conn "SELECT ARRAY (SELECT a.atttypid FROM pg_attribute AS a  WHERE a.attrelid = t.typrelid AND a.attnum > 0 AND NOT a.attisdropped  ORDER BY a.attnum) as arr FROM pg_type AS t LEFT JOIN pg_type AS d ON t.typbasetype = d.oid LEFT JOIN pg_range AS r ON r.rngtypid = t.oid OR r.rngmultitypid = t.oid OR (t.typbasetype <> 0 AND r.rngtypid = t.typbasetype) WHERE (t.typrelid = 0) AND (t.typelem = 0 OR NOT EXISTS (SELECT 1 FROM pg_catalog.pg_type s WHERE s.typrelid <> 0 AND s.oid = t.typelem))")]
```
jarohen: can this be pretty-printed?

```sql
SELECT ARRAY (SELECT a.atttypid
              FROM pg_attribute AS a
              WHERE a.attrelid = t.typrelid AND a.attnum > 0 AND NOT a.attisdropped
              ORDER BY a.attnum)
         as arr
FROM pg_type AS t
  LEFT JOIN pg_type AS d ON t.typbasetype = d.oid
  LEFT JOIN pg_range AS r ON r.rngtypid = t.oid OR r.rngmultitypid = t.oid OR (t.typbasetype <> 0 AND r.rngtypid = t.typbasetype)
WHERE (t.typrelid = 0)
  AND (t.typelem = 0
         OR NOT EXISTS (SELECT 1
                        FROM pg_catalog.pg_type s
                        WHERE s.typrelid <> 0 AND s.oid = t.typelem))
```

src/test/clojure/xtdb/pgwire_test.clj
```diff
@@ -1982,8 +1985,36 @@ ORDER BY t.oid DESC LIMIT 1"
...
+
+      (with-open [stmt (.prepareStatement conn "SELECT ARRAY(SELECT _id FROM docs ORDER BY _id) AS ids")]
+        (with-open [rs (.executeQuery stmt)]
+          (t/is (= [{"ids" "_int8"}]
+                   (result-metadata stmt)
+                   (result-metadata rs)))
+
+          (t/is (= [{"ids" [1 2]}] (rs->maps rs))))))))
+
+(deftest test-complex-array
```
jarohen: nit: I'd probably call this `test-elixir-client-complex-array` and lose the `t/testing`?

src/test/clojure/xtdb/pgwire_test.clj
```diff
@@ -1982,8 +1985,36 @@ ORDER BY t.oid DESC LIMIT 1"
 
         (t/is (= [{"_id" 2, "f" 2.0, "i" 2} {"_id" 1, "f" 1.0, "i" 1}] (rs->maps rs)))))))
 
+(deftest test-primitive-array
+  (t/testing "primitive array"
```
jarohen: nit: don't need the `t/testing` here - you'll get the same info from the test name

src/dev/clojure/dev.clj
```diff
@@ -45,8 +46,11 @@
   (do
     (halt)
     (util/delete-dir (util/->path dev-node-dir))
+    (set-log-level! 'xtdb.types :trace)
```
jarohen: assume this wasn't intended to be committed? :wink:

jarohen (changes requested):
Cheers @mpisanko, nice one :pray: 

Suggestions below - give me a shout if any are unclear :slightly_smiling_face: 

Could you also:

- push nrepl bump to `main` independently
- squash down your WIP commits so that each commit is a minimal, atomic change

Thanks!

core/src/main/clojure/xtdb/types.clj
mpisanko: do we have a common code formatter / editorconfig?
that might help reduce irrelevant changes. 
but point taken.

core/src/main/clojure/xtdb/types.clj
```diff
@@ -1191,7 +1200,99 @@
...
+                           (utf8 (.toString sb))))
+           :write-binary (fn [_env ^IVectorReader list-rdr idx]
+                           (let [^IVectorReader el-rdr (.listElementReader list-rdr)
+                                 sb (StringBuilder. "{")]
+                             (if (zero? (.getListCount list-rdr idx))
+                               (.append sb "}")
+                               (do
+                                 (loop [lidx (.getListStartIndex list-rdr idx)]
+                                   (when (< lidx (+ (.getListCount list-rdr idx)
+                                                    (.getListStartIndex list-rdr idx)))
```
FiV0: When you replace with `dotimes` or something else, can you let bind these, so they don't get called every time.

core/src/main/clojure/xtdb/types.clj
```diff
@@ -1191,7 +1200,99 @@
...
+                                     (.append sb (if (.isNull el-rdr lidx)
+                                                   ""
+                                                   (.getInt el-rdr lidx)))
+                                     (.append sb ",")
+                                     (recur (inc lidx))))
+                                 (.setCharAt sb (dec (.length sb)) \})))
+                             (utf8 (.toString sb))))}
```
FiV0: I don't think this is the binary representation of arrays in postgres. 
See https://github.com/postgres/postgres/blob/4b652692e97923298d0c7b9c1c303fc648984d92/src/include/utils/array.h
If you think you don't need this for elixir remove the binary read/writers for now.

core/src/main/clojure/xtdb/types.clj
```diff
@@ -1212,13 +1313,19 @@
...
-   :transit :transit})
+   #_#_; FIXME not supported by pgjdbc until we sort #3683 and #3212
+       :transit :transit})
 
 (defn col-type->pg-type [fallback-pg-type col-type]
   (get col-types->pg-types
        (cond-> col-type
+         ;; ignore lists
+         (is-list-type col-type)
+         (subvec 0 2)
```
FiV0: I don't think you want to ignore arbitrary lists here.

core/src/main/clojure/xtdb/types.clj
```diff
@@ -1230,21 +1337,48 @@
...
 
   ([fallback-pg-type ^Field field]
    (let [field-name (.getName field)
          col-type (field->col-type field)
-         col-types (-> (flatten-union-types col-type)
-                       (disj :null)
+         col-types (-> ((if (is-list-type col-type)
+                          flatten-list-unions
+                          flatten-union-types) col-type)
+                       remove-nulls
```
FiV0: I find this together with the above functions hard to parse and understand. Especially the ignoring of lists seems incorrect. I understand that you want to get rid of inner nulls. Could we maybe have:
- a function `remove-nulls` that takes an arbitrary col-type and removes any nulls (even arbitrary nested ones)
- together with `flatten-union-types` you remove unnessary unions and then only in the `->unified-col-type` code path you match against an "pure" pgwire array

FiV0 (changes requested):

core/src/main/clojure/xtdb/types.clj
FiV0: We can also pair if that would help.

core/src/main/clojure/xtdb/types.clj
mpisanko: nice! thanks!

core/src/main/clojure/xtdb/types.clj
mpisanko: fair enough - think this must have been an overhang from before.

core/src/main/clojure/xtdb/types.clj
mpisanko: yeah, definitely - let's review this - if it's hard to parse - definitely needs improving 👍

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -598,7 +598,9 @@
 (def ^:private io-uint32
   "An unsigned 32bit integer"
   {:read #(.readInt ^DataInputStream %)
-   :write #(.writeInt ^DataOutputStream %1 (int %2))})
+   :write (fn [^DataOutputStream out v]
+              (when v
+                (.writeInt out (int v))))})
```
FiV0: I think this function should fail rather than not write anything in case of nil. Why is this change needed anyway?

core/src/main/clojure/xtdb/types.clj
FiV0: Resolved offline.

core/src/main/clojure/xtdb/types.clj
FiV0: @mpisanko Let's write a test with `(jdbc-conn "binaryTransfer" true)` to show that the binary transfer works for int arrays. Then I am happy to bring this in.

core/src/main/clojure/xtdb/types.clj
```diff
@@ -1191,7 +1191,95 @@
...
+                          (mapv #(Long/parseLong %) elems)))
+                      ;; :read-binary #()
+           :write-text (fn [_env ^IVectorReader list-rdr idx]
+                         (let [^IVectorReader el-rdr (.listElementReader list-rdr)
+                               sb (StringBuilder. "{")
+                               len (.getListCount list-rdr idx)]
+                           (if-not (zero? len)
+                             (do
+                               (dotimes [i len]
+                                 (let [offset (.getListStartIndex list-rdr idx)]
```
FiV0: Can we move this `let` outside the `dotimes`.

mpisanko: yeah, dunno why i was under impression it would be changing every time.

core/src/main/clojure/xtdb/types.clj
mpisanko: only works 'for supported types' - these seem not suppported.

FiV0: Yes, then let's do the test through postgrex.

core/src/main/clojure/xtdb/types.clj
```diff
@@ -785,6 +785,48 @@
...
+        ;; _payload-size (.getInt bb)
+        ;; _dimensions (.getInt bb 4)
+        data-offset (.getInt bb 8)
+        oid (.getInt bb 12)
+        length (.getInt bb 16)
+        [typlen get-elem] (case oid
+                            (23 1007) [4 (fn [bb idx] (.getInt bb idx))]
+                            (20 1016) [8 (fn [bb idx] (.getLong bb idx))])
+        start  (+ data-offset 20)]
+    (persistent!
```
jarohen: nit: think this can be `(into [] (map (fn [idx] (get-elem ...)) (range length))`, which will do the reduce/conj, transient/persistent dances for you?

jarohen (approved):
nice one @mpisanko, thanks! :pray:

PR #3814, opened 2024-10-23
Fix MultiIndirectVectorReader.listElementReader

FiV0:
We are currently constructing the ElementReader for MultiIndirectVectorReader incorrectly.
Let's walk through an example. Assuming the following underlying vectors + indirections
```
[[1 2 3] [4 5 6]] first underlying vector list vector
[[7 8] [9 10]] second underlying vector list vector
rdr indirection [0 0 1 1]
vec indirection [0 1 0 1]
```
The corresponding represented vector
```
[[1 2 3] [4 5 6] [7 8] [9 10]]
```
Then the corresponding element list 
```
[1 2 3 4 5 6 7 8 9 10]
```
The real underlying element list vectors look as follows
```
[1 2 3 4 5 6] first underlying element list vector
[7 8 9 10] second underlying element list vector
```
The corresponding indirection vectors
``` 
[0 0 0 0 0 0 1 1 1 1]
[0 1 2 3 4 5 0 1 2 3]
```
These last indirections are currently not constructed correctly. The PR fixes this naively.
Normally getting a sub reader from an MIVR is constant in the number of underlying vectors. In this case the construction takes $O(n)$ in the accumulated number of elements in all lists.

### DIFF
--- a/core/src/main/kotlin/xtdb/vector/IndirectMultiVectorReader.kt
+++ b/core/src/main/kotlin/xtdb/vector/IndirectMultiVectorReader.kt
@@ -44,9 +44,7 @@ class IndirectMultiVectorReader(
         return readers[readerIndirection[idx]] ?: throw unsupported()
     }
 
-    override fun valueCount(): Int {
-        return readerIndirection.valueCount()
-    }
+    override fun valueCount(): Int =  readerIndirection.valueCount()
 
     override fun getName(): String {
         return name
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/xtdb/vector/IndirectMultiVectorReader.kt
```diff
@@ -100,11 +98,27 @@ class IndirectMultiVectorReader(
...
+            val listCount = rdr.getListCount(vectorIndirections[i])
+            readerIndirectionArray += repeat(listCount, readerIndirection[i])
+            vectorIndirectionArray += range(rdr.getListStartIndex(vectorIndirections[i]), listCount)
+        }
+        return IndirectMultiVectorReader(
+            listElementReaders, VectorIndirection.selection(readerIndirectionArray), VectorIndirection.selection(vectorIndirectionArray))
+    }
 
-    override fun getListStartIndex(idx: Int): Int = safeReader(idx).getListStartIndex(vectorIndirections[idx])
+    override fun getListStartIndex(idx: Int): Int = (0 until idx).map { safeReader(it).getListCount(vectorIndirections[it]) }.sum()
```
FiV0, line 121: The above construction is $O(n)$ which i guess is ok, but this is also $O(n)$ which is bad, as this will likely get called more often and caculates the same partial sums all over again.

jarohen (approved):
Yeah, agreed! :sweat_smile:

PR #3830, opened 2024-11-01
add elixir lang test - uses binary encoding

mpisanko:


lang/elixr/lib/elixr.ex
```diff
@@ -0,0 +1,18 @@
...
+  ## Examples
+
+      iex> Elixr.hello()
+      :world
+
+  """
+  def hello do
+    :world
+  end
+end
```
FiV0: What is this file for?

core/src/main/clojure/xtdb/types.clj
```diff
@@ -818,13 +825,14 @@
         oid (.getInt bb 12)
         length (.getInt bb 16)
         [typlen get-elem] (case oid
-                            (23 1007) [4 (fn [bb idx] (.getInt bb idx))]
-                            (20 1016) [8 (fn [bb idx] (.getLong bb idx))])
-        start  (+ data-offset 20)]
+                            (23 1007) [4 (fn [bb ^long start ^long typlen ^long idx] (.getInt bb (+ start (* idx typlen))))]
+                            (20 1016) [8 (fn [bb ^long start ^long typlen ^long idx] (.getLong bb (+ start (* idx typlen))))])
```
FiV0: These function still has reflection warnings on the `.getInt` and `.getLong`.

FiV0 (changes requested):
Just two tiny remarks otherwise LGTM.

lang/elixr/lib/elixr.ex
mpisanko: removed- not needed

PR #3834, opened 2024-11-04
Add gauges for disk cache and memory cache

### LINKED ISSUE #3832: Exposing MemoryCache and DiskCache metrics
Following on from #3828:

* the Arrow `BufferAllocator`'s metrics are no longer representative of what impact we're actually having on the system.
* we now have disk cache metrics that would be helpful to expose

Let's get gauges in for these, and added to our monitoring dashboards/guides

tggreene:
Closes #3832

### DIFF
--- a/core/src/main/kotlin/xtdb/cache/MemoryCache.kt
+++ b/core/src/main/kotlin/xtdb/cache/MemoryCache.kt
@@ -3,6 +3,8 @@
 package xtdb.cache
 
 import com.github.benmanes.caffeine.cache.RemovalCause
+import io.micrometer.core.instrument.Gauge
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/xtdb/cache/MemoryCache.kt
```diff
@@ -25,6 +27,7 @@ class MemoryCache
 
     @Suppress("MemberVisibilityCanBePrivate")
     val maxSizeBytes: Long,
+    val meterRegistry: MeterRegistry,
```
tggreene, line 30: Wasn't sure about arg ordering here, this seemed most sensible to me given optionality of pathLoader and non-optionality of meterRegistry

PR #3835, opened 2024-11-04
(WIP) Implementing txId Lag

danmason:
Currently heavily WIP - needs to potentially be implemented for each log.

Idea being this could be used like this within K8s, for example:
```
readinessProbe:
  exec:
    command:
      - /bin/sh
      - -c
      - |
        LAG=$(curl -s http://localhost:3000/txLag)
        if [ "$LAG" -lt 5 ]; then
          exit 0
        else
          exit 1
        fi
  initialDelaySeconds: 30    # Wait 30 seconds before starting the first probe
  periodSeconds: 20          # Check every 20 seconds
  failureThreshold: 5        # Allow up to 5 consecutive failures before marking as unhealthy
  successThreshold: 1        # One successful probe will mark the container as ready
  timeoutSeconds: 10         # Allow 10 seconds for each probe to complete
```

### DIFF
--- a/api/src/main/clojure/xtdb/api.clj
+++ b/api/src/main/clojure/xtdb/api.clj
@@ -216,6 +216,12 @@
   (xtp/status node))
 
 #_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
+(defn tx-lag
+  "Returns the Transaction ID lag of the node as a long"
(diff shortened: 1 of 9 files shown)


jarohen:
@danmason I'm assuming this is superseded, feel free to reopen if not

PR #3836, opened 2024-11-04
XT Authentication

### LINKED ISSUE #3626: Authentication/authorization support for pgwire clients
superseded by #4489

----

Proposal A:

- [x] SSL
- [x] https://github.com/xtdb/xtdb/issues/3852
- [ ] `xtdb` user, password configured in YAML (possibly via `XTDB_PASSWORD` env var). 
  - If this is specified, then we reject all other connections.
  - This user has full superuser authz.
  - Transaction log to carry details of the currently authenticated user (OIDs, so that users can be renamed or deleted/re-created).
- [ ] Ability for the `xtdb` user to create/delete other users. For now, just read-write and read-only policies.
  - Use Postgres's `CREATE USER` etc for guidance. for example:
    - built-in role `read_write` (read-only for everyone, nologin)
    - `CREATE USER dave IN ROLE read_write PASSWORD '...'`
    - `GRANT|REVOKE read_write TO dave`
    - only superusers can administrate authz/authn for now.
  - State saved within XTDB itself (`pg_user`). Prevent updates to this table via regular SQL - just `GRANT` and `REVOKE`.
  - Also need to consider revocation. On the transaction log, this is straightforward enough because everything is serialise - this is more about what happens when a user is making queries when their authn/authz is revoked. 
  - Authorization is to be read at the latest-completed-tx regardless of what basis the query specifies.
- [ ] Supporting `SET ROLE ...` 
- [ ] ACL for writes to a table

Out of scope:

1. Integration with third-party authentication (e.g. Microsoft Entra)

FiV0:
This adds pg-wire authentication. 

There are now two parts that play a role in user authentication. Authentication rules of the form 
```clj
[{:user "xtdb" :method :password :address "127.0.0.1"}
 {:user "ada" :method :trust :address "127.0.0.1"}]
```
which are passed to the server (http and pgwire) on startup.

There is also the possibility to `nil` the user or the address which means the the authentication rules applies to all users or all addresses.

Authentication options of the form 
```clj
{:user "xtdb" :password "xtdb"}
```
these are send via http or pgwire connection initialization. 

The auth algorithm is then as follows:
- We find the first matching authentication rule.
- If a record matches we check if a password is required or not. If not we authenticate the user.
- If a record matches and a password is required we check the password (http) or ask for one (pgwire).
- If the password matches the request is handled or the connection accepted.
- If a record matches, but we can't find the user in the database we also reject the request. For example auth-record `{:user "all" :method :password :address "127.0.0.1"}` and authentication options `{:user "fin" :password "foobar"}`.

The default authentication record (without any specification in the config) are currently:
```clj
[{:user nil :method :trust  :address nil}]
```
which allows anybody to connect.

The default user + password in the database is 
```clj
{:user "xtdb" :password "xtdb"}
```

The PR also adds the functionality to inject certain static data (initial database data) into scan for which subsequent operations on the specific table are dealt with as usual. The initial data is joined similarly to how the live trie pages are joined in scan. This static arrow data is loaded once at the start of the system.

Secondly it adds `CREATE/ALTER USER` which I deal exactly the same by just adding a new row to the pg_user table. One could potentially do a check for `ALTER USER` to make sure the user exists before writing the new row.

Todo:
- [ ] allow for passwords with special characters

Out of scope:
- autherization


core/src/main/clojure/xtdb/log.clj
```diff
@@ -381,15 +382,21 @@
...
+          default-tz-writer (vw/->writer (.getVector root "default-tz"))
+          user-writer (vw/->writer (.getVector root "user"))]
 
       (when system-time
         (.writeObject (vw/->writer (.getVector root "system-time")) system-time))
 
+      (if user
+        (.writeObject user-writer user)
+        ;; anonymous user
+        (.writeNull user-writer))
```
FiV0: Not sure if we should make `anon` explicit.

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -994,6 +992,34 @@
     (cmd-write-msg conn msg-backend-key-data {:process-id (:cid conn), :secret-key 0})
     (cmd-send-ready conn)))
 
+(defn user+password [node user]
+  (first (xt/q node "SELECT passwd AS password, username AS user FROM pg_user WHERE username = ?" {:args [user]})))
+
+(defn cmd-startup-pg30 [conn startup-params]
+  (let [{:keys [node conn-state !closing?]} conn]
+    (if node
```
FiV0: There is sort of a chicken/egg problem for the playground as I need a node to get the password, but the node only gets attached after startup. I could do some rejuggling if we want authn for the playground as well or always skip authn for the playground. Might be a bit weird to not be able to test authn via the playground.

src/test/java/xtdb/api/XtdbHelloWorld.kt
```diff
@@ -10,7 +10,7 @@ class XtdbHelloWorld {
     @Test
     fun `hello world`() {
         Xtdb.openNode(Xtdb.Config(ServerConfig(0))).use { xtdb ->
-            DriverManager.getConnection("jdbc:postgresql://localhost:${xtdb.serverPort}/xtdb").use { conn ->
+            DriverManager.getConnection("jdbc:postgresql://localhost:${xtdb.serverPort}/xtdb?user=anonymous").use { conn ->
```
FiV0: Lots of places I now need to add `anonymous` explicitly as JDBC o/w passes some other user. A bit annoying, but not too much.

core/src/main/clojure/xtdb/information_schema.clj
```diff
@@ -62,7 +62,8 @@
       pg_catalog/pg_settings {name :utf8, setting :utf8}
 
       pg_catalog/pg_range {rngtypid :i32, rngsubtype :i32, rngmultitypid :i32
-                           rngcollation :i32, rngsubopc :i32, rngcanonical :utf8, rngsubdiff :utf8}})
+                           rngcollation :i32, rngsubopc :i32, rngcanonical :utf8, rngsubdiff :utf8}
+      pg_catalog/pg_user {username :utf8 usesuper :bool passwd [:union #{:utf8 :null}]}})
```
FiV0: I make this field nullable so a user without a password doesn't need to got through authentication. Maybe not ideal...

jarohen: it's nullable in postgres too, IIRC - in fact, I think this is a view over pg_role there?

tl;dr I suspect it's fine, but I'd probably err on the side of not authenticating someone who doesn't have a password (maybe you've already done that, I'll continue scrolling)

core/src/main/clojure/xtdb/log.clj
jarohen: agreed to leave it as null :+1: 

you also don't need to explicitly writeNull here - I'd just go `(when user ...)` (like `(when system-time ...)` above

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1544,20 +1573,43 @@
...
+      (if (and password (= (util/md5 supplied-passwd) password))
+        (do (swap! conn-state dissoc :auth-req :startup-params)
+            ;; asking for a password
+            (cmd-write-msg conn msg-auth {:result 0})
+            (startup-ok conn startup-params))
+
+        (do
+          (reset! !closing? true)
+          (cmd-send-error conn (err-invalid-passwd "Invalid password")))))
+    ;; just ignore it?
```
jarohen: do you need to check `auth-req` here? if the user provides a password, regardless of whether it's required, we can still authenticate them - main thing is that if it's required and they _don't_, to kick them out

core/src/main/clojure/xtdb/pgwire.clj
jarohen: does pg30 come through before the password message?

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -994,6 +992,34 @@
...
+
+(defn cmd-startup-pg30 [conn startup-params]
+  (let [{:keys [node conn-state !closing?]} conn]
+    (if node
+      (let [{:keys [password] :as res} (user+password node (or (get startup-params "user") "anonymous"))]
+        (cond
+          (nil? res)
+          (do
+            (reset! !closing? true)
+            (cmd-send-error conn (err-invalid-auth-spec "User not found")))
```
jarohen: we should neither confirm nor deny the existence of a user :detective:

FiV0: Postgres does seem to make a difference here (copying from chatgpt):

28000 - invalid_authorization_specification:
This error code indicates a general issue with the provided authentication specification. It can occur when there’s a problem with the user or authentication mechanism itself, such as:

- The user does not 
- The client is attempting to connect using an unsupported authentication 
- There is a configuration mismatch in pg_hba.conf, the PostgreSQL host-based authentication configuration file.

Essentially, 28000 signals that the user authorization setup is invalid or not configured correctly, rather than specifically pointing to a password issue.

28P01 - invalid_password:
This error code specifically indicates that the password provided for an existing user is incorrect. It means that:

- The username exists and authentication has reached the password verification stage.
- The password provided does not match the one stored for the specified user.

In other words, 28P01 is raised only when the provided password is wrong for an existing user.

FiV0:
How this now works with default values `[{:user "xtdb" :password "xtdb"} {:user "anonymous" :password nil}]`
- User `fin` tries to authenticate - we ask for a password, when a password gets send we reject in all cases as this user doesn't exist
- User `anonymous` tries to authenticate - we just let him through as no password is required
- User `xtdb` tries to authenticate - we ask for a password, when the password gets send we check it and authenticate if correct

I also added `read-client-msg!` to the Frontend as it made testing in `pgwire_protocol_testing.clj` easier. I had to forward `declare` some stuff, maybe it would make sense to move some things to another namespace.

src/test/java/xtdb/api/XtdbHelloWorld.kt
jarohen: hmm, this doesn't feel right - what user does it pass?

(guessing) current username on the machine?

was hoping 'anonymous' isn't an actual user, more the absence of one - but if that assumption doesn't apply, it's probably worth reconsidering

FiV0:
Stuff to still address on this PR:
- [ ] https://github.com/xtdb/xtdb/issues/3852
- [ ] some unique identifer for the transaction log, otherwise usernames get mangled when getting recreated, deleted, superseded

FiV0:
So the latest status:
There are now two parts that play a role in user authentication. Authentication records of the form 
```clj
[{:user "xtdb" :method :password :address "127.0.0.1"}
 {:user "ada" :method :trust :address "127.0.0.1"}]
```
which are passed to the server (http and pgwire) on startup and authentication options of the form 
```clj
{:user "xtdb" :password "xtdb"}
```
these are send via http or connection initilization in pgwire. The auth algorithm is then as follows:
- Is there some authentication record for the combination of user and address? We fall back to an `all` authentication record if a non-user specific one doesn't exist. If there is no record that matches we reject the request/connection.
- If a record matches we check if a password is required or not. If not we authenticate the user.
- If a record matches and a password is required we check the password (http) or ask for one (pgwire).
- If the password matches the request is handled or the conneciton accepted.
- If a record matches, but we can't find the user in the database we also reject the request. For example auth-record `{:user "all" :method :password :address "127.0.0.1"}` and authentication options `{:user "fin" :password "foobar"}`.

The default authentication records (without any specification in the config) are currently:
```clj
[{:user "all" :method :trust  :address "0:0:0:0:0:0:0:1"}
 {:user "all" :method :trust :address "127.0.0.1"}]
```
which allows any `localhost` ip4 and ip6 to connect without password.

The default user + password in the database is 
```clj
{:user "xtdb" :password "xtdb"}
```

There is no need to tweak the clojure API (apart from allowing `xt/status` with options).  If no user name gets passed through we fall back to an `all` authentication record. The question is of course if we want to tweak the default to `xtdb` for non local connections.

Stuff still todo to decide on before this goes in:
- [ ] update the docs
   - wait for #3859 
- [x] regarding: some unique identifer for the transaction log, otherwise usernames get mangled when getting recreated, deleted, superseded. I wonder if we just say that the user-name is a unique identifier.
   - going with user name for now, we will need to switch to OID later
- [x] What to do about the `all` username? Either the "all" records are marked differently or we forbid the `all` username. Other options?
   - nullable user and addresss
- [x] I need to test how this behaves with the docker image and what should be the default authentication record for non `localhost` connections.
   - with the new default of `{:user nil :password nil :method :trust}` not relevant.
- [ ] Do we want to add subnet masks?
    - depends

FiV0:
@jarohen I forgot to mention that `xt/status` now takes options for the clojure API as well as the http request as one needs to (potentially) pass authentication for that. I did not implement that protocol part for the "standard"/in-process node as their is no need for authentication in that case.

FiV0:
I realized that I am not testing much that the user actually ends up on the log. Not super important as we are not doing authorization currently.

docs/shared
FiV0, line 1: I have seen this. Will try to squash it out.

docs/src/content/docs/ops/config/authentication.adoc
```diff
@@ -0,0 +1,52 @@
...
+  authn:
+   authnRules:
+      # admin always requires a password
+      - user: admin
+        method: PASSWORD
+      # We trust local connections
+      - address: 127.0.0.1
+        method: TRUST
+      # Everything else requires a password.
+      - method: PASSWORD
```
jarohen, line 34: worth double checking if this is the pg_hba behaviour - i.e. does it fall through if _either_ the rule doesn't match or the authn fails, or, if an earlier rule matches but the authn fails, does it then fail the whole authn regardless of later rules?

docs/src/content/docs/ops/config/authentication.adoc
```diff
@@ -0,0 +1,52 @@
...
+until the first one matches. If none matches the connection is rejected.
+
+Below is a pgwire server configuration with 3 different rules.
+
+[source,yaml]
+----
+server:
+  port: 5432
+  authn:
+   authnRules:
```
jarohen: nit: indentation (extra space?)

core/src/main/clojure/xtdb/indexer.clj
```diff
@@ -455,9 +455,34 @@
...
+                 (.writeObject user))
+
+               (doto (.structKeyWriter doc-writer "username" (FieldType/notNullable #xt.arrow/type :utf8))
+                 (.writeObject user))
+
+               (doto (.structKeyWriter doc-writer "usesuper" (FieldType/notNullable #xt.arrow/type :bool))
+                 (.writeObject false))
+
+               (doto (.structKeyWriter doc-writer "passwd" (FieldType/nullable #xt.arrow/type :utf8))
+                 (.writeObject (util/md5 password)))
```
jarohen: let's go straight for bcrypt here otherwise it'll be a migration to do so

FiV0: Just so we have it in writing:
- store the password in bcrypt
- this closes the door for `AuthenticationMD5Password` from https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-START-UP
- meaning drivers can't send a password encrypted over the wire (via pgwire)

docs/src/content/docs/ops/config/authentication.adoc
FiV0: From the docs
>There is no “fall-through” or “backup”: if one record is chosen and the authentication fails, subsequent records are not considered. If no record matches, access is denied.

api/src/main/clojure/xtdb/api.clj
```diff
@@ -73,6 +73,7 @@
     - `:current-time`: override wall-clock time to use in functions that require it
     - `:args`: arguments to pass to the query.
     - `:default-tz`: overrides the default time zone for the query
+    - `:auth-opts`: authentication options
```
jarohen: subjective: these are all options, shall we just go with `:authn`?

core/src/main/clojure/xtdb/indexer.clj
jarohen: > meaning drivers can't send a password encrypted over the wire (via pgwire)

we have TLS support if this is a concern :ok_hand:

core/src/test/kotlin/xtdb/api/YamlSerdeTest.kt
```diff
@@ -289,4 +289,27 @@ class YamlSerdeTest {
 
         unmockkObject(EnvironmentVariableProvider)
     }
+
+    @Test
+    fun testAuthnConfigDecoding() {
+        val input = """
+        authnRules:
```
jarohen: you're already in the context of a block called `authn` - this could just be `rules`?

docs/src/content/docs/ops/config/authentication.adoc
jarohen: sweet :ok_hand:

docs/src/content/docs/ops/config.adoc
```diff
@@ -2,15 +2,16 @@
...
 
-An XTDB node consists of a number of components. 
+An XTDB node consists of a number of components.
 
 For information on how to configure these and switch out implementations, see their reference documents:
 
 * link:config/tx-log[Transaction Log]
 * link:config/storage[Storage]
 * link:config/modules[Optional Modules]
+* link:config/authentication[Authentication]
```
jarohen: I know authn's a component in our config map; I don't think I'd include it here - could it have its own (small!) section?

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1746,12 +1821,31 @@
        (.start accept-thread)
        server))))
 
-(defmethod xtn/apply-config! ::server [^Xtdb$Config config, _ {:keys [port num-threads ssl] :as server}]
+(defn ->authn-settings ^AuthnSettings [auth-rules]
```
jarohen: is pgwire the right place for these, if they affect the HTTP server too? i.e. do we need a separate xtdb.authn ns?

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1746,12 +1821,31 @@
...
+     :address (.getAddress auth-rule)}))
+
+(defmethod xtn/apply-config! ::server [^Xtdb$Config config, _ {:keys [port num-threads ssl authn-rules] :as server}]
   (if server
     (cond-> (.getServer config)
       (some? port) (.port port)
       (some? num-threads) (.numThreads num-threads)
-      (some? ssl) (.ssl (util/->path (:keystore ssl)) (:keystore-password ssl)))
+      (some? ssl) (.ssl (util/->path (:keystore ssl)) (:keystore-password ssl))
+      (some? authn-rules) (.authn (->authn-settings authn-rules)))
```
jarohen: I'm not sure about these going under `server` in the config, which may be what's influencing the rest of the authn code going in the pgwire file. What'd be the impact of putting this at the top level?

core/src/main/clojure/xtdb/indexer.clj
jarohen: apparently the state of the art is argon2 today, Java's 'bouncy castle' lib supports it, which is generally a good shout

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -997,6 +1048,61 @@
     (cmd-write-msg conn msg-backend-key-data {:process-id (:cid conn), :secret-key 0})
     (cmd-send-ready conn)))
 
+(defn user+password [node user]
+  (first (xt/q node "SELECT passwd AS pw, username AS usr FROM pg_user WHERE username = ?" {:args [user]})))
+
+(defn most-specific-rule [authn-rules address user]
```
jarohen: nit (subjective): this isn't necessarily the most-specific rule, I don't think - most-specific (to me, at least) would check all the rules to see if a more specific one matches?

jarohen (changes requested):
nice one @FiV0 :bow: couple of mostly small requests, as always :innocent:

core/src/main/clojure/xtdb/indexer.clj
FiV0: - Should passwords be stored as bytes or as Strings (these seems to be some argument for bytes as Strings stick around longer on the JVM)
- What about the salt? Both bcrypt as well as argon2 need one.

FiV0: Should I also store the salt in `pg_user`?

core/src/main/clojure/xtdb/pgwire.clj
FiV0: That name was still from my first take

core/src/main/clojure/xtdb/pgwire.clj
FiV0: AuthnSettings at the toplevel which get passed to the pg-wire server and the http-server? 

I admit that I just made it work. It's confusing to me that for example we have the pg-wire config [here](https://github.com/xtdb/xtdb/blob/b7ad5796e151448eee22c201fbe9339438695756/core/src/main/kotlin/xtdb/api/Xtdb.kt#L32) but not the http-config.

core/src/main/clojure/xtdb/indexer.clj
FiV0: Resolved by using http://funcool.github.io/buddy-hashers/latest/user-guide.html

core/src/main/clojure/xtdb/pgwire.clj
jarohen: it's because pgwire is included as standard but HTTP isn't, if that makes more sense?

jarohen:
merged ^^

refset:
@FiV0 is the `admin` user (per the example config) always named `xtdb` (i.e. they're the same thing)? Alternatively, what's the difference and how do you configure the admin password? For context, we would ideally like to configure the admin password via an env var during startup.

jarohen:
@refset wider plan for this is to ship the authn off to an OIDC provider - if we do allow admin credentials, they should only really be for local dev

refset:
Thanks, make sense. Currently just hoping to replicate the journey of a typical Postgres k8s config setup, e.g. https://github.com/GoogleCloudPlatform/click-to-deploy/blob/acfe1a6741dc23b1a6b8e84a01520fe05522a688/k8s/postgresql/chart/postgresql/templates/postgresql-statefulset.yaml#L59-L63 where POSTGRES_PASSWORD is a ~standard container env var, per https://hub.docker.com/_/postgres#:~:text=Environment%20Variables,rest%20are%20optional.

FiV0:
@refset There is currently no way to create a superuser and also nothing there to configure the passwords (for standard users) in the config. If this is essential we should create a card for it.

PR #3839, opened 2024-11-05
Disallow writing to reserved xt tables

### LINKED ISSUE #3245: Disallow DML against xt$txs
e.g. `INSERT INTO xt$txs (xt$id, foo) VALUES (10, 10)` (currently halts ingestion, prevents shutting down without memory leaks, user probably needs to start fresh etc.)

FiV0:
Mainly just what it says in the title. I don't know if there is some way to check this more efficiently.

### DIFF
--- a/core/src/main/clojure/xtdb/indexer.clj
+++ b/core/src/main/clojure/xtdb/indexer.clj
@@ -83,6 +83,7 @@
         system-time-µs (time/instant->micros system-time)
         tables (->> (.legs docs-rdr)
                     (into {} (map (fn [table-name]
+                                    (when (xt-log/forbidden-table? table-name) (throw (xt-log/forbidden-table-ex table-name)))
                                     (let [table-docs-rdr (.legReader docs-rdr table-name)
                                           doc-rdr (.listElementReader table-docs-rdr)
                                           ks (.structKeys doc-rdr)]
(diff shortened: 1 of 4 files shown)


FiV0:
Do I actually need to check for explict `put/delete/erase` or should this happen client side?

core/src/main/clojure/xtdb/indexer.clj
```diff
@@ -62,6 +62,13 @@
 
 (def ^:private abort-exn (err/runtime-err :abort-exn))
 
+(def ^:private forbidden-tables #{"xt/txs" "xt/tx_fns"})
```
jarohen: could do with checking the schema for this one - I'd assume `xt.*` (_except_ xt/tx_fns, ironically - I would guess that this should break the tx-fn tests as new tx-fns are just inserted as put-docs), `information_schema.*` and `pg_catalog.*`?

src/test/clojure/xtdb/sql_test.clj
```diff
@@ -1,7 +1,7 @@
 (ns xtdb.sql-test
   (:require [clojure.java.io :as io]
             [clojure.string :as str]
-            [clojure.test :as t]
+            [clojure.test :as t :refer [deftest]]
```
jarohen: noo, I've only just gotten rid of that! :stuck_out_tongue:

jarohen:
> Do I actually need to check for explict `put/delete/erase` or should this happen client side?

I'd check explicitly, yep - chances are the tx-log can be trusted in practice but no reason not to make sure. I might be tempted to check it submit-side too, for early feedback to the user.

> I don't know if there is some way to check this more efficiently.

wdym, sorry? efficiently as in code-size, it doesn't feel egregious to me, fwiw

src/test/clojure/xtdb/sql_test.clj
FiV0: Yeah, is there a reason for that? I always never refer but for `clojure.test` it seems idiomatic to me.

FiV0:
>wdym, sorry? efficiently as in code-size, it doesn't feel egregious to me, fwiw

Checking every time we go through the indexing fns.

PR #3840, opened 2024-11-05
Modifiable pg user table

FiV0:
This is on top of #3836. 

This adds the functionality to inject certain static data (initial database data) into scan for which subsequent operations on the specific table are dealt with as usual. The initial data is joined in similar to how the live trie pages are joined in scan. This static arrow data is loaded once at the start of the system.

Secondly it adds `CREATE/ALTER USER` which I deal exactly the same by just adding a new row to the pg_user table. One could potentially do a check for `ALTER USER` to make sure the user exists before writing the new row.


FiV0:
Added on top of #3836.

Issue #3844, opened 2024-11-06
Clojure user experience working with Transit-encoded PgObjects


jarohen:
We currently support [Transit](https://github.com/cognitect/transit-format) as an extension type/encoding over the pg-wire protocol, for better round-tripping of values that don't have a native JSON encoding.

Current state:

* Users can (3081f915b92a46762ffcf4a4470c187ec971237d) set 'fallback_output_format' on the connection to return Transit objects
* We have `transit->pgobject` which means that users can `["INSERT INTO table RECORDS ?" (transit->pgobject doc)]`, but this is not exposed to users.

Is there anything we can do to both make this more accessible, and publicise it?

* Could this support be added to next.jdbc? (and would there be appetite to do so?)
* Consider adding `->pgobject` and `<-pgobject` to `xtdb.serde` (already bundled in `xtdb-api`), or maybe create `xtdb.api.transit` if we don't want to inadvertently make `xtdb.serde` a public API
* Document it up in the Clojure documentation

refset:
Working sample code that might be helpful (borrowing from pgwire.clj/serde.clj, also implements `pgobect->transit`):
```clj
(ns main
  (:require [next.jdbc :as jdbc]
            [cognitect.transit :as transit])
  (:import org.postgresql.util.PGobject
           [java.io ByteArrayInputStream ByteArrayOutputStream]))

(def db
  {:dbtype "postgresql"
   :dbname "xtdb"
   :user "xtdb"
   :password "xtdb"
   :host "localhost"
   :port 5432})

;; this is relatively low-level code - the usual connection pooling
;; and SQL abstraction libraries can be used too.

(with-open [conn (jdbc/get-connection db)]
  (jdbc/execute! conn ["INSERT INTO users RECORDS {_id: 'jms', name: 'James'}, {_id: 'joe', name: 'Joe'}"])

  (prn (jdbc/execute! conn ["SELECT * FROM users"])))

(defn write-transit
  (^bytes [v] (write-transit v nil))
  (^bytes [v fmt]
   (with-open [baos (ByteArrayOutputStream.)]
     (-> (transit/writer baos (or fmt :msgpack) {#_#_:handlers transit-write-handlers})
         (transit/write v))
     (.toByteArray baos))))

(defn transit->pgobject [v]
  (doto (PGobject.)
    (.setType "transit")
    (.setValue (String. (write-transit v :json)))))

(defn read-transit
  ([bytes] (read-transit bytes nil))
[...]

PR #3857, opened 2024-11-11
Testing transit params on pg2

egg-juxt:
Related: https://github.com/igrishaev/pg2/issues/18

These are some pg2-specific tests for testing the transit type. There are 2 tests, one providing the encoded string directly, another using pg2 custom types.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -260,7 +260,7 @@ dependencies {
     testImplementation("org.xerial", "sqlite-jdbc", "3.39.3.0")
     testImplementation("org.clojure", "test.check", "1.1.1")
     testImplementation("clj-kondo", "clj-kondo", "2023.12.15")
-    testImplementation("com.github.igrishaev", "pg2-core", "0.1.17")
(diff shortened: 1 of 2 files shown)


egg-juxt:
@jarohen , are these pg2-specific tests interesting to have in our test battery?

egg-juxt:
- [x] We should test transit decoding using connection option "fallback_output_format=transit"

jarohen:
> @jarohen , are these pg2-specific tests interesting to have in our test battery?

absolutely, yep :slightly_smiling_face:

jarohen (approved):

egg-juxt:
Added testing with transit as format for the output results, as well as for binary transit.

@jarohen , @mbutlerw , do we really want to use the transit+json format when doing pg binary encoding, or should we switch to transit+msgpack for that case?

jarohen:
> do we really want to use the transit+json format when doing pg binary encoding, or should we switch to transit+msgpack for that case?

probably, yep - but I wouldn't couple that change into here :) and tbh I'd look into getting #3844 solved first, see how people use it, and then we're in a better place to make that call

PR #3859, opened 2024-11-12
docs restructuring, separating SQL/XTQL

### LINKED ISSUE #3765: Separate documentation for the standard library into `XTQL` and `SQL`
Currently the standard lib documentation mixes XTQL and SQL which makes it especially confusing if certain functions/bits are not available in SQL.

jarohen:
* ~slimmed down intro section to just getting-started, everything else can come separately~ separate PR
* create operations section where we can start adding ops guides/reference
* industry use cases -> tutorials until they can be moved to the main site (xtdb/xtdb-website#114)
* ~lose appendices - mission to the main intro (although arguably this should be main site too). others - unless/until we can find a more natural home for them, I don't want a dumping ground~
* separate out SQL and XTQL reference to avoid confusion for SQL users


refset:
Initial thoughts:

1. The 'mission' text has been inlined to the welcome page, but it's likely far too much detail for that point in the journey - I think it was @Akeboshiwind's suggestion to have the Play example prominently on the welcome page currently, maybe it can be changed/elaborated if you want it to be more mission-oriented?
2. The proposed flow suggests people download a docker container _before_ they're presented with any examples of what XTDB can actually do (the SQL Quickstart was created for people to interactively learn about XTDB as quickly as possible)
3. At least some of the contents within How XTDB Works (/ 'Data Model') ought to live in the docs/reference *somewhere*, e.g. noting the mandatory `_id` and built-in columns per https://docs.xtdb.com/intro/data-model.html#_temporal_columns (also somewhere where we should be noting guidance on UUIDs)

jarohen:
thanks @refset - still in flux, but good to get feedback nonetheless :pray: 

> The 'mission' text has been inlined to the welcome page, but it's likely far too much detail for that point in the journey

Agreed - I'm much more in favour of moving that to the main site. I'd propose that we work to the assumption that, once people hit the reference docs site, their aims are now very much more about using/learning the finer details of XTDB rather than deciding whether it's a good fit - if that means we move some of the more tutorial-like 'what is XTDB' articles over too, I'd see that as a good thing.

Ideally, this introduction page could be reduced to a high-level map of the docs site, pointing people to where they need to go next, or back to the main site if they've landed here unsure of what XT is/why they should use it.

> The proposed flow suggests people download a docker container before they're presented with any examples of what XTDB can actually do (the SQL Quickstart was created for people to interactively learn about XTDB as quickly as possible)

Again, this PR is assuming that the main site has already gotten users to the point of knowing they want to try XTDB - at this stage, they're given the choice of Play vs Docker, and then pointed to the initial tutorials (including the SQL quick-start).

The intention here is more to decouple the 'how to try it' vs the 'what to try' instructions - in trying to do both of those at once, we tend to lose clarity whereas, by having a separate article with these instructions, the rest of the site can assume the user's already able to try things out.

> At least some of the contents within How XTDB Works (/ 'Data Model') ought to live in the docs/reference _somewhere_

I agree with this too - my main aim here (as mentioned in the top message) is to avoid a dumping ground of everything that we couldn't quite place, because we've only ever seen pages like that become more incoherent over time. If this forces us to actually decide where these belong, that's can only be a good thing :)

Cheers @refset :pray:

docs/src/content/docs/reference/main/sql/queries.adoc
```diff
@@ -2,7 +2,7 @@
 title: SQL Queries
 ---
 
-For examples on how to run SQL queries in each client library, see the link:/clients[individual driver documentation].
+For examples on how to run SQL queries in each client library, see the link:/drivers[individual driver documentation].
 
 == Top-level queries
```
refset, line 8: (Can't leave a comment at the bottom of this file, but...) There's a stray NOTE section on XTQL/Clojure mappings

refset:
Thanks for the explanations. Happy to see this merged as-is then and use it as motivation to revamp the main site :+1:

FiV0:
I have not fully read all the changes, but the `Reference` section is collapsing whenever you click on something.

jarohen:
> Thanks for the explanations. Happy to see this merged as-is then and use it as motivation to revamp the main site 👍

Sorry @refset, didn't spot this one, my PR page didn't refresh. I've backed out the changes to intro so that I can land the changes for #3765 - didn't want to block this one on getting buy-in for the intro changes :slightly_smiling_face: 

Let's chat separately about the intro refactor?

PR #3867, opened 2024-11-13
[docs] restructure intro section

jarohen:
(from #3859)


jarohen:
closing for now - @refset let's re-open if we come back to it at some point

PR #3874, opened 2024-11-15
Add txId and txIdLag gauges

### LINKED ISSUE #3871: Adding `tx-id`/`tx-lag` gauges.
To better monitor the indexing process, want to be able to easily get gauges against the following:
- `latest-submitted-tx`.
- `latest-completed-tx`.
- `tx-lag` (`latest-submitted-tx` - `latest-completed-tx`)

danmason:
Resolves #3871 

Adds in some basic txId gauges for the sake of monitoring:
- latestSubmittedTxId
- latestCompletedTxId
- txId lag.

Based on recent changes, behavior is as follows:
- Return whatever we get for `submittedTxId` (ie, even if it's -1/not yet set)
- Return whatever we get for `completedTxId`, or `0` if not found.
- When calculating the txId lag:
  - If submittedTxId < 0 (ie, it is unset), return lag as 0.
  - If completedTxId not set, return lag as 0.
  - Otherwise, return txId lag as `(submittedTxId - completedTxId)`.


jarohen (changes requested):
> Return whatever we get for completedTxId, or 0 if not found.

txId = 0 is potentially a valid transaction ID - could we return -1 here?

jarohen (approved):

PR #3885, opened 2024-11-20
XTDB JDBC module

jarohen:
Don't get too excited about the JDBC Driver here - it's literally just wrapping the Postgres one :sweat_smile: 

Main point of this is to create the `xtdb-jdbc` module, to which we'll add the next.jdbc helpers for #3844 

...

but you can now replace `jdbc:postgres://localhost/your_db` with `jdbc:xtdb://localhost/your_db`, so that's nice I guess :relaxed:

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -225,6 +225,7 @@ dependencies {
 
     projectDep(":xtdb-api")
     projectDep(":xtdb-core")
+    projectDep(":xtdb-jdbc")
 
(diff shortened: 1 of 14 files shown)


seancorfield:
Am I correct that there isn't a 2.0 SNAPSHOT of this new module yet, that could be used in place of a PG dependency to test against with `:dbtype "xtdb"`? See https://github.com/seancorfield/next-jdbc/issues/288

Issue #3909, opened 2024-11-26
Comprehensive native type handling over pgwire


refset:
- [X] UUID - https://www.postgresql.org/docs/current/datatype-uuid.html Currently, attempting to return a UUID causes a server error: `Unexpected type encountered by pgwire (class java.util.UUID)`
- [x] VARBINARY - example in the docs is `X('41af8e01')` gives `java.lang.AbstractMethodError: Receiver class xtdb.sql.plan.ExprPlanVisitor does not define or inherit an implementation of the resolved method 'abstract java.lang.Object visitBinaryStringLiteral(xtdb.antlr.Sql$BinaryStringLiteralContext)' of interface xtdb.antlr.SqlVisitor.`
- [x] URI - example in the docs `SELECT URI 'https://xtdb.com';` gives `- Errors parsing SQL statement:  - line 1:11 extraneous input ''https://xtdb.com'' expecting {<EOF>, ';'}`
- [ ] ???

The [json-clj](https://github.com/xtdb/xtdb/blob/069aa683a28e9c1be8a9a9c309463cc328010ebf/core/src/main/clojure/xtdb/pgwire.clj#L541) function needs to be extended. However, in light of this and [another recent example](https://github.com/xtdb/xtdb/issues/3866) we should exhaustively make sure that all native types can be handled (/ be clear on which ones can't).

refset:
UUID now support fixed in https://github.com/xtdb/xtdb/pull/3911

jarohen:
@refset - what's the gap between current state and calling this done?

I'll optimistically close given it was moved to 'Done' but feel free to re-open and flesh out the checklist

refset:
Just added two more - again, not exhaustive. Might be safest to comment out problematic types in the docs until we're ready to support them.

FiV0:
We should now support everything on https://docs.xtdb.com/reference/main/data-types.html with all the caveats of pgwire conversions (Is there an equivalent pg type? Is it nested? etc...).

PR #3911, opened 2024-11-26
support UUID #3909

mpisanko:
Elixir test now also works


lang/elixr/test/elixr_test.exs
```diff
@@ -70,19 +70,28 @@ defmodule ElixrTest do
...
+    assert 1 == id
+    assert ~D[2020-01-01] == dat
+  end
 
   # test "should be able to insert a JSON field and read it back", state do
   #   pid = state[:xt]
   #   json_field = %{"foo" => "bar"}
-  #   Postgrex.query!(pid, "INSERT INTO foo8 (_id, json) VALUES (1, '{\"foo\": \"bar\"}'::JSON)", [])
-  #   result = Postgrex.query!(pid, "SELECT * FROM foo8", [])
+  #   Postgrex.query!(pid, "INSERT INTO foo9 (_id, json) VALUES (1, '{\"foo\": \"bar\"}'::JSON)", [])
```
FiV0, line 93: Try to avoid these changes. They are o/w just distracting to the reviewer.

FiV0 (approved):
LGTM

core/src/main/clojure/xtdb/expression.clj
```diff
@@ -240,6 +240,10 @@
 (defmethod codegen-cast [:num :utf8] [_]
   {:return-type :utf8, :->call-code #(do `(resolve-utf8-buf (str ~@%)))})
 
+(defmethod codegen-cast [:utf8 :uuid] [_]
+  {:return-type :uuid, :->call-code #(do `(util/uuid->byte-buffer
```
FiV0, line 244: @mpisanko Was just relooking at this. I think you wan to return a UUID here.

mpisanko: looks like that - but having raised the inner form - a test fails (https://github.com/xtdb/xtdb/blob/main/src/test/clojure/xtdb/pgwire_test.clj#L1726) - so i guess it's allright - or the test is wrong?

FiV0: Yes, forget about this comment. I think we are dealing with multiple types (including UUID) in byte buffer form in the EE.

PR #3919, opened 2024-11-27
Make RecordBatch (Page) the main unit of memory handed out by the BufferPool

### LINKED ISSUE #3891: Memory Management
The buffer pool (BP henceforth) deals with reading and writing files. The ticket concerns the reading side of the BP. As XT is an append only system the reading of pages does not interfere with the writing of pages. The buffer pool also deals with management of disk caches which we will not be concerned with here. In the following record batch and page is somewhat used interchangeably. 

The memory management (read-side) as it currently stands:
- queries and also DML (if it is not a simple put statement) require to read from certain files when scanning
- the BP memory maps the file and scan then reads the footer and loads a page accordingly
- If multiple queries take out the same buffer the ref count is increased and only after the last query closes the buffer pool may decide to evict the mmap buffer from the cache.
- The upside of this approach is that due to memory-mapped files we do not have to deal with memory management too closely.
- The downside is that due to memory mapping we actually have very little good information of how much memory is used and are somewhat at the whim of the OS paging policies (which might not be very good for databases).
- The BP is also currently non blocking (as blocking introduces all kinds of other issues, more on that below). This means that files are memory mapped even if the maximum cache size is exceeded (see for example https://github.com/xtdb/xtdb/issues/3872). 

The following is an outline of some changes to the buffer pool to potential deal with the memory management finer grained ourselves:
- Allow for maximum size record batch (not necessarily in intermediate operators (join, sort etc..)) but for scan, ie. on disk.
  - The preliminary work would be to get some estimate on the size in bytes of a Arrow vector/relation.
- A query then has an upper bound on how much memory it will be using from the buffer pool (roughly the number of scans)
- Based on this upper bound a query gets queued until the buffer pool can serve the required number of record batches.
   - Issue: A query in the indexer should never be blocked.
- This introduces a finer hierarchy in the buffer pool. The unit of memory handed out is then a record batch + header rather than a whole memory mapped file. The record batch memory is than reused for subsequent record batch loading. There are two options here:
   - The buffer pool only deals in record batches and can hand out the same record batch to multiple queries. This means the buffer pool is mainly concerned with reusing the memory for handed back record batches.
   - The buffer pool hands out something that can hold a record batch and the footer and the queries (scan) themselves deal with loading the appropriate data into the record batch. This kind of closes the door on sharing pages as any query loads whatever it needs into the slot for the record batch.
- The blocking of BP requests might be a Pandora's box. Do we need to be able to preempt other BP usage for high priority requests?  Should there be timeouts for BP requests? Do we then need a query schedular?
   
Potential issues:
- The meta file (trie file) is currently always one record batch, so we might also need to buffer that if the tries become large. As it currently stands this likely secondary, as these files are << compared to data files. 
- An indexing thread query should essentially never be blocked. Do we need to preempt other queries? Do we need priorities for requests to the buffer pool?

Stuff to maybe look into:
- content-addressable file-systems/caches
- `O_Direct`
- `fadvise`

Other databases approaches:
- timeout and freeup the resources - this kind of assumes this scenario does not happen very often
- deadlock detection
- query schedular
- lock the the whole db (sqllite) - I don't think this is an option

Out of scope:
- batching and memory mangement in operators

Progress
- [x] Passed to memory-mapped record batches in #3919

FiV0:
The BufferPool interface essentially passes from:
```kotlin
    fun getBuffer(key: Path): ArrowBuf
```
to 
```kotlin
    fun getByteBuffer(key: Path): ByteBuffer

    fun getFooter (key: Path): ArrowFooter

    fun getRecordBatch(key: Path, blockIdx: Int): ArrowRecordBatch
```

Most of the heavy lifting happens in `getRecordBatch`.  For building up a `xtdb.arrow.Relation` we sometimes need a `Schema`, this is the main reason for the extra `getFooter` call.  Some files (only `transit` as far as I know) don't know anything about arrow, hence the `getByteBuffer` method. The `getByteBuffer` method should only be used on smallish files and for testing. Both `getFooter` and `getByteBuffer` don't have any impact on the direct memory footprint of the buffer pool and allocate their object on the heap (albeit they have an impact on disk cache). 

Things to consider / to do:
- I mostly just added stuff to `xtdb.arrow.Relation` to make things work. It would probably be good to figure out what we want from `Relation` and from which objects it can be instantiated.
- The buffer pool code has grown quite a bit and I suspect that things can be refactored, but I want to do that once the interfaces have settled.
- I am really unhappy about the interaction between `BufferPool`, `MemoryCache` and `DiskCache`. The way their respective interfaces interact with one another is very non transparent and one has to essentially look at the implementations to figure things out. I wonder if there is not some way to make things simpler (you should not need to know anything about the disk cache when writing some function for the memory cache). I might open a another card for that.
- Apart from some tests in `buffer-pool-test` nothing is really testing the interaction of `MemoryCache` and `DiskCache` in the standard suite. It would probably be good to add another test for that.

### DIFF
--- a/core/src/main/clojure/xtdb/buffer_pool.clj
+++ b/core/src/main/clojure/xtdb/buffer_pool.clj
@@ -7,26 +7,32 @@
             [xtdb.object-store :as os]
             [xtdb.util :as util])
   (:import (clojure.lang PersistentQueue)
+           (com.github.benmanes.caffeine.cache AsyncCache Cache Caffeine)
            [io.micrometer.core.instrument MeterRegistry]
            [io.micrometer.core.instrument.simple SimpleMeterRegistry]
            [java.io ByteArrayOutputStream Closeable]
            [java.lang AutoCloseable]
            (java.nio ByteBuffer)
-           (java.nio.channels Channels ClosedByInterruptException)
-           [java.nio.file FileVisitOption Files LinkOption Path]
+           (java.nio.channels Channels ClosedByInterruptException FileChannel$MapMode SeekableByteChannel)
+           [java.nio.file FileVisitOption Files LinkOption OpenOption Path StandardOpenOption]
            [java.nio.file.attribute FileAttribute]
            [java.util NavigableMap SortedMap TreeMap]
            (java.util NavigableMap)
+           (java.util.function Function)
            [java.util.concurrent CompletableFuture ConcurrentSkipListMap]
            kotlin.Pair
+           (org.apache.arrow.flatbuf Footer Message RecordBatch)
            [org.apache.arrow.memory ArrowBuf BufferAllocator]
            (org.apache.arrow.vector VectorSchemaRoot)
-           (org.apache.arrow.vector.ipc.message ArrowBlock ArrowFooter ArrowRecordBatch)
+           (org.apache.arrow.vector.ipc SeekableReadChannel)
+           (org.apache.arrow.vector.ipc.message ArrowBlock ArrowFooter ArrowRecordBatch )
            (xtdb ArrowWriter IBufferPool)
            (xtdb.api.log FileListCache FileListCache$Subscriber)
            (xtdb.api.storage ObjectStore Storage Storage$Factory Storage$LocalStorageFactory Storage$RemoteStorageFactory)
            xtdb.api.Xtdb$Config
+           (xtdb.arrow Relation)
            (xtdb.cache DiskCache DiskCache$Entry MemoryCache)
+           (org.apache.arrow.vector.ipc.message ArrowBlock ArrowFooter MessageSerializer)
            (xtdb.multipart IMultipartUpload SupportsMultipart)))
 
 (set! *unchecked-math* :warn-on-boxed)
@@ -50,6 +56,12 @@
       (metrics/add-allocator-gauge metrics-registry "buffer-pool.allocator.allocated_memory" child-allocator))
     child-allocator))
 
+(defn copy-bb ^ByteBuffer [^ByteBuffer bb]
+  (let [copy (ByteBuffer/allocate (.remaining bb))]
+    (.put copy bb)
+    (.flip copy)
+    copy))
+
 (defrecord MemoryBufferPool [allocator, ^NavigableMap memory-store]
   IBufferPool
   (getBuffer [_ k]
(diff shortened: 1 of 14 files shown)


core/src/main/kotlin/xtdb/IBufferPool.kt
```diff
@@ -1,12 +1,33 @@
...
 import java.nio.file.Path
 
 interface IBufferPool : AutoCloseable {
-    fun getBuffer(key: Path): ArrowBuf
+    /**
+     * Get the whole file as an on-heap ByteBuffer.
+     *
+     * This should only be used for small files (metadata) or testing purposes.
+     */
+    fun getByteBuffer(key: Path): ByteBuffer
```
jarohen: Apart from tests, it seems like the only usages you have of this would all prefer to have a byte-array, which is easier to get hold of from an ArrowBuf, and also more obviously on-heap

core/src/main/clojure/xtdb/buffer_pool.clj
```diff
@@ -42,21 +48,45 @@
...
-(defn- retain [^ArrowBuf buf] (.retain (.getReferenceManager buf)) buf)
-
 (defn ->buffer-pool-child-allocator [^BufferAllocator allocator, ^MeterRegistry metrics-registry]
   (let [child-allocator (.newChildAllocator allocator "buffer-pool" 0 Long/MAX_VALUE)]
     (when metrics-registry
       (metrics/add-allocator-gauge metrics-registry "buffer-pool.allocator.allocated_memory" child-allocator))
     child-allocator))
 
+(defn copy-bb ^ByteBuffer [^ByteBuffer bb]
+  (let [copy (ByteBuffer/allocate (.remaining bb))]
```
jarohen: ensure this doesn't mutate the position of the original bb - if so, you might want to `duplicate` it before you use it.

core/src/main/clojure/xtdb/buffer_pool.clj
```diff
@@ -116,18 +146,74 @@
...
+            (util/with-open [arrow-buf (.get memory-cache (PathSlice. k (.getOffset block)
+                                                                      (+ (.getMetadataLength block) (.getBodyLength block)))
+                                             (fn [^PathSlice path-slice]
+                                               (let [buffer-cache-path (.resolve disk-store (.getPath path-slice))]
+                                                 (when-not (util/path-exists buffer-cache-path)
+                                                   (throw (os/obj-missing-exception k)))
+
+                                                 (CompletableFuture/completedFuture
+                                                  (Pair. (->resolved-path-slice path-slice buffer-cache-path) nil)))))]
+              (let [prefix-size (if (= (.getInt arrow-buf 0) MessageSerializer/IPC_CONTINUATION_TOKEN) 8 4)
```
jarohen: can this come out into a shared fn with the remote store? it's a pretty gnarly one to duplicate.

I think (although may be wrong) this is already written in Kotlin - if it's not quite there, would it be worth extracting there so that you can use it here?

jarohen (commented):
Hey @FiV0, have had a first pass, looking good - worth a chat tomorrow regarding the AM performance?

core/src/main/clojure/xtdb/buffer_pool.clj
FiV0: When I apply the `ByteBuffer` to `ByteArray` chang above, this will go away anyway.

core/src/main/kotlin/xtdb/arrow/Relation.kt
```diff
@@ -284,6 +286,14 @@ class Relation(val vectors: SequencedMap<String, Vector>, override var rowCount:
 
         @JvmStatic
         fun fromRoot(vsr: VectorSchemaRoot) = Relation(vsr.fieldVectors.map(Vector::fromArrow), vsr.rowCount)
+
+        @JvmStatic
+        fun fromRecordBatch (schema: Schema, recordBatch: ArrowRecordBatch) : Relation {
+            val rel = Relation(recordBatch.buffers.first().referenceManager.allocator, schema)
+            // this load retains the buffers
+            rel.load(recordBatch)
+            return rel
```
FiV0, line 283: Pass in the allocator.

PR #3920, opened 2024-11-27
add PREPARE/EXECUTE statements

### LINKED ISSUE #3855: Implement pgwire PREPARE
`PREPARE` is useful for making `.sql` data loading scripts run fast (by avoiding re-parsing and re-compiling ~identical statements).

https://www.postgresql.org/docs/current/sql-prepare.html

jarohen:
This PR adds `PREPARE` and `EXECUTE` statements, mostly compatible with Postgres's equivalents.

* Can be used for both DML and DQL - DML buffers up transaction operations in the same way. We add batching here such that if (exactly) the same SQL string is submitted multiple times in a row within the same transaction, the arguments are combined into one operation.
* Implementation-wise, PREPARE and EXECUTE both run through the full parse/bind/execute lifecycle individually, as they're SQL statements in their own right - I'll use capitals vs lower to distinguish between the SQL operator and the pgwire message respectively.
  * When PREPARE is parsed, we also prepare the inner query, so there are two statements prepared here - we'll call them outer (the whole query, `PREPARE foo AS ...`) and inner (just the `...`).
  * There are no params in the outer PREPARE query, so bind does very little.
  * Execute for PREPARE saves the prepared query in the session - after this, we're in the same state as using the same statement prep over through the PG wire-protocol.
  * The client then sends through parse/bind/execute for the EXECUTE statement too. For the EXECUTE statement, we have to run two queries - one to evaluate the arguments, the second to run the actual prepared statement.
  * Parse for the EXECUTE statement creates the arguments prepared statement, essentially a `:table` operator so that literals in the EXECUTE statement are evaluated as they normally would be.
    * Users could, for example, call `EXECUTE foo($2, $1)`, or mix literals/params (`EXECUTE foo('literal', $1)`), so we can't just pass the arguments from the EXECUTE statement straight to the statement we prepared - hence this mapping.
  * Binding the EXECUTE statement does more work than the usual bind. Here, we bind and execute the args query, and then bind the result of that to the original prepared statement. From bind, we return a normal bound query - the prepared statement with the mapped args.
  * Executing the EXECUTE statement is just executing this prepared statement as before - indeed, there's no special treatment here.
* One caveat is with queries like `PREPARE foo AS SELECT $1` - when this is planned there's no way of knowing what types the query will return. In fact, 'real' Postgres will always return $1 here as a string regardless of what's passed in. Thankfully, this is only an artifact of the client - within the database itself, in the EXECUTE statement, the values are correctly typed - so INSERTs et al propagate values through correctly.
  * We don't support Postgres's ability to pass the parameter data types to the `PREPARE` statement.

resolves #3855


core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1587,6 +1656,7 @@
               (cmd-write-msg conn msg-command-complete {:command "COMMIT"}))
 
     :query (cmd-exec-query conn portal)
+    :prepare (cmd-prepare conn portal)
```
mpisanko, line 1659: are we missing a case for `:execute` here? or is that covered by handle-simple-query? (but handle simple query only deals with executing queries - not DMLs?

jarohen: heh, no, we don't need one now 🙂

so the bind for EXECUTE binds and returns the PREPAREd statement rather than the EXECUTE (`assoc inner` rather than `assoc stmt`), which means that rather than `statement-type :execute`, it now carries the statement-type of the underlying statement - `:query` or `:dml`, the `:execute` never makes it that far.

saved me a whole bunch of duplicate code, that one 😊

PR #3921, opened 2024-11-28
Clear leftover `openArrowWriter` temp files on close

### LINKED ISSUE #3913: `java.io.IOException: No space left on device` within Auctionmark Benchmark
Seen this on the 24hour runs - some notes on the config where this happens:
- Each node has a 50GiB persistent volume claim attached.
- One node failing with the error - running out of space within the PvC.
- The nodes are configured with a "maxdiskcachepercentage" of 75% - so seemingly we must go over that by over 12.5GiB.

## Investigation

```
java.util.concurrent.ExecutionException: clojure.lang.ExceptionInfo: Ingestion stopped: java.util.concurrent.ExecutionException: java.io.IOException: No space left on device {}
    at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
...
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
    at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: clojure.lang.ExceptionInfo: Ingestion stopped: java.util.concurrent.ExecutionException: java.io.IOException: No space left on device
    at xtdb.await$__GT_ingester_ex.invokeStatic(await.clj:17)
    at xtdb.await$__GT_ingester_ex.invoke(await.clj:16)
    at xtdb.await$await_tx_async$fn__21938.invoke(await.clj:27)
    at xtdb.await$await_tx_async.invokeStatic(await.clj:20)
    at xtdb.await$await_tx_async.invoke(await.clj:19)
    at xtdb.indexer.Indexer.awaitTx(indexer.clj:749)
    ... 30 common frames omitted
```

# Notes / Graphs

## Failing Run

**Here's a view of the diskcache usage from the run where the node died with the above error:**
image

Sure enough, the cache was pretty full when the no disk space error hit, though don't see any immediate problems here unless there was 10gigs of other "stuff" hanging around.

## Current Run

Decided to look at the CURRENT run we have going, with a 100GiB persisent volume claim instead.

**Here's some views over the current diskcache for node 1:**
image

As we can see:
- All stacks up to 75GiB (as expected with a volume claim of 100GiB and 75% max disk cache percentage)
- Current "in use" bytes of that should be pinned bytes + evictable bytes
- This currently totals around 44GiB.

Keeping the above in mind, decided to run `df -h` against that node, and saw the following:
```
Filesystem      Size  Used Avail Use% Mounted on
...
/dev/sde         98G   59G   40G  60% /var/lib/xtdb/buffers
...
```

59GiB! There's a gap between that and the current "in use" bytes of the bufferpool by about 15GiB. Looking at the disk usage of the `buffer` file directory:
```
15G	/var/lib/xtdb/buffers/disk-cache-1/.tmp
104K	/var/lib/xtdb/buffers/disk-cache-1/chunk-metadata
45G	/var/lib/xtdb/buffers/disk-cache-1/tables
59G	/var/lib/xtdb/buffers/disk-cache-1
```

Sure enough, 15GiB of usage in the `.tmp` folder! That WOULD be enough to push our original node over the edge, ie:
- 37.5GiB buffer pool (at total max capacity)
- 15GiB temp directory.
=
52.5GiB total size! Would exceed the total size of the persisent volume claim. 

## Summarized Problem

- We have around 15GiB of temp currently, which in the 50GiB persistent volume case would have pushed us into "run out of disk space" territory.
- All of the 15GiB of current temp are "upload<num>.arrow" files, created by the disk cache.
- Usual behaviour of "open arrow writer":
  - Call to upload-arrow-file on the object store ->
  - We write a file notification to the file notification topic ->
  - We call to .put on the diskcache, with the key and tmp-path.
    - put on the diskcache is responsible for doing the atomic move from the tmp-path to the disk cache directory.
  - So only put on the diskcache is going to clear up the tmp-path made by open-arrow-writer.

Now, we've seen a recurring info/warn of `15:47:29.039 [xtdb.compactor-558] INFO  xtdb.azure.object-store - Blob already exists for x - aborting multipart upload` type errors - Ie, when another node has already uploaded a file:
- In THAT case, we abort the multipart and throw an error in upload-arrow-file,
- Ie, we wouldnt call .put on diskcache - wouldnt make sense to!  

In that case - I don't think anything clears up the temp file, indeed, looking at the most recent temp file that has stuck around:
```
15:47 upload11723359101008666313.arrow
```

This being so - should be able to write a test against this and fix it fairly easily.

danmason:
Resolves #3913 

Essentially, we were relying on the behaviour of the DiskCache to atomically move the temp file to the disk cache location - so if there was an error thrown inside of `.end` (ie, the file already existed on the object store when we called `upload-arrow-file`), the temp files would never get cleared up, leading to use eventually running out of disk space even with a generously low `maxDiskCachePercentage` size set. 

The approach taken to fix this is to rely on `close` being called on the `ArrowWriter` (which it should be, as we rely on it to close any open filechannels as well, and usages of open-arrow-writer are inside of `with-close-on-catch` blocks). 

Keen to test this against auctionmark (ie, setting persistent volume claim sizes smaller once again and keeping an eye on the size of `.tmp` directory).

### DIFF
--- a/core/src/main/clojure/xtdb/buffer_pool.clj
+++ b/core/src/main/clojure/xtdb/buffer_pool.clj
@@ -159,7 +159,7 @@
   (objectSize [_ k] (Files/size (.resolve disk-store k)))
 
   (openArrowWriter [_ k rel]
-    (let [tmp-path (create-tmp-path disk-store)]
+    (let [^Path tmp-path (create-tmp-path disk-store)]
       (util/with-close-on-catch [file-ch (util/->file-channel tmp-path util/write-truncate-open-opts)
                                  unl (.startUnload rel file-ch)]
         (reify ArrowWriter
@@ -180,7 +180,9 @@
           (close [_]
             (util/close unl)
             (when (.isOpen file-ch)
-              (.close file-ch)))))))
+              (.close file-ch))
+            ;; If tmp path still present/hasn't been atomic moved, delete it on close
+            (Files/deleteIfExists tmp-path))))))
 
   EvictBufferTest
   (evict-cached-buffer! [_ k]
@@ -317,7 +319,7 @@
     (.get !os-files k))
 
   (openArrowWriter [_ k rel]
-    (let [tmp-path (.createTempPath disk-cache)]
+    (let [^Path tmp-path (.createTempPath disk-cache)]
       (util/with-close-on-catch [file-ch (util/->file-channel tmp-path util/write-truncate-open-opts)
                                  unl (.startUnload rel file-ch)]
 
@@ -337,7 +339,9 @@
           (close [_]
             (util/close unl)
             (when (.isOpen file-ch)
-              (.close file-ch)))))))
+              (.close file-ch))
+            ;; If tmp path still present/hasn't been atomic moved, delete it on close
+            (Files/deleteIfExists tmp-path))))))
 
   (putObject [_ k buffer]
     @(-> (.putObject object-store k buffer)
--- a/src/test/clojure/xtdb/buffer_pool_test.clj
+++ b/src/test/clojure/xtdb/buffer_pool_test.clj
@@ -329,3 +329,41 @@
   (tu/with-tmp-dirs #{tmp-dir}
     (with-open [bp (bp/open-local-storage tu/*allocator* (Storage/localStorage tmp-dir) (SimpleMeterRegistry.))]
       (test-list-objects bp))))
+
+(defn count-tmp-files [^Path tmp-dir]
+  (count (.listFiles (.toFile tmp-dir))))
+
+(t/deftest buffer-pool-clears-up-arrow-writer-temp-files
+  (with-open [bp (remote-test-buffer-pool)]
+    (let [^Path root-path (.getRootPath ^DiskCache (:disk-cache bp))
+          ^Path tmp-dir (.resolve root-path ".tmp")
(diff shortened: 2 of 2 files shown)


core/src/main/clojure/xtdb/buffer_pool.clj
```diff
@@ -180,7 +180,9 @@
           (close [_]
             (util/close unl)
             (when (.isOpen file-ch)
-              (.close file-ch)))))))
+              (.close file-ch))
+            ;; If tmp path still present/hasn't been atomic moved, delete it on close
+            (Files/deleteIfExists tmp-path))))))
```
FiV0, line 185: Do we always call close when the `end` call errors (or `writeBatch`)? If that holds then that seems reasonable.

FiV0: Your tests suggest so, so all good.

danmason: Admittedly my test does explicitly use `with-open`, so that does explicitly close it - I believe in the places where we are using `OpenArrowWriter` (ie, in trie.clj) it's within either a `with-close-on-catch` or a `with-open`, so in both cases it should get closed.

I'll update the test to use `with-close-on-catch` since it's a bit more accurate.

jarohen (approved):

PR #3922, opened 2024-11-28
Add `cleanUp` method to metadata Manager

FiV0:
We are removing any retaining of buffers from the Metadata Manager and only cache the table metadata indices. 

Superseded metadata buffers were not getting cleared from the metadata manager. As the metadata manager does no longer hold on to those this should no longer be a problem. We are also adding a `cleanUp` method to the metadata manager to clear superseded metadata indices which are retained on heap. 

The first commit is just some reshuffling of metadata functions to not get a cyclic dependency with the trie namespace.

Todo:
- We could also pass to soft references for the table metadata indices to better deal with heap memory pressure.

### DIFF
--- a/core/src/main/clojure/xtdb/indexer/live_index.clj
+++ b/core/src/main/clojure/xtdb/indexer/live_index.clj
@@ -355,7 +355,12 @@
             (.finishChunk metadata-mgr chunk-idx
                           {:latest-completed-tx latest-completed-tx
                            :next-chunk-idx next-chunk-idx
-                           :tables table-metadata}))))
+                           :tables table-metadata})
+            (future
+              (try
+                (.cleanUp metadata-mgr)
+                (catch Exception e
+                  (log/error e "Error while cleaning up the metadata manager.")))))))
 
       (.nextChunk row-counter)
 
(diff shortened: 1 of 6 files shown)


core/src/main/clojure/xtdb/indexer/live_index.clj
```diff
@@ -355,7 +355,12 @@
             (.finishChunk metadata-mgr chunk-idx
                           {:latest-completed-tx latest-completed-tx
                            :next-chunk-idx next-chunk-idx
-                           :tables table-metadata}))))
+                           :tables table-metadata})
+            (future
+              (try
+                (.cleanUp metadata-mgr)
+                (catch Exception e
+                  (log/error e "Error while cleaning up the metadata manager.")))))))
```
FiV0, line 363: Not sure this is the right place or way to do it, but wanted to start with something.

core/src/main/clojure/xtdb/metadata.clj
```diff
@@ -402,6 +217,9 @@
   (columnFields [_ table-name] (get fields table-name))
```
FiV0, line 207: Try without cache.

FiV0:
After https://github.com/xtdb/xtdb/commit/611fb3d32a830766e9a33b1e9c70835ab7f9f7f7 (which was the first part of ths PR) this might not make much of a difference, so closing for now unless this shows up as a bottleneck again.

PR #3925, opened 2024-11-29
add a test for keyword transformation #3910

### LINKED ISSUE #3910: xtdb.next.jdbc should transform/keywordize nested return values
e.g. with the `:builder-fn` [configured](https://github.com/xtdb/xtdb/pull/3886/files#diff-23eaea0515eaa95e7e73b79d824b6cea2b1676e3491e587ad1ef2857a5e3da85R41-R47) the following is observed:

```
SELECT p.*, 
       NEST_MANY(
           SELECT bs.bdr$submission$state
           FROM bdr_submission FOR ALL SYSTEM_TIME bs
           WHERE bs.project$identifier = p._id
           
       ) AS submission_history
FROM project p limit 5
=>
{:submission-history
 [{"bdr$submission$state" "Submission complete"}
  {"bdr$submission$state" "Submission in progress"}],
 :project/comment nil,
 :project/start-date #inst "1984-10-31T13:00:00.000-00:00",
 :project/date-modified nil,
 :project/contact-point nil,
 :project/name "Quadra Forest Block",
 :xt/id #uuid "0004f309-728f-479d-a93d-3743ac5ea9dc",
 :project/description nil,
 :project/date-created #inst "2011-08-15T14:00:00.000-00:00",
 :project/vba-identifier 132,
 :project/license nil,
 :project/status "Published",
 :project/end-date #inst "1984-10-31T13:00:00.000-00:00"}
```

mpisanko:
resolves #3910

### DIFF
--- a/jdbc/src/main/clojure/xtdb/next/jdbc.clj
+++ b/jdbc/src/main/clojure/xtdb/next/jdbc.clj
@@ -6,7 +6,7 @@
   (:import clojure.lang.Keyword
            java.nio.charset.StandardCharsets
            [java.sql ResultSet]
-           [java.util Map]
+           [java.util List Map Set]
            org.postgresql.util.PGobject
            xtdb.api.Xtdb
            xtdb.api.query.IKeyFn
@@ -20,11 +20,21 @@
                    (String. StandardCharsets/UTF_8)))))
 
 (defn- denormalize-keys [v, ^IKeyFn key-fn]
-  (w/prewalk (fn [v]
-               (cond-> v
-                 (instance? Map v)
-                 (update-keys #(.denormalize key-fn %))))
-             v))
+  (letfn [(process-map [m]
+            (if (keyword? (first (keys m)))
+              m
+              (update-keys m #(.denormalize key-fn %))))]
+    (w/prewalk (fn [v]
+                 (cond-> v
+                   (instance? Map v)
+                   (update-keys #(.denormalize key-fn %))
+
+                   (instance? List v)
+                   vec
+
+                   (instance? Set v)
+                   set))
+               v)))
 
 (defn ->sql-col [^Keyword k]
   (NormalForm/normalForm k))
@@ -46,6 +56,16 @@
           (denormalize-keys key-fn)))))
 
 (def col-reader (->col-reader :kebab-case-keyword))
+
+
+(def ^{:doc "Builder function which converts result-set rows' map keys to kebab-cased keywords (iteratively)"}
+  builder-fn
+  (nj-rs/as-maps-adapter
+   (fn [rs opts]
+     (nj-rs/as-unqualified-modified-maps
+      rs
+      (assoc opts :label-fn label-fn)))
+   col-reader))
 
 (defmulti <-pg-obj
   (fn [^PGobject obj]
--- a/docs/src/content/docs/drivers/clojure.adoc
+++ b/docs/src/content/docs/drivers/clojure.adoc
@@ -41,11 +41,7 @@ Then, once you've link:/intro/installation-via-docker[started the XTDB node], fo
 
   ;; optional: use the XT col-reader to transform nested values too
   (prn (jdbc/execute! conn ["SELECT * FROM users"]
-         {:builder-fn (jdbc-rs/as-maps-adapter
-                        (fn [rs opts]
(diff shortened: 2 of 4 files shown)


mpisanko:
happy to drop the second commit - just thought it might be nice for users to have a shorthand.

jarohen:
@mpisanko I'm fine with both here, but let's just call it `builder-fn` and document what it does separately - that name's quite a mouthful :sweat_smile:

mpisanko:
a side effect is that nil columns are now returned - if that's ok will squash these commits, otherwise easy to take it out.

jdbc/src/main/clojure/xtdb/next/jdbc.clj
```diff
@@ -47,6 +58,16 @@
 
 (def col-reader (->col-reader :kebab-case-keyword))
 
+
+(def ^{:doc "Builder function which converts result-set rows' map keys to kebab-cased keywords (iteratively)"}
+  builder-fn
+  (nj-rs/as-maps-adapter
+   (fn [rs opts]
+     (nj-rs/as-unqualified-modified-maps
```
jarohen, line 65: let's stick with the one from optional (as the test-util was) - we're looking to round-trip documents as they were inserted, and I think I'd want to err on the side of dropping null values than inserting them (although it's obviously a subjective decision).

jdbc/src/main/clojure/xtdb/next/jdbc.clj
```diff
@@ -20,11 +20,22 @@
...
+            (if (keyword? (first (keys m)))
+              m
+              (update-keys m #(.denormalize key-fn %))))]
+    (w/prewalk (fn [v]
+                 (cond-> v
+                   (instance? Map v)
+                   process-map
+
+                   (instance? Collection v)
+                   (->>
```
jarohen: in theory, given it's a pre-walk, we should be able to just have `(instance? List v) vec` here, and have the pre-walk deal with iterating it?

(also worth us adding `(instance? Set v) set`, if the cause of this one was prewalk not recognising Java collections)

jdbc/src/main/clojure/xtdb/next/jdbc.clj
```diff
@@ -20,11 +20,22 @@
                    (String. StandardCharsets/UTF_8)))))
 
 (defn- denormalize-keys [v, ^IKeyFn key-fn]
-  (w/prewalk (fn [v]
-               (cond-> v
-                 (instance? Map v)
-                 (update-keys #(.denormalize key-fn %))))
-             v))
+  (letfn [(process-map [m]
+            (if (keyword? (first (keys m)))
```
jarohen, line 24: what's the reasoning (/assumption) behind only checking one key of the map here?

jarohen (changes requested):

jdbc/src/main/clojure/xtdb/next/jdbc.clj
mpisanko: resolved by delegating to prewalk which in practice confirmed your theory :)

jdbc/src/main/clojure/xtdb/next/jdbc.clj
```diff
@@ -20,11 +20,21 @@
...
+    (w/prewalk (fn [v]
+                 (cond-> v
+                   (instance? Map v)
+                   (-> (update-keys #(.denormalize key-fn %)))
+
+                   (instance? List v)
+                   vec
+
+                   (instance? Set v)
+                   vec))
```
jarohen: nit: `set`

jdbc/src/main/clojure/xtdb/next/jdbc.clj
```diff
@@ -20,11 +20,21 @@
...
-                 (update-keys #(.denormalize key-fn %))))
-             v))
+  (letfn [(process-map [m]
+            (if (keyword? (first (keys m)))
+              m
+              (update-keys m #(.denormalize key-fn %))))]
+    (w/prewalk (fn [v]
+                 (cond-> v
+                   (instance? Map v)
+                   (-> (update-keys #(.denormalize key-fn %)))
```
jarohen: nit: don't need `->` here

jarohen (approved):
with the two remaining nits addressed, good to go - thanks :pray:

PR #3938, opened 2024-12-03
Option for disabling log catchup on node start

FiV0:
The possibility to disable log catchup on startup
```clj
(xtn/start-node {:node {:log-catchup? false}})
```
Maybe that should also be at the toplevel. I added it at a node key to be in line with all the other components.


jarohen:
closing until we've resolved the open discussion around whether we should (not ruling out re-opening)

I'm not keen on a flag here, because I'm not keen on the behaviour itself and I'm especially not keen on supporting both behaviours through a flag.

Let's take a step back and understand the context before making further decisions.

PR #3940, opened 2024-12-03
Display tables

### LINKED ISSUE #3860: Display tables using the psql `\d` or `\dt` command
The shorthand `\dt` psql command sends this query (`\d` is essentially the same too):
```sql
SELECT n.nspname as "Schema",
  c.relname as "Name",
  CASE c.relkind WHEN 'r' THEN 'table' WHEN 'v' THEN 'view' WHEN 'm' THEN 'materialized view' WHEN 'i' THEN 'index' WHEN 'S' THEN 'sequence' WHEN 't' THEN 'TOAST table' WHEN 'f' THEN 'foreign table' WHEN 'p' THEN 'partitioned table' WHEN 'I' THEN 'partitioned index' END as "Type",
  pg_catalog.pg_get_userbyid(c.relowner) as "Owner"
FROM pg_catalog.pg_class c
     LEFT JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace
     LEFT JOIN pg_catalog.pg_am am ON am.oid = c.relam
WHERE c.relkind IN ('r','p','')
      AND n.nspname <> 'pg_catalog'
      AND n.nspname !~ '^pg_toast'
      AND n.nspname <> 'information_schema'
  AND pg_catalog.pg_table_is_visible(c.oid)
ORDER BY 1,2;
```

The query currently fails because both `pg_catalog.pg_get_userbyid` and `pg_catalog.pg_table_is_visible` are not being recognised and implemented as functions. We could consider hard-coding these functions to return static values.

mpisanko:
The shorthand \dt psql command sends this query (\d is essentially the same too)
The query currently fails because both pg_catalog.pg_get_userbyid and pg_catalog.pg_table_is_visible are not being recognised and implemented as functions. We could consider hard-coding these functions to return static values.


mpisanko:
blocked by #3962

mpisanko:
This PR adds grammar required to use postgres compatible \d, \dt, \d foo
\d and \dt work
to get `\d foo` to work - we'd need to be able to cast string expressed number to proper number (something like: '123'::int - except w/o explicit cast) as that's what's used in the underlying query (issued by psql-client).
hence the comment that this one is blocked by #3962 
currently we will not support `\d foo` - describing a specific table.

mpisanko:
not really sure why the information_schema/pg_catalog tables are showing along the 'public' schema tables (so it appears):
```
mpi@mpimbp ~ [2]> psql -p 5432 -h localhost -U xtdb
psql (17.0, server 16)
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, compression: off, ALPN: none)
Type "help" for help.

xtdb=> \d
              List of relations
 Schema |        Name         | Type  | Owner
--------+---------------------+-------+-------
 public | columns             | table | xtdb
 public | foo                 | table | xtdb
 public | pg_am               | table | xtdb
 public | pg_attribute        | table | xtdb
 public | pg_class            | table | xtdb
 public | pg_database         | table | xtdb
 public | pg_description      | table | xtdb
 public | pg_matviews         | table | xtdb
 public | pg_namespace        | table | xtdb
 public | pg_proc             | table | xtdb
 public | pg_range            | table | xtdb
 public | pg_settings         | table | xtdb
 public | pg_stat_user_tables | table | xtdb
 public | pg_tables           | table | xtdb
 public | pg_type             | table | xtdb
 public | pg_user             | table | xtdb
 public | pg_views            | table | xtdb
 public | schemata            | table | xtdb
 public | tables              | table | xtdb
 public | txs                 | table | xtdb
(20 rows)
```

FiV0:
> not really sure why the information_schema/pg_catalog tables are showing along the 'public' schema tables (so it appears):

I recently added metadata tables to metadata tables https://github.com/xtdb/xtdb/pull/3956. I suspect this is the reason they show up.

core/src/main/clojure/xtdb/information_schema.clj
```diff
@@ -199,6 +216,28 @@
      :nspowner nspowner
      :nspacl nil}))
 
+(def pg-ams [{:oid 2
+              :amname "heap"
+              :amhandler "heap_tableam_handler"
+              :amtype "t"
+             }
+             ;; FIXME this should prolly be something else (arrow-HMT-handler?)
```
FiV0, line 224: Can we try to use correct spelling.

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -1668,11 +1668,27 @@
           data-type (-> (.dataType ctx) (.accept (->CastArgsVisitor env)))]
       (handle-cast-expr ve data-type)))
 
+  (visitCollateExpr [_ ctx]
+    ;; TODO - apply collation
```
FiV0, line 1633: What does this mean? We are now parsing `COLLATE` but are not really applying the logic?

core/src/main/clojure/xtdb/information_schema.clj
FiV0: Likely `HAMTrie`

FiV0 (changes requested):

core/src/main/clojure/xtdb/sql/plan.clj
mpisanko: yeah - no idea what to do. in my understanding collation affects eg string ordering - but have no idea how and where to apply that. also - when trying to apply it - it might have to spread thru the system?

core/src/main/clojure/xtdb/information_schema.clj
mpisanko: probably you mean `prolly`? it's kind of like i18n - so not really keen on changing that. sorry.

core/src/main/clojure/xtdb/sql/plan.clj
FiV0: If a query is using `COLLATE` then it's likely expecting results in some order. I am not saying we need to fully implement it, but it would be good to understand why a query is using `COLLATE` and then decide if or if not to honor the meaning.

mpisanko: the query in this case is:
```sql
 SELECT c.oid,
  n.nspname,
  c.relname
FROM pg_catalog.pg_class c
     LEFT JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace
WHERE c.relname OPERATOR(pg_catalog.~) '^(foo)$' COLLATE pg_catalog.default
  AND pg_catalog.pg_table_is_visible(c.oid)
ORDER BY 2, 3;
```
and default collation is used - so it would affect sort order - but given this is purely psql 'sugar' - think we can get away even in some cases when the tables are slightly 'misordered' (according to default collation)?
but then if users start using COLLATE - it might spell trouble. so maybe it should be documented that it's not supported?

FiV0: But `foo` is just an example here, right? This currently wouldn't work for `FOO` even though that is what the "real" pg would also work with. I am not saying you need to implement it, but lets' document it somewhere. Preferably some top-level comment on this PR and link to it from that `TODO`.

FiV0: If you want you can even open a new ticket for the feature `COLLATE` where you document what we currently do and what would need to happen to fully support `COLLATE`.

mpisanko: opened #4020 to implement COLLATE

jarohen:
prefer rebases to merges, please, and link PR to card :pray: 

https://github.com/xtdb/xtdb/blob/main/dev/GIT.adoc

jarohen:
https://github.com/xtdb/xtdb/pull/3940#issuecomment-2567446761

@FiV0 these should be in the `pg_catalog` schema - not urgent, but one we should card up?

core/src/main/antlr/xtdb/antlr/Sql.g4
```diff
@@ -226,6 +232,7 @@ exprPrimary
     | exprPrimary '.' fieldName #FieldAccess
     | exprPrimary '[' expr ']' #ArrayAccess
     | exprPrimary '::' dataType #PostgresCastExpr
+    | exprPrimary '::' schemaName '.' 'REGTYPE' '::' schemaName '.' 'TEXT' # PgCatalogResolveOid
```
jarohen, line 234: this looks like two distinct expressions? i.e. schemaName.regtype and `expr :: schema.text`?

in fact, the latter looks like it's standard cast syntax, so do we need to make that change there instead?

core/src/main/antlr/xtdb/antlr/SqlLexer.g4
```diff
@@ -265,9 +267,12 @@ PASSWORD: 'PASSWORD' ;
 PATCH : 'PATCH' ;
 PERCENT_RANK : 'PERCENT_RANK' ;
 PERIOD : 'PERIOD' ;
+PG_CATALOG_DEFAULT : 'pg_catalog.default' ;
```
jarohen, line 271: I doubt this is a token in the language, unless we're going to specify every collation as a token - I suspect collation wants to be `(schemaName '.')? identifier`?

core/src/main/clojure/xtdb/information_schema.clj
```diff
@@ -45,7 +45,10 @@
                           typtype :utf8, typbasetype :i32, typnotnull :bool, typtypmod :i32
                           typsend :utf8, typreceive :utf8, typinput :utf8, typoutput :utf8
                           typrelid :i32, typelem :i32}
-      pg_catalog/pg_class {oid :i32, relname :utf8, relnamespace :i32, relkind :utf8}
+      pg_catalog/pg_class {oid :i32, relname :utf8, relnamespace :i32, relkind :utf8, relam :i32, relchecks :i32
+                          relhasindex :bool, relhasrules :bool, relhastriggers :bool, relrowsecurity :bool
```
jarohen, line 49: nit: indent

core/src/main/clojure/xtdb/information_schema.clj
```diff
@@ -199,6 +216,28 @@
      :nspowner nspowner
      :nspacl nil}))
 
+(def pg-ams [{:oid 2
+              :amname "heap"
+              :amhandler "heap_tableam_handler"
+              :amtype "t"
+             }
+             ;; FIXME this should prolly be something else (arrow-HAMTrie-handler?)
```
jarohen: this is the sort of comment to either resolve beforehand, or explicitly decide not to handle now

jarohen (changes requested):

core/src/main/antlr/xtdb/antlr/Sql.g4
mpisanko: right, but then - it's kind of context dependent: regtype can be an int or text (which is the typename - so it's a special case really - where something like [regtype=]23::TEXT - returns 'int4' as opposed to '23'). 
Also - making the cast expression recursive would complicate it w/o much benefit? happy to be told otherwise if you have a strong feeling about this.

jarohen: > Also - making the cast expression recursive would complicate it w/o much benefit?

having two expressions in one rule is more complicated, we don't do this anywhere else - in a grammar, you decompose the syntax down into its constituent parts.

> happy to be told otherwise if you have a strong feeling about this.

yes please :pray:

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -1629,11 +1629,27 @@
...
+    (log/trace "visitCollateExpr" (.getText (.collation ctx)))
+    (.getText (.collation ctx)))
+
+  (visitPostgresOperatorExpr [this ctx]
+    (.visitPostgresRegexPredicate this ctx))
```
jarohen, line 1638: presumably `operator` can take any kind of operator, not just regex?

core/src/main/clojure/xtdb/information_schema.clj
jarohen: > probably you mean prolly? it's kind of like i18n

no it isn't, please fix it.

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -1834,7 +1850,17 @@
...
-        (vary-meta assoc :identifier (->col-sym 'version)))))
+        (vary-meta assoc :identifier (->col-sym 'version))))
+
+  (visitPostgresTableIsVisibleFunction [_ ctx]
+    ;; FIXME - when we have authorization - need to check permissions
+    ;; (let [column-oid (.getText (.columnOid ctx))])
+    true)
+
+  (visitPostgresGetUserbyidFunction [_ ctx]
+    (let [_rel-owner (.getText (.relOwner ctx))]
```
jarohen, line 1861: is this let-binding dead code? if so, could it be removed

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -1629,11 +1629,27 @@
           data-type (-> (.dataType ctx) (.accept (->CastArgsVisitor env)))]
       (handle-cast-expr ve data-type)))
 
+  (visitCollateExpr [_ ctx]
+    ;; TODO - apply collation
+    (log/trace "visitCollateExpr" (.getText (.collation ctx)))
+    (.getText (.collation ctx)))
```
jarohen, line 1635: I doubt the right thing to do is return the collation part here, even if we're not planning to fully implement it. Once implemented, won't this expression return the `expr` string rather than the name of the collation?

tbh I'm surprised this doesn't break something else?

core/src/main/antlr/xtdb/antlr/Sql.g4
```diff
@@ -266,6 +273,10 @@ exprPrimary
 
     | (schemaName '.')? 'VERSION' '(' ')' #PostgresVersionFunction
 
+    | (schemaName '.')? 'PG_GET_USERBYID' '(' relOwner=columnReference ')' #PostgresGetUserbyidFunction
```
jarohen, line 276: please make this casing consistent with the rest of the file: `PostgresGetUserByIdFunction`

jarohen (changes requested):
this still seems a fair way off I'm afraid :slightly_frowning_face: 

I didn't see any tests for any of the added functionality - even if it doesn't fully support displaying tables yet (because of collate) I would have expected to see tests added for `operator.~` or visitPgCatalogResolveOid etc.

Could you land the pg_am change independently? this is an atomic change that can be added separately, and would make this PR much easier to read :slightly_smiling_face: :pray:

mpisanko:
Closing as this started with supporting \d and then added quite a lot of stuff to support \d foo - but that is not implemented, so no point having all those changes hanging there. will submit another isolated PR

PR #3943, opened 2024-12-04
Add more memory cache statistics

### LINKED ISSUE #3929: Finer grained MemoryCache statistics
Add finer grained statistics on the memory cache:

For both data as well as meta files:
- number of files
- percentage of cache
- total size 
- count of level of files (for meta files)
- average size
- average level size
- pinned vs unpinned
- percentage of change/turnover

FiV0:
This adds some more statistics to the memory cache as well as the buffer pool. 
The node debugging dashboard gets a memory cache section.

### DIFF
--- a/core/src/main/clojure/xtdb/metrics.clj
+++ b/core/src/main/clojure/xtdb/metrics.clj
@@ -15,10 +15,9 @@
 (defn add-counter
   ([reg name] (add-counter reg name {}))
   ([reg name {:keys [description]}]
-   (let [cnt (cond-> (Counter/builder name)
-               description (.description description)
-               :always (.register reg)) ]
-     cnt)))
+   (cond-> (Counter/builder name)
+     description (.description description)
+     :always (.register reg))))
 
 (def percentiles [0.75 0.85 0.95 0.98 0.99 0.999])
 
--- a/core/src/main/clojure/xtdb/metrics.clj
+++ b/core/src/main/clojure/xtdb/metrics.clj
@@ -10,12 +10,15 @@
            java.util.List
            (java.util.stream Stream)
            (org.apache.arrow.memory BufferAllocator)
-           (xtdb.cache Stats)))
+           (xtdb.cache Stats MemoryCacheStats)))
 
-(defn add-counter [reg name {:keys [description]}]
-  (cond-> (Counter/builder name)
-    description (.description description)
-    :always (.register reg)))
+(defn add-counter
+  ([reg name] (add-counter reg name {}))
+  ([reg name {:keys [description]}]
+   (let [cnt (cond-> (Counter/builder name)
+               description (.description description)
+               :always (.register reg)) ]
+     cnt)))
 
 (def percentiles [0.75 0.85 0.95 0.98 0.99 0.999])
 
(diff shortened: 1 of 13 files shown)


core/src/main/clojure/xtdb/metrics.clj
```diff
@@ -52,12 +54,34 @@
                #(.getFreeBytes ^Stats (get-stats))
                {:unit "bytes"})))
 
+(defn add-mem-cache-gauges [reg meter-name get-stats]
+  (add-cache-gauges reg meter-name get-stats)
+  (doto reg
+    (add-gauge (str meter-name ".metaSliceCount")
```
jarohen, line 63: suspect meta vs data should be a dimension here rather than part of the meter name (if that's exposed within micrometer?)

core/src/main/kotlin/xtdb/cache/Stats.kt
```diff
@@ -1,3 +1,22 @@
...
 
-data class Stats(val pinnedBytes: Long, val evictableBytes: Long, val freeBytes: Long)
+interface Stats {
+    val pinnedBytes: Long
+    val evictableBytes: Long
+    val freeBytes: Long
+
+    companion object {
+
+        fun new (pinnedBytes: Long, evictableBytes: Long, freeBytes: Long): Stats {
```
jarohen: I'd probably avoid the name `new` here - I know it's not a keyword in Kotlin but still probably a name I'd avoid on the JVM :sweat_smile: 

one idea - I've used `operator fun invoke(...)` here in the past, so that you can call it as `Stats(...)`.

core/src/main/kotlin/xtdb/cache/MemoryCache.kt
```diff
@@ -9,17 +9,30 @@ import org.apache.arrow.memory.BufferAllocator
...
 import java.nio.channels.FileChannel
 import java.nio.file.Path
 import java.util.concurrent.CompletableFuture
 
-
 data class PathSlice(val path: Path, val offset: Long? = null, val length: Long? = null) {
     constructor(path: Path) : this(path, null, null)
 }
 
+data class MemoryCacheStats(
```
jarohen, line 25: rather than the double-up and the weakly-typed map here, I'd probably create something like

```kotlin
data class CacheSectionStats(sliceCount, weightBytes, pinned, unpinned)
```

then have one of those for meta and data?

```kotlin
data class MemoryCacheStats(pinningCacheStats, metaFileStats: CacheSectionStats, dataFileStats: CacheSectionStats)
```

(it's relatively unusual IME in Java/Kotlin to use Maps with keys that you know at compile-time)

core/src/main/kotlin/xtdb/cache/MemoryCache.kt
```diff
@@ -34,7 +47,23 @@ class MemoryCache
 ) : AutoCloseable {
     private val pinningCache = PinningCache<PathSlice, Entry>(maxSizeBytes)
 
-    val stats get() = pinningCache.stats
+    val stats: MemoryCacheStats
+        get()  {
```
jarohen, line 50: I'm assuming you considered keeping running tallies of these as mutable state? don't know how expensive this might be if the cache's large

(might not be an issue, of course)

jarohen (changes requested):

core/src/main/kotlin/xtdb/cache/MemoryCache.kt
FiV0: I didn't want to to do that on the hot paths.

FiV0: What exactly do you mean by running tallies? Like for example a diff of files previously in the cache and now in the cache? This could be done on `getStats`.

core/src/main/clojure/xtdb/metrics.clj
FiV0: I tried with:
- `Gauge` - can only be registered with the same name + same set of tags (in the `PrometheusMeterRegistry`)
- `Counter` - needs to be dealt with explicitly (ie `.increment()`)
- `FunctionCounter` - needs to be a monotonic increasing metric

PR #3954, opened 2024-12-09
Flip from `mmap` to allocating memory with netty direct buffers 

### LINKED ISSUE #3952: Removing `mmap` and use `readRecordBatch` directly.
Let's try to remove the usage of mmap (which uses native memory) and only use `readRecordBatch` (which uses direct memory).

FiV0:
The PR replaces memory mapping Memory cache buffers with direct allocations through netty. The reason we are going through Netty is so all direct memory (there being direct and netty) usage is counted under the netty counters. Underneath Arrow Java does allocations using netty (when vectors get created and expanded).  It also adds metrics for direct memory usage (`(.maxMemory (Runtime/getRuntime))`) and netty usage. The direct memory usage should always be rather low (single digit MB) as XT is not really using it and we are giving all specified direct memory to Netty.

See also the comment at https://github.com/xtdb/xtdb/issues/3891#issuecomment-2536189581

I looked into making usage of existing methods like `ArrowFileReader.loadRecordBatch` and `AFR.readRecordBatch` but the main problem is the reference counting again (no access to `release0`). 

Todo:
- [x] Try to break the node with 2gb heap, 3gb direct, 500mb meta space
- [x] Have a look at native usage

### DIFF
--- a/core/src/main/clojure/xtdb/metrics.clj
+++ b/core/src/main/clojure/xtdb/metrics.clj
@@ -1,7 +1,8 @@
 (ns xtdb.metrics
   (:require [clojure.tools.logging :as log]
             [integrant.core :as ig]
-            [xtdb.node :as xtn])
+            [xtdb.node :as xtn]
+            [xtdb.util :as util])
   (:import (io.micrometer.core.instrument Counter Gauge MeterRegistry Tag Timer Timer$Sample)
            (io.micrometer.core.instrument.binder MeterBinder)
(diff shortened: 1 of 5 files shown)


jarohen (commented):
@FiV0 do we need to load into a bbuf and then pass it to Arrow? IIRC we could go directly from channel to ArrowBuf

FiV0:
The main reason we are not using a method from `channel` to `ArrowBuf` directly from Java Arrow is that we are not having `release0` available which makes sharing the buffers difficult.

jarohen (approved):
thanks @FiV0 🙏 once you've seen this through the benchmarks and confirmed it has the desired effect, good to merge

PR #3956, opened 2024-12-10
Add meta-tables schema to meta-tables, resolves #3550

### LINKED ISSUE #3550: Add metadata tables to metadata tables
Currently metadata tables (pg_catalog/info_schema etc.) don't contain entries for their own data, only user space tables.

FiV0:


src/test/clojure/xtdb/information_schema_test.clj
```diff
@@ -536,3 +544,12 @@
              (count results)))
     (t/is (= types
              results))))
+
+(deftest schema-for-meta-tables-3550
+  (t/is (= [{:table-catalog "xtdb",
+             :table-schema "information_schema",
+             :table-name "tables",
+             :table-type "BASE TABLE"}]
```
FiV0, line 556: The table type in Postgres is a `VIEW`, some others are basic `BASE TABLE`'s. I don't know if we want to emulate the exact mapping or if `VIEW` and `BASE TABLE` should reflect our own internal structure. Everything is a `VIEW` apart from `pg_user`.

jarohen (approved):

PR #3988, opened 2024-12-27
SQL clause pipelining - multiple WHEREs/SELECTs in one query

jarohen:
(based atop #3985, will merge that first)

This PR adds support for creating a pipeline of SQL operators (like in XTQL) - in addition to spec-compliant SQL top-level queries (SELECT, FROM, WHERE, GROUP BY, etc), we also now support:

- FROM/JOIN
- 0..N of ( WHERE? (GROUP BY?, HAVING?, SELECT)? )
- ORDER BY, LIMIT, OFFSET

This reduces the need for subqueries when you want to do operations in a different order to the normal SQL clause order.

e.g.:

```sql
-- double aggregate - the frequency distribution one
FROM foo
SELECT v, COUNT(*) AS v_count
SELECT v_count, COUNT(*) AS freq

-- (before)
SELECT v_count, COUNT(*) AS freq
FROM (SELECT v, COUNT(*) AS v_count
      FROM foo
      GROUP BY v) counts
GROUP BY v_count


-- allowing filtering by the results of window functions (again, requires subqs in standard SQL):
FROM foo
SELECT _id, ROW_NUMBER () OVER (PARTITION BY v ORDER BY _id) row_num
WHERE row_num = 0 -- N.B. WHERE applied after SELECT
SELECT _id
```

### DIFF
--- a/core/src/main/antlr/xtdb/antlr/Sql.g4
+++ b/core/src/main/antlr/xtdb/antlr/Sql.g4
@@ -682,11 +682,16 @@ queryExpressionBody
 
 queryTerm
     : selectClause fromClause? whereClause? groupByClause? havingClause? windowClause? # QuerySpecification
-    | fromClause whereClause? (groupByClause? havingClause? selectClause)? windowClause? # QuerySpecification
+    | fromClause queryTail* windowClause? # QuerySpecification
     | tableValueConstructor # ValuesQuery
     | recordsValueConstructor # RecordsQuery
     | '(' queryExpressionNoWith ')' # WrappedQuery
     | queryTerm 'INTERSECT' (ALL | DISTINCT)? queryTerm # IntersectQuery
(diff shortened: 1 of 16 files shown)


FiV0:
The semantics here is always top to bottom?
```sql
SELECT v, COUNT(*) AS v_count
FROM foo
SELECT v_count, COUNT(*) AS freq
```

jarohen:
> The semantics here is always top to bottom?
> 
> ```sql
> SELECT v, COUNT(*) AS v_count
> FROM foo
> SELECT v_count, COUNT(*) AS freq
> ```

@FiV0 if `SELECT` is at the start, we disallow multiple tails - I (like you, I suspect) felt like the behaviour in this case wouldn't have been immediately obvious, so it's only allowed if the query starts with `FROM`

PR #3989, opened 2024-12-31
Order-by :nulls not taking effect for system-to/valid-to #3278

### LINKED ISSUE #3278: Order-by `:nulls` not taking effect for system-to/valid-to
(from Shane on [Clojurians](https://clojurians.slack.com/archives/CG3AM2F7V/p1712196728923369))

> I can't get `:nulls` to work when sorting by the system/valid columns. When I switch between `:dir :desc` and `:dir :asc` the nils flip sides but when I change between `:nulls :last` and `:nulls :first` they do nothing. I think they are stuck on `:first`.

```clojure
 (order-by {:val xt$valid-from
            :dir :desc :nulls :last})
```

This'll be a positive infinity under the hood rather than a null (it's only rendered as null on output). Options:

1. We document this as a behaviour of system-to/valid-to
2. We adapt the comparison logic to treat positive infinity timestamps as null
3. We return new scalar type(s) to represent infinity, but could still optionally return null for compatibility reasons

mpisanko:


src/test/clojure/xtdb/sql_test.clj
```diff
@@ -2514,3 +2514,17 @@ UNION ALL
 
     (t/is (= {:app_count 2}
              (jdbc/execute-one! tu/*node* ["SELECT COUNT(*) app_count FROM applications"])))))
+
+(t/deftest test-order-nulls-first-last
+  (xt/execute-tx tu/*node* ["INSERT INTO foo (_id, name) VALUES (1, 'foo'), (2, 'bar'), (3, 'baz')"])
+  (xt/execute-tx tu/*node* ["INSERT INTO foo (_id, name) VALUES (1, 'fooz'), (2, 'barz'), (3, 'bazz')"])
+  (t/is (=
+         [nil nil nil]
```
jarohen: incorrect code (e.g. returning all nulls, or not sorting non-nulls times correctly) would also pass this test - could you explicitly specify enough to prove this is behaving correctly?

it should be the same valid-to every time, if that's what you were concerned about - I think this namespace has a mock-clock fixture

jarohen (changes requested):

PR #3990, opened 2025-01-01
add pg_sleep and pg_sleep_for functions #3983

### LINKED ISSUE #3983: Add `pg_sleep`
The following should be possible
```sql
SELECT pg_sleep(30)
```

mpisanko:
Add support for `pg_sleep` and `pg_sleep_for` functions.
chose to not do much parsing in ANTLR and instead do it in the EE (upon feedback).
also learned that actual sleeping should be done in EE (if done at plan level -  it would get cached and not sleep on calls with same arguments).


core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -1502,6 +1510,31 @@
   (visitPgExpandArrayFunction [_ _] nil)
   (visitPgGetExprFunction [_ _] nil)
   (visitPgGetIndexdefFunction [_ _] nil)
+  (visitPgSleepFunction [this ctx]
+    (list 'sleep (-> (.sleepSeconds ctx) (.accept this))))
+
+  (visitPgSleepForFunction [_ ctx]
+    (let [period (.getText (.sleepPeriod ctx))]
+      ;; Would it be cleaner to do it in ANTLR? or EE maybe?
```
jarohen: this is the kind of thing to flag in the PR description, for example.

as it happens, I _would_ put this in the EE - pass it through to sleep as a string (`(list 'sleep (.accept ... this))`), then have sleep accept a string and cast to duration-micros (see usages of `recall-with-cast`).

core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -1502,6 +1510,31 @@
   (visitPgExpandArrayFunction [_ _] nil)
   (visitPgGetExprFunction [_ _] nil)
   (visitPgGetIndexdefFunction [_ _] nil)
+  (visitPgSleepFunction [this ctx]
+    (list 'sleep (-> (.sleepSeconds ctx) (.accept this))))
```
jarohen: I'd emit this as a duration, e.g. 
```clojure
`(Duration/ofSeconds ~(.accept ... this))
```
so that the assumption that the user's passing seconds in is dealt with as early as possible.

core/src/main/clojure/xtdb/expression.clj
```diff
@@ -1203,6 +1206,22 @@
      :->call-code (fn [[include-implicit?]]
                     `(current-schemas ~include-implicit?))})
 
+(defn sleep [^double seconds-expr]
+  (Thread/sleep (long (* 1000 seconds-expr))))
+
+(defmethod codegen-call [:sleep :f64] [_]
```
jarohen: I would make the standard `:sleep` EE function be `[:sleep :duration]` here, so that the units are clear (see below comments)

jarohen (changes requested):
comments below. 

also, PR bureaucracy please:

- assign it to yourself, you're responsible for seeing it through until it's landed (see git.adoc in the repo)
- ensure that the PR is correctly linked to the original card

cheers :pray:

core/src/main/clojure/xtdb/sql/plan.clj
mpisanko: this actually is not what the [pg_sleep](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-DELAY) specifies - it calls for a `double precision` - while `Duration/ofSeconds` accepts a `long`. so to make it happen it would require additional parsing beforehand..? is that what you were thinking? - is that where recall-with-cast  comes in?

core/src/main/clojure/xtdb/sql/plan.clj
mpisanko: did try - but really can't seem to be able to use recall-with-cast..

core/src/main/clojure/xtdb/sql/plan.clj
jarohen: ok, I misinterpreted from `sleepSeconds` that it'd be integer seconds, apologies.

maybe something like

```
`(cast ~(.accept ... this) [:duration :millis])
```
?

core/src/main/clojure/xtdb/expression.clj
```diff
@@ -1223,6 +1226,41 @@
...
+(defn sleep [millis]
+  (Thread/sleep millis))
+
+(defmethod codegen-call [:sleep :f64] [_expr]
+  (let [{convert :->call-code} (codegen-cast {:source-type :f64 :target-type [:duration :milli]})]
+    {:return-type :null
+     :->call-code (fn [lit]
+                    `(sleep ~(convert lit)))}))
+
+(defn string->duration [input]
```
jarohen: I'd:

* have this as `(defmethod codegen-cast [:utf8 :duration] ...)` so that anyone casting from string to duration gets to use it
* in the planner, you can then emit `(cast ~<expr> [:duration :millis])`
* `(defmethod codegen-call [:sleep :duration])` should then be the only one you need - looks like you can have both emitting to the same EE function

jarohen (changes requested):
this seems to still have outstanding issues to address.

please also link to the issue as requested - thanks :pray:

src/test/clojure/xtdb/expression_test.clj
```diff
@@ -1573,6 +1573,17 @@
                                   #(->dur-vec "x" TimeUnit/SECOND 10)
                                   #(->bigint-vec "y" 3))))))))
 
+(t/deftest cast-duration-test
```
jarohen, line 1576: I don't think this will work with anything other than `:milli`?

core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -583,12 +613,21 @@
               :arg-types [:utf8 :utf8 :int :utf8 :int]}]
     (expr/codegen-call expr)))
 
+(defn duration-from-seconds [^double seconds]
+  ;; apparently EE pases durations as long - so here we do as well
+  (long (* 1000 seconds)))
+
+(defmethod expr/codegen-cast [:f64 :duration] [_expr]
+  {:return-type :duration
```
jarohen: `:duration` isn't a return-type on its own, needs to be `[:duration <time-unit>]` (should be other examples of this in this file). You'll be able to extract the requested time-unit from the `_expr` (again, see other examples in the file).

core/src/main/clojure/xtdb/expression.clj
```diff
@@ -1224,6 +1227,14 @@
    :->call-code (fn [[setting-name]]
                   `(current-setting (resolve-string ~setting-name)))})
 
+(defn sleep [^long duration]
+  (Thread/sleep duration))
+
+(defmethod codegen-call [:sleep :duration] [_]
```
jarohen: this will only work if the input is millis - same handling as comment below

core/src/main/clojure/xtdb/expression/temporal.clj
```diff
@@ -583,12 +613,21 @@
               :arg-types [:utf8 :utf8 :int :utf8 :int]}]
     (expr/codegen-call expr)))
 
+(defn duration-from-seconds [^double seconds]
+  ;; apparently EE pases durations as long - so here we do as well
```
jarohen: don't need to keep this type of comment, just adds noise to the namespace.

separately, I'd be tempted to have this function handle the different units - e.g. 1000 for millis, 1e6 for micros etc

jarohen (changes requested):
still a few bugs to go here I'm afraid - mostly around dealing with different duration time-units. pop a post on the dev channel if it's unclear?

src/test/clojure/xtdb/expression_test.clj
mpisanko: hopefully now after checking and using desired resolution this should work

jarohen: much better, thanks :relaxed: :pray:

Issue #3432, opened 2024-06-21
Add support for current_schemas() function


mbutlerw:
https://www.postgresql.org/docs/current/functions-info.html

mpisanko:
looks like there is support for current_schemas already - the only kink is the return format - square brackets - as JSON (while postgres returns squigly brackets) - do we want to change it?

mpisanko:
that would probably require supporting a `_text` (text array) type - think that would suffice

mpisanko:
@jarohen the question above - do we want to be compliant to the bracket level?
(ie return `{public}` rather than `[public]`)

jarohen:
it won't just be brackets - for jdbc, for example, it'll return a different type?

(the brackets are just rendering)

mpisanko:
> that would probably require supporting a `_text` (text array) type - think that would suffice

as per this comment. the function description:
```
current_schemas ( include_implicit boolean ) → name[]

Returns an array of the names of all schemas presently in the effective search path, in their priority order. (Items in the current [search_path](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-SEARCH-PATH) setting that do not correspond to existing, searchable schemas are omitted.) If the Boolean argument is true, then implicitly-searched system schemas such as pg_catalog are included in the result.
```

mpisanko:
i guess adding _text will make us a bit more compliant.

mpisanko:
related #3991

mpisanko:
``` 
mpi@mpimbp ~> psql -p 5432 -h localhost -U xtdb
psql (17.0, server 16)
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, compression: off, ALPN: none)
Type "help" for help.

xtdb=> select current_schemas(true);
      _column_1
---------------------
 {pg_catalog,public}
(1 row)
```

refset:
(it looks like this was meant to have been closed already)

PR #3991, opened 2025-01-01
Text array

mpisanko:
This PR adds support for flat (single dimension) text arrays - also known as text[] or datatype _text (oid 1009). it supports text and binary formats.
Anything that looks like [:list :utf8] will be treated as text array. A corresponding entry is added to pg_type to reflect this type.
To match postgres behaviour we're using the curly braces around arrays (unlike JSON square brackets).
This started off with matching behaviour of postgres current_schemas function https://github.com/xtdb/xtdb/issues/3432
It also fixes read-binary for integer arrays.

### DIFF
--- a/core/src/main/clojure/xtdb/types.clj
+++ b/core/src/main/clojure/xtdb/types.clj
@@ -13,7 +13,7 @@
            [java.nio.charset StandardCharsets]
            [java.time Duration LocalDate LocalDateTime OffsetDateTime ZoneId ZoneOffset ZonedDateTime]
            [java.time.format DateTimeFormatter DateTimeFormatterBuilder]
-           [java.util UUID]
+           [java.util Arrays UUID]
            (org.apache.arrow.vector.types DateUnit FloatingPointPrecision IntervalUnit TimeUnit Types$MinorType UnionMode)
            (org.apache.arrow.vector.types.pojo ArrowType ArrowType$Binary ArrowType$Bool ArrowType$Date ArrowType$Decimal ArrowType$Duration ArrowType$FixedSizeBinary ArrowType$FixedSizeList ArrowType$FloatingPoint ArrowType$Int ArrowType$Interval ArrowType$List ArrowType$Map ArrowType$Null ArrowType$Struct ArrowType$Time ArrowType$Time ArrowType$Timestamp ArrowType$Union ArrowType$Utf8 Field FieldType)
            (xtdb JsonSerde Types)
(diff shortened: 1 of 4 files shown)


core/src/main/clojure/xtdb/types.clj
```diff
@@ -836,6 +852,29 @@
...
            (range length)))))
 
+(defn read-binary-text-array [^bytes ba]
+  (let [^ByteBuffer bb (ByteBuffer/wrap ba)
+        ;; _payload-size (.getInt bb 0)
+        ;; _dimensions (.getInt bb 4)
+        data-offset (.getInt bb 8)
+        ;; _oid (.getInt bb 12)
+        length (.getInt bb 16)
+        get-start (fn ^long [^long header-length ^long data-offset] (+ header-length data-offset))
```
jarohen: IIRC Clojure won't be able to keep these as primitives, that only works with functions under a var, not a locally-defined function. Given they're small and used once (AFAICT?), I'd recommend inlining here?

core/src/main/clojure/xtdb/types.clj
```diff
@@ -836,6 +852,29 @@
...
+        start (get-start 20 data-offset)
+        start-of-str (fn ^long [^long start] (+ start 4))
+        next-start (fn ^long [^long start ^long str-len] (+ start 4 str-len))]
+    (:arr
```
jarohen: `loop` probably more in keeping with the rest of the codebase here, save creating a map on every iteration just to unwrap it on the next.

src/test/clojure/xtdb/pgwire_test.clj
```diff
@@ -2105,6 +2103,29 @@ ORDER BY t.oid DESC LIMIT 1"
         (with-open [rs (.executeQuery stmt)]
           (t/is (= [{"arr" [1 2 3]}] (rs->maps rs))))))))
 
+(deftest test-text-array
```
jarohen, line 2120: test for reading the text array? assume we should be able to pass one in as a parameter

core/src/main/clojure/xtdb/types.clj
```diff
@@ -805,6 +805,10 @@
   ^long [^long length ^long typlen]
   (+ 4 (* 4 4) (* typlen length) (* 4 length)))
 
+(defn calculate-variable-size-array-buffer-size
+  ^long [^long element-count ^long payload-size]
+  (+ 4 (* 4 4) payload-size (* 4 element-count)))
```
jarohen: subjective: I'd inline this so that it's closer to the knowledge of what these 4s represent

jarohen (changes requested):
Thanks @mpisanko - feedback in comments :pray: 

Please could you also get into the habit of filling out implementation notes in the PR description - this is a record for future-us of what you've changed and some of the context/intuition behind it.

mpisanko:
This PR adds support for flat (single dimension) text arrays - also known as `text[]` or datatype `_text` (oid 1009). it supports text and binary formats. 
Anything that looks like [:list :utf8] will be treated as text array. A corresponding entry is added to pg_type to reflect this type.
To match postgres behaviour we're using the curly braces around arrays (unlike JSON square brackets). 
This started off with matching behaviour of postgres current_schemas function #3432

core/src/main/clojure/xtdb/types.clj
mpisanko: this is purely to circumvent reflection warnings - if i inline they come back.. we don't like them. do we?

jarohen:
https://github.com/xtdb/xtdb/pull/3991#issuecomment-2568415378
@mpisanko can this go in the PR description please so that readers don't have to scroll through the comments to understand the current state of the PR - thanks

PR #4001, opened 2025-01-07
Add Implementation-Version and use in xtdb-server-version

### LINKED ISSUE #3695: `xt.version()` should not require git to be installed for in-process JAR users
[Play](https://play.xtdb.com/?version=2.0.0-SNAPSHOT&type=sql&txs=W3sic3lzdGVtLXRpbWUiOm51bGwsInR4cyI6IklOU0VSVCBJTlRPIGRvY3MgKF9pZCwgZm9vKSBWQUxVRVMgKDEsICdiYXInKSJ9XQ%3D%3D&query=U0VMRUNUIHh0LnZlcnNpb24oKSwgZm9vIEZST00gZG9jcw%3D%3D) (the Lambda underpinning Play doesn't have git installed)

```
Error: java.io.IOException

Cannot run program "git": error=2, No such file or directory
```

tggreene:
This resolves #3695 by putting the version under `Implementation-Version` in the META-INF/MANIFEST.MF of compiled jars, falling back to git or unknown otherwise.

tasks.jar has been added to the docker gradle files explicitly as anything at the root level triggered errors in the build, I haven't found a better approach for this.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -62,6 +62,14 @@ allprojects {
             withSourcesJar()
         }
 
+        tasks.jar {
+            manifest {
+                attributes(
+                    "Implementation-Version" to project.version,
+                )
+            }
+        }
+
         if (plugins.hasPlugin("org.jetbrains.dokka"))
             tasks.register<Jar>("dokkaJavadocJar") {
                 dependsOn(tasks.dokkaJavadoc)
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -39,6 +39,8 @@
   ;; TODO this won't work if the user has XT in-process in their own git repo.
   (or (System/getenv "XTDB_VERSION")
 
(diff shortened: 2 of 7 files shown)


build.gradle.kts
```diff
@@ -57,6 +57,16 @@ allprojects {
         maven { url = uri("https://repo.clojars.org/") }
     }
 
+    if (plugins.hasPlugin("java")) {
```
jarohen: this can probably go under the existing `plugins.hasPlugin("java-library")` guard

jarohen (approved):
thanks @tggreene - just one minor change here, if you can squash that into your commit it's good to merge without further review :pray: 

Do we need a check around the git call to guard against IOException, as mentioned in the original card?

tggreene:
From my testing we don't (replacing git with another unknown executable) it just defaults to `unknown`. Plus the error mentioned in the original issue on play doesn't appear to be the case anymore.

Scratch that I'll just add a catch around as you say.

PR #4003, opened 2025-01-08
PGWire msg buffering

### LINKED ISSUE #3968: I/O performance in pgwire server
Seems like OutputStream writes/InputStream reads are dominating our pgwire server time on the tx-ingest benchmark when the batches are set to 1000 docs/tx:

Image

(For comparison, local Postgres does this in about 1/8th of the time, and presumably has to send all the same messages over a wire 😅)

Let's spike out including something like vert.x here?

FiV0:
I am adding a `BufferedStream` (of size 8kb) around the socket input/output streams and ~also set the send and receive buffer of the socket to 1kb~ (I removed the setting on the socket itself, as it didn't seem to have any positive effect. I assume this is more about tcp buffering than application level buffering) . For an in-memory node this reduces the `tx-ingest-benchmark` from ~90secs to ~30secs with batchsize 1000 and 1M documents. 

Messages I am now forcing a flush:
- Initial negatiations
- `error`
- `notice`
- `parameter-status`
- `auth` 
- `ready`
- `portal-suspended` (although we currently don't seem to use this).

Not sure if others need a flush. Most of this relies on `ready` flushing a lot of stuff I assume.

### DIFF
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -18,7 +18,7 @@
             [xtdb.util :as util]
             [xtdb.vector.writer :as vw])
   (:import [clojure.lang MapEntry]
-           [java.io ByteArrayInputStream ByteArrayOutputStream Closeable DataInputStream DataOutputStream EOFException IOException InputStream OutputStream PushbackInputStream]
+           [java.io ByteArrayInputStream ByteArrayOutputStream Closeable DataInputStream DataOutputStream EOFException IOException InputStream OutputStream PushbackInputStream BufferedInputStream BufferedOutputStream]
            [java.lang AutoCloseable Thread$State]
            [java.net ServerSocket Socket SocketException URI]
            [java.nio ByteBuffer]
@@ -151,6 +151,7 @@
 (declare client-err)
 (declare err-protocol-violation)
 (declare ->socket-frontend)
+(declare flush-messages)
 
 (defrecord SocketFrontend [^Socket socket, ^DataInputStream in, ^DataOutputStream out]
   Frontend
@@ -158,7 +159,9 @@
     (log/trace "Writing server message" (select-keys msg-def [:char8 :name]))
 
     (.writeByte out (byte (:char8 msg-def)))
-    (.writeInt out 4))
+    (.writeInt out 4)
+    (when (flush-messages (:name msg-def))
+      (.flush out)))
 
   (send-client-msg! [_ msg-def data]
     (log/trace "Writing server message (with body)" (select-keys msg-def [:char8 :name]))
@@ -168,7 +171,9 @@
           arr (.toByteArray bytes-out)]
       (.writeByte out (byte (:char8 msg-def)))
       (.writeInt out (+ 4 (alength arr)))
-      (.write out arr)))
+      (.write out arr)
+      (when (flush-messages (:name msg-def))
+        (.flush out))))
 
   (read-client-msg! [_]
     (let [type-char (char (.readUnsignedByte in))
@@ -193,6 +198,7 @@
         (log/trace "upgrading to SSL")
 
         (.writeByte out (byte \S))
+        (.flush out)
 
         (let [^SSLSocket ssl-socket (-> (.getSocketFactory ^SSLContext ssl-ctx)
                                         (.createSocket socket
@@ -213,6 +219,7 @@
       ;; unsupported - recur and give the client another chance to say hi
       (do
         (.writeByte out (byte \N))
+        (.flush out)
         this)))
 
   (flush! [_] (.flush out))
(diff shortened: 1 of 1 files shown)


src/test/clojure/xtdb/pgwire_test.clj
```diff
@@ -1592,6 +1592,7 @@
       (jdbc/execute! tx ["INSERT INTO foo RECORDS ?" (xt-jdbc/->pg-obj {:xt/id 4, :a "four"})])
 
       ;; HACK.
+      #_
```
FiV0: Commented out for now, as this didn't work. I suspect the flushing behavior might be the cause of this. Will look into it, if we are happy with the approach.

FiV0:
Number for different sizes.
```clj
{:batch-size 10, :buffer-size-kb 1, :elapsed-ms 194083}
{:batch-size 10, :buffer-size-kb 8, :elapsed-ms 190521}
{:batch-size 10, :buffer-size-kb 32, :elapsed-ms 192038}
{:batch-size 100, :buffer-size-kb 1, :elapsed-ms 41764}
{:batch-size 100, :buffer-size-kb 8, :elapsed-ms 41996}
{:batch-size 100, :buffer-size-kb 32, :elapsed-ms 41745}
{:batch-size 1000, :buffer-size-kb 1, :elapsed-ms 24490}
{:batch-size 1000, :buffer-size-kb 8, :elapsed-ms 24623}
{:batch-size 1000, :buffer-size-kb 32, :elapsed-ms 25096}
```
A bit weird that there is no correlation between buffer size and time improvements.

FiV0:
Changed to 1kb.

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1878,6 +1892,10 @@
           (try
             (let [conn-socket (.accept accept-socket)]
               (.setTcpNoDelay conn-socket true)
+              (.setSendBufferSize conn-socket socket-buffer-size)
+              (.setReceiveBufferSize conn-socket socket-buffer-size)
+              (log/info "Accepted connection" {:send-buffer-size (.getSendBufferSize conn-socket),
```
jarohen: we usually recommend that users run XT at info, so I'd guess this is debug?

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1878,6 +1892,10 @@
           (try
             (let [conn-socket (.accept accept-socket)]
               (.setTcpNoDelay conn-socket true)
+              (.setSendBufferSize conn-socket socket-buffer-size)
```
jarohen: potentially silly question on my part - do we need both `.setSendBufferSize` and `BufferedOutputStream` here? i.e. are they both having the same effect?

jarohen (approved):
cheers @FiV0, good result :pray: minor non-blocking (pardon the pun) qs below

core/src/main/clojure/xtdb/pgwire.clj
FiV0: Good catch. That is a leftover from me as the OS sometimes can set them as well.

core/src/main/clojure/xtdb/pgwire.clj
FiV0: Yes, maybe I should do a run with just these and also check the underlying implementation.

core/src/main/clojure/xtdb/pgwire.clj
FiV0: Ah, while removing it, I actually saw that the 1kb is not exactly honoured:
```
Accepted connection {:send-buffer-size 2304, :receive-buffer-size 1152}
```

FiV0:
> Number for different sizes.
> 
> ```clojure
> {:batch-size 10, :buffer-size-kb 1, :elapsed-ms 194083}
> {:batch-size 10, :buffer-size-kb 8, :elapsed-ms 190521}
> {:batch-size 10, :buffer-size-kb 32, :elapsed-ms 192038}
> {:batch-size 100, :buffer-size-kb 1, :elapsed-ms 41764}
> {:batch-size 100, :buffer-size-kb 8, :elapsed-ms 41996}
> {:batch-size 100, :buffer-size-kb 32, :elapsed-ms 41745}
> {:batch-size 1000, :buffer-size-kb 1, :elapsed-ms 24490}
> {:batch-size 1000, :buffer-size-kb 8, :elapsed-ms 24623}
> {:batch-size 1000, :buffer-size-kb 32, :elapsed-ms 25096}
> ```
> 
> A bit weird that there is no correlation between buffer size and time improvements.

This is complete garbage, so please ignore.

core/src/main/clojure/xtdb/pgwire.clj
FiV0: From the docs
```java
    /**
     * Sets the {@link SocketOptions#SO_SNDBUF SO_SNDBUF} option to the
     * specified value for this {@code Socket}.
     * The {@link SocketOptions#SO_SNDBUF SO_SNDBUF} option is used by the
     * platform's networking code as a hint for the size to set the underlying
     * network I/O buffers.
```
It seems to get honored on Linux. ~With just `setBufferSize` and without `BufferedOutputStream` I don't see any performance reduction. So maybe we just set that.~

FiV0: It seems that 
```
socket.setSendBufferSize(bufferSize);
socket.setReceiveBufferSize(bufferSize);
``` 
has no effect for performance on the ingest benchmark. I guess this is more on tcp socket level.

FiV0:
Merged with 1kb buffers and no setting on the socket itself.

PR #4016, opened 2025-01-14
Change azure instrumentation key to connection string

### LINKED ISSUE #3837: Fix deprecation warning in AzureMonitorMetrics
(`instrumentationKey`)

`AzureMonitorMetrics.kt:24:30 This declaration overrides deprecated member but not marked as deprecated itself. Please add @Deprecated annotation or suppress. See https://youtrack.jetbrains.com/issue/KT-47902 for details`

Can we confirm that we're using this as it's meant to be used now, or add `@Deprecated` ourselves?

tggreene:
This resolves a deprecation notice on our azure metrics impl #3837, and looks like the right way ™ to do it with connection string.

Tested on the azure benchmark.

Closes #3837

### DIFF
--- a/cloud-benchmark/azure/README.adoc
+++ b/cloud-benchmark/azure/README.adoc
@@ -220,11 +220,11 @@ There is a helper script to do _all_ of the above, which can be run by:
 
 === Setup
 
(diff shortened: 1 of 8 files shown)


danmason (commented):
@tggreene Assuming we have tested this and verified that metrics are still being reported to application insights? If so, code changes all look sensible to me.

PR #4044, opened 2025-01-17
Plan `SELECT` subqueries before group-by

### LINKED ISSUE #3821: Unknown symbol error when using `IN` in a `COUNT` aggregate
Came across this from a query ChatGPT generated for me (I then minimised the query)

[play](https://play.xtdb.com/?version=2.0.0-b1&type=sql&txs=W3sic3lzdGVtLXRpbWUiOm51bGwsInR4cyI6IklOU0VSVCBJTlRPIGRvY3MgUkVDT1JEU1xue19pZDogMH0sXG57X2lkOiAxfTsifV0%3D&query=U0VMRUNUIENPVU5UKENBU0UgV0hFTiBfaWQgSU4gKDEsIDIpIFRIRU4gX2lkIEVORCkKRlJPTSBkb2Nz)
```sql
INSERT INTO docs RECORDS
{_id: 0},
{_id: 1};

SELECT COUNT(CASE WHEN _id IN (1, 2) THEN _id END)
FROM docs
```

Gives me:
```
Unknown symbol: '_sq_3'
```

FiV0:
Maybe we are wrong, but @mbutlerw and me couldn't think of an example where the select subqueries need to run after the group by.

### DIFF
--- a/core/src/main/clojure/xtdb/sql/plan.clj
+++ b/core/src/main/clojure/xtdb/sql/plan.clj
@@ -2166,6 +2166,8 @@
              plan]
             plan)
 
+          (-> plan (apply-sqs (:subqs select-plan)))
+
           (cond-> plan
             grouped-table? (wrap-aggs aggs group-invariant-cols))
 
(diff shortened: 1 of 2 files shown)


jarohen:
Also unconfirmed, and out of ctx, but maybe if the subq is in a group-invariant position?

e.g. 

```sql
SELECT (SELECT ...) AS foo, 
       COUNT(*) 
FROM ... 
GROUP BY foo
```

FiV0:
Maybe I'll come back to this at some point, but will try a more incremental approach for now.

PR #4051, opened 2025-01-20
Plan aggregate subqueries before group-by, fix #3821

### LINKED ISSUE #3821: Unknown symbol error when using `IN` in a `COUNT` aggregate

FiV0:
This adds a new hash map `agg-subqs` to the SQL planner that saves subqueries inside aggregates. These then need to run before the `groub-by`. Other subqueries are understood to be correlated to invariant grouping columns and can run after the group by.

### DIFF
--- a/core/src/main/clojure/xtdb/sql/plan.clj
+++ b/core/src/main/clojure/xtdb/sql/plan.clj
@@ -683,6 +683,7 @@
     (let [sl-ctx (.selectList ctx)
           !subqs (HashMap.)
           !aggs (HashMap.)
+          !agg-subqs (HashMap.)
           !windows (HashMap.)
 
           explicitly-projected-cols (->> (.selectSublist sl-ctx)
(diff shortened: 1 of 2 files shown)


core/src/main/clojure/xtdb/sql/plan.clj
```diff
@@ -1633,11 +1635,11 @@
     (let [ve (-> (.exprPrimary ctx) (.accept this))
           data-type (-> (.dataType ctx) (.accept (->CastArgsVisitor env)))]
       (handle-cast-expr ve data-type)))
-  
+
   (visitAggregateFunctionExpr [{:keys [!aggs] :as this} ctx]
     (if-not !aggs
       (add-err! env (->AggregatesDisallowed))
-      (-> (.aggregateFunction ctx) (.accept this))))
+      (-> (.aggregateFunction ctx) (.accept (assoc this :agg-expr true)))))
```
mbutlerw: nitpick: `?` for bool?

mbutlerw (approved):
lgtm, only question would be if there's a more idiomatic way to tell you are within an agg-expr in this visitor pattern planner, but not familiar enough with it myself to comment.

FiV0:
>lgtm, only question would be if there's a more idiomatic way to tell you are within an agg-expr in this visitor pattern planner, but not familiar enough with it myself to comment.

Yes agree. One for @jarohen.

core/src/main/clojure/xtdb/sql/plan.clj
jarohen: try `(assoc this :!subqs !agg-subqs)` :crossed_fingers:

PR #4052, opened 2025-01-21
handle exceptions and show them to user with adequate messages

### LINKED ISSUE #4026: Kafka RecordTooLargeException should be handled properly
Currently requires debug logging turned on to show:
```
20:23:27 | DEBUG xtdb.pgwire | Error on execute-tx
org.apache.kafka.common.errors.RecordTooLargeException: The message is 7772320 bytes when serialized which is larger than 1048576, which is the value of the max.request.size configuration.
```

Users only see `ERROR: unexpected error on tx submit (report as a bug)`

mpisanko:
(coming from exceptions)

### DIFF
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -363,11 +363,12 @@
 
 (defn err-pg-exception
   "Returns a pg specific error for an XTDB exception"
-  [^Throwable ex generic-msg]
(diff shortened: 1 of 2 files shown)


refset:
As discussed, let's confirm with a real Kafka fixture. It might be simplest to create a new test ns drawing inspiration from both `xtdb.pgwire-test` and `xtdb.kafka-test`

mpisanko:
the same works in EDGE so this would be a dupe

PR #4058, opened 2025-01-22
add pg_am and additional columns to pg_class

mpisanko:
This supports the display tables story by adding missing table (pg_am) and missing columns to pg_class table


jarohen (approved):
thanks @mpisanko :pray:

PR #4077, opened 2025-01-27
Add monitoring for query failures

### LINKED ISSUE #3898: Add monitoring for Query Failures/Errors
Want to add some form of [counter](https://docs.micrometer.io/micrometer/reference/concepts/counters.html) against Query Failures/Errors:
- Keeping a general count of these query errors/failures within XTDB.
- Want to display these over time in a graph on the ["XTDB Monitoring Dashboard"](https://github.com/xtdb/xtdb/tree/main/monitoring/grafana/dashboards).

tggreene:
Adds `query.error` and `query.warning` metrics to pgwire and direct queries. Also adds `query.error` to the xtdb monitoring dashboard.

Tested with a variety of queries for each different phase `SLECT * FROM foo` for parsing, `SELECT 1/0;` for execution and `SELECT * FROM missing_table` for planning warnings.

Closes #3898

### DIFF
--- a/core/src/main/clojure/xtdb/node/impl.clj
+++ b/core/src/main/clojure/xtdb/node/impl.clj
@@ -16,6 +16,7 @@
             [xtdb.vector.writer :as vw]
             [xtdb.xtql :as xtql])
   (:import io.micrometer.core.instrument.composite.CompositeMeterRegistry
+           io.micrometer.core.instrument.Counter
            (java.io Closeable Writer)
            (java.util.concurrent ExecutionException)
            java.util.HashMap
@@ -53,11 +54,11 @@
                              :latest-completed-tx latest-completed-tx
                              :snapshot-time snapshot-time}))))
 
-(defn- then-execute-prepared-query [^PreparedQuery prepared-query, allocator query-timer {:keys [args], :as query-opts}]
+(defn- then-execute-prepared-query [^PreparedQuery prepared-query, allocator {:keys [args], :as query-opts} {:keys [query-timer] :as metrics}]
   (util/with-close-on-catch [bound-query (util/with-close-on-catch [args-rel (vw/open-args allocator args)]
                                            (.bind prepared-query (assoc query-opts :args args-rel)))]
     ;;TODO metrics only currently wrapping openQueryAsync results
-    (-> (q/open-cursor-as-stream bound-query query-opts)
+    (-> (q/open-cursor-as-stream bound-query query-opts metrics)
         (metrics/wrap-query query-timer))))
 
 (defn- ->TxOps [tx-ops]
@@ -75,7 +76,8 @@
                  ^CompositeMeterRegistry metrics-registry
                  default-tz
                  system, close-fn,
-                 query-timer]
+                 query-timer
+                 ^Counter query-error-counter]
   Xtdb
   (getServerPort [this]
     (or (:port (util/component this :xtdb.pgwire/server))
@@ -114,13 +116,24 @@
 
   (open-sql-query [this query query-opts]
     (let [query-opts (-> query-opts (with-query-opts-defaults this))]
-      (-> (xtp/prepare-sql this query query-opts)
-          (then-execute-prepared-query allocator query-timer query-opts))))
+      ;; We catch exceptions here to count errors outside of query execution e.g. parsing errors
+      (try
+        (-> (xtp/prepare-sql this query query-opts)
+            (then-execute-prepared-query allocator query-opts {:query-timer query-timer :query-error-counter query-error-counter}))
+        (catch Exception e
+          (when query-error-counter
+            (.increment query-error-counter))
+          (throw e)))))
 
   (open-xtql-query [this query query-opts]
     (let [query-opts (-> query-opts (with-query-opts-defaults this))]
-      (-> (xtp/prepare-xtql this query query-opts)
-          (then-execute-prepared-query allocator query-timer query-opts))))
+      (try
+        (-> (xtp/prepare-xtql this query query-opts)
+            (then-execute-prepared-query allocator query-opts {:query-timer query-timer :query-error-counter query-error-counter}))
+        (catch Exception e
+          (when query-error-counter
+            (.increment query-error-counter))
+          (throw e)))))
 
   xtp/PStatus
   (latest-completed-tx [_] (.getLatestCompletedTx live-idx))
@@ -189,7 +202,8 @@
 (defmethod ig/init-key :xtdb/node [_ {:keys [metrics-registry] :as deps}]
   (let [node (map->Node (-> deps
                             (assoc :query-timer (metrics/add-timer metrics-registry "query.timer"
-                                                                   {:description "indicates the timings for queries"}))))]
+                                                                   {:description "indicates the timings for queries"}))
+                            (assoc :query-error-counter (metrics/add-counter metrics-registry "query.error"))))]
 
     (doto metrics-registry
       (metrics/add-gauge "node.tx.latestSubmittedTxId" (fn [] (xtp/latest-submitted-tx-id node)))
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -8,6 +8,7 @@
             [xtdb.api :as xt]
             [xtdb.authn :as authn]
             [xtdb.expression :as expr]
+            [xtdb.metrics :as metrics]
             [xtdb.node :as xtn]
             [xtdb.node.impl]
             [xtdb.protocols :as xtp]
@@ -29,6 +30,7 @@
            [java.util List Map Set UUID]
            [java.util.concurrent ConcurrentHashMap ExecutorService Executors TimeUnit]
            [javax.net.ssl KeyManagerFactory SSLContext SSLSocket]
+           io.micrometer.core.instrument.Counter
            (org.antlr.v4.runtime ParserRuleContext)
            (org.apache.arrow.memory BufferAllocator RootAllocator)
            [org.apache.arrow.vector PeriodDuration]
@@ -1029,7 +1031,7 @@
 
   If the connection is operating in the :extended protocol mode, any error causes the connection to skip
   messages until a msg-sync is received."
-  [{:keys [conn-state] :as conn} err]
+  [{:keys [conn-state ^Counter query-error-counter] :as conn} err]
 
   ;; error seen while in :extended mode, start skipping messages until sync received
   (when (= :extended (:protocol @conn-state))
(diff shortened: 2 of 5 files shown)


core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1502,14 +1504,20 @@
...
 
-    (interpret-sql query {:default-tz (.getZone clock)
-                          :watermark-tx-id watermark-tx-id
-                          :session-parameters session-parameters})))
+    (try
+      (interpret-sql query {:default-tz (.getZone clock)
+                            :watermark-tx-id watermark-tx-id
+                            :session-parameters session-parameters})
+      (catch Exception e
+        (when metrics-registry
```
jarohen: You probably don't want to be adding the counter every time here - likely not the most costly of things but we tend to get in the habit of creating as much as we might need ahead of time rather than at runtime.

I'd probably stick the Counter object next to conn-state (i.e. on server, IIRC) so that you don't have to pass a separate opts thing all the way down too :slightly_smiling_face:

core/src/main/clojure/xtdb/node/impl.clj
```diff
@@ -111,8 +112,12 @@
 
   (open-sql-query [this query query-opts]
     (let [query-opts (-> query-opts (with-query-opts-defaults this))]
-      (-> (xtp/prepare-sql this query query-opts)
-          (then-execute-prepared-query allocator query-timer query-opts))))
+      (try
+        (-> (xtp/prepare-sql this query query-opts)
+            (then-execute-prepared-query allocator query-timer query-opts))
+        (catch Exception e
+          (.increment ^Counter (metrics/add-counter metrics-registry "sql-query-error"))
```
jarohen: (see other comment) I'd create the counter with the node here

core/src/main/clojure/xtdb/pgwire.clj
tggreene: Ah, you mean add both the counter and the metrics-registry to conn-state?

jarohen: should just need the counter, I don't think you need the registry after you've used it to add the counter?

tggreene: Yup, not thinking that one through; thanks James

core/src/main/clojure/xtdb/node/impl.clj
```diff
@@ -111,8 +113,13 @@
...
+      (try
+        (-> (xtp/prepare-sql this query query-opts)
+            (then-execute-prepared-query allocator query-timer query-opts))
+        (catch Exception e
+          (when query-error-counter
+            (.increment ^Counter query-error-counter))
+          (throw e)))))
 
   (open-xtql-query [this query query-opts]
     (let [query-opts (-> query-opts (with-query-opts-defaults this))]
```
FiV0, line 129: Even though this is currently not priority, could we add the metrics also for XTQL?

core/src/main/clojure/xtdb/node/impl.clj
```diff
@@ -111,8 +113,13 @@
...
   (open-sql-query [this query query-opts]
     (let [query-opts (-> query-opts (with-query-opts-defaults this))]
-      (-> (xtp/prepare-sql this query query-opts)
-          (then-execute-prepared-query allocator query-timer query-opts))))
+      (try
+        (-> (xtp/prepare-sql this query query-opts)
+            (then-execute-prepared-query allocator query-timer query-opts))
+        (catch Exception e
+          (when query-error-counter
+            (.increment ^Counter query-error-counter))
```
FiV0: I think you are double counting here. If you follow `then-execute-prepared-query` you end up in `wrap-query` at some point. With all the different phases this might be a bit hard to see, I agree. Maybe let's pair on this, so we can best figure out where to put the counters on the node and in pgwire.

FiV0 (commented):
I left two comments. In general I would try to wrap the exceptions in pgwire closer to the action (SQL string (for `interpret-sql` this is already good I think), preparedQuery, node etc..) as we don't want to count other pgwire issues in these counters.

core/src/main/clojure/xtdb/query.clj
```diff
@@ -77,14 +79,19 @@
...
         (throw (InterruptedException.)))
 
       (binding [expr/*clock* clock]
-        (.tryAdvance cursor c)))
+        (try
+          (.tryAdvance cursor c)
+          (catch Exception e
+            (when query-error-counter
+              (.increment query-error-counter))
+            (throw e)))))
```
FiV0: Let's remove this one and try to catch it everywhere above.

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1505,32 +1507,38 @@
...
-                                                                                    (into [] (comp (filter (comp nil? :col-type))
-                                                                                                   (map :typname)
-                                                                                                   (distinct)))))))})))
+    (try
+      (interpret-sql query {:default-tz (.getZone clock)
+                            :watermark-tx-id watermark-tx-id
+                            :session-parameters session-parameters})
+      (catch Exception e
+        (when query-error-counter
+          (.increment query-error-counter))
```
FiV0: We keep this error counter because an error here does never go through the node.

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1505,32 +1507,38 @@
...
+                (and (= :prepare statement-type)
+                     (= :query (:inner-statement-type stmt))))
+          (let [param-col-types (mapv :col-type param-types)]
+            (when (some nil? param-col-types)
+              (throw (ex-info "unsupported param-types in query"
+                              {::client-error (err-protocol-violation (str "Unsupported param-types in query: "
+                                                                           (pr-str (->> param-types
+                                                                                        (into [] (comp (filter (comp nil? :col-type))
+                                                                                                       (map :typname)
+                                                                                                       (distinct)))))))})))
```
FiV0: In here it's just that protocol violation, so let's increment the `error-counter` here.

core/src/main/clojure/xtdb/node/impl.clj
FiV0: So lets add the `error-counter` on the `PLocalNode` methods. `prepare-sql`, `prepare-sql` and `prepare-ra`.

core/src/main/clojure/xtdb/pgwire.clj
FiV0: Let's scrap this one for now until we come back to adding transaction-error counters. So now error counting at all at this place.

core/src/main/clojure/xtdb/pgwire.clj
FiV0: We still keep `query-error-counter`

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1606,65 +1618,70 @@
...
+                                        (finally
+                                          (util/close args-rel))))))))
+                       (finally
+                         (util/close bq))))
+
+          stmt)))
+    (catch Exception e
+      (when query-error-counter
+        (.increment query-error-counter))
+      (throw e))))
```
FiV0: This one seems reasonable.

core/src/main/clojure/xtdb/node/impl.clj
```diff
@@ -183,7 +190,8 @@
 (defmethod ig/init-key :xtdb/node [_ {:keys [metrics-registry] :as deps}]
   (let [node (map->Node (-> deps
                             (assoc :query-timer (metrics/add-timer metrics-registry "query.timer"
-                                                                   {:description "indicates the timings for queries"}))))]
+                                                                   {:description "indicates the timings for queries"}))
```
FiV0, line 205: Let's add the `warning` counter on node level. On the `PreparedQuery` object you can call the `warnings` method.

FiV0:
As this seems super complicated. We decided to try to do a more radical simple approach. 
We count an error if we send out an error in pgwire.
We count an error if it goes through `open-sql-query`, as that it is the code path taken by all other APIs.

core/src/main/clojure/xtdb/node/impl.clj
```diff
@@ -111,13 +113,23 @@
...
     (let [query-opts (-> query-opts (with-query-opts-defaults this))]
-      (-> (xtp/prepare-sql this query query-opts)
-          (then-execute-prepared-query allocator query-timer query-opts))))
+      (try
+        (-> (xtp/prepare-sql this query query-opts)
+            (then-execute-prepared-query allocator query-timer (assoc query-opts :query-error-counter query-error-counter)))
+        (catch Exception e
+          (when query-error-counter
+            (.increment query-error-counter))
+          (throw e)))))
```
FiV0, line 126: As we are now counting the errors in `wrapping-error-count-cursor` this `try/catch` is no longer needed.

FiV0 (changes requested):
All looks good to me. Can we add a metrics test namespace where we do the query
```sql
SELECT 0/1
```
and 
```sql
SELECT non_existing_column FROM non_existing_table
```
against the node and via pgwire and check that the registry counters match what we expect afterwards.

core/src/main/clojure/xtdb/node/impl.clj
tggreene: It doesn't appear to pick up parsing errors

FiV0: Ah I see. Let's keep it then. Can you add a short comment about it.

core/src/main/clojure/xtdb/node/impl.clj
```diff
@@ -114,13 +116,24 @@
 
   (open-sql-query [this query query-opts]
     (let [query-opts (-> query-opts (with-query-opts-defaults this))]
-      (-> (xtp/prepare-sql this query query-opts)
-          (then-execute-prepared-query allocator query-timer query-opts))))
+      ;; We catch exceptions here to count errors outside of query execution e.g. parsing errors
+      (try
+        (-> (xtp/prepare-sql this query query-opts)
+            (then-execute-prepared-query allocator query-timer (assoc query-opts :query-error-counter query-error-counter)))
```
FiV0: Sorry to be a nitpicker. Can we try to pass the `query-timer` and `query-error-counter` around the same way as they are both metrics. Somethin like a metrics map as second arg here.

FiV0: And also pass that metrics-map, just call it `metrics` down to `open-cursor-as-stream` as well.

FiV0 (approved):
We are counting warnings always on Node level. For errors there are two parts.
- PGWire: We count an error if we send an error response via pgwire.
- PNode protocol: Everything else should go via the `PNode` protocol and this is where we are counting errors for all other queries to a node.

PR #4091, opened 2025-01-30
Expression Engine consistent hashing in arrow vectors

### LINKED ISSUE #4061: Outer join fails to join two date, when only one is calculated
If we have these records:
```sql
INSERT INTO docs RECORDS
  {_id: 1, value: TIMESTAMP '2024-01-01T00:00:00Z'},
  {_id: 2, value: TIMESTAMP '2024-01-01T00:10:00Z'};
```

The following query *should* join the table onto itself, but doesn't:
```sql
FROM docs AS d1
LEFT OUTER JOIN docs AS d2
  ON d1.value = d2.value + INTERVAL 'PT0M' -- Same with DATE_BIN(INTERVAL 'PT5M', d2.value)
SELECT
  d1._id AS _id1,
  d1.value AS value1,
  d2._id AS _id2,
  d2.value AS value2
```

Even though the values are still `=` to eachother:
```sql
FROM docs
SELECT *, value = (value + INTERVAL 'PT0M') AS eq
```

---

The query without the calculation joins as expected:
```sql
FROM docs AS d1
LEFT OUTER JOIN docs AS d2
  ON d1.value = d2.value
SELECT
  d1._id AS _id1,
  d1.value AS value1,
  d2._id AS _id2,
  d2.value AS value2
```

Also, if both values have a calculation, the join works as expected:
```sql
FROM docs AS d1
LEFT OUTER JOIN docs AS d2
  ON d1.value + INTERVAL 'PT0M' = d2.value  + INTERVAL 'PT0M'
SELECT
  d1._id AS _id1,
  d1.value AS value1,
  d2._id AS _id2,
  d2.value AS value2
```

The above is a good workaround, if you're running into this issue.

### LINKED ISSUE #4078: Fixing joins across polymorphic columns
As unearthed by #4061 we are currently not supporting joins across polymorphic columns where the value of the joined columns only differs in the type. Best illustrated by the following test
```clj
(t/deftest joins-across-polymorphic-data-4061
  (xt/submit-tx tu/*node* [[:put-docs :docs {:xt/id 1 :a 1}]
                           [:put-docs :docs2 {:xt/id 1 :a 1.0}]])

  (t/is (= [{:a1 1, :a2 1.0}]
           (xt/q tu/*node* "SELECT d1.a a1, d2.a a2 FROM docs d1, docs2 d2 WHERE d1.a = d2.a"))))
```
Most of them time users will likely run into this is when joining on temporal data as we support different resolutions for `:timestamp-tz`. It's unlikely that you are joining on a double and an int, but given our logic in the expression engine we should support it correctly. 

The fix for the join logic is to always hash the widest supertype (double for any number, `:timestamp-tz :nano` etc...). This means we need to update all vectors with custom hashing logic based on our column type hierarchy. 

There is also some work to be done in the metadata comparison logic. Currently we hash for different `num-types`, but if logically equivalent `num-type` values hash to the same value all that can go as well. This part is also the bit that requires the index-bump, as the blooms on disk will change.

TODO:
- [x] Writing tests for joins
- [x] Wrtiting test for group-by
- [x] Writing tests for pushdown-blooms

FiV0:
This adds custom hashing in the Java `IVectorReader` classes and the new vectors so that values that are equal according to our Expression Engine hash to the same value and consequently joining, grouping and set operators should work correctly.

I did a small benchmark joining 1M ints/doubles across two tables and the performance was no different to `main`.

Todo:
- [ ] Check that all types that are equal according to the EE, hash to the same thing (numbers, date-time, dates)
     - [ ] ~Intervals~
     - I am not hashing decimals correctly as they are currently poorly supported anyway
- [x] Write a couple of more tests
     - [x] other types
     - [x] other operators


core/src/main/java/xtdb/vector/ValueVectorReader.java
```diff
@@ -77,9 +81,14 @@ public Field getField() {
...
+        return HashUtilKt.byteBufferToBytes(getBytes(idx));
+    }
+
     @Override
-    public int hashCode(int idx, ArrowBufHasher hasher) {
-        return vector.hashCode(idx, hasher);
+    public int hashCode(int idx, XtByteHasher hasher) {
+        if (isNull(idx)) return ArrowBufPointer.NULL_HASH_CODE;
+        return hasher.hashValue(getRawBytes(idx));
     }
```
FiV0, line 94: My idea was that for all vector types for which we don't specialize `hashCode` we could default to hashing the raw bytes. This would mean that for example `:keyword` and `:string` still hash to the same value, but they get filtered afterwards by the EE. The other option is to use the vector type + raw data in the default hashing. The main reason I was thinking of `getRawBytes` was so that no objects get created.

core/src/main/clojure/xtdb/expression/metadata.clj
```diff
@@ -27,32 +28,50 @@
...
-                            :meta-value meta-value
-                            :col-type col-type
-                            :field field,
-                            :value-expr value-expr
-                            :bloom-hash-sym (when (= meta-value :bloom-filter)
-                                              (gensym 'bloom-hashes))}))}))))
+              (let [base-col-types (->> (get col-types field)
+                                        types/flatten-union-types)]
+
+                (if (= meta-value :bloom-filter)
```
FiV0, line 34: I only updated this bit for bloom filters. I don't know if the min and max logic should also be updated to do something smarter with regards to num types hierarchy or date-time hierarchy. Maybe one for another ticket.

jarohen: yes, if we're converting (e.g.) numbers to doubles for hashing we should probably use the same process for inequalities. let's figure out how much work it is to make the decision of whether to include it here or not?

core/src/main/kotlin/xtdb/util/HashUtil.kt
```diff
@@ -0,0 +1,13 @@
...
+
+fun ByteBuffer.toByteArray(): ByteArray {
+    return if (hasArray() && arrayOffset() == 0 && array().size == remaining()) {
+        array()
+    } else {
+        ByteArray(remaining()).also { get(it) }
+    }
+}
+
+fun byteBufferToBytes(buf: ByteBuffer): ByteArray = buf.toByteArray()
```
jarohen: this can probably go as a default impl on the Hasher interface?

i.e. `fun hash(value: ByteBuffer) = ...`

core/src/main/kotlin/xtdb/util/Hasher.kt
```diff
@@ -0,0 +1,44 @@
...
+import java.nio.ByteBuffer
+
+
+interface Hasher {
+
+    fun hash(value: ByteArray): Int
+    fun hash(value: Double): Int
+    fun hash(value: Long): Int
+
+    abstract class DefaultHasher : Hasher {
```
jarohen: yep, fair - if we need this we might as well make Hasher the abstract class instead, save the level of indirection

src/test/clojure/xtdb/node_test.clj
```diff
@@ -917,3 +917,10 @@ VALUES(1, OBJECT (foo: OBJECT(bibble: true), bar: OBJECT(baz: 1001)))"]])
   (t/is (thrown-with-msg? IllegalArgumentException
                           #"Port value out of range: 99999"
                           (xtn/start-node {:server {:port 99999}}))))
+
+
+(t/deftest backfill-test
```
jarohen: doesn't look like it's testing anything?

jarohen (commented):
nice one @FiV0 :clap:

let's have a chat about your metadata comment, and lmk if you want a quick pair on any of the Kotlin suggestions

cheers :pray:

core/src/main/clojure/xtdb/expression/metadata.clj
FiV0: Adding an extra card for min/max/bloom logic cleanup in metadata/expression.clj.

src/test/clojure/xtdb/node_test.clj
FiV0: Yep, leftover.

core/src/main/clojure/xtdb/expression/metadata.clj
FiV0: Carded up at https://github.com/xtdb/xtdb/issues/4116.

jarohen:
@FiV0: pushed the rebased branch to [my remote](https://github.com/jarohen/xtdb/tree/fix-joins-4078), commit f3fbf9effd3a0a99397554e4bff708db67703536 - could you push this commit to yours?

cheers mate :pray:

jarohen (approved):
@FiV0: pushed the rebased branch to [my remote](https://github.com/jarohen/xtdb/tree/fix-joins-4078), commit f3fbf9effd3a0a99397554e4bff708db67703536 - could you push this commit to yours and then merge?

cheers mate :pray:

PR #4095, opened 2025-01-30
Dropping file notification log

jarohen:
After #4079 we now have no need for the file notification log, so this PR drops it.

Breaking changes:

* Kafka configuration:
  * `txTopic` -> `topic`, `filesTopic` gone
  * `autoCreateTopics` -> `autoCreateTopic`
* Docker image env var `XTDB_TX_TOPIC` -> `XTDB_LOG_TOPIC`, `XTDB_FILES_TOPIC` removed.

I've also removed references to 'transaction log' in the documentation now that there's more than just transactions on there - it's now either just 'log' or, when that's ambiguous (i.e. compared to stdout-logging), 'message log'.

Also in this PR, I've changed the behaviour of listing objects on the buffer pool and object store. 

* Previously, we had a recursive `listAllObjects` and a shallow `listObjects(dir)`; the only usages of these were in tests and to list the chunk-metadata files (only a single directory anyway), so I've taken the opportunity to make this much more consistent.
* Now we have `listObjects()` and `listObjects(dir)` on both BP and obj-store, all are recursive. 
* Remote BP now delegates straight through to the obj-store (previously it kept the file-list cache).


jarohen:
@danmason: mind checking through your Helm stuff?
@refset: could you give the docs changes a quick glance to make sure they're consistent and clear?

cheers gents :pray:

danmason (approved):
I've not reviewed the whole PR, but `helm` and docker config changes all look good to me :)

PR #4111, opened 2025-02-03
Check earlier pages of tries for overlapping system time. Fix #4097

### LINKED ISSUE #4097: Older recency buckets can affect splitting in later buckets
Raised by @egg-juxt and distilled together with @mbutlerw.

The current assumption when calculating merge tasks is that earlier pages (earlier pages in the trie, i.e. smaller recency) that don't match the metadata of a query can only affect the bounding of  valid-to. The following test shows that this assumption is wrong and that earlier events even if they don't match the metadata, can split relevant rows in "later" matching pages.
```clj
(t/deftest recency-bucketing-bug
  (binding [c/*page-size* 16 
            c/*ignore-signal-block?* true
            #_#_c/*l1-file-size-rows* 10]
    (dotimes [x 1000]
      (xt/submit-tx tu/*node*
                    [[:put-docs {:into :docs
                                 :valid-from #xt/zoned-date-time "2024-01-01T00:00Z[UTC]" ,
                                 :valid-to #xt/zoned-date-time "2025-01-01T00:00Z[UTC]" }
                      {:xt/id x :col1 "yes"}]])
      (xt/submit-tx tu/*node* [[:put-docs {:into :docs
                                           :valid-from #xt/zoned-date-time "2024-01-03T00:00Z[UTC]" ,
                                           :valid-to #xt/zoned-date-time "2024-01-04T00:00Z[UTC]"}
                                {:xt/id x :col1 "no"}]]))
    (tu/finish-chunk! tu/*node*)

    (c/compact-all! tu/*node*)
    (t/is (= 1
             (xt/q tu/*node* "SELECT *, _valid_time FROM docs FOR ALL VALID_TIME WHERE _id = 0 AND col1 = 'yes'")))))
```

There is an event `{:xt/id 1 :col1 "yes"}` which is valid for a year (recency 2025) and a later event (in sys time) `{:xt/id 1 :col1 "no"}` wihch splits the earlier event but which is only valid for a day (recency 2024). We also ingest a bunch of other data to exhibit the bug, I suspect this has something to do with filling up the appropriate levels in the trie (more investigation needed). As the "no" event sits in earlier bucket and doesn't match the metadata it is not taken into account when resolving the valid time history of the "yes" event.

Image

As for a solution, to be on the safe side we need to include all pages that have events in system-time that come after the first matching metadata page. You could of course also check that the iid overlap is non empty with the matching metadata pages.

FiV0:
There are two levels of granularity when resolving data for a query, files and pages within those files. With our current approach we should always be able to strictly order files (for a given IID because that is what is important) according to system time. Meaning there is a list of files $f_1$ to $f_n$ where system time increases across these files, i.e. max_system_from($f_{n-1}$) <= min_system_from($f_n$). As #4097 shows this does not hold for pages within a given file $f_n$. It's possible that max_system_from($p_{n-1}$) > min_system_from($p_n$). If a metadata predicate than matches $p_n$ we must take $p_{n-1}$ to assure correct bitemporal splitting. The commit tries to address this issue.

The approach currently taken:
1. Find all pages that match metadata and query bound filters. Let that be set $S$
2. Find all pages not yet taken that can bound valid-to from $S$ (up to here this is what we do on `main`). We take $p_n$ if it is a non taken page and $p_n$ intersects the query bounds and can bound an event of $S$ in the query range.
3. Find all pages not yet taken that can affect resolution of pages taken so far, i.e. take page $p_n$ if min_system_from($S$) <= max_system_from($p_n$).

Do 2. and 3. need to be applied recursively until they are exhausted or can one apply 2. + 3. in any order and only on $S$?

There are lots of optimizations one can do here. For example 3 only needs to be applied on per file level where as 2 needs to be applied across files. One could do the steps only based on the original $S$ (I think this is sound) or apply them on expanding sets (definitely not incorrect, maybe just more work later).

### **UPDATE**
I am now doing the following:
1.  Find all pages that match metadata and query bound filters. Let that be set $S$

All other pages that still need to be taken only get taken to potentially split rows in $S$. To split rows in $S$ the following two requirements are needed for a page $P$ not yet in $S$.
- min_system_from($S$) <= max_system_from($P$)
-  valid_time($S$) $\cap$ valid_time($P$) $\not= \emptyset$  

Where `valid_time` is the interval on min_valid_from to max_valid_to across all rows in the set.

I am also now using `max_system_from` which is not part of `TemporalBounds` so I am wondering if we have some wrong abstraction here?

### DIFF
--- a/core/src/main/clojure/xtdb/metadata.clj
+++ b/core/src/main/clojure/xtdb/metadata.clj
@@ -341,7 +341,8 @@
           ^long valid-from-idx (.get page-idx-cache (PageIndexKey. "_valid_from" page-idx))
           ^long valid-to-idx (.get page-idx-cache (PageIndexKey. "_valid_to" page-idx))]
       (TemporalBounds. (TemporalDimension. (.getLong min-rdr valid-from-idx) (.getLong max-rdr valid-to-idx))
-                       (TemporalDimension. (.getLong min-rdr system-from-idx) Long/MAX_VALUE))))
+                       (TemporalDimension. (.getLong min-rdr system-from-idx) Long/MAX_VALUE)
+                       (.getLong max-rdr system-from-idx))))
 
   AutoCloseable
   (close [_]
--- a/core/src/main/clojure/xtdb/trie.clj
+++ b/core/src/main/clojure/xtdb/trie.clj
@@ -14,7 +14,7 @@
            (xtdb.arrow Relation)
            xtdb.BufferPool
            (xtdb.trie ArrowHashTrie$Leaf HashTrie$Node ISegment MemoryHashTrie MemoryHashTrie$Leaf MergePlanNode TrieWriter)
-           (xtdb.util TemporalBounds)))
+           (xtdb.util TemporalBounds TemporalDimension)))
 
 (defn ->l0-l1-trie-key [^long level, ^long block-idx]
   (assert (<= 0 level 1))
@@ -134,13 +134,21 @@
 (defn min-valid-from ^long [^TemporalBounds tb]
   (.getLower (.getValidTime tb)))
 
+(defn min-system-from ^long [^TemporalBounds tb]
+  (.getLower (.getSystemTime tb)))
+
+(defn max-system-from ^long [^TemporalBounds tb]
+  (.getMaxSystemFrom tb))
+
 (defn ->merge-task
   ([mp-pages] (->merge-task mp-pages (TemporalBounds.)))
   ([mp-pages ^TemporalBounds query-bounds]
    (let [leaves (ArrayList.)]
      (loop [[mp-page & more-mp-pages] mp-pages
             node-taken? false
+            smallest-valid-from Long/MAX_VALUE
             largest-valid-to Long/MIN_VALUE
+            smallest-system-from Long/MAX_VALUE
             non-taken-pages []]
        (if mp-page
          (let [^TemporalBounds page-bounds (temporal-bounds mp-page)
@@ -152,28 +160,30 @@
                (.add leaves mp-page)
                (recur more-mp-pages
                       true
+                      (min smallest-valid-from (min-valid-from page-bounds))
                       (max largest-valid-to (max-valid-to page-bounds))
+                      (min smallest-system-from (min-system-from page-bounds))
                       non-taken-pages))
 
              (recur more-mp-pages
                     node-taken?
+                    smallest-valid-from
                     largest-valid-to
+                    smallest-system-from
                     (cond-> non-taken-pages
-                      (let [page-valid-time (.getValidTime page-bounds)]
-                        (and node-taken?
-                             (.intersects (.getSystemTime page-bounds) (.getSystemTime query-bounds))
-                             ;; this page can bound a page in the query set
-                             (< (.getLower (.getValidTime query-bounds)) (.getUpper page-valid-time))
-                             (< (.getLower page-valid-time) largest-valid-to)))
+                      (.intersects (.getSystemTime page-bounds) (.getSystemTime query-bounds))
                       (conj mp-page)))))
 
          (when node-taken?
-           (loop [[page & more-pages] non-taken-pages]
-             (when page
-               (let [smallest-valid-from (min-valid-from (temporal-bounds page))]
-                 (when (< smallest-valid-from largest-valid-to)
-                   (.add leaves page))
-                 (recur more-pages))))
+           (let [valid-time (TemporalDimension. smallest-valid-from largest-valid-to)]
+             (loop [[page & more-pages] non-taken-pages]
+               (when page
+                 (let [page-valid-time (.getValidTime ^TemporalBounds (temporal-bounds page))
+                       page-largest-system-from (max-system-from (temporal-bounds page))]
+                   (when (and (<= smallest-system-from page-largest-system-from)
+                              (.intersects valid-time page-valid-time))
+                     (.add leaves page))
+                   (recur more-pages)))))
            (vec leaves)))))))
 
 (definterface IDataRel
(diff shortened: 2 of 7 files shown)


core/src/main/kotlin/xtdb/util/TemporalBounds.kt
```diff
@@ -27,8 +27,11 @@ data class TemporalDimension(
 
 data class TemporalBounds(
     val validTime: TemporalDimension = TemporalDimension(),
-    val systemTime: TemporalDimension = TemporalDimension()
+    val systemTime: TemporalDimension = TemporalDimension(),
+    val maxSystemFrom: Long = Long.MAX_VALUE,
```
FiV0, line 31: This is just to make things work for now. We likely want a different abstraction here if `maxSystemFrom` is important.

jarohen:
Thanks @FiV0, found that write-up very helpful :pray: 

So IIRC when you first implemented this there were two distinct phases:

1. we find all the pages that can match the temporal bounds of the query
2. we find all the pages that can intersect with the temporal bounds of the pages in group 1.

specifically, we didn't then need to do this recursively (i.e. find all the pages that intersect with group 2) - group 2 were only included to bound the events in group 1.

That said, the page/file distinction is interesting here. I wouldn't be surprised if there was the assumption you mentioned that the pages are totally ordered, and I believe you're right in saying it's only a partial order. I find it a little weird that the solution here appears to be asymmetric - I'd (perhaps naively) expect us to need to check both directions (past and future) in the same way.

Anyway, all of this is to say that I'm now suitably in context for a chat :slightly_smiling_face:

core/src/main/clojure/xtdb/trie.clj
```diff
@@ -260,25 +267,24 @@
...
 
          (when node-taken?
            (loop [[page & more-pages] non-taken-pages]
              (when page
-               (let [smallest-valid-from (min-valid-from (temporal-bounds page))]
-                 (when (< smallest-valid-from largest-valid-to)
+               (let [smallest-valid-from (min-valid-from (temporal-bounds page))
+                     largest-valid-from (max-system-from (temporal-bounds page))]
+                 (when (or (< smallest-valid-from largest-valid-to)
+                           (<= smallest-system-from largest-valid-from))
```
FiV0: Correct naming.

core/src/main/clojure/xtdb/metadata.clj
```diff
@@ -347,7 +347,8 @@
           ^long valid-from-idx (.get page-idx-cache (PageIndexKey. "_valid_from" page-idx))
           ^long valid-to-idx (.get page-idx-cache (PageIndexKey. "_valid_to" page-idx))]
       (TemporalBounds. (TemporalDimension. (.getLong min-rdr valid-from-idx) (.getLong max-rdr valid-to-idx))
-                       (TemporalDimension. (.getLong min-rdr system-from-idx) Long/MAX_VALUE))))
+                       (TemporalDimension. (.getLong min-rdr system-from-idx) Long/MAX_VALUE)
+                       (.getLong max-rdr valid-from-idx))))
```
FiV0: `system-from-idx`

FiV0:
Updated the description once more to reflect the PR. I think one could still write more tests to gain more confidence in the approach.

src/test/clojure/xtdb/trie_test.clj
```diff
@@ -173,23 +173,89 @@
...
+        [valid-to1 valid-to2] (mapv inst->micros [#inst "2021-01-05" #inst "2020-01-04"])]
+    (t/is (= [1 0] (->> (trie/->merge-task
+                         [(->mock-merge-plan-page 0 false valid-from1 valid-to1 system-from1)
+                          (->mock-merge-plan-page 1 true  valid-from2 valid-to2 system-from2)])
+                        (map :page)))
+          "earlier pages that contain data with later system time need to be taken")))
 
-(def ^:private inst->micros (comp time/instant->micros time/->instant))
+(deftest test-to-merge-task
+  (let [insts (tu/->instants :day)
```
jarohen: use numbers rather than instants, save the round-trip, then we can just reason about raw nums

src/test/clojure/xtdb/trie_test.clj
```diff
@@ -173,23 +173,89 @@
...
+                                          (tu/->min-max-page-bounds vf2 Long/MAX_VALUE sf2 Long/MAX_VALUE))
+                       ->pages))
+          "No need to take extra pages!")
+
+    (t/is (= #{1 2} (->> (trie/->merge-task [(->mock-merge-plan-page 0 false vf1 (inc vt1) sf1 Long/MAX_VALUE sf1)
+                                             (->mock-merge-plan-page 1 true vf2 (inc vt2) sf2 Long/MAX_VALUE sf2)
+                                             (->mock-merge-plan-page 2 false vf3 (inc vt3) sf3 Long/MAX_VALUE sf3)]
+                                            (tu/->min-max-page-bounds vf2 Long/MAX_VALUE sf2 Long/MAX_VALUE))
+                         ->pages))
+          "We make the valid time windows slightly larger so it overlaps. Third page needs to get taken.")
```
jarohen: let's express these messages in terms of allen intervals/similar

src/test/clojure/xtdb/trie_test.clj
```diff
@@ -359,56 +425,90 @@
...
+               micros-2021 8}]]
+
+      ;; valid-from - valid-to (if not otherwise specified valid-from = system-from)
+      (with-open [al (RootAllocator.)
+                  t1-root (tu/open-arrow-hash-trie-rel al t1)
+                  t2-root (tu/open-arrow-hash-trie-rel al t2)]
+
+        (let [t1-page-bounds {#{3}       [micros-2019 micros-2020]
+                              #{0 1 2 4} [micros-2020 micros-2021]
+                              ;; 5 has an retprspective update
```
FiV0: typo

src/test/clojure/xtdb/trie_test.clj
FiV0: I added some comments in that test. I wasn't able to purely write it in terms Allen intervals.

PR #4128, opened 2025-02-05
Add terraform and helm config for Google Cloud. 

### LINKED ISSUE #4113: Add well documented shared `terraform` and `helm` templates for Google Cloud.
Want to create and document a set of `terraform`/`helm` templates that are easy to share and use for Google cloud:
- Add a terraform file setting up all of the cloud infra XTDB requires and for setting up/running kubernetes:
  - Setting up an XTDB service account. 
  - Setting up workload federation.
  - Using [**GoogleCloud/storage-bucket**]( https://registry.terraform.io/modules/terraform-google-modules/cloud-storage/google/latest) for setting up storage bucket and perms.
  - Using [**GoogleCloud/network**](https://registry.terraform.io/modules/terraform-google-modules/network/google/latest) for setting up a simple VPC for the Kubernetes Engine infra.
  - Using [**GoogleCloud/kubernetes-engine**](https://registry.terraform.io/modules/terraform-google-modules/kubernetes-engine/google/latest) for setting up GKE with a system and application node pool.
  - Setting up everything else we need around those - using sensible variables to allow easy customisation with `terraform.tfvars`.
- Using `helm` for deployments: 
  - Make a copy of the `azure` helm template and update as necessary for setting up on google cloud.
- Test deployment of both of the above.
- Add google cloud helm to helm deploy task, release version 2.0.0-SNAPSHOT to GHCR.
- Update the `Google Cloud` docs to include notes on using both terraform and helm.

Extension:
- Add "getting started with google cloud" guide.

danmason:
Resolves #4113 
Docs Build: https://google-cloud-tf-and-helm.d2zcybuz6k9g4m.amplifyapp.com/

- Add a terraform file setting up all of the cloud infra XTDB requires and for setting up/running kubernetes:
  - Setting up an XTDB service account. 
  - Using [**GoogleCloud/storage-bucket**]( https://registry.terraform.io/modules/terraform-google-modules/cloud-storage/google/latest) for setting up storage bucket and perms.
  - Using [**GoogleCloud/network**](https://registry.terraform.io/modules/terraform-google-modules/network/google/latest) for setting up a simple VPC for the Kubernetes Engine infra.
  - Using [**GoogleCloud/kubernetes-engine**](https://registry.terraform.io/modules/terraform-google-modules/kubernetes-engine/google/latest) for setting up GKE with a system and application node pool.
  - Setting up everything else we need around those - using sensible variables to allow easy customisation with `terraform.tfvars`.
- Add helm config for google cloud:
  - Based on azure helm config.
  - Added to helm deploy task.
  - Deploying all necessary config to kubernetes,  
- Update the `Google Cloud` docs to include notes on using both terraform and helm.
- Add a "getting started with google cloud" guide.

### DIFF
--- a/docs/src/content/docs/ops/guides/starting-with-gcp.adoc
+++ b/docs/src/content/docs/ops/guides/starting-with-gcp.adoc
@@ -1,0 +1,302 @@
+---
+title: Setting up a cluster on Google Cloud
+---
+
+This guide will walk you through the process of configuring and running an XTDB Cluster on Google Cloud. This setup includes:
+
+* Using **Google Cloud Storage** as the remote storage implementation.
+* Utilizing **Apache Kafka** as the shared message log implementation.
+* Exposing the cluster to the internet via a Postgres wire-compatible server and HTTP.
+ 
+The required Google Cloud infrastructure is provisioned using **Terraform**, and the XTDB cluster and it's resources are deployed on link:https://cloud.google.com/kubernetes-engine?hl=en[**Google Kubernetes Engine**^] using **Helm**.
+
+Although we provide numerous parameters to configure the templates, you are encouraged to edit them, use them as a foundation for more advanced use cases, and reuse existing infrastructure when suitable. 
+These templates serve as a simple starting point for running XTDB on Google Cloud and Kubernetes, and should be adapted to meet your specific needs, especially in production environments.
+
+This guide assumes that you are using the default templates.
+
+== Requirements 
+
+Before starting, ensure you have the following installed:
+
+* The **Google Cloud CLI** - See the link:https://cloud.google.com/sdk/docs/install[**Installation Instructions**^].
+* **Terraform** - See the link:https://developer.hashicorp.com/terraform/tutorials/aws-get-started/install-cli[**Installation Instructions**^].
+* **kubectl** - The Kubernetes CLI, used to interact with the AKS cluster. See the link:https://kubernetes.io/docs/tasks/tools/install-kubectl/[**Installation Instructions**^].
+* **Helm** - The Kubernetes package manager, used to deploy numerous components to AKS. See the link:https://helm.sh/docs/intro/install/[**Installation Instructions**^].
+
+=== Requirements on Google Cloud
+
+Within Google Cloud, ensure that you have an existing Google Cloud project you can deploy into, and ensure the following APIs are enabled on the project:
+
+* Cloud Storage API
+* IAM API
+* Compute Engine API
+* Kubernetes Engine API
+
+Additionally, ensure that the following permissions are granted to the logged-in user (at a minimum):
+
+* `Storage Admin` - Required for creating and managing Google Cloud Storage buckets.
+* `Service Account Admin` - Required for creating and managing service accounts.
+* `Kubernetes Engine Admin` - Required for creating and managing Google Kubernetes Engine clusters and their resources.
+
+To login into the Google cloud project using the command line, run the following command and run through the steps to authenticate:
+
+```bash
+gcloud init
+```
+
+This allows you to perform necessary operations on Google Cloud - primarily, creating and managing infrastructure using Terraform.
+
+[#terraform]
+== Getting started with Terraform
+
+The following assumes that you are authenticated on the Google Cloud CLI, have Terraform installed on your machine, and are located within a directory that you wish to use as the root of the Terraform configuration.
+
+First, make the following `terraform init` call:
+```
+terraform init -from-module github.com/xtdb/xtdb.git//google-cloud/terraform
+```  
+
+This will download the Terraform files from the XTDB repository, and initialize the working directory.
+
+NOTE: For the sake of this guide, we store Terraform state locally. 
+However, to persist the state onto Google Cloud, you will need to configure a remote backend using Google Cloud Storage. 
+This allows you to share the state file across teams, maintain versioning, and ensure consistency during deployments. 
+For more info, see the link:https://developer.hashicorp.com/terraform/language/backend/gcs[**Terraform gcs backend**^] documentation.
+
+== What is being deployed on Google Cloud?
+
+The sample Terraform directory sets up a few distinct parts of the infrastructure required by XTDB. 
+If using the default configuration, the following will be created:
+
+* **IAM Service Account** for accessing required Google Cloud resources.
+* **Google Cloud Storage Bucket** for remote storage.
+** Configured with associated resources using the link:https://registry.terraform.io/modules/terraform-google-modules/cloud-storage/google/latest[**GoogleCloud/storage-bucket**^] Terraform module.
+** Adds required permissions to the Service Account.
+* **Virtual Private Cloud Network** for the XTDB GKE cluster.
+** Configured with associated resources using the link:https://registry.terraform.io/modules/terraform-google-modules/network/google/latest[**GoogleCloud/network**^] Terraform module.
+* **Google Kubernetes Engine Cluster** for running the XTDB resources.
+** Configured with associated resources using the link:https://registry.terraform.io/modules/terraform-google-modules/kubernetes-engine/google/latest[**GoogleCloud/kubernetes-engine**^] Terraform module.
+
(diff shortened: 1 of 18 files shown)


google-cloud/terraform/main.tf
```diff
@@ -0,0 +1,94 @@
+resource "google_service_account" "xtdb_service_account" {
+  project = var.project_id
+
+  account_id   = var.service_account_name
+  display_name = "XTDB Service Account"
+}
+
+resource "google_service_account_iam_binding" "xtdb_service_account_binding" {
```
danmason: We do ahead of time create a service account binding to allow workload identity federation on the Kubernetes Service Account (prior to actually deploying the kubernetes namespace or kubernetes service account) - we do similar within Azure, too. (See https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#kubernetes-sa-to-iam)

This is mentioned within the docs for both _but_ I wonder if there's a better way to go about it.

danmason: We _may_ be able to auth the whole cluster/namespace as a bucket admin (from [here](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#verify)) - though I'm not particularly sure about that, either. Feels like IAM service account -> K8s service account makes sense.

google-cloud/helm/values.yaml
```diff
@@ -0,0 +1,127 @@
+nodeCount: 3
+
+image:
+  repository: ghcr.io/xtdb/xtdb-google-cloud
+  # tag: 2.0.0-beta5
```
jarohen, line 5: to be uncommented/deleted?

google-cloud/terraform/terraform.tfvars
```diff
@@ -0,0 +1,42 @@
...
+## GKE System Node Pool Settings
+system_node_pool_machine_type = "n2-standard-2"
+system_node_pool_locations    = "us-central1-a"
+system_node_pool_min_count    = 1
+system_node_pool_max_count    = 1
+system_node_pool_disk_size_gb = 50
+system_node_pool_disk_type    = "pd-standard"
+
+## GKE Application Node Pool Settings
+application_node_pool_machine_type = "n2-highmem-4"
```
jarohen: assume you've confirmed this is the most appropriate default node type for XT on GCP?

google-cloud/helm/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+  # See https://docs.oracle.com/cd/E13150_01/jrockit_jvm/jrockit/jrdocs/refman/optionX.html
+  jdkOptions: "-Xmx2000m -Xms2000m -XX:MaxDirectMemorySize=2000m -XX:MaxMetaspaceSize=500m"
+
+  # We mount and use a YAML file for configuring the XTDB nodes -
+  # We create a default file, setup wihin templates/xtdbyamlconfig.yaml using values set above.
+  #
+  # Setting `yamlConfigOverride/xtdbconfig.yaml` will override the contents of the config file 
+  # Ensure that you inline any values/set them using !Env tags
+  # See the XTDB configuration documentation for options: https://docs.xtdb.com/ops/config.html
+  yamlConfigOverride:
```
jarohen: do we need this to be an override? i.e. if we specify xtdbconfig.yaml here every time, and drop the override logic in xtdbyamlconfig, what happens?

google-cloud/terraform/main.tf
jarohen: openly, I'm not following the implications here - lmk if you'd like to go into it in more detail :duck:

jarohen (approved):
cheers @danmason - clear, concise, precise docs, well done :clap: 

happy for you to merge without further review, lmk if you want to discuss any of the below :slightly_smiling_face:

google-cloud/terraform/terraform.tfvars
danmason: - Take an action to set this up with `c4` machine type, make note about quotas.

google-cloud/helm/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+    #   healthz:
+    #     port: 8080
+
+    #   modules:
+    #   - !HttpServer
+    #     port: 3000
+
+  # Persisent Volume Claim settings for XTDB local disk cache (one per node)
+  localDiskCache:
+    volumeClaimTemplate:
```
danmason: Switch these over to using disk storage local to the node.

google-cloud/terraform/terraform.tfvars
```diff
@@ -0,0 +1,42 @@
...
+application_node_pool_locations    = "us-central1-a"
+application_node_pool_min_count    = 1
+application_node_pool_max_count    = 1
```
danmason: Minimum production viable infra here would be at least 3 nodes over 3 availability zone.

docs/src/content/docs/ops/guides/starting-with-gcp.adoc
```diff
@@ -0,0 +1,258 @@
...
+The GKE cluster is now ready for deployment,
+
+'''
+
+=== Deploying an example Kafka 
+
+To deploy a basic set of Kafka resources within GKE, you can make use of the `bitnami/kafka` Helm chart. Run the following command:
+
+```bash
+helm install kafka oci://registry-1.docker.io/bitnamicharts/kafka \
```
danmason, line 161: If we wanted to make this guide more "production like" could use better kafka setup defaults.

google-cloud/helm/values.yaml
danmason: - Continue using xtdbConfig to set necessary values
- Remove extra contents of xtdbyamlconfig
- Inside of values.yaml have the `xtdbconfig.yaml` parameter which is used by the configmap.
- YAML Config under `nodeConfig` in values, using Env Tags, setting ENV on statefulset etc.

google-cloud/terraform/main.tf
danmason: ## DONE
- Extract namespace and service account out of helm.
- Remove iambinding from terraform,
- Add instructions for setting up namespace/serviceaccount/iam binding via CLI to the doc & guide

danmason: Perhaps as a pre-install hook?

google-cloud/helm/values.yaml
danmason: https://cloud.google.com/kubernetes-engine/docs/concepts/local-ssd

danmason:
^ Number of tasks above to be completed - we'll want to similarly do these changes within Azure/S3 :)

google-cloud/helm/values.yaml
danmason: ## DONE
- Have done this within the google cloud config

google-cloud/helm/values.yaml
danmason: Seems easy enough using `ephemeral-storage` and `emptyDir` - means that the storage is very much only for the pods lifetime, but I think that is fine for the diskcache. 

## DONE

- I've added in ephemeral disk and empty dir in statefulset for google cloud.
- Setup ephemeral local ssd to nodes https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/local-ssd#1st-or-2nd-generation

google-cloud/helm/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+  repository: ghcr.io/xtdb/xtdb-google-cloud
+  # tag: 2.0.0-beta5
+  pullPolicy: IfNotPresent
+
+nodeSelector:
+  iam.gke.io/gke-metadata-server-enabled: "true"
+
+tolerations: []
+
+affinity: {}
```
danmason, line 15: ## DONE

- Nodeselector using xtdbpool.

google-cloud/terraform/terraform.tfvars
danmason: ## DONE
- Setup ephemral local ssds, see storage notes.
- Setup multiple nodes spread over AZs
- Setup quotas as necessary (if possible?) - or test with `n2` nodes.

PR #4150, opened 2025-02-11
Add transaction error counter

### LINKED ISSUE #3899: Add monitoring for Transaction Timeouts/Failures/Errors.
Very similar to #3898 , but for Transactions.

Adding some form of [counter](https://docs.micrometer.io/micrometer/reference/concepts/counters.html) against Failures/Errors:
  - Transaction errors/failures within XTDB.
- Want to display these over time in a graph on the ["XTDB Monitoring Dashboard"](https://github.com/xtdb/xtdb/tree/main/monitoring/grafana/dashboards).

tggreene:
Adds `tx.error`, picks up errors in indexer, submit-tx and pgwire when dealing with DML.

There is the possibility that bad queries like `INSERT foo` will be treated as query errors in pgwire but this might be acceptable as there's no inherent meaning (i.e. this is definitely DML) to `INSERT foo`.

We're catching IllegalArugmentExceptions on the node level (in submit-tx) to ensure we capture errors that occur during pre-submit phase simplifying inserts.

### DIFF
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -300,8 +300,10 @@
 
 (defn- client-err
   ([client-msg] (client-err client-msg nil))
-  ([client-msg data]
-   (ex-info client-msg (assoc data ::client-error (err-protocol-violation client-msg)))))
+  ([client-msg {:keys [error-type] :as data}]
+   (ex-info client-msg (assoc data ::client-error
+                              (cond-> (err-protocol-violation client-msg)
+                                error-type (assoc :error-type error-type))))))
 
 ;;TODO parse errors should return a PSQL parse error
 ;;this code is generic, but there are specific ones a well
@@ -1052,7 +1054,7 @@
 
   If the connection is operating in the :extended protocol mode, any error causes the connection to skip
   messages until a msg-sync is received."
-  [{:keys [conn-state ^Counter query-error-counter] :as conn} err]
+  [{:keys [conn-state ^Counter query-error-counter ^Counter tx-error-counter] :as conn} {:keys [error-type] :as err}]
 
   ;; error seen while in :extended mode, start skipping messages until sync received
   (when (= :extended (:protocol @conn-state))
@@ -1061,8 +1063,9 @@
   ;; mark a transaction (if open as failed), for now we will consider all errors to do this
   (swap! conn-state util/maybe-update :transaction assoc :failed true, :err err)
 
-  (when query-error-counter
-    (.increment ^Counter query-error-counter))
+  (when (and (not= :dml error-type)
+             query-error-counter)
+    (.increment query-error-counter))
 
   (cmd-write-msg conn msg-error-response {:error-fields err}))
 
@@ -1245,7 +1248,7 @@
   (try
     (xt/execute-tx node dml-ops tx-opts)
     (catch xtdb.IllegalArgumentException e
-      (throw (client-err (ex-message e))))
+      (throw (client-err (ex-message e) {:error-type :dml})))
     (catch Throwable e
       (log/debug e "Error on execute-tx")
       (let [msg "unexpected error on tx submit (report as a bug)"]
@@ -1282,7 +1285,8 @@
     (do (log/error "Missing types for params in DML statement")
         (cmd-send-error
          conn
-         (err-protocol-violation "Missing types for args - client must specify types for all params in DML statements")))
+         (-> (err-protocol-violation "Missing types for args - client must specify types for all params in DML statements")
+             (assoc :error-type :dml))))
 
     (let [{:keys [session transaction]} @conn-state
           ^Clock clock (:clock session)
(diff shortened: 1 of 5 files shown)


src/test/clojure/xtdb/metrics_test.clj
```diff
@@ -22,3 +24,29 @@
...
+    (t/is (thrown-with-msg? Exception #"" (jdbc/execute! tu/*conn* ["INSERT INTO foo (a) VALUES (42)"])))
+
+    (t/testing "DML enabled with BEGIN READ WRITE"
+      (jdbc/execute! tu/*conn* ["START TRANSACTION"])
+      (jdbc/execute! tu/*conn* ["INSERT INTO foo (a) VALUES (42)"])
+      (t/is (thrown-with-msg? Exception #"" (jdbc/execute! tu/*conn* ["COMMIT"]))))
+
+    (t/is (= 0.0 (.count ^Counter (.counter (.find registry "query.error")))))
+    (t/is (= 0.0 (.count ^Counter (.counter (.find registry "query.warning")))))
+    (t/is (= 2.0 (.count ^Counter (.counter (.find registry "tx.error")))))))
```
FiV0: What happens here if you add something that is an async error in the indexer in the above test going through jdbc. Will it get counted twice? I think this will show you if the whole approach is sound.

FiV0 (changes requested):
Once again I don't like it as we are adding stuff all over the place, but I think there is no way around it. I am wondering if we should just scrape the part in `pgwire`. Not the query-errer-counting part of course.

src/test/clojure/xtdb/metrics_test.clj
tggreene: I'm not honestly sure how to test that. Is there something we can pass in that parses correctly but causes an error in the indexer?

FiV0: Try something like `INSERT INTO docs SELECT 1/0 AS _id`

tggreene: Yeah this does double count, I'm not really sure how to prevent it

tggreene: I've added another condition (not ideal I know) for a :dml-result error type so that we can do nothing in that case where we know the indexer will have already counted the error.

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -300,8 +300,10 @@
 
 (defn- client-err
   ([client-msg] (client-err client-msg nil))
-  ([client-msg data]
-   (ex-info client-msg (assoc data ::client-error (err-protocol-violation client-msg)))))
+  ([client-msg {:keys [error-type] :as data}]
+   (ex-info client-msg (assoc data ::client-error
+                              (-> (err-protocol-violation client-msg)
```
FiV0, line 305: Can you make this a `cond->` conditional on `error-type`.

FiV0: Con you make this a `cond->` conditional on `error-type`.

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1061,8 +1063,9 @@
   ;; mark a transaction (if open as failed), for now we will consider all errors to do this
   (swap! conn-state util/maybe-update :transaction assoc :failed true, :err err)
 
-  (when query-error-counter
-    (.increment ^Counter query-error-counter))
+  (when (and (not (#{:dml :transaction} error-type))
```
FiV0: As there is no differentiation between `transaction` and `dml` in logic, can we make it `:dml` througout.

src/test/clojure/xtdb/metrics_test.clj
```diff
@@ -2,23 +2,53 @@
...
+    (t/is (thrown? Exception (xt/execute-tx tu/*node* ["INSERT INTO foo (a) VALUES (42)"])))
+    (t/is (thrown? Exception (jdbc/execute! tu/*conn* ["INSERT INTO docs SELECT 1/0 AS _id"])))
+
+    (t/is (= 0.0 (.count ^Counter (.counter (.find registry "query.error")))))
+    (t/is (= 0.0 (.count ^Counter (.counter (.find registry "query.warning")))))
+    (t/is (= 5.0 (.count ^Counter (.counter (.find registry "tx.error")))))))
+
+(t/deftest test-transaction-exception-counter-on-submit-tx
+  (let [registry ^CompositeMeterRegistry (tu/component tu/*node* :xtdb.metrics/registry)]
+    (t/is (thrown? Exception (jdbc/execute! tu/*conn* ["INSERT INTO foo (a) VALUES (42)"])))
```
FiV0: `submit-tx` is never used via `pgwire` so this can go.

FiV0 (changes requested):

FiV0:
@tggreene Could you merge this in https://github.com/FiV0/xtdb/commit/d11dd637700e584e594a643e47710487f70025d9. I annotated the assertions with some comments.

src/test/clojure/xtdb/metrics_test.clj
tggreene: I believe it is via execute-tx, no?

FiV0: Yes agree, but we have the same assertion in the test above. From the test description this is explicitly about `submit-tx` so I didn't see the value.

tggreene: Okay, that's fair

FiV0 (approved):
LGTM. Have ofcourse not checked the dashboard.

PR #4164, opened 2025-02-11
Add `XTDB_SKIP_TXS` and implement/test transaction skipping.

### LINKED ISSUE #4154: Add 'skip-txs' so that a user has a chance to recover from ingestion-stopped
Context: we stop ingestion because we **cannot** run the risk of one node committing a transaction and another rolling the same transaction back - while ingestion-stopped might be bad, this would be worse.

That said, we'd like to give users a chance to recover from ingestion-stopped - we propose adding a skip-txs configuration so that users can elect to skip the errant tx. This likely won't be perfect - whatever's caused the transaction to halt ingestion in the first place may well halt ingestion on the next tx instead - but it's a tool in the toolbox.

* Users _must_ apply this change to all nodes atomically - i.e. take all the nodes down, add the config, start all the nodes back up.
* I'd probably do this as an env var (e.g. `XTDB_SKIP_TXS=412,541`), then impl in LogProcessor - but final decision left to the implementor.
* log warn when you skip the transactions
* log info when you get to the end of the block, informing the user that they are safe to remove the skip-txs env var.

danmason:
Resolves #4154 

- Could do with a check over the "block finished" behavior. 
  - This log may/will be returned multiple times - I decided that was likely fine, XTDB_SKIP_TXS shouldn't really be kept around much longer than it has to.  
- Will require documenting along with Caveats.

### DIFF
--- a/core/src/main/clojure/xtdb/indexer/live_index.clj
+++ b/core/src/main/clojure/xtdb/indexer/live_index.clj
@@ -17,7 +17,7 @@
   (:import [clojure.lang MapEntry]
            (java.lang AutoCloseable)
            (java.time Duration)
-           (java.util ArrayList HashMap Map)
+           (java.util ArrayList HashMap List Map)
            (java.util.concurrent StructuredTaskScope$ShutdownOnFailure StructuredTaskScope$Subtask)
            (java.util.concurrent.locks StampedLock)
            (java.util.function Function)
@@ -218,7 +218,8 @@
 
                     ^RowCounter row-counter, ^long rows-per-block
 
-                    ^long log-limit, ^long page-limit]
+                    ^long log-limit, ^long page-limit 
+                    ^List skip-txs]
   xtdb.indexer.LiveIndex
   (getLatestCompletedTx [_] latest-completed-tx)
   (getLatestCompletedBlockTx [_] latest-completed-block-tx)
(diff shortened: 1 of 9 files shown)


core/src/main/clojure/xtdb/indexer/live_index.clj
```diff
@@ -218,7 +218,8 @@
 
                     ^RowCounter row-counter, ^long rows-per-block
 
-                    ^long log-limit, ^long page-limit]
+                    ^long log-limit, ^long page-limit 
+                    ^List txs-to-skip]
```
jarohen: nit: skip-txs for consistency

core/src/main/kotlin/xtdb/api/IndexerConfig.kt
```diff
@@ -13,9 +14,17 @@ data class IndexerConfig(
     var pageLimit: Long = 1024L,
     var rowsPerBlock: Long = 102400L,
     var flushDuration: Duration = Duration.ofHours(4),
+    var skipTxs: List<LogOffset> = System.getenv("XTDB_SKIP_TXS")?.let(::parseSkipTxsEnv).orEmpty()
```
danmason, line 17: Is there a good place to test this? @jarohen

jarohen (approved):
all good Dan, cheers mate :pray: 

> This log may/will be returned multiple times - I decided that was likely fine, `XTDB_SKIP_TXS` shouldn't really be kept around much longer than it has to.

It shouldn't, but it probably will be - previously I've simply wrapped this in a `defonce`

```clojure
(when condition
  ;; usually needs a kondo ignore
  (defonce log-msg
    (log/... )))
```

> Will require documenting along with Caveats.

can you stick this in the existing section in this PR and I'll move it as appropriate?

core/src/main/kotlin/xtdb/api/IndexerConfig.kt
jarohen: I probably wouldn't worry tbh, but if you want you can make the private fun `internal` and run a few cases through it

PR #4195, opened 2025-02-21
Add table block files

### LINKED ISSUE #4171: Split block file, creating table-level metadata per block
Current state:

* At the end of every block, we create an L0 data/meta file for each table, and then a single block-metadata transit/json file. 
* That block-level metadata contains the latest-completed-tx and, for each table, a list of columns and their types.

Future state:

* L0 files uploaded in the same way
* New table-block file uploaded (e.g. `/tables/my-table-name/blocks/b04.binpb`) containing the cumulative types of the columns
  * could represent the column types using Arrow Schema serialized to bytes in its usual way (rather than our current col-type representation)
* Block file becomes significantly smaller - only needs to contain the latest-completed-tx and a list of tables. (e.g. `/blocks/b04.binpb`)

Out of scope for this one, but eventually this table-block file would contain a list of trie files for each table with their trie-level metadata

FiV0:
As per #4171 this splits the toplevel block files into a toplevel block file and table level block files. Additionally the metadata in the block files becomes a snaphshot up to that point, so when the metadata arrives for the new block we merge it with the existing block file data. The metadata manager then also only needs to know about the latest block metadata.

There are still some decisions to be made:
- [x] ~protobuf vs. transit:~ pbuf
- [x] ~protobuf utitilies similar to json?~ yep


core/src/main/clojure/xtdb/indexer/live_index.clj
```diff
@@ -149,6 +149,7 @@
             (let [data-file-size (trie/write-live-trie! allocator buffer-pool table-name trie-key
                                                         live-trie data-rel)]
               (MapEntry/create table-name
+                               ;; TODO remove trie-key/data-file-size ?
                                {:fields (live-rel->fields live-rel)
                                 :trie-key trie-key
                                 :data-file-size data-file-size
```
FiV0: Not sure if `trie-key` and `data-file-size` should be removed?

core/src/main/proto/xtdb/meta/block/proto/block.proto
```diff
@@ -0,0 +1,15 @@
...
+
+message TableBlock {
+  bytes arrow_schema = 1;
+  int64 row_count = 2;
+}
+
+message Block {
+  int64 latest_completed_tx_id = 1;
+  repeated string table_blocks = 2;
+}
```
FiV0, line 20: This could go if we decide to stay with transit at the toplevel.

src/test/clojure/xtdb/test_json.clj
```diff
@@ -102,7 +104,10 @@
        (check-arrow-json-file expected actual)
 
        (.endsWith file-name ".transit.json")
-       (check-transit-json-file expected actual)))))
+       (check-transit-json-file expected actual)
+
+       (.endsWith file-name ".binpb")
+       (= (seq (Files/readAllBytes expected)) (seq (Files/readAllBytes actual)))))))
```
FiV0: These changes have nothing todo with `json`. We could add some utilities for `binpb` if we would like to look at the `binpb`. I think we would need to know what kind of protobuf we are talking about.

core/src/main/clojure/xtdb/indexer/live_index.clj
jarohen: yes please!

FiV0: So you don't want them in the block level metadata? They are currently still being used to create messages for the trie catalog.

src/test/clojure/xtdb/test_json.clj
jarohen: binpbs aren't necessarily ordered, they're a map (same as Transit) so need to be parsed to check for equality. The objects you get back after parsing are comparable, so I think it'd be a case of passing the parsing fn through (which you can now do with Clojure 1.12)

reckon it's best we split this out into `check-pbuf`?

core/src/main/clojure/xtdb/indexer/live_index.clj
jarohen: > So you don't want them in the block level metadata? 

nope :+1: 

> They are currently still being used to create messages for the trie catalog.

so sounds like we might still need to return them here and then elide them later?

core/build.gradle.kts
```diff
@@ -76,6 +76,7 @@ dependencies {
     api(libs.kaml)
 
     api(libs.protobuf.kotlin)
+    api(libs.protobuf.java)
```
jarohen: do we need the Java one too? I _thought_ we didn't... :thinking: will have a look at the rest of the PR :slightly_smiling_face:

core/src/main/clojure/xtdb/indexer/live_index.clj
```diff
@@ -384,7 +384,8 @@
    :config config})
 
 (defmethod ig/init-key :xtdb.indexer/live-index [_ {:keys [allocator buffer-pool metadata-mgr log trie-catalog compactor ^IndexerConfig config metrics-registry]}]
-  (let [{:keys [latest-completed-tx next-block-idx], :or {next-block-idx 0}} (meta/latest-block-metadata metadata-mgr)]
+  (let [{:keys [latest-completed-tx block-idx]} (meta/latest-block-metadata metadata-mgr)
+        next-block-idx (or (some-> block-idx inc) 0)]
```
jarohen: let's see if we can remove this, or Just™ pass the `block-idx`

core/src/main/clojure/xtdb/metadata.clj
```diff
@@ -76,22 +80,72 @@
...
 
-(defn- write-block-metadata ^java.nio.ByteBuffer [block-meta]
+(def ^Path block-table-metadata-path (util/->path "blocks"))
+
+(defn- write-block-metadata ^java.nio.ByteBuffer [latest-completed-tx table-block-paths]
   (with-open [os (ByteArrayOutputStream.)]
     (let [w (transit/writer os :json {:handlers metadata-write-handler-map})]
-      (transit/write w block-meta))
+      (transit/write w {:latest-completed-tx latest-completed-tx
+                        :table-block-paths table-block-paths}))
```
jarohen: worth using table-names here and deriving the table-block paths?

core/src/main/clojure/xtdb/metadata.clj
```diff
@@ -76,22 +80,72 @@
...
+                         (.setArrowSchema (ByteString/copyFrom (.serializeAsMessage table-schema)))
+                         (.setRowCount row-count))
+                       (.build)
+                       (.toByteArray))))
+
+(defn- <-table-block [^TableBlock table-block]
+  (let [^Schema schema
+        (Schema/deserializeMessage (ByteBuffer/wrap (.toByteArray (.getArrowSchema table-block))))]
+    {:row-count (.getRowCount table-block)
+     :fields (->> (for [^Field field (.getFields schema)]
```
jarohen, line 140: you could arguably keep the `:schema` here rather than transforming it...? depends how it's used

core/src/main/clojure/xtdb/metadata.clj
```diff
@@ -375,65 +429,86 @@
...
-  (let [blocks-metadata (load-blocks-metadata deps)
+  (let [last-block-metadata (load-latest-block-metadata deps)
         table-metadata-cache (-> (Caffeine/newBuilder)
                                  (.maximumSize cache-size)
                                  (.build))]
     (MetadataManager. buffer-pool
                       table-metadata-cache
-                      blocks-metadata
-                      (->> (vals blocks-metadata) (reduce merge-fields {})))))
+                      ;; TODO is nil here an issue?
```
jarohen: shouldn't be so long as everything that needs to be defaulted is (i.e. block-idx, empty set of tables)

core/src/main/proto/xtdb/meta/block/proto/block.proto
jarohen: on balance I'm still in favour of protobuf here, given the direction of travel for low-level IO is towards Kotlin - transit in Kotlin is relatively painful; pbuf in Clojure is acceptable

src/test/clojure/xtdb/test_json.clj
```diff
@@ -84,13 +84,15 @@
            :let [file-name (str (.getFileName path))
                  file-type (cond
                              (str/ends-with? file-name ".arrow") :arrow
-                             (str/ends-with? file-name ".transit.json") :transit)]
+                             (str/ends-with? file-name ".transit.json") :transit
+                             ;; TODO this should like not go in test-json
```
jarohen: agreed, yep :)

to be honest, Transit probably shouldn't have done either, in hindsight

core/src/main/clojure/xtdb/indexer/live_index.clj
FiV0: Yes, if all of it changes anyway in the next steps, lets just remove them in a later step.

jarohen (changes requested):
nice one @FiV0 - minor changes to make then we're good to go, happy to pair

> There are still some decisions to be made:
> 
> - protobuf vs. transit

pbuf

> - protobuf utitilies similar to json?

yep :ok_hand:

core/src/main/clojure/xtdb/metadata.clj
FiV0: I think I initially thought of putting `table-block-paths` here, so that if a table doesn't get updated in a block we can use the old table block file and not need to upload a new table block file that only differs in name.

core/src/main/clojure/xtdb/indexer/live_index.clj
FiV0: RowCounter takes a `long` on initializaion

core/src/main/clojure/xtdb/metadata.clj
FiV0: The interface of the metadata manager has `(columnField [_ table-name col-name]`. I think I would then need to destructure the schema for every call to this method.

core/src/main/clojure/xtdb/metadata.clj
FiV0: Moved to `table-name`, idea from above is still relevant though.

src/test/clojure/xtdb/indexer_test.clj
```diff
@@ -100,10 +101,13 @@
...
                    (-> (meta/latest-block-metadata mm)
-                       (dissoc :tables :next-block-idx))))
+                       (dissoc :tables :block-idx))))
 
           (tj/check-json (.toPath (io/as-file (io/resource "xtdb/indexer-test/can-build-block-as-arrow-ipc-file-format")))
-                         (.resolve node-dir "objects")))))))
+                         (.resolve node-dir "objects"))
+
+          (cpb/check-pbuf  (.toPath (io/as-file (io/resource "xtdb/indexer-test/can-build-block-as-arrow-ipc-file-format")))
+                           (.resolve node-dir "objects")))))))
```
FiV0, line 110: I wasn't sure where to add the `check-pbuf` checks and where not. Most tests using `check-json` seem to be testing something different.

core/src/main/clojure/xtdb/indexer/live_index.clj
jarohen: does it still need to?

core/src/main/clojure/xtdb/metadata.clj
```diff
@@ -376,68 +448,87 @@
...
-                          ^:volatile-mutable ^Map fields]
+                          ^:volatile-mutable ^Map last-block-metadata
+                          ^:volatile-mutable ^Map table->fields]
   IMetadataManager
+  ;; the new-block-metadata is only the delta for the new block
   (finishBlock [this block-idx new-block-metadata]
-    (.putObject buffer-pool (->block-metadata-obj-key block-idx) (write-block-metadata new-block-metadata))
-    (set! (.fields this) (merge-fields fields new-block-metadata))
-    (.put blocks-metadata block-idx new-block-metadata))
+    (when (or (nil? (:block-idx last-block-metadata)) (< ^long (:block-idx last-block-metadata) block-idx))
```
jarohen, line 456: not reviewing this in any great detail as I assume it's moving about in the next PR :slightly_smiling_face:

src/test/clojure/xtdb/check_pbuf.clj
```diff
@@ -0,0 +1,39 @@
...
+           (xtdb.block.proto Block TableBlock)))
+
+(defn multi-parse-fn
+  "Tries to parse a byte array from a given list of parse-fns. Throws if non parses."
+  [parse-fns]
+  (fn [& args]
+    (or (some #(try (apply % args) (catch Exception _e nil)) parse-fns)
+        (throw (RuntimeException. "No matching parse-fn!")))))
+
+(defn check-pbuf
```
jarohen, line 14: we can probably make this a bit simpler than the JSON equivalent - we're looking at bringing a flat directory full of binpbs into a Clojure map, and comparing them against a known dir

src/test/clojure/xtdb/indexer_test.clj
jarohen: for indexer test: I'd add them in most places here tbh - this is the main place that we really test what block-level files get written to disk.

jarohen (approved):
cheers @FiV0, good to merge without further review (minor comments below).

lmk if the rebase on main is a PITA :)

core/src/main/clojure/xtdb/indexer/live_index.clj
FiV0: Make `RowCounter` take an optional? I am not fully following.

src/test/clojure/xtdb/check_pbuf.clj
FiV0: I am not following. We have two directories with `binpb`s in them. How should I compare them?

jarohen: sorry, not explaining well over text :sweat_smile: 

check them directory-at-a-time, have the util function only know how to check one directory of files against another, rather than the util being able to work recursively. that way, you can specify the exact function rather than needing to do the 'try-parse' bit - the try-parse _might_ work when it shouldn't, with pbuf

core/src/main/clojure/xtdb/indexer/live_index.clj
jarohen: again, not explaining well over text :sweat_smile: 

RowCounter always used to have to know about the chunk-idx because the chunk-idx depended on the number of rows within the chunk. now we've broken that dependency, there's less of a reason for RowCounter to get involved in what the next block idx will be, it might be able to just count rows, with the block-idx being maintained as a simple mutable var in the live-idx or similar

PR #4222, opened 2025-02-27
Move table block metadata to new table-catalog component

### LINKED ISSUE #4218: Separate trie level metadata from block level metadata
In the context of https://github.com/xtdb/xtdb/issues/4171 and https://github.com/xtdb/xtdb/issues/4173 we will soon have block level metadata. The block level metadata is relevant for the trie-catalog (mainly initialization). To avoid cyclic dependencies and also with a linux philosophy-like mindset (one component does one thing well), lets split out the block level metadata from the metadata manager.

Current status:
- Page level metadata is written on a per table basis on `finishBlock` of the live-index.
- Page level metadata is read via the metadata manager.
- We currently only have a toplevel block file (this will already change in https://github.com/xtdb/xtdb/pull/4195) which gets written on `finishBlock` of the metadata manager.
- All these block files get reread when the node starts back up.

Option 1:
- Create a BlockMetadataManager. All methods apart from `openTableMetadata` move to this new component.
- It deals with reading / writing block level metadata. It depends on the trie catalog to get a list of currently relevant tries.

Option2: 
- Merge the block metadata logic with the trie catalog. This might be easier to avoid cyclic dependecies (for example trie catalog currently uses the metadata manager to get a list of tables).
- The relevant methods (and their logic) move to the trie-catalog.

FiV0:
On top of #4195. This contains no functional change, but moves table block metadata creation from metadata manager to the trie catalog.

### DIFF
--- a/core/src/main/clojure/xtdb/indexer.clj
+++ b/core/src/main/clojure/xtdb/indexer.clj
@@ -808,6 +808,7 @@
                     ^IMetadataManager metadata-mgr
                     ^IQuerySource q-src
                     ^LiveIndex live-idx
+                    table-catalog
                     ^Timer tx-timer
                     ^Counter tx-error-counter]
   IIndexer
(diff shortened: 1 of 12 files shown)


FiV0:
We now have a hierarchy of `block` -> `table` -> `trie` -> `page` catalog (where page catalog is currently still named Metadata Manager).

PR #4227, opened 2025-03-03
Add trie list to table block file

### LINKED ISSUE #4173: Store trie list in the per-table block files
(see #4171)

1. On finish-block, we should store all of the table-tries from the trie catalog and include them in the table-level block file.
2. On node startup, rather than listing all the objects in the object store to re-populate the trie catalog, we should:
   1. read the block file to get the list of tables
   2. read the block table files to reconstruct the trie-catalog for each table.

N.B. there may be some slight discrepancies between the content of the that the different nodes store here: 
* in the live-index/compactor, we eagerly send the notification to the current node so that it doesn't have to wait for the notification to make it all the way back around.
  * i.e. in the node that the notification is raised on, the trie might get included before the block boundary; on all the others, it might only arrive afterwards.
* all that should happen here is that any node restoring from this file will see a duplicate notification - trie notifications are (deliberately) idempotent.
* tl;dr we accept some eventual consistency here.

FiV0:
This is on top of #4222. All active trie files get added to the table block file and the trie catalog reconstructed from these table block files on startup. The most important (!!!) bit of this PR is that files are no longer guaranteed to be content addressable (for the lack of a better word), i.e. unique name does not guarantee unique content. As the active trie files depend on the state of the node when the block finishes. Fhe active tries can therefore differ across nodes writing the latest table block file. 

There are a couple of quirks in the code that I am going to comment on.


core/src/main/proto/xtdb/meta/block/proto/block.proto
```diff
@@ -2,11 +2,14 @@ edition = "2023";
...
 package xtdb.block.proto;
 
+import "xtdb/log/proto/log.proto";
+
 option java_multiple_files = true;
 
 message TableBlock {
   bytes arrow_schema = 1;
   int64 row_count = 2;
+  repeated xtdb.log.proto.AddedTrie current_tries = 3;
```
FiV0: Not sure we want to reuse `AddedTrie` here, but it was close enough for my purposes so that I reused it.

src/test/clojure/xtdb/trie_catalog_test.clj
```diff
@@ -101,3 +112,55 @@
...
+          (tu/finish-block! node)
+
+          (t/is (= ["l02-rc-p0-b01" "l02-rc-p1-b01" "l02-rc-p2-b01" "l02-rc-p3-b01"
+                    "l01-rc-b02"
+                    "l00-rc-b03"]
+                   (->> (.getByteArray bp (util/->path "tables/public$foo/blocks/b02.binpb"))
+                        TableBlock/parseFrom
+                        trie/<-table-block
+                        :current-tries
+                        (mapv (comp :trie-key added-trie->edn))))))))))
```
FiV0: This last part of the test is kind of weird as I am artificially adding trie notifications.

core/src/main/clojure/xtdb/table_catalog_test.clj
```diff
@@ -0,0 +1,64 @@
...
+                      table-cat/<-table-block
+                      :current-tries
+                      (mapv added-trie->edn ))))
+
+        (t/testing "artifically adding tries (simulating another node finishing and compacting these)"
+          (->> [["l00-rc-b00" 1] ["l00-rc-b01" 1] ["l00-rc-b02" 1] ["l00-rc-b03" 1]
+                ["l01-rc-b00" 2] ["l01-rc-b01" 2] ["l01-rc-b02" 2]
+                ["l02-rc-p0-b01" 4] ["l02-rc-p1-b01" 4] ["l02-rc-p2-b01" 4] ["l02-rc-p3-b01"4]]
+               (map #(apply (partial trie/->added-trie "public/foo") %))
+               (.addTries trie-catalog))
```
FiV0: This last part of the test is kind of weird as I am artificially adding trie notifications.

core/src/main/proto/xtdb/meta/block/proto/block.proto
jarohen: yeah, seems reasonable. maybe a rename then? `TrieDetails`? :thinking:

core/src/main/clojure/xtdb/table_catalog_test.clj
jarohen: yeah, I think that's fine - I do the same in the compactor-test to test what jobs it would schedule in a given state

:golf: : with `apply` you don't need the `partial` - `#(apply trie/->added-trie "public/foo" %)` would also work

also Clojure style - I don't think I'd fully thread this here, because the final clause is a side-effect, you're not threading/transforming a value. 

```clojure
(.addTries trie-catalog
           (->> [["l00-rc-b00" 1] ["l00-rc-b01" 1] ["l00-rc-b02" 1] ["l00-rc-b03" 1]
                 ["l01-rc-b00" 2] ["l01-rc-b01" 2] ["l01-rc-b02" 2]
                 ["l02-rc-p0-b01" 4] ["l02-rc-p1-b01" 4] ["l02-rc-p2-b01" 4] ["l02-rc-p3-b01"4]]
                (map #(apply trie/->added-trie "public/foo" %))))
```

core/src/main/clojure/xtdb/indexer/live_index.clj
```diff
@@ -319,14 +319,21 @@
...
               (.addTries trie-cat added-tries)
-              @(.appendMessage log (Log$Message$TriesAdded. added-tries))))))
+              @(.appendMessage log (Log$Message$TriesAdded. added-tries)))
+
+            (let [all-tables (set (concat (keys table-metadata) (.getAllTableNames block-cat)))
+                  table->current-tries (->> all-tables
+                                            (map (fn [table-name]
+                                                   (MapEntry/create table-name
+                                                                    (->> (trie-cat/trie-state trie-cat table-name)
+                                                                         trie-cat/current-tries))))
```
jarohen: I suspect we want all the tries stored in here, live and garbage - idea being to reflect what's in the object store. Maybe not an issue now, but when we add GC to the LSM we'll want to know all the tries, not just the live ones.

core/src/main/proto/xtdb/meta/block/proto/block.proto
jarohen: although see other comment re live + garbage tries - might be that we need a different type

jarohen (changes requested):
thanks @FiV0 - all the structure's there but just the below comments to address on the data we're storing. cheers mate :pray:

core/src/main/proto/xtdb/meta/block/proto/block.proto
FiV0: Table Name is kind of redundant in that message.

PR #4240, opened 2025-03-07
Add trie level metadata to table blocks and trie catalog

### LINKED ISSUE #4174: Include trie-level metadata in the trie-list
(following on from #4173)

Goal is that we don't then need to read individual trie-level metadata files just to determine whether they're temporally valid, or contain a given IID.

* Min/max valid-from, valid-to, system-from, recency
* IID bloom filter (consider whether we want a larger bloom filter here to ~preserve the false positive rate)
* Row count (count of events)
* No content-col information here.

FiV0:
This addresses #4174. The new metadata is not (yet) used anywhere for pruning.

I am also moving a bunch of bloom logic to Kotlin.

Todo:
- [ ] figure out correct iid bloom size for larger files
- [x] ~recency? wasn't sure if recency still plays a role in the new plans.~
      - #4257 
- [x] trie metadata test for compactor?
- [x] move kotlin work directly to `main`


core/src/main/kotlin/xtdb/bloom/BloomBuilder.kt
```diff
@@ -0,0 +1,41 @@
+package xtdb.bloom
+
+import com.carrotsearch.hppc.IntArrayList
+import org.roaringbitmap.RoaringBitmap
+import xtdb.arrow.VectorReader
+import xtdb.util.Hasher
+
+interface BloomBuilder {
```
jarohen: I'd make this a class - not the sort of thing we're likely to have multiple implementations of, nor mock

core/src/main/kotlin/xtdb/bloom/BloomBuilder.kt
```diff
@@ -0,0 +1,41 @@
...
+            return create(col, BloomUtils.BLOOM_K, BloomUtils.BLOOM_BIT_MASK)
+        }
+
+        @JvmStatic
+        fun create(col: VectorReader, k: Long, mask: Long): BloomBuilder {
+            val buf = IntArrayList((col.valueCount * k).toInt())
+            val xxHasher = Hasher.Xx()
+            val murmur3Hasher = Hasher.Murmur3()
+
+            return object : BloomBuilder {
```
jarohen: once this is a class, most of this method can be move into the class body

core/src/main/kotlin/xtdb/bloom/BloomUtils.kt
```diff
@@ -0,0 +1,84 @@
...
+import xtdb.util.Hasher
+import java.nio.ByteBuffer
+
+object BloomUtils {
+    const val BLOOM_BITS = 1L shl 20
+    const val BLOOM_BIT_MASK = BLOOM_BITS - 1
+    const val BLOOM_K = 3L
+
+    @JvmStatic
+    fun bloomToBitmap(bloomRdr: VectorReader, idx: Int): ImmutableRoaringBitmap {
```
jarohen: can be more terse: given you're in an object you can probably drop the `bloom` prefix on these methods, because you get the namespacing from the class name: `BloomUtils/toBitmap`, `BloomUtils/contains` is still clear

core/src/main/kotlin/xtdb/bloom/BloomUtils.kt
```diff
@@ -0,0 +1,84 @@
...
+        return ImmutableRoaringBitmap(nioBuffer)
+    }
+
+    @JvmStatic
+    fun bloomContains(bloomRdr: VectorReader, idx: Int, hashes: IntArray) =
+        bloomContains(bloomToBitmap(bloomRdr, idx), hashes)
+
+    @JvmStatic
+    fun bloomContains(bloom: ImmutableRoaringBitmap, hashes: IntArray): Boolean {
+        for (i in hashes.indices) {
```
jarohen: `hashes.all { }`

core/src/main/kotlin/xtdb/bloom/BloomUtils.kt
```diff
@@ -0,0 +1,84 @@
...
+            if (!bloom.contains(hashes[i])) {
+                return false
+            }
+        }
+        return true
+    }
+
+    @JvmStatic
+    fun bloomHashes(col: VectorReader, idx: Int): IntArray {
+        return bloomHashes(col, idx, BLOOM_K, BLOOM_BIT_MASK)
```
jarohen: expression body

core/src/main/kotlin/xtdb/bloom/BloomUtils.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    // Cassandra-style hashes:
+    // https://www.researchgate.net/publication/220770131_Less_Hashing_Same_Performance_Building_a_Better_Bloom_Filter
+
+    @JvmStatic
+    fun bloomHashes(col: VectorReader, idx: Int, k: Long, mask: Long): IntArray {
+        val hash1 = col.hashCode(idx, Hasher.Xx())
+        val hash2 = col.hashCode(idx, Hasher.Murmur3())
+        val acc = IntArray(k.toInt())
```
jarohen: IntArray has an overload that takes an optional closure which initialises the array - something like this (written blind, YMMV):

```kotlin
return IntArray(k.toInt()) { n -> (mask and (hash1 + hash2 * n)).toInt() }
```

core/src/main/kotlin/xtdb/bloom/BloomUtils.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    @JvmStatic
+    fun ByteBufferToRoaringBloom(buf: ByteBuffer): RoaringBitmap {
+        val bloom = RoaringBitmap()
+        bloom.deserialize(buf)
+        return bloom
+    }
+
+    fun roaringBloomToByteBuffer(bloom: RoaringBitmap): ByteBuffer {
+        val buf = ByteBuffer.allocate(bloom.serializedSizeInBytes())
```
jarohen: `also` is your friend here:

```kotlin
return ByteBuffer.allocate(...).also {
  bloom.serialize(it)
  it.clear()
}
```

(then it's an expression body too)

core/src/main/kotlin/xtdb/bloom/BloomUtils.kt
```diff
@@ -0,0 +1,84 @@
...
+        val hash2 = col.hashCode(idx, Hasher.Murmur3())
+        val acc = IntArray(k.toInt())
+        for (n in 0 until k) {
+            acc[n.toInt()] = (mask and (hash1 + hash2 * n)).toInt()
+        }
+        return acc
+    }
+
+    @JvmStatic
+    fun ByteBufferToRoaringBloom(buf: ByteBuffer): RoaringBitmap {
```
jarohen: I'd probably make this an extension method - `fun ByteBuffer.toRoaringBitmap()` - so that users then write `buf.toRoaringBitmap`.

Essentially, when your callers would likely have used a `->` in Clojure, because you're transforming one value to another, I tend to use extension methods.

(if you need this from Clojure as well, I'm reasonably sure it stays as `(BloomUtils/toRoaringBloom buf)`, but check `javap` if not)

core/src/main/kotlin/xtdb/bloom/BloomUtils.kt
```diff
@@ -0,0 +1,84 @@
...
+    fun writeBloom(bloomWtr: VectorWriter, col: VectorReader) {
+        val bloomBuilder = BloomBuilder.create(col)
+        for (idx in 0 until col.valueCount) {
+            if (!col.isNull(idx)) {
+                bloomBuilder.add(idx)
+            }
+        }
+
+        val bloom = bloomBuilder.build()
+        bloomWtr.writeBytes(roaringBloomToByteBuffer(bloom))
```
jarohen: `roaringBloomToByteBuffer(bloom)` -> `bloom.toByteBuffer()`

core/src/main/kotlin/xtdb/indexer/LiveTable.kt
```diff
@@ -26,6 +28,16 @@ class LiveTable
 ) : AutoCloseable {
 
     private val trieWriter = TrieWriter(al, bp, false)
+    private val trieMetadataBuilder = TrieMetadata.newBuilder()
+        .setMinValidFrom(MAX_LONG)
```
jarohen: may be able to default these in the protobuf instead...? :thinking:

core/src/main/kotlin/xtdb/indexer/LiveTable.kt
```diff
@@ -68,10 +85,10 @@ class LiveTable
         fun openWatermark(): Watermark = openWatermark(transientTrie)
         val docWriter: IVectorWriter = putWtr
         val liveRelation: IRelationWriter = this@LiveTable.liveRelation
+        val startPos = liveRelation.writerPosition().position
+        private var pos = startPos
```
jarohen: I'd be a little cautious here about representing the same mutable state in two different places - it adds the assumption that the two will remain in sync

(`val startPos` is fine, it's `var pos` that alerted my radar)

core/src/main/kotlin/xtdb/trie/TrieWriter.kt
```diff
@@ -64,12 +81,28 @@ class TrieWriter(
...
+            val validFrom = dataRel["_valid_from"]!!
+            val validTo = dataRel["_valid_to"]!!
+            val iidVec = dataRel["_iid"]!!
+
+            val temporalCols = listOf(systemFrom, validFrom, validTo, iidVec)
+
+            if(writeTrieMetadata) {
+                val (minValidFrom, maxValidFrom) = validFrom.fold(trieMetadataBuilder.minValidFrom to trieMetadataBuilder.maxValidFrom) { (min, max) , v : Long -> minOf(min, v) to maxOf(max, v) }
+                trieMetadataBuilder.minValidFrom = minValidFrom
+                trieMetadataBuilder.maxValidFrom = maxValidFrom
```
jarohen, line 98: give this some newlines? :mag: :slightly_smiling_face:

core/src/main/clojure/xtdb/table_catalog_test.clj
```diff
@@ -29,21 +41,43 @@
...
-                   :data-file-size 1966}
-                  {:table-name "public/foo",
-                   :trie-key "l00-rc-b01",
-                   :data-file-size 1966}]
-                 (->> (.getByteArray bp (util/->path "tables/public$foo/blocks/b01.binpb"))
-                      TableBlock/parseFrom
-                      table-cat/<-table-block
-                      :current-tries
-                      (mapv added-trie->edn ))))
+        (let [current-tries (->> (.getByteArray bp (util/->path "tables/public$foo/blocks/b01.binpb"))
```
jarohen, line 48: (out of scope of this PR) these are the tests that I'd like us to move to Kotlin too, given the code it's testing is. I understand we don't yet have the same level of fixtures there that we do in Clojure (which is why I have also done the same as you here), so when we next come up for air let's figure out what we need to make this easier

jarohen (changes requested):
Good PR, @FiV0, thanks :pray: 

- next time, can you land the bloom change separately - if need be, stash/commit what you have
- a few Kotlin idiom suggestions in here - happy to pair up and walk through if that'd help

core/src/main/kotlin/xtdb/indexer/LiveTable.kt
FiV0: I put it in the proto for now. The important bit to keep in mind is that if you build a TrieMetadata without calculating anything on it, the contained data doesn't make much sense.

PR #4269, opened 2025-03-13
Add HyperLogLog to TableBlock metadata 

### LINKED ISSUE #4175: Add HLLs to the table-block files
This increment will add HyperLogLog statistics for each table to the per-table block files, so that we can eventually use these stats to better inform our query planning.

(This card only involves saving the stats, given that we have a index-format change going through atm - incrementing the query planning is out of scope.)

On each block, we should store the cumulative HLLs for each column - i.e. add the previous HLL to the stats from the new L0 data file to get the new HLL.

FiV0:
HyperLogLog implementation (mostly copied straight from v1) and the data added to the ~TrieMetadata~ TableBlocks.

~I have also added it the live trie (and consequently also to L0), but given the size of it and likely the speed implications it might be better to only add it at L1 level. We can also not reuse the HLL from L0 in L1 as some rows stay in the `current` dichotomy and others go to the `historical`.~

We are keeping a keeping cumulative HLL's per column in the table block file. When finish chunk gets called on the live table, we pass the HLL deltas for the new block to the new TableBlock creation, which merges the existing HLL with the new delta.


core/src/main/kotlin/xtdb/util/HyperLogLog.kt
```diff
@@ -0,0 +1,136 @@
...
+     */
+    fun estimateDifference(hllA: ByteBuf, hllB: ByteBuf): Double {
+        return estimate(hllA) - estimateIntersection(hllA, hllB)
+    }
+
+    @JvmStatic
+    fun createHLL(bufferSize: Int = DEFAULT_BUFFER_SIZE) = Unpooled.buffer(bufferSize)
+    @JvmStatic
+    fun toHLL(bytes: ByteArray) = Unpooled.wrappedBuffer(bytes)
+}
```
FiV0: @jarohen  I am starting to like this style the most in Kotlin. I am guessing because it's the closest to Clojure. Function where the first argument is the main data structure worked on. Happy to also go with extension functions or other idioms if you prefer.

core/src/main/proto/xtdb/log/proto/log.proto
```diff
@@ -28,6 +28,8 @@ message TrieMetadata {
     int64 max_system_from = 6  [default = -9223372036854775808];
     int64 row_count = 7 [default = 0];
     bytes iid_bloom = 8;
+    repeated string hll_column_name = 9;
+    repeated bytes hyper_log_log = 10;
 }
```
FiV0: I don't if you prefer a composite object here.

core/src/main/clojure/xtdb/table_catalog_test.clj
```diff
@@ -41,8 +42,8 @@
         (xt/execute-tx node [[:put-docs :foo {:xt/id 2}]])
         (tu/finish-block! node)
 
-        (t/is (= [(os/->StoredObject "tables/public$foo/blocks/b00.binpb" 286)
-                  (os/->StoredObject "tables/public$foo/blocks/b01.binpb" 415)]
+        (t/is (= [(os/->StoredObject "tables/public$foo/blocks/b00.binpb" 4392)
+                  (os/->StoredObject "tables/public$foo/blocks/b01.binpb" 8627)]
```
FiV0: A 10x increase in size. Might be better to only add them (HLL's) at L1.

core/src/main/kotlin/xtdb/indexer/LiveTable.kt
```diff
@@ -134,7 +138,12 @@ class LiveTable
                 trieMetadataBuilder.minSystemFrom = minOf(trieMetadataBuilder.minSystemFrom, systemFromRdr.getLong(i))
                 trieMetadataBuilder.maxSystemFrom = maxOf(trieMetadataBuilder.maxSystemFrom, systemFromRdr.getLong(i))
                 trieMetadataBuilder.rowCount += (pos - startPos)
+
                 it.iidBloom.add(*bloomHashes(VectorReader.from(iidRdr) , i))
+
+                for ((col, rdr) in legRdrs) {
+                    hyperLogLogs.compute(col) { _, hll -> (hll ?: HyperLogLog.createHLL()).also { HyperLogLog.add(it, rdr, i) } }
+                }
```
FiV0, line 139: This is the hot path, so might also be quite the slow down to do it here.

core/src/main/clojure/xtdb/table_catalog_test.clj
jarohen: can't - HLLs aren't part of the LSM tree; they only apply to L0 files?

core/src/main/kotlin/xtdb/indexer/LiveTable.kt
```diff
@@ -125,6 +126,9 @@ class LiveTable
         }
 
         fun commit() = this@LiveTable.also {
+            val putRdr = putWtr.asReader
+            val columns = putRdr.structKeys()
+            val legRdrs = columns.map { it to VectorReader.from(putRdr.structKeyReader(it)!!) }
```
jarohen: these are key readers rather than leg readers: DUV -> leg; struct -> key

core/src/main/kotlin/xtdb/util/HyperLogLog.kt
```diff
@@ -0,0 +1,136 @@
...
+
+    const val DEFAULT_BUFFER_SIZE: Int = Integer.BYTES * 1024
+
+    /**
+     * Add a value to the HyperLogLog counter
+     * @param hll The ByteBuf containing HLL registers
+     * @param hash The value to add
+     * @return The same ByteBuf after modification
+     */
+    fun add(hll: ByteBuf, hash: Int): ByteBuf {
```
jarohen: Kotlin uses operator overloading for this - I'd go for `operator fun HLL.plus(hash: Int): ByteBuf`, then used as `hll + hash`.

core/src/main/kotlin/xtdb/util/HyperLogLog.kt
```diff
@@ -0,0 +1,136 @@
...
+        hll.setInt(position, newValue)
+
+        return hll
+    }
+
+    @JvmStatic
+    @JvmOverloads
+    fun add(hll: ByteBuf, vec: VectorReader, idx: Int, hasher: Hasher = Hasher.Xx()) = add(hll, vec.hashCode(idx, hasher))
+
+    private val twoTo32 = Integer.toUnsignedLong(-1)
```
jarohen: `private const val TWO_TO_32`, probably. I'd probably also write it as `2 shl 32` given the choice?

core/src/main/kotlin/xtdb/util/HyperLogLog.kt
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * Estimate the cardinality of the set represented by this HyperLogLog
+     * @param hll The ByteBuf containing HLL registers
+     * @return The estimated cardinality
+     */
+    @JvmStatic
+    fun estimate(hll: ByteBuf): Double {
```
jarohen: `fun HLL.estimate(): Double`

core/src/main/kotlin/xtdb/util/HyperLogLog.kt
```diff
@@ -0,0 +1,136 @@
...
+        return result
+    }
+
+    /**
+     * Estimate the cardinality of the union of two sets
+     * @param hllA First HLL buffer
+     * @param hllB Second HLL buffer
+     * @return The estimated cardinality of the union
+     */
+    fun estimateUnion(hllA: ByteBuf, hllB: ByteBuf): Double {
```
jarohen: exprs rather than block

core/src/main/kotlin/xtdb/util/HyperLogLog.kt
```diff
@@ -0,0 +1,136 @@
...
+     * @param hllA First HLL buffer
+     * @param hllB Second HLL buffer
+     * @return The estimated cardinality of the difference
+     */
+    fun estimateDifference(hllA: ByteBuf, hllB: ByteBuf): Double {
+        return estimate(hllA) - estimateIntersection(hllA, hllB)
+    }
+
+    @JvmStatic
+    fun createHLL(bufferSize: Int = DEFAULT_BUFFER_SIZE) = Unpooled.buffer(bufferSize)
```
jarohen: probably just `create` here - you're already in context of HLL so callers will look like `HyperLogLog.create`

core/src/main/proto/xtdb/log/proto/log.proto
jarohen: yes please - this modelling is encoding a zipmap assumption. I assume protobuf has a map type?

core/src/main/kotlin/xtdb/indexer/LiveTable.kt
```diff
@@ -31,6 +31,7 @@ class LiveTable
     private val trieWriter = TrieWriter(al, bp, false)
     private val trieMetadataBuilder = TrieMetadata.newBuilder()
     private var iidBloom = RoaringBitmap()
+    private val hyperLogLogs = mutableMapOf<String, HLL> ()
```
jarohen: I'd be fine with calling these `hlls`, I think, save both your fingers and my eyes :sweat_smile:

core/src/main/kotlin/xtdb/util/HyperLogLog.kt
jarohen: Agreed that this is closer to Clojure, but Kotlin != Clojure :slightly_smiling_face: 

I have a slight preference for extension functions wherever you'd typically use threading in Clojure (i.e. where the first parameter is the object being acted on) because it then allows you to use (e.g.) `?.` and dot-chaining in Kotlin, rather than needing more and more nesting - same as threading in Clojure, now I think about it.

core/src/main/proto/xtdb/log/proto/log.proto
jarohen: Is this message the right place to put the HLLs? they're tied to the table, rather than to the trie itself (just happens to be the L0 trie?)

When we come to read them back in, I suspect we'll want them as part of the table-block file, and written cumulatively (so that we don't have to scan the history)

core/src/main/clojure/xtdb/table_catalog_test.clj
jarohen: do we get any choice over the sizing parameters here, if you're concerned about the size?

core/src/main/kotlin/xtdb/indexer/LiveTable.kt
jarohen: how slow?

jarohen (changes requested):
Thanks @FiV0, good first pass :slightly_smiling_face: 

I suspect the biggest change here is regarding where we store these - now that you've got more of an idea about what they are and how they work, could you take a mo to put yourself in the shoes of us using them, and you'd store them as a result? (If you don't have enough context to confidently do that, that's always something to raise please :slightly_smiling_face:).

Let's have a chat if that'd be easier

FiV0:
As discussed only have HLL's on per table/column basis. The HLL per table/file/column is overkill and likely not really useful.

core/src/main/kotlin/xtdb/util/HyperLogLog.kt
FiV0:
```
rlwrap kotlin
Welcome to Kotlin version 2.1.0 (JRE 21.0.5+11)
Warning: this REPL implementation is deprecated and will be removed soon.
Warning: REPL is not compatible with the Kotlin version 2.1.0, using '-language-version 1.9'.
Type :help for help, :quit for quit
>>> Integer.toUnsignedLong(-1)
res0: kotlin.Long = 4294967295
>>> 2 shl 32
res1: kotlin.Int = 2
>>> 2L shl 32
res2: kotlin.Long = 8589934592
>>> 2L shl 31
res3: kotlin.Long = 4294967296
```
Not sure what to make of that.

FiV0: Probably need to recheck the logic around here:
https://github.com/xtdb/xtdb/blob/06132bc1fad63944d144d79a90053b05d78a41c8/core/src/xtdb/hyper_log_log.clj#L46-L50. 
That `Math.pow(-2.0, 32.0)` seems weird.

core/src/main/kotlin/xtdb/indexer/LiveTable.kt
FiV0: Compared `ingest_tx_overhead` of this with `main` and couldn't see any significant difference.

PR #4282, opened 2025-03-18
Use temporal file metadata

### LINKED ISSUE #4260: Using file-level temporal metadata to elide whole files in scan

FiV0:
~With #4281 the change of using metadata for filtering files becomes a breeze. We rename `->merge-task` to `filter-meta-objects` which means it should works for anything that implements `PMetadta` (pages + files for now).~

I have first added a function `filter-tries` to the trie-catalog which takes the tries as well as the query bounds and using the metadata + recency calculates which tries are relevant for the query.

In a second pass I have consolidated `->merge-task` (page level) and `filter-tries` (file level) into a single function (`filter-meta-objects`) as the logic is identical. The only difference is that trie filtering also makes use of recency which then becomes an extra step in `filter-tries`.

TODO:
- [x] The big todo is currently in `trie_test.clj`. There are quite a few tests that mainly test the logic around which pages to take while we were still having recency in the tries. ~Should this be moved to end to end tests with appropriate data ingestion (things going into historical as well as current) + compaction + range scans~ or should this remain ~higher~ lower level test on the trie_catalog?

Decision:
- Try to remove all recency logic testing from `trie_test.clj` (keep the correct IID resolving)
- Add appropriate tests for `filter-meta-objects` to trie_catalog
- Actually use `recency` in `filter-meta-objects` on file level

### DIFF
--- a/src/test/clojure/xtdb/operator/scan_test.clj
+++ b/src/test/clojure/xtdb/operator/scan_test.clj
@@ -13,6 +13,7 @@
             [xtdb.vector.writer :as vw])
   (:import java.util.Date
            (java.util.function IntPredicate)
+           (xtdb.compactor RecencyPartition)
            xtdb.vector.RelationReader))
 
 (t/use-fixtures :each tu/with-mock-clock tu/with-allocator tu/with-node)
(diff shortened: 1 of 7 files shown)


jarohen (changes requested):
thanks @FiV0 :pray: 

code change looks good - as discussed, we have the new scan-test to fix, then let's re-review

src/test/clojure/xtdb/operator/scan_test.clj
```diff
@@ -820,59 +821,79 @@
...
-                                               (Date. (+ (.getTime date) delta))))))
-                             [range1 range2])]
-            (t/testing (format "date: '%s'" (pr-str date))
-              (doseq [q ["SELECT * FROM foo FOR VALID_TIME AS OF ? WHERE _id = ?"
-                         "SELECT *, _valid_from, _valid_to FROM foo FOR VALID_TIME AS OF ? WHERE _id = ?"]]
-                (t/testing (format "query: '%s'" q)
-                  ;; params the wrong way around - see #4305
-                  (t/is (= (xt/q n1 [q idx date])
-                           (xt/q n2 [q idx date]))))))))))))
+  (doseq [bucketing  [RecencyPartition/WEEK RecencyPartition/YEAR]
```
FiV0, line 825: This test now runs in ~20 secs. Could take out some factors if you find it too long.

FiV0: Could also make it an integration test.

jarohen (approved):

PR #4287, opened 2025-03-19
Migration tool for beta7, #4176

### LINKED ISSUE #4176: Migration tool for beta7
We should provide a migration tool for users migrating their data to beta7.

If users have the full history of their database in the log still (either through infinite retention, or maybe they're deriving the XT log from another upstream log) it may be easier for them to simply replay - i.e. remove the contents of the object store - but an infinite retention log is no longer a hard requirement, so we should provide a tool to migrate users' object stores.

Migration process could be as follows:

* User runs a one-shot task while the beta6 nodes are still up
* Task copies over L0 files and block files (see below)
* Task notifies the user that the migration is complete and exits.
* User performs a green/blue migration of their XT nodes from beta6 -> beta7 in the usual way
* beta7 compactors compact L0 files as normal
* Later, user cleans up `/v05` from their object store.

As for the tool:

* Connect to the user's object store.
* For each block-metadata transit/json file in `/v05/blocks`
  * create a chunk-idx -> block-idx mapping
  * for each table:
    * copy L0 files over, renaming as appropriate (`l00-b04.arrow`)
      * ~this should only need a rename, because L0 file contents are unchanged~ no, because 97940ac3 and #4014 
    * create table-block files (see #4171)
  * create block file.
* Log a message to the user letting them know that:
  * the migration has finished successfully
  * they can now scale up the new nodes and scale down the old ones in the normal green/blue manner.

jarohen:
resolves #4176 

(below to be added to release notes, but also serves as a description of the PR)

---

2.0.0-beta7 is a breaking index-format change - users will need to ensure their storage is migrated before standing up beta7 nodes.

There are two options:

1. If you have an infinite-retention log, you can simply stand beta7 nodes up in the usual green/blue manner. These will then re-index the log from scratch, which may take a while on large datasets.

   This is the _easiest_ approach, assuming you have a relatively small dataset.
2. If you have a larger dataset, or a finite-retention log, you'll need to run our migration tool. This is executed as a one time task - you can leave your beta6 nodes running and run the migration tool in the background.  

   To run the migration tool:

   * If you run XTDB through its CLI, run the beta7 artifact supplying your usual configuration arguments, but additionally add `--migrate-from 5`.
   * If you run XTDB through a Docker container orchestration platform (e.g. Kubernetes), run the beta7 version of your usual Docker image as a one-shot task, and override the Docker command to add the `--migrate-from 5` flag.

   Once the migration tool has finished (it'll notify you, and then exit), you can deploy your beta7 nodes and decommission your beta6 nodes in the usual green/blue manner.

Rollback procedure:

1. Turn off any beta7 nodes if you've started them
2. Turn on beta6 nodes
3. Delete `v06` directory in your object store.


refset:
> Once the migration tool has finished (it'll notify you, and then exit), you can deploy your beta7 nodes

@jarohen is there a finite window of time that the deploy of the beta7 nodes needs to happen within?

PR #4290, opened 2025-03-20
Add capability to restart node from storage when you have an empty/stale log topic

### LINKED ISSUE #3928: Support & document restoring a node with intact storage and empty/missing log topic.
XTDB currently assumes the transaction log (Kafka) is always available, relying a single 'epoch' of offsets from the Kafka topic. However, if the Kafka topic is lost, nodes must be able to reset their offsets and recover data from object storage. A manual recovery process must be defined to guide users through restoring an XTDB node with a missing tx-log:
- Implement a mechanism for XTDB to be able to start from an empty Log Topic / reset it's expected offsets and start entirely from object storage.
- Document a manual recovery procedure for users who have lost/reset their Log topic - creating a new topic and resetting the node offsets.
  - Documentation will also need to make clear what data will be missing - ie, contents of the LogTopic.
- Improve error messages in the case where offset = 0 and last indexed tx > 0 to point to this process and document what steps the user has to take to rectify.
- Ensure we have some form of automated testing against this.

### LINKED ISSUE #4248: Support & document restoring a node with intact storage and stale log topic.
Related to and following on from work done within #3928.

When XTDB encounters a log topic where the current transactions are older than those already indexed (i.e., current offset is lower than last indexed tx but not zero), it currently throws an ingestion stopped error and blocks further progress. In this scenario, XTDB must handle the situation similarly to a total loss of Kafka (as outlined in #3928), and include a manual recovery process to allow users to restore their node safely:
- Document a manual recovery procedure for users who have a stale Log topic, covering the process of creating a new topic and resetting the node offsets and starting the node from storage.
  - Documentation will also need to make clear what data will be missing on node restart - ie, contents of the LogTopic.
- Improve error messages in the case where `current offset > 0` and `current offset < last indexed tx` to point to this process and document what steps the user has to take to rectify.
- Ensure we have some form of automated testing against this.

danmason:
Resolves #3928 and #4248
Github actions: https://github.com/danmason/xtdb/actions?query=branch%3Arestart-node-empty-log-topic++

### Summary

Stores an epoch value into the message id, which allows us to have a "version" of the offsets, and start reading from the start of the log when we need to (ie, the log component is now pointed at any empty log) and a new currentEpoch has been set. 
- Of the 64bit message ID, we use:
  - The first 16 bits to represent the epoch (giving us a potential 65536 epochs)  
  - The remaining 48 bits represent the offset (allowing us a total of 2^48-1 offsets per epoch).
- Since the first epoch is 0, messageId remains as it is, assuming nobody has more than 2^48-1 offsets currently (ie, those first 16 bits will always be zero currently)

When an out of sync log with intact storage is detected, we log an error at the point of starting up log and prevent the node from starting, providing a chunk of this information to the user within the error itself. This handles:
- Empty log -  (ie, current epoch is the same as last written epoch, and last submitted offset is -1/empty)
- Stale log - (ie, current epoch is the same as last written epoch, and last submitted offset is earlier than latest completed one)

This PR also starts our "backup and restore" docs section, with the following:
- An overview document explaining what data is stored where, impact of losing said data and linking out to specific recovery scenarios.
- A document explaining the recovery process for an out of sync log.

### DIFF
--- a/core/src/main/clojure/xtdb/log.clj
+++ b/core/src/main/clojure/xtdb/log.clj
@@ -1,5 +1,6 @@
 (ns xtdb.log
   (:require [clojure.string :as str]
+            [clojure.tools.logging :as log]
             [integrant.core :as ig]
             [xtdb.api :as xt]
             [xtdb.error :as err]
@@ -17,12 +18,14 @@
            (org.apache.arrow.vector VectorSchemaRoot)
            (org.apache.arrow.vector.types.pojo ArrowType$Union FieldType Schema)
            org.apache.arrow.vector.types.UnionMode
-           (xtdb.api Xtdb$Config)
+           (xtdb.api Xtdb$Config TransactionKey)
            (xtdb.api.log Log Log$Factory Log$Message$Tx)
            (xtdb.api.tx TxOp$Sql)
            (xtdb.arrow Relation VectorWriter)
+           xtdb.catalog.BlockCatalog
            xtdb.indexer.LogProcessor
-           (xtdb.tx_ops Abort DeleteDocs EraseDocs PatchDocs PutDocs SqlByteArgs)))
+           (xtdb.tx_ops Abort DeleteDocs EraseDocs PatchDocs PutDocs SqlByteArgs)
+           (xtdb.util TxIdUtil)))
 
 (set! *unchecked-math* :warn-on-boxed)
 
@@ -229,15 +232,17 @@
 
       (.getAsArrowStream rel))))
 
-(defmethod xtn/apply-config! ::memory-log [^Xtdb$Config config _ {:keys [instant-src]}]
+(defmethod xtn/apply-config! ::memory-log [^Xtdb$Config config _ {:keys [instant-src epoch]}]
   (doto config
     (.setLog (cond-> (Log/getInMemoryLog)
-               instant-src (.instantSource instant-src)))))
+               instant-src (.instantSource instant-src)
+               epoch (.epoch epoch)))))
 
-(defmethod xtn/apply-config! ::local-directory-log [^Xtdb$Config config _ {:keys [path instant-src]}]
+(defmethod xtn/apply-config! ::local-directory-log [^Xtdb$Config config _ {:keys [path instant-src epoch]}]
   (doto config
     (.setLog (cond-> (Log/localLog (util/->path path))
-               instant-src (.instantSource instant-src)))))
+               instant-src (.instantSource instant-src)
+               epoch (.epoch epoch)))))
 
 (defmethod xtn/apply-config! :xtdb/log [config _ [tag opts]]
   (xtn/apply-config! config
@@ -247,8 +252,38 @@
                        :kafka :xtdb.kafka/log)
                      opts))
 
-(defmethod ig/init-key :xtdb/log [_ ^Log$Factory factory]
-  (.openLog factory))
+(defmethod ig/prep-key :xtdb/log [_ factory]
+  {:block-cat (ig/ref :xtdb/block-catalog)
+   :factory factory})
+
+(def out-of-sync-log-message
+  "Node failed to start due to an invalid transaction log state (%s) that does not correspond with the latest indexed transaction (epoch=%s and offset=%s).
+   
+   Please see https://docs.xtdb.com/ops/ops/backup-and-restore/out-of-sync-log for more information and next steps.")
+
+(defn ->out-of-sync-exception [latest-completed-offset ^long latest-submitted-offset ^long epoch]
+  (let [log-state-str (if (= -1 latest-submitted-offset)
+                        "the log is empty"
+                        (format "epoch=%s, offset=%s" epoch latest-submitted-offset))]
+    (IllegalStateException.
+     (format out-of-sync-log-message log-state-str epoch latest-completed-offset))))
+
+(defn validate-offsets [^Log log ^TransactionKey latest-completed-tx]
+  (when latest-completed-tx
+    (let [latest-completed-tx-id (.getTxId latest-completed-tx)
+          latest-completed-offset (TxIdUtil/txIdToOffset latest-completed-tx-id)
+          latest-completed-epoch (TxIdUtil/txIdToEpoch latest-completed-tx-id)
+          epoch (.getEpoch log)
+          latest-submitted-offset (.getLatestSubmittedOffset log)]
+      (if (= latest-completed-epoch epoch)
+        (cond
+          (< latest-submitted-offset latest-completed-offset)
+          (throw (->out-of-sync-exception latest-completed-offset latest-submitted-offset epoch)))
+        (log/info "Starting node with a log that has a different epoch than the latest completed tx (This is expected if you are starting a new epoch) - Skipping offset validation.")))))
+
+(defmethod ig/init-key :xtdb/log [_ {:keys [^BlockCatalog block-cat ^Log$Factory factory]}]
+  (doto (.openLog factory)
+    (validate-offsets (.getLatestCompletedTx block-cat))))
 
 (defmethod ig/halt-key! :xtdb/log [_ ^Log log]
   (util/close log))
--- a/core/src/main/kotlin/xtdb/api/log/InMemoryLog.kt
+++ b/core/src/main/kotlin/xtdb/api/log/InMemoryLog.kt
@@ -15,14 +15,18 @@ import java.time.temporal.ChronoUnit.MICROS
 import kotlin.time.Duration.Companion.minutes
 import kotlin.time.Duration.Companion.seconds
 
-class InMemoryLog(private val instantSource: InstantSource) : Log {
+class InMemoryLog(private val instantSource: InstantSource, override val epoch: Int) : Log {
 
     @SerialName("!InMemory")
     @Serializable
-    data class Factory(@Transient var instantSource: InstantSource = InstantSource.system()) : Log.Factory {
+    data class Factory(
+        @Transient var instantSource: InstantSource = InstantSource.system(),
+        var epoch: Int = 0
+    ) : Log.Factory {
         fun instantSource(instantSource: InstantSource) = apply { this.instantSource = instantSource }
+        fun epoch(epoch: Int) = apply { this.epoch = epoch }
 
-        override fun openLog() = InMemoryLog(instantSource)
+        override fun openLog() = InMemoryLog(instantSource, epoch)
     }
 
     private val scope: CoroutineScope = CoroutineScope(Dispatchers.Default)
--- a/core/src/main/kotlin/xtdb/indexer/LogProcessor.kt
+++ b/core/src/main/kotlin/xtdb/indexer/LogProcessor.kt
@@ -7,13 +7,14 @@ import kotlinx.coroutines.runBlocking
 import org.apache.arrow.memory.BufferAllocator
 import org.apache.arrow.vector.ipc.ArrowStreamReader
 import org.slf4j.LoggerFactory
-import xtdb.api.log.Log
+import xtdb.api.log.*
 import xtdb.api.log.Log.Message
-import xtdb.api.log.MessageId
-import xtdb.api.log.Watchers
 import xtdb.api.storage.Storage
 import xtdb.arrow.asChannel
 import xtdb.trie.TrieCatalog
+import xtdb.util.TxIdUtil.offsetToTxId
+import xtdb.util.TxIdUtil.txIdToEpoch
+import xtdb.util.TxIdUtil.txIdToOffset
 import xtdb.util.error
 import xtdb.util.logger
 import java.nio.channels.ClosedByInterruptException
@@ -36,7 +37,18 @@ class LogProcessor(
         private val LOG = LogProcessor::class.logger
     }
 
-    private val watchers = Watchers(liveIndex.latestCompletedTx?.txId ?: -1)
+    private val epoch = log.epoch
+
+    override var latestProcessedMsgId: MessageId =
+        liveIndex.latestCompletedTx?.txId?.let {
+            if (txIdToEpoch(it) == epoch) it else offsetToTxId(epoch, 0) - 1
+        } ?: -1
+        private set
+
+    override val latestSubmittedMsgId: MessageId
+        get() = offsetToTxId(epoch, log.latestSubmittedOffset)
+
+    private val watchers = Watchers(latestProcessedMsgId)
     private val LOGGER = LoggerFactory.getLogger(LogProcessor::class.java)
 
     val ingestionError get() = watchers.exception
@@ -89,26 +101,21 @@ class LogProcessor(
         allocator.close()
     }
 
-    override val latestSubmittedMsgId: MessageId
-        get() = log.latestSubmittedOffset
-
-    override var latestProcessedMsgId: MessageId =
-        liveIndex.latestCompletedTx?.txId ?: -1
-        private set
-
     private val flusher = Flusher(flushTimeout, liveIndex)
 
-    private val latestProcessedOffset = latestProcessedMsgId
+    private val latestProcessedOffset = liveIndex.latestCompletedTx?.txId?.let {
+        if (txIdToEpoch(it) == epoch) txIdToOffset(it) else -1
+    } ?: -1
     private val subscription = log.subscribe(this, latestProcessedOffset)
 
     override fun processRecords(records: List<Log.Record>) = runBlocking {
         flusher.checkBlockTimeout(liveIndex)?.let { flushMsg ->
-            val msgId = log.appendMessage(flushMsg).await()
-            flusher.flushedTxId = msgId
+            val offset = log.appendMessage(flushMsg).await()
+            flusher.flushedTxId = offsetToTxId(epoch, offset)
         }
 
         records.forEach { record ->
-            val msgId = record.logOffset
+            val msgId = offsetToTxId(epoch, record.logOffset)
 
             try {
                 val res = when (val msg = record.message) {
@@ -122,7 +129,6 @@ class LogProcessor(
                                 ArrowStreamReader(txOpsCh, allocator).use { reader ->
                                     reader.vectorSchemaRoot.use { root ->
                                         reader.loadNextBatch()
-
                                         indexer.indexTx(msgId, record.logTimestamp, root)
                                     }
                                 }
@@ -152,12 +158,12 @@ class LogProcessor(
                 throw CancellationException(ie)
             } catch (e: Throwable) {
                 watchers.notify(msgId, e)
-                LOG.error(e, "Ingestion stopped: error processing log record at id $msgId.")
+                LOG.error(e, "Ingestion stopped: error processing log record at id $msgId (epoch: $epoch, logOffset: ${record.logOffset})")
                 LOG.error(
                     """
                     XTDB transaction processing has encountered an unrecoverable error and has been stopped to prevent corruption of your data.
                     This node has also been marked unhealthy, so if it is running within a container orchestration system (e.g. Kubernetes) it should be restarted shortly.
-                    
+
                     Please see https://docs.xtdb.com/ops/troubleshooting#ingestion-stopped for more information and next steps.
                 """.trimIndent()
                 )
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -151,6 +151,13 @@ export default defineConfig({
                         { label: 'AWS', link: '/ops/aws' },
                         { label: 'Azure', link: '/ops/azure' },
                         { label: 'Google Cloud', link: '/ops/google-cloud' },
+                        { 
+                            label: 'Backup and Restore',
+                            items: [
+                                { label: 'Overview', link: '/ops/backup-and-restore/overview' },
+                                { label: 'Out of Sync Log & Intact Storage', link: '/ops/backup-and-restore/out-of-sync-log' },
+                            ]
+                        },
                         { label: 'Troubleshooting', link: '/ops/troubleshooting' },
 
                     ]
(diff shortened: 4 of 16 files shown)


core/src/main/clojure/xtdb/log.clj
```diff
@@ -247,8 +250,56 @@
...
+     (format template latest-offset (inc current-epoch)))))
+
+(defn validate-offsets [^Log log ^TransactionKey latest-completed-tx]
+  (when latest-completed-tx
+    (let [log-type (class->log-type log)
+          latest-completed-tx-id (.getTxId latest-completed-tx)
+          latest-offset (tx-id->offset latest-completed-tx-id)
+          latest-epoch (tx-id->epoch latest-completed-tx-id)
+          current-epoch (.getCurrentEpoch log)]
+      (when (= latest-epoch current-epoch)
```
danmason: Could perhaps include a log when not validating offsets (ie, because epochs differ)

core/src/main/clojure/xtdb/log.clj
```diff
@@ -247,8 +251,58 @@
...
+  (when latest-completed-tx
+    (let [log-type (class->log-type log)
+          latest-completed-tx-id (.getTxId latest-completed-tx)
+          latest-completed-offset (tx-id->offset latest-completed-tx-id)
+          latest-completed-epoch (tx-id->epoch latest-completed-tx-id)
+          current-epoch (.getCurrentEpoch log)
+          latest-submitted-offset (.getLatestSubmittedOffset log)]
+      (if (= latest-completed-epoch current-epoch)
+        (cond
+          (= -1 latest-submitted-offset)
```
refset: Rather than handling an empty log as a special case, could it be worth widening the scope of this PR to cover "empty" _and_ "stale" states? i.e. simply swap this logic to `(< latest-submitted-offset latest-completed-offset)` (and adjust the various error text etc. accordingly)

This would then also cover the scenario of a log that has been restored from a backup, but is too old to be useful.

docs/src/content/docs/ops/backup-and-restore/empty-log.adoc
```diff
@@ -0,0 +1,68 @@
...
+
+If the storage backend remains intact, then all previously compacted and indexed data is still available.
+
+* Historical data in all compacted Arrow files (Level-0 and higher)
+* The full queryable state of the database, up to the last successfully indexed offset
+
+== What Is Lost
+
+* The transaction log and its message history, including any unindexed or recently submitted transactions
+* Internal system messages (e.g. coordination events) recorded only in the log
```
refset: Does the loss of internal system messages have any discernible impact to the user?

docs/src/content/docs/ops/backup-and-restore/empty-log.adoc
```diff
@@ -0,0 +1,68 @@
...
+== What Is Preserved
+
+If the storage backend remains intact, then all previously compacted and indexed data is still available.
+
+* Historical data in all compacted Arrow files (Level-0 and higher)
+* The full queryable state of the database, up to the last successfully indexed offset
+
+== What Is Lost
+
+* The transaction log and its message history, including any unindexed or recently submitted transactions
```
refset: How something slightly more meaningful to the user: "Any recently submitted transactions that were not indexed prior to the loss of the transaction log"

core/src/main/clojure/xtdb/log.clj
danmason: I was planning to cover this in a follow up PR, see issue https://github.com/xtdb/xtdb/issues/4248 - essentially, have a separate/specific error message and also specific restore documentation for it. (Since you will have to first point it at a new, empty log)

docs/src/content/docs/ops/backup-and-restore/empty-log.adoc
danmason: It shouldn't do - could definitely drop that from here!

docs/src/content/docs/ops/backup-and-restore/empty-log.adoc
danmason: :+1: Seems good, thank you for the suggestion!

core/src/main/clojure/xtdb/log.clj
refset: Thanks for the pointer. Still, I'm wondering whether a simpler UX here would be to have a single (universal) error message that covers both scenarios, e.g.

> Node failed to start due to an invalid transaction log state (epoch=N, offset=X <|| "the log is empty">) that does not correspond with the latest indexed transaction (epoch=M and offset=Y).
> Please check your log configuration and make sure the log's underlying storage/state has not been modified.
> If you wish to start the node anyway and resume transaction processing based on the indexed storage alone, (i.e. you are willing to accept some potential loss of previously written transactions since the most recently indexed transaction), you can configure a new epoch using an empty log.
> WARNING: Please note that beginning a new epoch will affect all nodes attempting to write to the log and therefore the configuration change should be applied universally and atomically (i.e. any new transactions submitted by other nodes that attempt to use a prior epoch will be rejected). Furthermore, beginning a new epoch will make the recovery of unindexed transactions from any previous epoch nearly impossible. Therefore, you should only increment the epoch when you are confident that the original log is irrecoverable and you understand the potential consequences. In any case, we strongly recommend ensuring you create a fresh backup of the indexed storage prior to changing the epoch config.
> Assuming you are satisfied with the above explanations, you can set the following in your config to begin a new epoch:
> [...] 
> See the XTDB documentation for more guidance and recovery information.

Otherwise it's more code and docs to maintain, and more docs for users to read/understand/feel concerned about, with little benefit.

Note also how the above message guides users to read the docs but more importantly equips them with the essential information they need _before_ they naively attempt to change the epoch.

danmason: Those are some really useful insights - I shall have a look at reworking my error logging here, then.

core/src/main/clojure/xtdb/log.clj
```diff
@@ -229,15 +231,17 @@
 
       (.getAsArrowStream rel))))
 
-(defmethod xtn/apply-config! ::memory-log [^Xtdb$Config config _ {:keys [instant-src]}]
+(defmethod xtn/apply-config! ::memory-log [^Xtdb$Config config _ {:keys [instant-src current-epoch]}]
```
jarohen: nit: `current` here is superfluous given I think we only ever deal with the current one? -> `epoch`?

core/src/main/kotlin/xtdb/indexer/LogProcessor.kt
```diff
@@ -36,7 +37,34 @@ class LogProcessor(
...
+    private val currentEpoch = log.currentEpoch
+
+    private fun offsetToTxId(offset: LogOffset): Long {
+        require(currentEpoch < epochLimit) { "Epoch value ($currentEpoch) exceeds the limit ($epochLimit)" }
+        require(offset < offsetLimit) { "Offset value ($offset) exceeds the limit ($offsetLimit)" }
+
+        return (currentEpoch.toLong() shl 48) + offset
+    }
+
+    private fun txIdToEpoch(txId: Long): Int {
```
jarohen: prefer expression bodies where possible (IntelliJ will give you this as an option if you Alt+Enter)

`private fun txIdToEpoch(txId: Long) = (txId shr 48).toInt()`

core/src/main/clojure/xtdb/log.clj
```diff
@@ -247,8 +251,53 @@
...
+   :factory factory})
+
+(defn tx-id->offset [^long tx-id]
+  (bit-and tx-id (dec (bit-shift-left 1 48))))
+
+(defn tx-id->epoch [^long tx-id]
+  (bit-shift-right tx-id 48))
+
+(def out-of-sync-log-message
+  "Node failed to start due to an invalid transaction log state (%s) that does not correspond with the latest indexed transaction (epoch=%s and offset=%s).
```
jarohen, line 260: can this detail go in the operational docs instead, and a link here? reasoning being we can point people to that if they don't spot the log message, and also gives us more flexibility with formatting/level of detail.

see bottom of LogProcessor.kt for prior art

core/src/main/kotlin/xtdb/api/log/InMemoryLog.kt
```diff
@@ -44,7 +48,7 @@ class InMemoryLog(private val instantSource: InstantSource) : Log {
             onCommit.complete(record.logOffset)
             record
         }
-        .shareIn(scope, SharingStarted.Eagerly)
+        .shareIn(scope, SharingStarted.Eagerly, 100)
```
jarohen: unrelated change?

docs/astro.config.mjs
```diff
@@ -150,6 +150,14 @@ export default defineConfig({
                         { label: 'AWS', link: '/ops/aws' },
                         { label: 'Azure', link: '/ops/azure' },
                         { label: 'Google Cloud', link: '/ops/google-cloud' },
+                        { 
+                            label: 'Backup and Restore',
+                            items: [
+                                { label: 'Overview', link: '/ops/backup-and-restore/overview' },
+                                { label: 'Empty Log & Intact Storage', link: '/ops/backup-and-restore/empty-log' },
```
jarohen: combine? especially given the solution is so similar

docs/src/content/docs/ops/backup-and-restore/overview.adoc
```diff
@@ -0,0 +1,58 @@
...
+=== Log
+
+A totally ordered, append-only log.
+
+* This contains all transactions (representing one or more record-level operations) and their metadata, and inter-node messages. 
+* Guarantees ACID-compliant serial execution of all writes.
+* Pluggable implementations include:
+** *In-memory* (ephemeral - any unindexed data is lost on restart)
+** *Local disk* (durable on a single node)
+** *Remote/Kafka* (distributed, durable across nodes)
```
jarohen: > (distributed, durable across nodes, high availability)

docs/src/content/docs/ops/backup-and-restore/overview.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+This document provides a high-level overview of XTDB’s stateful architecture and outlines what data is stored where, what typical backups include, and how to think about restoring XTDB in different failure scenarios.
+
+== Overview of Stateful Components
+
+XTDB consists of two primary stateful components:
+
+=== Log
+
+A totally ordered, append-only log.
```
jarohen: we already have a little bit of this under (e.g.) https://docs.xtdb.com/ops/config/log.html. Worth either bringing that over here, or talking about backup over there?

docs/src/content/docs/ops/backup-and-restore/overview.adoc
```diff
@@ -0,0 +1,58 @@
...
+** *Remote/Kafka* (distributed, durable across nodes)
+* Every transaction is recorded immutably and linked to a durable log offset.
+
+=== Storage Module
+
+A immutable store containing the data and indexes that make up the database.
+
+* Structured as https://arrow.apache.org[Apache Arrow^] columnar files for performance and flexibility.
+* Data is periodically compacted into increasingly efficient layers.
+** *Level-0* files contain full snapshots of each table.
```
jarohen: (weak opinion) I'm not sure this (details of how compaction works) is what a reader reading this page would necessarily care about...? :thinking: from the PoV of the storage, it doesn't know nor need to care about what files are going in there, maybe only that these are snapshots of the log.

docs/src/content/docs/ops/backup-and-restore/overview.adoc
```diff
@@ -0,0 +1,58 @@
...
+|===
+| Component | Risk if Lost
+
+| Transaction Log
+| All transactions on the log since the last indexed offset would be lost.
+
+| Storage Module
+| Without a backup or infinite retention on the log, all indexed data would be lost.
+
+| Local Caches (disk/in-memory)
```
jarohen: I'd probably call these the 'XTDB compute nodes' as the headline, to keep it higher-level for the reader here - you're right, in that it's the caches that are ephemeral, but putting myself in a user's shoes, it's probably not the level of detail that I'm thinking about? :thinking:

docs/src/content/docs/ops/backup-and-restore/stale-log.adoc
```diff
@@ -0,0 +1,71 @@
+---
+title: Scenario - Stale Log & Intact Storage
+---
+
+If the log contains transaction entries but the most recent offset is **lower than the last indexed transaction**, XTDB will refuse to start. This typically indicates the log has been truncated, rotated, or replaced unexpectedly, resulting in **missing historical entries** that the system relies on to maintain consistency.
```
jarohen: sentence per line in adoc :wink:

docs/src/content/docs/ops/backup-and-restore/stale-log.adoc
```diff
@@ -0,0 +1,71 @@
...
+* Historical data in all compacted Arrow files (Level-0 and higher)
+* The full queryable state of the database, up to the last successfully indexed offset
+
+== What Is Lost
+
+Any recently submitted transactions _after_ the last indexed offset that were truncated or lost from the log:
+
+* These are **unrecoverable** unless separately backed up or tracked externally.
+
+== Recovery Strategy: Using Epochs
```
jarohen: Rather than including details about epochs within the individual scenarios, I'd include this as reference documentation of the log component? Good to have a guide for the scenarios, granted - but it seems a little inside-out to _primarily_ document epochs here.

src/test/clojure/xtdb/kafka_test.clj
```diff
@@ -89,3 +96,207 @@
...
+              (doseq [^ConsumerRecord r (take remaining recs)]
+                (.send producer (ProducerRecord. dest-topic (.key r) (ByteBuffer/wrap (.value r)))))
+              (when (< (count recs) remaining)
+                (recur (- remaining (count recs)))))))
+        (.flush producer))
+      (finally
+        (.close consumer)
+        (.close producer)))))
+
+  (t/deftest ^:integration test-stale-log-recovery
```
jarohen, line 191: good to test this here - I'd guess that this is probably easier written against the memory/local logs?

might also be easier to test this using the existing XT implementations - e.g. against our `Log` interface - then you can test all three logs using the same test infra?

core/src/main/kotlin/xtdb/indexer/LogProcessor.kt
```diff
@@ -36,7 +37,34 @@ class LogProcessor(
...
+        return (txId shr 48).toInt()
+    }
+
+    private fun txIdToOffset(txId: Long): Long {
```
jarohen: let's see a few unit tests for these? just a round-trip should suffice, checking the encoding and decoding?

core/src/main/clojure/xtdb/log.clj
```diff
@@ -247,8 +251,53 @@
                        :kafka :xtdb.kafka/log)
                      opts))
 
-(defmethod ig/init-key :xtdb/log [_ ^Log$Factory factory]
-  (.openLog factory))
+(defmethod ig/prep-key :xtdb/log [_ factory]
+  {:block-cat (ig/ref :xtdb/block-catalog)
+   :factory factory})
+
+(defn tx-id->offset [^long tx-id]
```
jarohen: try to re-use the Kotlin one here, save us the risk of the two implementations being out of sync

jarohen (changes requested):
Thanks @danmason - there was evidently a lot to think about with this change and demonstrating good attention to detail :clap:

have a series of suggestions below, give me a shout if/when you'd like to run through them :slightly_smiling_face:

src/test/clojure/xtdb/kafka_test.clj
danmason: Perhaps instead of the node level for this - could essentially unit test validate-offsets using the Log itself?

danmason: Perhaps do end-to-end with memory log instead? Or create a new log implementation (ie, SimulatedObjectStore type stuff)

danmason:
As a note here - `status`, `submit-tx` and `execute-tx` still act as if they are talking in terms of txIds, whereas they are currently passing around offsets. We will need to fix this prior to merging this.

danmason:
Noting down some thoughts on the api passing around offsets and the UX of MessageId in general:

# TxIds on Edges of the System

## Current State

### Log Talks in Terms of Offsets Only

- The Log tracks its offsets.
- `appendMessage` on the Log returns the submitted offset.
- As such, `log/submit-tx` returns offsets from `appendMessage`. The impl's `submit-tx` returns these offsets.
- `execute-tx` calls `submit-tx`, and then `LogProcessor.awaitAsync` with the offset.

### Rest of the System Uses MessageIds

- MessageIds are created and managed by the `LogProcessor`.
- MessageIds are `Long` values, which are currently analogous to offsets.
- The `LogWatcher` is notified with these MessageIds.
- `LogProcessor` functions (e.g., `awaitAsync`) operate in terms of MessageIds.
- `await-tx` calls down to `awaitAsync` and expects MessageIds.
- The Indexer receives and uses MessageIds from the `LogProcessor`.
- Node status / health checks also talk in MessageIds.
- `SKIP_TXES` also operates on MessageIds.
- In all these places, the user will get back or work with a MessageId.

### As We Introduce Epochs

- MessageIds with epoch `= 0` behave as they do today — users can read them as raw offsets (e.g., in an `IngestionStopped` message, the user can infer the exact offset that failed).
- MessageIds with epoch `≥ 1` will encode the epoch into a large `Long` value.
  - This makes it **non-obvious** what the corresponding offset is.
  - We'll need to ensure **all subsystems** talking in MessageIds are actually **receiving MessageIds**, not raw offsets.

### So, Issues

- We're calling down to the Log with raw offsets — and we must preserve this, as the Log should remain offset-native.
- But we need to reassess what APIs like `submit-tx`, `execute-tx`, and `await-tx` return/use:
  - Should they return and use `MessageIds`?
  - Especially important as epoch-aware MessageIds start diverging from raw offsets.

### Surface-Level UX Problem

- Consider a failure scenario:  
  `"Ingestion stopped: error processing log record at id <MessageId>"`
- With epoch `= 0`, this is readable.
- With epoch `> 0`, this becomes opaque — a large, unintuitive number.
- We still want to return `MessageIds` to the user, but they must be:
  - **Human-friendly**
  - **Inspectable/debuggable**
  - Ideally: `MessageId(epoch=3, offset=1234)` or similar formatting.
- Open question on whether SKIP_TXES becomes SKIP_OFFSETS instead.

danmason:
Following up on the above:
- Spent some time clarifying a few areas of the code where we were passing MessageId
- Ensured **submit-tx** on the API returns an epoch encoded message id.
- Ensured the LogFlusher uses epoch encoded message ids.

docs/src/content/docs/ops/backup-and-restore/overview.adoc
```diff
@@ -0,0 +1,54 @@
...
+| Component | Risk if Lost
+
+| Log
+| All transactions on the log since the last indexed offset would be lost.
+
+| Storage Module
+| Without a backup or infinite retention on the log, all indexed data would be lost.
+
+| XTDB compute nodes
+| All in-memory state would be lost, but the contents of the log and storage module remain intact.
```
jarohen: Let's go a bit more reassuring here - something along the lines that the compute nodes only contain transient/cache data which is replicated durably elsewhere, and so compute nodes can come and go without data loss?

docs/src/content/docs/ops/config/log.adoc
```diff
@@ -55,3 +55,29 @@ A multi-node persistent log implementation that uses a remote service to store t
...
+log: !<LogType>
+  epoch: <new-epoch>
+----
+
+Where:
+
+* `<LogType>` is the chosen log implementation (e.g., `!Kafka`, `!Local`).
+* `<new-epoch>` is a positive integer greater than the previous epoch.
+
+All nodes within the same cluster **must** use an identical epoch value at startup.
```
jarohen, line 83: could add the process here for bumping an epoch? i.e. shut everything down, update config, turn nodes back on again

jarohen (approved):
thanks @danmason :clap: :pray: minor changes, then good to go without further review :rocket:

docs/src/content/docs/ops/backup-and-restore/out-of-sync-log.adoc
```diff
@@ -0,0 +1,62 @@
...
+XTDB will skip offset validation and begin writing to the log under the new epoch.
+
+. *(Optional) Re-submit any lost transactions* +
+If you are aware of any missing transactions, manually reissue them or restore them via your application’s recovery layer.
+
+== Notes & Warnings
+
+* Always take a backup of storage before making epoch changes.
+* The new epoch marks a clean slate for the log but does *not* delete any existing storage files.
+* All nodes must use the same epoch value. Mismatched epochs will cause startup failures.
```
refset, line 64: I suggest we still need to spell this out (per my earlier comment): "Beginning a new epoch will make the recovery of unindexed transactions from any previous epoch nearly impossible. Therefore, it is recommended to only increment the epoch after you are confident that the original log is irrecoverable and you understand the potential consequences. "

refset, line 62: I suggest we still need to spell this out (per my earlier comment): "Beginning a new epoch will make the recovery of unindexed transactions from any previous epoch nearly impossible. Therefore, it is recommended to only increment the epoch after you are confident that the original log is irrecoverable and you understand the potential consequences. "

docs/src/content/docs/ops/config/log.adoc
refset: Unrelated suggestion: `All nodes within the same cluster` => `All nodes connected to the same object store` ('cluster' is a bit ambiguous)

PR #4308, opened 2025-03-27
Bitemporal resolution for events in the same transaction when compacting L0 -> L1

### LINKED ISSUE #4303: Temporal resolution for entities overriden in the same transaction
Previously when an entity got overridden or partially superseded in valid-time in the same transaction, we could rely on there being some deterministic ordering of the transaction parts. With the recency split on file level, the events might have a different order in the live index than when they are sitting in multiple recency files. The following tests illustrate the issue:

(To make the following deterministically fail, one might has to tweak the order in which files are scanned in `scan`.)
```clj
(t/deftest test-same-entity-same-transaction
  ;; depends on historical coming before historical in scan
  (t/testing "same st, vt"
    (with-open [node (xtn/start-node)]
      (xt/execute-tx node [[:put-docs :docs {:xt/id 1, :version 1} {:xt/id 1, :vervsion 2}]])

      (t/is (= [{:xt/id 1, :vervsion 2}] (xt/q node "SELECT * FROM docs")))

      (tu/finish-block! node)
      (c/compact-all! node)

      (t/is (= [{:xt/id 1, :vervsion 2}] (xt/q node "SELECT * FROM docs")))))

  ;; depends on current coming before historical
  (t/testing "overridden in valid-time"
    (with-open [node (xtn/start-node (merge tu/*node-opts* {:log [:in-memory {:instant-src (tu/->mock-clock (tu/->instants :year))}]}))]
      (xt/execute-tx node [[:put-docs {:into :docs :valid-from #inst "2023" :valid-to #inst "2024"}
                            {:xt/id 1, :version 1}]
                           [:put-docs {:into :docs :valid-from #inst "2022" :valid-to #inst "2025"}
                            {:xt/id 1, :version 2}]])

      (t/is (= [{:xt/id 1, :version 2}] (xt/q node "SELECT *, FROM docs FOR ALL VALID_TIME")))

      (tu/finish-block! node)

      (c/compact-all! node)

      ;; v2 split by v1
      (t/is (= [{:xt/id 1, :version 2}] (xt/q node "SELECT * FROM docs FOR ALL VALID_TIME")))))


  ;; depends on current coming before historical
  (t/testing "split in valid-time"
    (with-open [node (xtn/start-node {:log [:in-memory {:instant-src (tu/->mock-clock (tu/->instants :year))}]})]
      (xt/execute-tx node [[:put-docs {:into :docs :valid-from #inst "2022" :valid-to #inst "2025"}
                            {:xt/id 1, :version 2}]
                           [:put-docs {:into :docs :valid-from #inst "2023" :valid-to #inst "2024"}
                            {:xt/id 1, :version 1}]])

      (t/is (= [{:xt/id 1, :version 1} {:xt/id 1, :version 2} {:xt/id 1, :version 2}]
               (xt/q node "SELECT * FROM docs FOR ALL VALID_TIME")))

      (tu/finish-block! node)

      (c/compact-all! node)

      ;; only version 2 is visible
      (t/is (= [{:xt/id 1, :version 1} {:xt/id 1, :version 2} {:xt/id 1, :version 2}]
               (xt/q node "SELECT * FROM docs FOR ALL VALID_TIME")))))

  (t/testing "overlapping valid-time"
    (with-open [node (xtn/start-node {:log [:in-memory {:instant-src (tu/->mock-clock (tu/->instants :year))}]})]
      (xt/execute-tx node [[:put-docs {:into :docs :valid-from #inst "2022" :valid-to #inst "2025"}
                            {:xt/id 1, :version 2}]
                           [:put-docs {:into :docs :valid-from #inst "2023" :valid-to #inst "2026"}
                            {:xt/id 1, :version 1}]])

      ;; v2 cut of by v1
      (t/is (= nil
               (xt/q node "SELECT *, _valid_from, _valid_to FROM docs FOR ALL VALID_TIME")))

      (tu/finish-block! node)

      (c/compact-all! node)

      ;; v1 cut of by v2
      (t/is (= nil
               (xt/q node "SELECT *, _valid_from, _valid_to FROM docs FOR ALL VALID_TIME"))))))
```

FiV0:
As per #4303, to guarantee we always do the same resolution for events in the same transaction, we need to do the bi-temporal resolution of these events when they get potentially moved to different files (different recencies). 

In L0 events sit in a certain order in the same file and that order (later events in the tx come earlier in the bitemporal resolution) determines the bitemporal resolution. When events get moved to higher levels the order is no longer guaranteed as events with the same system time get moved to different files. To keep that order across all levels we do bi-temporal resolution for events in the same transaction when they move from L0 to L1. This means that we might have more events then were originally in the transaction in L1. The increase in events should only happen when there are overlapping valid times for the same entity id in the same transaction. We are doing this extra bitemporal resolution when we are loading a page from L0.

This PR also updates TPCH 0.001 results as in 0.001 there are some overwrites in the SF 0.001 transactions. We were previously storing the "wrong" result which can be verified on main by comparing non compacted TPCH 0.001 against compacted TPCH 0.001.

### DIFF
--- a/core/src/main/kotlin/xtdb/compactor/SegmentMerge.kt
+++ b/core/src/main/kotlin/xtdb/compactor/SegmentMerge.kt
@@ -20,6 +20,8 @@ import java.nio.file.Path
 import java.time.LocalDate
 import java.util.*
 import java.util.function.Predicate
+import kotlin.Long.Companion.MAX_VALUE
+import kotlin.Long.Companion.MIN_VALUE
 import kotlin.io.path.deleteExisting
 import kotlin.math.min
 import kotlin.Long.Companion.MAX_VALUE as MAX_LONG
@@ -37,6 +39,67 @@ private fun ByteArray.toPathPredicate() =
 
 private fun <N : HashTrie.Node<N>, L : N> MergePlanNode<N, L>.loadDataPage(): RelationReader? =
     segment.dataRel?.loadPage(node)
+
+/**
+ * A function to do bitemporal resolution for events with the same system-time (same transaction). See #4303
+ */
+fun resolveSameSystemTimeEvents(al: BufferAllocator, dataReader: RelationReader, path: ByteArray = byteArrayOf()) : Relation {
+    val isValidPtr = ArrowBufPointer()
+    val startIidPtr = ArrowBufPointer()
+    val curIidPtr = ArrowBufPointer()
+    var curSystemFrom : Long
+
+    val relWriter = Relation(al, dataReader.schema)
+    val iidVec = dataReader["_iid"].rowCopier(relWriter["_iid"])
+    val sysFromVec= dataReader["_system_from"].rowCopier(relWriter["_system_from"])
+    val validFromVec = relWriter["_valid_from"]
+    val validToVec = relWriter["_valid_to"]
+    val opCopier = dataReader["op"].rowCopier(relWriter["op"])
+
+    val evPtr = EventRowPointer(dataReader, path)
+    val polygonCalculator = PolygonCalculator()
+
+    var seenErase = false
+
+    while (evPtr.isValid(isValidPtr, path)) {
+        evPtr.getIidPointer(startIidPtr)
+        curSystemFrom = evPtr.systemFrom
+
+        while(evPtr.isValid(isValidPtr, path) && evPtr.systemFrom == curSystemFrom && startIidPtr == evPtr.getIidPointer(curIidPtr)) {
+
+            when (val polygon = polygonCalculator.calculate(evPtr)) {
+                // here we are only taking care of an erase that happens within the same transaction
+                null -> {
+                    if (!seenErase) {
+                        iidVec.copyRow(evPtr.index)
+                        sysFromVec.copyRow(evPtr.index)
+                        validFromVec.writeLong(MIN_VALUE)
+                        validToVec.writeLong(MAX_VALUE)
+                        opCopier.copyRow(evPtr.index)
+                        relWriter.endRow()
+                    }
+                    seenErase = true
+                }
+                else -> {
+                    repeat(polygon.validTimeRangeCount) { i ->
+                        if (polygon.getSystemTo(i) > curSystemFrom) {
+                            iidVec.copyRow(evPtr.index)
+                            sysFromVec.copyRow(evPtr.index)
+                            validFromVec.writeLong(polygon.getValidFrom(i))
+                            validToVec.writeLong(polygon.getValidTo(i))
+                            opCopier.copyRow(evPtr.index)
+                            relWriter.endRow()
+                        }
+                    }
+                }
+            }
+            evPtr.nextIndex()
+        }
+        polygonCalculator.reset()
+        seenErase = false
+    }
+    return relWriter
+}
 
 internal class SegmentMerge(private val al: BufferAllocator) : AutoCloseable {
 
--- a/core/src/main/kotlin/xtdb/trie/DataRel.kt
+++ b/core/src/main/kotlin/xtdb/trie/DataRel.kt
@@ -5,6 +5,7 @@ import org.apache.arrow.vector.types.pojo.Schema
 import xtdb.BufferPool
 import xtdb.arrow.Relation
 import xtdb.arrow.RelationReader
+import xtdb.compactor.resolveSameSystemTimeEvents
 import xtdb.trie.Trie.dataFilePath
 import xtdb.util.closeAll
 import java.nio.file.Path
@@ -24,8 +25,9 @@ interface DataRel<L> : AutoCloseable {
             mutableListOf<Arrow>().also { res ->
                 try {
                     trieKeys.forEach { trieKey ->
+                        val parsedKey = Trie.parseKey(trieKey)
                         val dataFile = tableName.dataFilePath(trieKey)
-                        res.add(Arrow(al, bp, dataFile, bp.getFooter(dataFile).schema))
+                        res.add(Arrow(al, bp, dataFile, bp.getFooter(dataFile).schema, parsedKey.level))
                     }
                 } catch (e: Throwable) {
                     res.closeAll()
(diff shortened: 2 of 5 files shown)


src/test/clojure/xtdb/compactor_test.clj
```diff
@@ -630,3 +630,22 @@
   (c/compact-all! tu/*node*)
 
   (t/is (= [] (xt/q tu/*node* "SELECT _id FROM foo"))))
+
+
+(t/deftest test-same-entity-same-transaction-4303
```
FiV0, line 645: @jarohen  The bad thing about this test is that I needed to add a `reverse` to `trie-catalog/current-tries` to make most of these fail in the non fixed version. So it's no really testing much in this case as the natural order of the `current-tries` gives the same splitting behavior.

core/src/main/kotlin/xtdb/compactor/SegmentMerge.kt
```diff
@@ -76,48 +78,120 @@ internal class SegmentMerge(private val al: BufferAllocator) : AutoCloseable {
 
     private val outWriters = OutWriters(al)
 
-    private fun MergePlanTask.merge(outWriter: OutWriter, pathFilter: ByteArray?) {
+    private fun resolveSameSystemTimeEvents(dataReader: RelationReader, path: ByteArray) : RelationReader {
```
FiV0: This does a full copy of the data which adds of course memory pressure.

core/src/main/kotlin/xtdb/compactor/SegmentMerge.kt
```diff
@@ -135,16 +209,16 @@ internal class SegmentMerge(private val al: BufferAllocator) : AutoCloseable {
     ): Results {
         val schema = logDataRelSchema(segments.map { it.dataRel!!.schema })
 
-        val outWriter = when(recencyPartitioning) {
-            RecencyPartitioning.Partition -> outWriters.PartitionedOutWriter(schema, recencyPartition)
-            is RecencyPartitioning.Preserve -> outWriters.OutRel(schema, recency = recencyPartitioning.recency)
+        val (outWriter, bitemporalResolution)= when(recencyPartitioning) {
```
FiV0: This assumes `RecencyPartitioning.Partition` is only used for L1, I could make the `bitemporalResolution` option more explicit, by passing `mergeSegments` if you prefer that.

FiV0:
@jarohen Left 3 comments with things to discuss. Otherwise good for a first review or chat through.

core/src/main/kotlin/xtdb/compactor/SegmentMerge.kt
jarohen: I can't make a call on this one without knowing _how much_ memory pressure we're talking - is it keeping 2 pages in memory rather than one? or do we now need to bring the whole file into memory again?

src/test/clojure/xtdb/compactor_test.clj
jarohen: I'm assuming you've cherry-picked this against your TPC-H failure on your other branch? if so, once your other branch is merged too we'll have a regression test for this one

core/src/main/kotlin/xtdb/compactor/SegmentMerge.kt
jarohen: yeah, let's do that - while it might just work this way it's probably better to be explicit, save the next person working in this area missing the assumption

core/src/main/kotlin/xtdb/compactor/SegmentMerge.kt
```diff
@@ -76,48 +78,120 @@ internal class SegmentMerge(private val al: BufferAllocator) : AutoCloseable {
...
         val isValidPtr = ArrowBufPointer()
 
         val mpNodes = mpNodes
         val path = path.let { if (pathFilter == null || it.size > pathFilter.size) it else pathFilter }
         val mergeQueue = PriorityQueue(Comparator.comparing(QueueElem::evPtr, EventRowPointer.comparator()))
 
-        for (dataReader in mpNodes.mapNotNull { it.loadDataPage() }) {
-            val evPtr = EventRowPointer(dataReader, path)
-            val rowCopier = outWriter.rowCopier(dataReader)
+        val dataReaders = mutableListOf<RelationReader>()
```
jarohen: it seems that this `bitemporalResolution` flag is quite intertwined in the below logic, to the extent I found it relatively onerous to figure out which parts happen during which mode. did you consider duplicating it? if so, how'd it go?

jarohen (changes requested):
thanks @FiV0 :pray: fwiw having seen your implementation I much prefer this approach - it might not be the easiest of changes but it's very much isolated to this one part of the system; the rest of the system continues unaware.

couple of questions below to address, then give me a shout and I'll take another look through :slightly_smiling_face:

core/src/main/kotlin/xtdb/trie/DataRel.kt
```diff
@@ -39,14 +41,24 @@ interface DataRel<L> : AutoCloseable {
...
 
         override fun loadPage(leaf: ArrowHashTrie.Leaf): RelationReader =
             bp.getRecordBatch(dataFile, leaf.dataPageIndex).use { rb ->
-                Relation.fromRecordBatch(al, schema, rb).also { relsToClose.add(it) }
+                Relation.fromRecordBatch(al, schema, rb).let { standardRel ->
+                    if (level == 0L) {
+                        resolveSameSystemTimeEvents(al, standardRel).also {
+                            standardRel.close()
+                            relsToClose.add(it)
+                        }
```
FiV0, line 56: @mbutlerw This extra bitemporal resolution is actually only needed when compacting and not when scanning. I am now wondering if it should be skipped in scanning.

jarohen: unless this would ease the complexity in the code you've already written, I wouldn't worry about this - in practice, we're not going to be scanning from L0s for long before someone compacts them into L1

jarohen (approved):
thanks @FiV0 - lot to consider here but you seem to have handled it well :clap: 

I'm easy either way on whether you move the code to the compactor or not, your call :slightly_smiling_face:

core/src/main/kotlin/xtdb/trie/DataRel.kt
FiV0: Just checked this assumption again. `MergePlanPage` is used in scan.clj instead of `DataRel` so actually not issue for now. Maybe something we can come back to once we have completely switched to the new `xtdb.arrow` implementation and the merging happens the same way in both scanning and compacting.

PR #4310, opened 2025-03-27
Add Microsecond Precision Intervals (MDM) - 4187

### LINKED ISSUE #4187: Remove Default Nanosecond support
I think its worth reviewing the current support for nanosecond precision types (as of #4188):

- At a vector/arrow level both timestamp and interval have nanosecond support. 
- In the case of timestamp there is explicit support for `micro` precision instead. This is not true for Intervals, which are explicitly `MONTH_DAY_NANO`
- The comparable pg types for both these only support `micro` precision. Therefore we need to truncate any nano values.
- The build in temporal columns (sys/valid_to/from) are in `micro`.
- EE converts all java timestamp objects (ZoneDT/Instant etc.) into `micros`
- MDN intervals are handled as `nanos` throughout
- Expressions that involve different precision units return values with the greatest precision. For example if you take a micro ts and a mdn interval we return a nano ts. Even though the user (especially via pgwire) isn't really aware of nanos.

The result is that we have a system that mostly deals (and assumes) timestamps are `micro` but that we always need to be aware of the potential for `nano`. While also meaning that because we conflate and truncate both concepts in pgwire, actual nanos are not available to pgwire users in any way. 

Proposal:

- [x]  Add a `MONTH_DAY_MICRO` interval type, which explicitly only has microsecond precision.
- [x]  Ensure all timestamp creation/handling uses `micro` units.
- [ ] Add custom types in pgwire for nanosecond ts/intervals and SQL syntax for their creation.

NOTE: The reason for explicit syntax is so that different precision literals don't create different types.

mbutlerw:
Adds MonthDayMicro interval on top off mdn and uses it as the default. Also adds some, but not complete support for casting between the durations/intervals etc. of various precisions as code already existed for these and wanted to avoid removing features. That being said, I think 99.9% of people are using micro durations and intervals, so a case could be made for simplifying the logic/API here.

Commit also chooses to implicitly assume `INTERVAL` SQL literals are in `micro` rather than inferring from context. This matches current timestamp behaviour, as we map ZDT -> tstz micro, however this happens implicitly, truncating any nanos, whereas we have a specific type for MDM and so we throw during planning.

 TODO:
 
- [x] Fix MDM literals allowing nano precision.
- [x] More tests around mdn now its not the default
- [x] More testing around the various precision based casting which would cause a change in interval type
- [x] See if we can clean up the usage of PD, so its limited to internal runtime EE rep, never infer a type from it.

### DIFF
--- a/api/src/main/clojure/xtdb/time.clj
+++ b/api/src/main/clojure/xtdb/time.clj
@@ -5,6 +5,7 @@
            [java.time.format DateTimeParseException]
            java.time.temporal.ChronoUnit
            (java.util Date)
+           (org.apache.arrow.vector PeriodDuration)
            xtdb.time.Time))
 
 (defn ->duration [d]
@@ -123,3 +124,15 @@
                               {::err/message (str "invalid timestamp: " (ex-message e))
                                :timestamp ts-str}
                               e)))))
+
+(defn alter-duration-precision ^Duration [^long precision ^Duration duration]
+  (if (= precision 0)
+    (.withNanos duration 0)
+    (.withNanos duration (let [nanos (.getNano duration)
+                               factor (Math/pow 10 (- 9 precision))]
+                           (* (Math/floor (/ nanos factor)) factor)))))
+
+(defn alter-md*-interval-precision ^PeriodDuration [^long precision ^PeriodDuration pd]
+  (PeriodDuration.
+   (.getPeriod pd)
+   (alter-duration-precision precision (.getDuration pd))))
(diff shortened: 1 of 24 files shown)


api/src/main/clojure/xtdb/time.clj
```diff
@@ -123,3 +124,15 @@
...
                               e)))))
+
+(defn alter-duration-precision ^Duration [^long precision ^Duration duration]
+  (if (= precision 0)
+    (.withNanos duration 0)
+    (.withNanos duration (let [nanos (.getNano duration)
+                               factor (Math/pow 10 (- 9 precision))]
+                           (* (Math/floor (/ nanos factor)) factor)))))
+
+(defn alter-md*-interval-precision ^Duration [^long precision ^PeriodDuration pd]
```
jarohen, line 135: type hint looks incorrect?

jarohen: type hint looks incorrect?

jarohen (approved):
thanks @mbutlerw - as discussed, to fix up the TODOs in the PR header, then good to merge :pray:

api/src/main/clojure/xtdb/time.clj
mbutlerw: Good spot :+1:

PR #4334, opened 2025-04-02
Add support for round-tripping decimals

### LINKED ISSUE #4330: Round-trip decimals through XTDB (incl. through pgwire)
* Decimal types in Arrow have three parameters: precision, scale and bit-width.
  * Our leg names need to reflect this - when asked for a vector from a field with these arguments, we should respect them.
  * When choosing an Arrow type for a JVM object, we should respect the scale of the incoming BigDecimal/BigInteger, and we should choose a sufficiently high precision to store the BigDecimal without loss of precision.
    * Arrow supports bit-widths of 32, 64, 128, 256. Arrow Java only supports 128 and 256, so let's limit ourselves to these two for now.
    * precision ≈ bit-width / 4, so if a number comes in with precision 25, we should use precision 32 and bit-width 128, in order to reduce the number of DUV legs for different decimal types.
    * we use the scale of the incoming numbers directly - e.g. 10.02 will have scale 2; 84.523 will have scale 3 - and we accept that we will have multiple DUV legs in this case (on the assumption that users won't have many distinct scales in one column).
* BigDecimals/BigIntegers should be converted to doubles when hashing/including in min-max metadata, so that they're comparable to the equivalent fixed-width integers.
* Bonus: SQL literal for big-decimals, like Clojure - e.g. `100.24M`

FiV0:
This adds support for decimal round-tripping. We now support `[:decimal #{32 64} scale #{128 256}]`. Where the first value is the precision and the last the bit-width. If not specified by the user we default to one of the two and only using  the larger when needed. This also adds the `xtdb.arrow.DecimalVector` implementation.

### DIFF
--- a/core/src/main/clojure/xtdb/types.clj
+++ b/core/src/main/clojure/xtdb/types.clj
@@ -20,7 +20,8 @@
            (xtdb JsonSerde Types)
            (xtdb.types IntervalMonthDayMicro IntervalMonthDayNano ZonedDateTimeRange)
            [xtdb.vector IVectorReader]
-           (xtdb.vector.extensions KeywordType RegClassType RegProcType SetType TransitType TsTzRangeType UriType UuidType IntervalMDMType)))
+           (xtdb.vector.extensions KeywordType RegClassType RegProcType SetType TransitType TsTzRangeType UriType UuidType IntervalMDMType)
+           (xtdb.pg.codec NumericBin)))
 
 (set! *unchecked-math* :warn-on-boxed)
 
@@ -1011,6 +1012,20 @@
                                 (.array bb)))
               :write-text (fn [_env ^IVectorReader rdr idx]
                             (utf8 (.getDouble rdr idx)))})
+
+   :numeric {:typname "numeric"
+             :col-type :decimal
+             :oid 1700
+             :typlen -1
+             :typsend "numeric_send"
+             :typreceive "numeric_recv"
+             :read-binary (fn [_env ba] (NumericBin/decode (ByteBuffer/wrap ba)))
+             :read-text (fn [_env ba] (-> ba read-utf8 bigdec))
+             :write-binary (fn [_env ^IVectorReader rdr idx]
+                             (.array (NumericBin/encode (.getObject rdr idx))))
+             :write-text (fn [_env ^IVectorReader rdr idx]
+                           (utf8 (.getObject rdr idx)))}
+
    :uuid (let [typlen 16]
            {:typname "uuid"
             :col-type :uuid
@@ -1457,6 +1472,7 @@
    #_#_:i8 :bit
    :f64 :float8
    :f32 :float4
+   :decimal :numeric
    :uuid :uuid
    :bool :boolean
    :null :text
@@ -1482,7 +1498,9 @@
        (cond-> col-type
          ;; ignore TZ
          (= (col-type-head col-type) :timestamp-tz)
-         (subvec 0 2))
+         (subvec 0 2)
+         (= (col-type-head col-type) :decimal)
+         col-type-head)
 
        (or fallback-pg-type :json)))
 
--- a/core/src/main/clojure/xtdb/types.clj
+++ b/core/src/main/clojure/xtdb/types.clj
@@ -95,7 +95,9 @@
   ArrowType$Binary (<-arrow-type [_] :varbinary)
   ArrowType$FixedSizeBinary (<-arrow-type [arrow-type] [:fixed-size-binary (.getByteWidth arrow-type)])
 
-  ArrowType$Decimal (<-arrow-type [_] :decimal)
+  ArrowType$Decimal
+  (<-arrow-type [arrow-type]
+    [:decimal (.getPrecision arrow-type) (.getScale arrow-type) (.getBitWidth arrow-type)])
 
   ArrowType$FixedSizeList (<-arrow-type [arrow-type] [:fixed-size-list (.getListSize arrow-type)])
 
@@ -139,9 +141,6 @@
     :f32 (.getType Types$MinorType/FLOAT4)
     :f64 (.getType Types$MinorType/FLOAT8)
 
-    ;; HACK we should support parameterised decimals here
-    :decimal (ArrowType$Decimal/createDecimal 38 19 (int 128))
-
     :utf8 (.getType Types$MinorType/VARCHAR)
     :varbinary (.getType Types$MinorType/VARBINARY)
 
@@ -170,6 +169,9 @@
 
       :date (let [[_ date-unit] col-type]
               (ArrowType$Date. (kw->date-unit date-unit)))
+
+      :decimal (let [[_ precision scale bitwidth] col-type]
+                 (ArrowType$Decimal/createDecimal precision scale (int bitwidth)))
 
       :timestamp-tz (let [[_ time-unit tz] col-type]
                       (ArrowType$Timestamp. (kw->time-unit time-unit) tz))
@@ -466,11 +468,11 @@
                      :utf8 Types$MinorType/VARCHAR, :varbinary Types$MinorType/VARBINARY)]
     (->field col-name (.getType minor-type) (or nullable? (= col-type :null)))))
 
-(defmethod col-type->field* :decimal [col-name nullable? _col-type]
-  ;; TODO decide on how deal with precision that is out of this range but still fits into
-  ;; a 128 bit decimal
-  (->field col-name (ArrowType$Decimal/createDecimal 38 19 (int 128)) nullable?))
+(defmethod col-type->field* :decimal [col-name nullable? [_ precision scale bit-width]]
+  (->field col-name (ArrowType$Decimal/createDecimal precision scale bit-width) nullable?))
 
+(defmethod col-type->field-name :decimal [[type-head precision scale bit-width :as v]]
+  (str (name type-head) "-" precision "-" scale "-" bit-width))
 
 (defmethod col-type->field* :tstz-range [col-name nullable? _col-type]
   (->field col-name TsTzRangeType/INSTANCE nullable?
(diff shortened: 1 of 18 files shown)


core/src/main/kotlin/xtdb/arrow/DecimalVector.kt
```diff
@@ -0,0 +1,52 @@
...
+class DecimalVector private constructor(
+    override var name: String, override var nullable: Boolean, override var valueCount: Int,
+    override val validityBuffer: ExtensibleBuffer, override val dataBuffer: ExtensibleBuffer,
+    val precision: Int, val scale: Int, val bitWidth: Int,
+) : FixedWidthVector(), MetadataFlavour.Number {
+
+    private val BIT_WITDTHS = setOf(32, 64, 128, 256)
+
+    init {
+       require(bitWidth in BIT_WITDTHS) { "Invalid bit width for DecimalVector: $bitWidth" }
```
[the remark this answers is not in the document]
FiV0: Good bot.

core/src/main/kotlin/xtdb/Types.kt
```diff
@@ -112,8 +113,17 @@ fun valueToArrowType(obj: Any?) = when (obj) {
...
-    // HACK we should support parameterised decimals here
-    is BigDecimal -> ArrowType.Decimal(38, 19, 128)
+    is BigDecimal -> {
+        val precision = when (obj.precision()) {
+            // Java Arrow only supports 128 and 256 bit widths
+//            in 0..7 -> 32
+//            in 8..16 -> 64
+            in 0..32 -> 32
+            in 33..64 -> 64
+            else -> throw IllegalArgumentException("unsupported precision: ${obj.precision()}")
```
FiV0, line 121: @jarohen FYI. There is a range of values we are not representing here
```clj
  (def bd 115792089237316195423570985008687907853269984665640564039457584007913129639935M)
  
  (-> (.unscaledValue bd) (.bitLength))
  ;;    256

  (count (str bd))
  ;; => 78
```

jarohen: I think I'll cope with that :sweat_smile: IIUC it's an Arrow restriction? if so, not a lot we're likely to be able to do about it unless we create extension vectors.

Let's wait for the first user with those precision requirements before implementing it :)

core/src/main/clojure/xtdb/types.clj
```diff
@@ -95,7 +95,8 @@
   ArrowType$Binary (<-arrow-type [_] :varbinary)
   ArrowType$FixedSizeBinary (<-arrow-type [arrow-type] [:fixed-size-binary (.getByteWidth arrow-type)])
 
-  ArrowType$Decimal (<-arrow-type [_] :decimal)
+  ArrowType$Decimal (<-arrow-type [arrow-type]
+                      [:decimal (.getPrecision arrow-type) (.getScale arrow-type) (.getBitWidth arrow-type)])
```
jarohen: if it's no longer a one-liner, preferred indentation:
```clojure
  ArrowType$Decimal 
  (<-arrow-type [arrow-type]
    [:decimal (.getPrecision arrow-type) (.getScale arrow-type) (.getBitWidth arrow-type)])
```

core/src/main/kotlin/xtdb/Types.kt
```diff
@@ -112,8 +113,17 @@ fun valueToArrowType(obj: Any?) = when (obj) {
     is Float -> MinorType.FLOAT4.type
     is Double -> MinorType.FLOAT8.type
 
-    // HACK we should support parameterised decimals here
-    is BigDecimal -> ArrowType.Decimal(38, 19, 128)
+    is BigDecimal -> {
+        val precision = when (obj.precision()) {
+            // Java Arrow only supports 128 and 256 bit widths
+//            in 0..7 -> 32
```
jarohen: drop commented code? doesn't look correct, would have expected `in 0..8 -> 8`, but also doesn't really add anything for readers

core/src/main/kotlin/xtdb/arrow/DecimalVector.kt
```diff
@@ -0,0 +1,52 @@
...
+            try {
+                dataBuffer.writeBigDecimal(value, byteWidth)
+            } catch (e: UnsupportedOperationException) {
+                throw InvalidWriteObjectException(fieldType, value)
+            }
+            writeNotNull()
+        } else throw InvalidWriteObjectException(fieldType, value)
+    }
+
+    override fun getMetaDouble(idx: Int): Double = (getObject(idx) as BigDecimal).toDouble()
```
jarohen: `getObject0` and then you don't need the cast :)

core/src/main/kotlin/xtdb/vector/FieldVectorWriters.kt
```diff
@@ -229,6 +230,22 @@ private class DecimalVectorWriter(override val vector: DecimalVector) : ScalarVe
     override fun writeValue0(v: ValueReader) = writeObject(v.readObject())
 }
 
+private class Decimal256VectorWriter(override val vector: Decimal256Vector) : ScalarVectorWriter(vector) {
+    @Throws(RuntimeException::class)
```
jarohen: dw about `@Throws`, especially not for runex - wouldn't even declare that in Java

core/src/main/kotlin/xtdb/Types.kt
FiV0: Sorry if I wasn't clear. We could replace the second `in` with `33..77` and make it precision 77 and that would I think "just work" as it fits a `Decimal256Vector`.

core/src/main/kotlin/xtdb/vector/FieldVectorWriters.kt
```diff
@@ -229,6 +230,22 @@ private class DecimalVectorWriter(override val vector: DecimalVector) : ScalarVe
...
+    override fun writeObject0(obj: Any) {
+        if (obj !is BigDecimal) throw InvalidWriteObjectException(field.fieldType, obj)
+        try {
+            vector.setSafe(valueCount++, obj.setScale(vector.scale))
+        } catch (e: Exception) {
+            throw RuntimeException(DECIMAL_ERROR_KEY, e.message, emptyMap<Keyword, Any>(), e)
+        } catch (e: ArithmeticException) {
```
jarohen, line 235: what exceptions does this actually throw? if it's just ArithmeticException, I'd drop Exception; if there are other exceptions, I'd be surprised if you don't need to (and it's idiomatic to, even if it's not a requirement) put more specific exception types above more general types

src/test/clojure/xtdb/expression_test.clj
```diff
@@ -1388,16 +1387,21 @@
...
              (run-test 'greatest (double 3.0) (bigdec 2.0))))
 
-
     ;; TODO decide on behaviour
     ;; out of fixed precision range
     (t/is (thrown? RuntimeException
-                   (run-test '+ (bigdec 1E+19M) (bigdec 1E+19M))))
+                   (run-test '+ (bigdec 1E+32M) (bigdec 1E+64M))))
+
+    ;; FIXME not sure why this doesn't throw
```
jarohen: fixme

core/src/main/kotlin/xtdb/arrow/DecimalVector.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.arrow.vector.types.pojo.ArrowType
+import xtdb.api.query.IKeyFn
+import xtdb.arrow.metadata.MetadataFlavour
+import xtdb.util.Hasher
+import java.math.BigDecimal
+
+class DecimalVector private constructor(
+    override var name: String, override var nullable: Boolean, override var valueCount: Int,
+    override val validityBuffer: ExtensibleBuffer, override val dataBuffer: ExtensibleBuffer,
+    val precision: Int, val scale: Int, val bitWidth: Int,
```
jarohen: I'd probably have the constructor take ArrowType.Decimal, then extract these three in the class body - chances are your caller already has the ArrowType in hand

core/src/main/kotlin/xtdb/arrow/DecimalVector.kt
```diff
@@ -0,0 +1,52 @@
...
+       dataBuffer.readBigDecimal(idx, scale, byteWidth)
+
+    override fun writeObject0(value: Any) {
+        if (value is BigDecimal) {
+            if (value.precision() > precision || value.scale() != scale) {
+                throw InvalidWriteObjectException(fieldType, value)
+            }
+            try {
+                dataBuffer.writeBigDecimal(value, byteWidth)
+            } catch (e: UnsupportedOperationException) {
```
jarohen: why might this throw UOE, sorry?

jarohen (approved):
thanks @FiV0 - with the comments addressed, good to merge without further review :pray:

core/src/main/kotlin/xtdb/arrow/DecimalVector.kt
FiV0: If your decimal doesn't fit your byteWidth.

jarohen: ah, kk - that's one for IllegalArgumentException - UOE is something the code doesn't support (yet); IAE is 'caller messed up'

jarohen: I'd stick that as a `require` in `writeBigDecimal` - that throws IAE if the condition isn't met

core/src/main/kotlin/xtdb/vector/FieldVectorWriters.kt
FiV0: I just copied the `DecimalVectorWriter` so I assume there is some reason we put it.

core/src/main/kotlin/xtdb/arrow/DecimalVector.kt
FiV0: I think this will likely never happen if we properly deal with checking the precision before. It's the `DecimalUtility` that throws this error. I think it can only happen if you create a arrow decimal type where the precision can't fit the bitwidth, but I guess the assumption here is that you then messed up elsewhere.

src/test/clojure/xtdb/expression_test.clj
FiV0: There is lots of stuff to address in the EE, I will fix these FIXME's when I am tackling https://github.com/xtdb/xtdb/issues/4331.

PR #4335, opened 2025-04-02
fix InMemoryLog race condition on subscriber channel closed

tKe:
Contains a few tweaks/fixes:

- Switches to `select` to select over the timeout-and-process-empty vs receive.
- Switches to using `receiveCatching` (via `onReceiveCatching`) to handle channel closure.
- Uses a common supervisor-scope for subscriber jobs, with correct parental association.
- Uses `receiveAsFlow` and `produceIn` operators for connection channel<->flow to ensure proper cancellation handling.
- Uses `shareIn` operator instead of manual MutableSharedFlow creation/management


jarohen (approved):
awesome @tKe, much cleaner - thanks! :pray:

PR #4339, opened 2025-04-03
Using idiomatic XTDB (and also Clojure) in Auctionmark

jarohen:
This PR iterates on our Auctiomark benchmark to write it as an XTDB user would do. Mainly:

* Sane column names (no more `ib_i_b_u_id` et al :sweat_smile:). I've also distinguished between the different `user_id`s, so we now have `buyer-id`, `seller-id`, `bidder-id`, etc. And kebab-case, because it's written in Clojure.
* Removing `item_purchase` in favour of a nullable field on the `item` table
* Inlined `item_images` in favour of a nested list on `item`.
* Using item-id as the PK on the item-max-bid table given they have a 0..1 relationship, saves creating more surrogate keys than we need
* Drop table prefixes in queries (previously these were required in XT SQL)
* Drop table prefixes in column names (e.g. `users.u_this, users.u_that`)
* Needless let-bound names that were then just put into a map (with a suitably named keyword key)
* Remove `custom-state` and `domain-state` in the bench worker, in favour of the individual benchmarks managing their own state (as opposed to trying to prematurely generalise).
* `sut` -> `node` throughout bench.
* a couple of mischievous `core2` references :laughing: 

Strictly speaking this means we aren't comparable with other databases that use this benchmark, but tbh we haven't been doing that anyway - in practice, we're using it more as a regression/optimisation test between XTDB versions.

### DIFF
--- a/api/src/main/clojure/xtdb/api.clj
+++ b/api/src/main/clojure/xtdb/api.clj
@@ -182,7 +182,6 @@
   (^TransactionKey [node, tx-ops tx-opts]
    (xtp/execute-tx node (vec tx-ops) tx-opts)))
 
-#_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
 (defn status
   "Returns the status of this node as a map,
   including details of both the latest submitted and completed tx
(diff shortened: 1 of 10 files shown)


danmason (approved):
Changes looking great :) Fairly sure run-auctionmark is looking just fine, but should be quick enough to verify all works as intended.

PR #4346, opened 2025-04-04
Ignore newer `TriesAdded` messages in beta6

FiV0:
This assures newer `TriesAdded` (storage version 6) messages get ignored in beta 6.
- We use  [UnknownFieldSet](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/UnknownFieldSet.html) to check if the TriesAdded message is of a newer type (i.e. containing fields we don't know yet about).
- Add a test that plays through a beta7 log (on beta6) to assure we are ignoring any newer `TriesAdded` messages.


danmason (approved):
Looking great :)

PR #4352, opened 2025-04-08
Support decimals in pgwire and the EE.

### LINKED ISSUE #4331: Expressions and casts involving BigDecimals
* Expressions as per Postgres - particularly, consider the return types of arithmetic operations containing either big-decs with different scale/precisions params, as well as big-decs vs ints etc.
  * Pick a runtime representation of big-decs within the EE.
  * Postgres (as an addition to the spec, IIRC) allows the user to elide precision and scale in decimal types - check how this then behaves.
* Casts:
  * Casting to/from ints, longs, doubles, strings
  * SQL allows users to cast to big-decs with different precisions/scales - then we return a big-dec with precision 5.
* User-facing documentation

FiV0:
I took the least invasive route to add support. This mainly means:
- The EE is working on standard Java big decimals. The EE now deals specially with the following operations:
   - Unless otherwise specified the precision is 32 if both operands have a precision less than or equal to 32, otherwise a precision of 64 is used.
   - `+` - scale = max of the two operands
   - `-` - scale = max of the two operands
   - `*` - scale = the sum of two operands
   - `/` - scale = `max(6, s1 + s2 + 1)`
        - Decimal devision might create infinite decimal expansion in the result. So we can only try to do something reasonable. Multiple databases like SQL Server and AWS Kinesis Data Analytics use `MAX (6, s1 + p2 +1)`. As we don't have the precision of the divisor at hand, this is not an option. We do something similar but only take the scales into account. We always set a minimum scale of 6 for devision.
        - The standard says:
        If the operator is division and the approximate mathematical result of the operation represented with the precision and scale of the result data type loses one or more leading significant digits after rounding or truncating if necessary, then an exception condition is raised: data exception — numeric value out of range. The choice of whether to round or truncate is implementation-defined.
        - DuckDB actually goes to floats in this case.
   - We use the standard clojure operands like `<`, `<=` etc... and by extension `greatest` and `least` for decimals. 
       - This means an operation on columns with multiple decimal legs (union) will also return a union for the latter ones (`greatest` and `least`).  
   - Operations between a decimal type and a non decimal type follow the standard widening hierarchy mechanics. 
- A Cast like `::DECIMAL(5,2)` gets honoured on disk.
- A Cast like `::DECIMAL(5)` gets honoured on disk with a scale of 0.
- A Cast like `::DECIMAL` gets defaulted to `[:decimal 64 9 256]` (this differs from the spec and likely postgres)
     - The exception to this is if the value you are casting from is already a decimal. In that case the `::DECIMAL` cast is a no-op.
     - The main reason for this quirck is so that a cast (which is actually a type annotation) in the following `["INSERT INTO table RECORDS {_id: 1, data: ?::decimal}" 1.111111111111M]` does not override the scale getting passed in via pgwire.
- There is one kind of "bad case" in the EE for decimal operations. If you add two large decimals at the edges of 32 precision types, you might create a decimal that needs a 64 precision vector. In this case the EE will throw although we support 64 precision in general. 
   
Todo:
- [x] Cleanup the throws to idiomatic EE errors

### DIFF
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -11,6 +11,7 @@
             [xtdb.vector.writer :as vw])
   (:import (clojure.lang IPersistentMap Keyword MapEntry)
            (java.lang NumberFormatException)
+           (java.math RoundingMode)
            [java.net URI]
            (java.nio ByteBuffer)
            (java.nio.charset StandardCharsets)
@@ -240,6 +241,33 @@
 (defmethod codegen-cast [:num :num] [{:keys [target-type]}]
   {:return-type target-type
    :->call-code #(do `(~(type->cast target-type) ~@%))})
+
+(defn- precision->bit-width [^long p]
+  (cond (<= p 32) (int 128)
+        (<= p 64) (int 256)
+        :else (throw (err/illegal-arg :xtdb.expression/unsupported-precision
+                                      {::err/message (format "Unsupported precision: %d" p)}))))
+
+(defmethod codegen-cast [:num :decimal] [{{:keys [precision scale]} :cast-opts}]
+  (let [scale (if precision (or scale 0) 9)
+        precision (or precision 64)]
+    {:return-type [:decimal precision scale (precision->bit-width precision)]
+     :->call-code #(do `(.setScale (bigdec ~@%) ~scale RoundingMode/HALF_EVEN))}))
+
+;; We don't apply the default cast (i.e. [:decimal 64 9 256]) to decimals themselves.
+;; This is to mainly preserve decimal scales passed via pgwire parameters with type annotations.
+(defmethod codegen-cast [:decimal :decimal] [{:keys [source-type] {:keys [precision scale]} :cast-opts}]
+  (if precision
+    {:return-type [:decimal precision (or scale 0) (precision->bit-width precision)]
+     :->call-code #(do `(.setScale (bigdec ~@%) ~scale RoundingMode/HALF_EVEN))}
+    {:return-type source-type
+     :->call-code first}))
+
+(defmethod codegen-cast [:utf8 :decimal] [{{:keys [precision scale]} :cast-opts}]
+  (let [scale (if precision (or scale 0) 9)
+        precision (or precision 64)]
+    {:return-type [:decimal precision scale (precision->bit-width precision)]
+     :->call-code #(do `(.setScale (bigdec (buf->str ~@%)) ~scale RoundingMode/HALF_EVEN))}))
 
 (defmethod codegen-cast [:num :utf8] [_]
   {:return-type :utf8, :->call-code #(do `(resolve-utf8-buf (str ~@%)))})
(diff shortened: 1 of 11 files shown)


core/src/main/kotlin/xtdb/arrow/DecimalVector.kt
```diff
@@ -39,7 +42,12 @@ class DecimalVector private constructor(
             if (value.precision() > precision || value.scale() != scale) {
                 throw InvalidWriteObjectException(fieldType, value)
             }
+            // HACK, we throw a RuntimeError here, but it should likely be dealt with in the EE if a object doesn't fit
+            try {
             dataBuffer.writeBigDecimal(value, byteWidth)
+            } catch (e: UnsupportedOperationException) {
+                throw RuntimeException(DECIMAL_ERROR_KEY, e.message, emptyMap<Keyword, Any>(), e)
+            }
```
FiV0, line 50: I had this initially in the previous Decimal PR, but James thought this check should be on the caller. The problem is then that I need to check decimals (if they still fit the specified type) on every EE operation. I prefer this for now.

src/main/clojure/xtdb/test_util.clj
```diff
@@ -541,3 +543,16 @@
...
+  ([node query] (q node query {}))
+  ([node query opts]
+   (let [old-then-execute-prepared-query node.impl/then-execute-prepared-query
+         p (promise)]
+     (with-redefs [node.impl/then-execute-prepared-query
+                   (fn [^PreparedQuery prepared-query & args]
+                     (deliver p (.columnFields prepared-query))
+                     (apply old-then-execute-prepared-query prepared-query args))]
+       {:res (xt/q node query opts)
+        :res-type (mapv (juxt #(.getName ^Field %) types/field->col-type) @p)}))))
```
FiV0: Not sure if this is the right approach. We could potentially also expose more of the node?

src/test/clojure/xtdb/expression_test.clj
```diff
@@ -1393,23 +1403,74 @@
...
              (run-test 'greatest (double 3.0) (bigdec 2.0))))
 
-    ;; TODO decide on behaviour
-    ;; out of fixed precision range
-    ;; FIXME see #4331, this is not correct
-    (t/is (= {:res 2E+128M, :res-type [:decimal 32 -128 128]}
-             (run-test '+ (bigdec 1E+128M) (bigdec 1E+128M))))
+    (t/is (= {:res-type [:union #{[:decimal 32 2 128] [:decimal 32 1 128]}],
+              :res 1.0M}
+             (run-test 'least (bigdec 1.0M) (bigdec 2.02))))
```
FiV0, line 1408: I am using the standard `<` in this case. So a union goes in, a union comes out, but we could do something similar to the arithmetic operations and also take the `[:decimal 32 2 128]` here. A tradeoff once more.

FiV0 (commented):
@mbutlerw Left a couple of comments, but otherwise good for a real review this time. Ignore the Postgres bit in the writup for now as I actually couldn't verify that claim.

core/src/main/clojure/xtdb/expression.clj
```diff
@@ -794,6 +821,37 @@
   {:return-type (types/least-upper-bound arg-types)
    :->call-code #(do `(* ~@%))})
 
+;; decimal
+
+(defn- combine-precision [^long p1 ^long p2]
+  (if (and (<= p1 32) (<= p2 32)) 32 64))
+
+(defn- precision->bit-width [^long p]
```
mbutlerw: `precision->bit-width` and `precision->bit-witdh` seem to be duplicates

src/main/clojure/xtdb/test_util.clj
mbutlerw: Node exposes `prepare-sql`, and you can get the result types and run the query of the returned prepared-query. I think wrapping those steps up in a test util might be a cleaner method.

src/test/clojure/xtdb/sql_test.clj
```diff
@@ -2837,3 +2837,45 @@ UNION ALL
...
+
+    (let [expected-res [2.02M 2.1M]
+          {:keys [res res-type]} (tu/q tu/*node* "SELECT d + 1::decimal(64,1) AS v FROM docs")
+          res (map :v res)]
+
+      (t/is (= [["v" [:union #{[:decimal 64 2 256] [:decimal 64 1 256]}]]]
+               res-type) )
+
+      (doseq [[expected-d reslut-d] (map vector expected-res res)]
+        (t/is (compare-decimals expected-d reslut-d)
```
mbutlerw: Typo

src/test/clojure/xtdb/expression_test.clj
mbutlerw: Seems reasonable for now.

core/src/main/kotlin/xtdb/arrow/DecimalVector.kt
mbutlerw: Understood, I agree this doesn't feel like the right place, but happy to defer to you.

mbutlerw (changes requested):
LGTM after requested changes :+1:

core/src/main/clojure/xtdb/expression.clj
```diff
@@ -241,6 +242,32 @@
...
+        precision (or precision 64)]
+    {:return-type [:decimal precision scale (precision->bit-witdh precision)]
+     :->call-code #(do `(.setScale (bigdec ~@%) ~scale RoundingMode/HALF_EVEN))}))
+
+;; We don't apply the default cast (i.e. [:decimal 64 9 256]) to decimals themselves.
+;; This is to mainly preserve decimal scales passed via pgwire parameters with type annotations.
+(defmethod codegen-cast [:decimal :decimal] [{:keys [source-type] {:keys [precision scale]} :cast-opts}]
+  (if precision
+    {:return-type [:decimal precision (or scale 0) (precision->bit-witdh precision)]
+     :->call-code #(do `(bigdec ~@%))}
```
FiV0: Incorrect. TODO

FiV0:
Only potential thing to come back to is the part where we deal with the non fitting decimals in the vectors rather than at the EE level.

PR #4368, opened 2025-04-17
Garbage collector component GCing old blocks.

### LINKED ISSUE #4307: GCing block files + table block files
With #4364 we can now assume any node is at most 5 blocks behind the furthest ahead node. 
- Create a new garbage collector component (good naming! - xtgc)
- The `xtgc` component runs a periodic job (every 10 minutes) that:
   - Choose 100 random distinct tables, list all table block files for these 100 tables, clean up all table block files apart from the last 10 blocks
   - List block files, clean up all block files apart from the last 10 block files

Note: 
- If these delete operations interleave with block and table block creation events it might be that we end up with block files that reference table block files which no longer exist.

Old card description (@jarohen):
We no longer need more than one block file/one table block file per table - given these are cumulative, their size adds up.

Let's keep (say) 5 for safe-keeping and GC the rest?

Image

danmason:
Resolves #4307 
Github actions: https://github.com/danmason/xtdb/actions?query=branch%3Ablock-garbage-collector++

Adds a separated out configurable "garbage collector" component to XTDB, which:
- Runs a job on a schedule, with some random jitter.
- Calls down to garbage collection logic implemented within block catalog.

Within block catalog, implement functions for garbage collection all but "blocksToKeep" files:
- Deletes old block and table block files, retaining up to a configurable number of files.
- Deletion logic validates that it can NOT delete the latest block. 
- Has a set of tests against it.

In terms of feedback, interested in:
- Anything else we think that we should test here? Is there a useful test we could run against this to see the garbage collection component in operation against a node? 
- Have already moved the garbage collection logic into BlockCatalog itself - should we move anything else? Ie, should table block GC be handled by the table catalog?

### DIFF
--- a/core/src/main/clojure/xtdb/garbage_collector.clj
+++ b/core/src/main/clojure/xtdb/garbage_collector.clj
@@ -1,0 +1,26 @@
+(ns xtdb.garbage-collector
+  (:require [integrant.core :as ig]
+            [xtdb.node :as xtn])
+  (:import [xtdb.api Xtdb$Config GarbageCollectorConfig]
+           [xtdb.garbage_collector GarbageCollector]))
+
+(defmethod xtn/apply-config! :xtdb/garbage-collector [^Xtdb$Config config _ {:keys [enabled? blocks-to-keep approx-run-interval]}]
+  (.garbageCollector config
+                     (cond-> (GarbageCollectorConfig.)
+                       enabled? (.enabled enabled?)
+                       blocks-to-keep (.blocksToKeep blocks-to-keep)
+                       approx-run-interval (.approxRunInterval approx-run-interval))))
+
+(defmethod ig/prep-key :xtdb/garbage-collector [_ ^GarbageCollectorConfig config]
+  {:block-catalog (ig/ref :xtdb/block-catalog)
+   :enabled? (.getEnabled config)
+   :blocks-to-keep (.getBlocksToKeep config)
+   :approx-run-interval (.getApproxRunInterval config)})
+
+(defmethod ig/init-key :xtdb/garbage-collector [_ {:keys [block-catalog enabled? blocks-to-keep approx-run-interval]}]
+  (when enabled?
+    (GarbageCollector. block-catalog blocks-to-keep approx-run-interval)))
+
+(defmethod ig/halt-key! :xtdb/garbage-collector [_ ^GarbageCollector gc]
+  (when gc
+    (.close gc)))
--- a/core/src/main/kotlin/xtdb/garbage_collector/GarbageCollector.kt
+++ b/core/src/main/kotlin/xtdb/garbage_collector/GarbageCollector.kt
@@ -1,0 +1,48 @@
+package xtdb.garbage_collector
+
+import kotlinx.coroutines.*
+import org.slf4j.LoggerFactory
+import xtdb.BufferPool
+import xtdb.api.storage.ObjectStore
+import xtdb.catalog.BlockCatalog
+import xtdb.util.asPath
+import java.io.Closeable
+import java.time.Duration
+import kotlin.random.Random
+import kotlin.time.Duration.Companion.seconds
+
+private val LOGGER = LoggerFactory.getLogger(GarbageCollector::class.java)
+
+class GarbageCollector(
+    private val blockCatalog: BlockCatalog,
+    private val blocksToKeep: Int,
+    private val approxRunInterval: Duration
+) : Closeable {
+    private val scope = CoroutineScope(Dispatchers.IO)
+
+    init {
+        LOGGER.info("Starting GarbageCollector with approxRunInterval: $approxRunInterval, blocksToKeep: $blocksToKeep")
+        scope.launch {
+            delay(Random.nextLong(approxRunInterval.toMillis()))
+            while (isActive) {
+                try {
+                    LOGGER.debug("Starting block garbage collection")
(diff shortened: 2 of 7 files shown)


core/src/main/clojure/xtdb/garbage_collector.clj
```diff
@@ -0,0 +1,26 @@
+(ns xtdb.garbage-collector
+  (:require [integrant.core :as ig]
+            [xtdb.node :as xtn])
+  (:import [xtdb.api Xtdb$Config GarbageCollectorConfig]
+           [xtdb.garbage_collector GarbageCollector]))
+
+(defmethod xtn/apply-config! :xtdb/garbage-collector [^Xtdb$Config config, _, [enabled blocks-to-keep approx-run-interval]]
```
jarohen: nit: `enabled?` in EDN

core/src/main/kotlin/xtdb/garbage_collector/GarbageCollector.kt
```diff
@@ -0,0 +1,45 @@
...
+) : Closeable {
+    private val scope = CoroutineScope(Dispatchers.IO)
+
+    init {
+        LOGGER.info("Starting GarbageCollector with approxRunInterval: $approxRunInterval, blocksToKeep: $blocksToKeep")
+        scope.launch {
+            delay(Random.nextLong(approxRunInterval.toMillis()))
+            while (isActive) {
+                LOGGER.debug("Next GC run scheduled in ${approxRunInterval.toMillis()}ms")
+                delay(approxRunInterval.toMillis())
```
jarohen: now that you have the delay above, I'd probably move this one after so that it first runs after the initial delay rather than initial delay + full delay?

core/src/main/kotlin/xtdb/garbage_collector/GarbageCollector.kt
```diff
@@ -0,0 +1,45 @@
...
+
+                try {
+                    LOGGER.debug("Starting block garbage collection")
+                    blockCatalog.garbageCollectBlocks(blocksToKeep)
+                    LOGGER.debug("Block garbage collection completed")
+                } catch (e: Exception) {
```
jarohen, line 32: fine to catch and log Exception, but should additionally catch Throwable, otherwise Errors (non-Exceptions) may not get logged with any context. 

```kotlin
try {
  ...
} catch (e: Exception) {
  LOGGER.warn(...)
} catch (e: Throwable) {
  throw RuntimeException("...", e)
}
```

src/test/kotlin/xtdb/catalog/BlockCatalogTest.kt
```diff
@@ -0,0 +1,105 @@
...
+        assertBlockCount(bufferPool, blocksPath, 10)
+        assertBlockCount(bufferPool, table1BlockPath, 10)
+        assertBlockCount(bufferPool, table2BlockPath, 10)
+
+        // Create BlockCatalog instance
+        val catalog = BlockCatalog(bufferPool)
+
+        // Validate that the latest block is correct
+        val latestBlockIndex = catalog.currentBlockIndex
+        assertTrue(latestBlockIndex == 10L, "Expected latest block index to be 10, but got $latestBlockIndex")
```
jarohen: `assertEquals`

jarohen: also message not adding a lot here over what `assertEquals` would do, best to elide

src/test/kotlin/xtdb/catalog/BlockCatalogTest.kt
```diff
@@ -0,0 +1,105 @@
...
+        catalog.garbageCollectBlocks(blocksToKeep = 3)
+
+        // Validate that the oldest blocks are deleted, leaving only the latest 3 blocks
+        assertBlockCount(bufferPool, blocksPath, 3)
+        assertBlockCount(bufferPool, table1BlockPath, 3)
+        assertBlockCount(bufferPool, table2BlockPath, 3)
+
+        // Validate latest block still exists
+        val latestBlockFile = bufferPool.listAllObjects(blocksPath).toList().find { it.key == blocksPath.resolve("b${10L.asLexHex}.binpb") }
+        assertTrue(latestBlockFile != null, "Expected latest block file to exist after GC, but it was deleted")
```
jarohen: `assertNotNull`

jarohen (approved):
Thanks @danmason - a few nits but nothing major, good work :clap: 

Longer term, I think I'd turn your GC component into something that handles all of the background task scheduling (or, at least, the top-level, even if it defers to the individual components straightaway) so that we can turn off all background tasks with one configuration entry (e.g. for tests - we currently do this a lot for the compactor)

core/src/main/kotlin/xtdb/garbage_collector/GarbageCollector.kt
```diff
@@ -0,0 +1,45 @@
...
+                    LOGGER.warn("Block garbage collection failed", e)
+                }
+            }
+        }
+    }
+
+    override fun close() {
+        scope.cancel()
```
jarohen: ah - you'll want `runBlocking { scope.cancelAndJoin(...) }` here, so that the close is synchronous

jarohen: (pseudo-code, grep `cancelAndJoin`)

PR #4375, opened 2025-04-21
Roundtripping keywords in next.jdbc

### LINKED ISSUE #4237: Roundtripping keywords through pgwire
Users should be able to insert a keyword as a column value via a parameter in a SQL query using our next.jdbc support, then receive keywords back from queries.

FiV0:
This adds a new oid, which is arbitrary, for keywords. Otherwise (de)serialeses keywords similar to strings. In our custom `next.jdbc` namespace we convert the string to keyword. We are hoping that drivers that are not able to deal with this type will just fall back to text.

Something like the following still doesn't work
```clj
(jdbc/execute! conn ["INSERT INTO docs (_id, foo) VALUES (1, ?)" (types/as-other :bar)])
```

### DIFF
--- a/core/src/main/clojure/xtdb/types.clj
+++ b/core/src/main/clojure/xtdb/types.clj
@@ -1210,6 +1210,24 @@
                                (.get bb ba)
                                ba))}
 
+   :keyword {:typname "keyword"
+             :col-type :keyword
+             :oid 11111
+             :typsend "keywordsend"
+             :typreceive "keywordrecv"
+             :read-text (fn [_env ba] (read-utf8 ba))
+             :read-binary (fn [_env ba] (read-utf8 ba))
+             :write-text (fn [_env ^IVectorReader rdr idx]
+                           (let [bb (.getBytes rdr idx)
+                                 ba ^bytes (byte-array (.remaining bb))]
+                             (.get bb ba)
+                             ba))
+             :write-binary (fn [_env ^IVectorReader rdr idx]
+                             (let [bb (.getBytes rdr idx)
+                                   ba ^bytes (byte-array (.remaining bb))]
+                               (.get bb ba)
+                               ba))}
+
    :bytea {:typname "bytea"
            :col-type :varbinary
            :oid 17
@@ -1481,6 +1499,7 @@
    :null :text
    :regclass :regclass
    :varbinary :bytea
+   :keyword :keyword
    [:date :day] :date
    [:timestamp-local :micro] :timestamp
    [:timestamp-tz :micro] :timestamptz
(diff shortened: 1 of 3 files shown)


src/test/clojure/xtdb/pgwire_test.clj
```diff
@@ -2704,3 +2704,26 @@ ORDER BY 1,2;")
...
+    (t/is (= [{:_id 1, :foo :bar}]
+             (jdbc/execute! conn ["FROM docs"])))
+
+    (t/is (= [{:xt/id 1, :foo :bar}]
+             (jdbc/execute! conn ["FROM docs"]
+                            {:builder-fn xt-jdbc/builder-fn}))))
+
+  (with-open [pg-conn (pg-conn {})]
+    (t/is (= [{:_id 1, :foo "bar"}] (pg/execute pg-conn "FROM docs"))
+          "if the driver doesn't know about keywords it might fall back to text")))
```
FiV0, line 2729: It might be that other drivers complain about this.

core/src/main/clojure/xtdb/types.clj
```diff
@@ -1210,6 +1210,24 @@
                                (.get bb ba)
                                ba))}
 
+   :keyword {:typname "keyword"
+             :col-type :keyword
+             :oid 11111
```
FiV0, line 1215: Arbitrarily chosen above 10000. It could be the case that this conflicts with existing types in other systems.

jarohen (approved):
Thanks @FiV0 :pray: 

for your example above, any idea why it doesn't work?

FiV0:
@jarohen I am getting following error when running 
```clj
(jdbc/execute! conn ["INSERT INTO docs (_id, foo) VALUES (1, ?)" (types/as-other :bar)])
```
Error:
```
1. Unhandled org.postgresql.util.PSQLException
   ERROR: Cannot invoke "clojure.lang.IFn.invoke(Object, Object)"
   because "read_text" is null

    QueryExecutorImpl.java: 2733  org.postgresql.core.v3.QueryExecutorImpl/receiveErrorResponse
    QueryExecutorImpl.java: 2420  org.postgresql.core.v3.QueryExecutorImpl/processResults
    QueryExecutorImpl.java:  372  org.postgresql.core.v3.QueryExecutorImpl/execute
          PgStatement.java:  517  org.postgresql.jdbc.PgStatement/executeInternal
          PgStatement.java:  434  org.postgresql.jdbc.PgStatement/execute
  PgPreparedStatement.java:  194  org.postgresql.jdbc.PgPreparedStatement/executeWithFlags
  PgPreparedStatement.java:  180  org.postgresql.jdbc.PgPreparedStatement/execute
            result_set.clj:  674  next.jdbc.result-set/stmt->result-set
            result_set.clj:  669  next.jdbc.result-set/stmt->result-set
            result_set.clj:  902  next.jdbc.result-set/eval18256/fn
```
I suspect that jdbc doesn't know how to encode the keyword.

jarohen:
reverted

FiV0:
> for your example above, any idea why it doesn't work?

This comes through as a unspecified parameter. So the main issue is how to tell `next.jdbc` (or other clients) the oid and how to encode/decode that oid type.

PR #4406, opened 2025-05-02
ability to inline XTQL in SQL using dollar-quoted strings

jarohen:
This PR adds the ability to query XTQL inlined within SQL.

* Grammar is `XTQL 'XTQL string'` - but most likely to be used with dollar-quoted strings (as is common in Postgres for things like `CREATE FUNCTION (...) AS $$ [SQL] $$` to avoid needing to escape quotes.
* Works as a standalone query: `XTQL $$ (-> (from ...) (aggregate ...) ...) $$`
* Also works as a FROM/JOIN relation: `FROM (XTQL $$ ... $$) foo SELECT ...`
* I'd imagine this'll be most often used from Clojure with a helper (which we may want to consider adding to `xt.next.jdbc`, but also see seancorfield/honeysql#572.

  ```clojure
  (defn xtql [q] 
    (format "XTQL $$ %s $$" (pr-str q)
  ```

### DIFF
--- a/src/test/clojure/xtdb/sql_test.clj
+++ b/src/test/clojure/xtdb/sql_test.clj
@@ -2934,3 +2934,28 @@ SELECT
     WHERE v.valid_time_intersection CONTAINS (dates.d + INTERVAL 'PT0M')
   ) AS member_count
 FROM dates"))))
+
+(t/deftest inline-xtql
+  (xt/submit-tx tu/*node* ["INSERT INTO foo RECORDS {_id: 1, x: 'foo'}"
+                           "INSERT INTO bar RECORDS {_id: 1, y: 'bar'}"
+                           "INSERT INTO bar RECORDS {_id: 2, y: 'baz'}"])
+
+  (t/is (= [{:xt/id 1, :x "foo"}] (xt/q tu/*node* "XTQL $$(from :foo [*])$$")))
+  (t/is (= [{:xt/id 1, :x "foo"}] (xt/q tu/*node* "XTQL $$(from :foo [xt/id x])$$")))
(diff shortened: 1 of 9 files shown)


src/test/clojure/xtdb/sql_test.clj
```diff
@@ -2934,3 +2934,28 @@ SELECT
...
+                           "INSERT INTO bar RECORDS {_id: 2, y: 'baz'}"])
+
+  (t/is (= [{:xt/id 1, :x "foo"}] (xt/q tu/*node* "XTQL $$(from :foo [*])$$")))
+  (t/is (= [{:xt/id 1, :x "foo"}] (xt/q tu/*node* "XTQL $$(from :foo [xt/id x])$$")))
+  (t/is (= [{:x "foo"}] (xt/q tu/*node* "XTQL $$ (from :foo [x]) $$")))
+
+  (t/is (= [{:x "foo", :y "bar"}]
+           (xt/q tu/*node* "SELECT * EXCLUDE _id
+                            FROM foo f1
+                              JOIN (XTQL $$(from :bar [_id y])$$) f2
```
refset, line 2950: So `_id` is an alias of `xt/id` ? TIL! Neat that you don't have to think about the kebab<->snake dance every time with this one at least

jarohen: I hadn't even noticed I'd done that :sweat_smile: 

Yep, symbols in XTQL get normalised, so `:xt/id` -> `_id` (with `_id` itself passing straight through). Then, if we'd queried for `_id`, it would be returned as `:xt/id` through the Clojure API

Issue #3447, opened 2024-06-26
Explicitly disallow column references in system/valid-time period specifications


jarohen:
§7.6 Syntax rule 1.c:

> QSTPS shall not contain a 'column reference' or an 'SQL parameter reference'

Currently, it passes silently through the planner:

```clojure
(plan-sql "SELECT bar FROM foo FOR VALID_TIME AS OF bar"
          {:table-info {"foo" #{"bar"}}})

;; => `[:at foo.1/bar]`, which isn't a valid column at the scan level
```

refset:
Noting that Flink's SQL supports column refs for `SYSTEM_TIME AS OF` (at least), e.g.:
```sql
SELECT s.machineid, MAX(r.temperature) AS maxTemp
FROM SensorReading AS r  JOIN Sensor
    FOR SYSTEM_TIME AS OF SensorReading.timestamp AS s
    ON r.sensorid = s.id GROUP BY s.machineid
```
https://www.datasqrl.com/blog/temporal-join/

PR #4429, opened 2025-05-09
Add expressions in temporal filters

### LINKED ISSUE #3306: Support for expressions in temporal filters
e.g. this would enable `... FOR VALID_TIME AS OF (CURRENT_TIMESTAMP - INTERVAL 'PT1M')`

* In scope: allowing this through syntax/planning
* Out of scope (of this card): #3849

FiV0:
Simply adds an extra EE call at runtime (before any values are scanned from disk) to evaluate the expression. No column references are allowed see #3447. Literals and params are "evaluated" without a call to the expression engine.

Todo:
- [x] Dealing with overflowing `start-of-time` and `end-of-time`.

### DIFF
--- a/api/src/main/clojure/xtdb/time.clj
+++ b/api/src/main/clojure/xtdb/time.clj
@@ -117,6 +117,8 @@
 (def ^java.time.Instant start-of-time
   (micros->instant Long/MIN_VALUE))
 
+(def ^long start-of-time-as-micros Long/MIN_VALUE)
+
 (defn max-tx [l r]
   (if (or (nil? l)
           (and r (neg? (compare l r))))
--- a/core/src/main/antlr/xtdb/antlr/Sql.g4
+++ b/core/src/main/antlr/xtdb/antlr/Sql.g4
@@ -577,17 +577,10 @@ queryValidTimePeriodSpecification
    ;
 
 tableTimePeriodSpecification
-    : 'AS' 'OF' periodSpecificationExpr # TableAsOf
+    : 'AS' 'OF' at=expr # TableAsOf
     | 'ALL' # TableAllTime
-    | 'BETWEEN' periodSpecificationExpr 'AND' periodSpecificationExpr # TableBetween
-    | 'FROM' periodSpecificationExpr 'TO' periodSpecificationExpr # TableFromTo
-    ;
-
-// replace usages of this with dateTimeExpr when possible - #3306
-periodSpecificationExpr
-    : literal #PeriodSpecLiteral
-    | parameterSpecification #PeriodSpecParam
-    | ('NOW' | 'CURRENT_TIMESTAMP') #PeriodSpecNow
+    | 'BETWEEN' from=expr 'AND' to=expr # TableBetween
+    | 'FROM' from=expr 'TO' to=expr # TableFromTo
     ;
 
 tableOrQueryName : tableName ;
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -13,7 +13,8 @@
             [xtdb.trie :as trie]
             [xtdb.trie-catalog :as cat]
             [xtdb.types :as types]
-            [xtdb.util :as util])
+            [xtdb.util :as util]
+            [xtdb.vector.writer :as vw])
   (:import (clojure.lang MapEntry)
            java.nio.ByteBuffer
            java.time.Instant
@@ -55,33 +56,44 @@
 
 (def ^:dynamic *column->pushdown-bloom* {})
 
-(defn- ->temporal-bounds [^RelationReader args, {:keys [for-valid-time for-system-time]}, ^Instant snapshot-time]
+(defn ->temporal-bounds [^BufferAllocator alloc, ^RelationReader args,
+                         {:keys [for-valid-time for-system-time]}, ^Instant snapshot-time]
   (letfn [(->time-μs
             ([arg] (->time-μs arg nil))
             ([[tag arg] default]
              (case tag
-               :literal (or (some-> arg (time/sql-temporal->micros expr/*default-tz*))
+               :literal (or (some-> arg
+                                    (time/sql-temporal->micros expr/*default-tz*))
                             default)
-               :param (or (some-> (-> (.vectorForOrNull args (name arg))
-                                      (.getObject 0))
+               :param (or (some-> (-> (.vectorForOrNull args (name arg)) (.getObject 0))
                                   (time/sql-temporal->micros expr/*default-tz*))
                           default)
-               :now (-> (expr/current-time)
-                        (time/instant->micros)))))
+               :now (-> (expr/current-time) (time/instant->micros))
+               :expr (let [param-types (expr/->param-types args)
+                           projection (expr/->expression-projection-spec "_temporal_expression"
+                                                                         (expr/form->expr arg {:param-types param-types})
+                                                                         {:param-types param-types})]
+                       (util/with-open [res (.project projection alloc vw/empty-args {} args)]
+                         (if-let [inst-like (.getObject res 0)]
+                           (do
+                             (time/expect-instant inst-like)
+                             (-> inst-like time/->instant time/instant->micros))
+                           default))))))
           (apply-constraint [constraint]
             (if-let [[tag & args] constraint]
               (case tag
                 :at (let [[at] args
-                          at-μs (->time-μs at)]
+                          at-μs (->time-μs at (-> (expr/current-time) (time/instant->micros)))]
                       (TemporalDimension/at at-μs))
 
                 ;; overlaps [time-from time-to]
                 :in (let [[from to] args]
-                      (TemporalDimension/in (->time-μs from Long/MIN_VALUE)
-                                            (some-> to (->time-μs))))
+                      ;; TODO asymmetry of defaulting start-of-time-here and end-of-time in TemporalBounds
+                      (TemporalDimension/in (->time-μs from time/start-of-time-as-micros)
+                                            (some-> to ->time-μs)))
 
                 :between (let [[from to] args]
-                           (TemporalDimension/between (->time-μs from Long/MIN_VALUE)
+                           (TemporalDimension/between (->time-μs from time/start-of-time-as-micros)
                                                       (some-> to ->time-μs)))
 
                 :all-time (TemporalDimension.))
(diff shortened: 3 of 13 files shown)


core/src/main/clojure/xtdb/sql.clj
```diff
@@ -2666,7 +2665,7 @@
                     {doc ~(into {} (map (juxt keyword identity)) known-cols)}]
           [:order-by [[_iid] [_valid_from]]
            [:scan {:table ~table,
-                   :for-valid-time [:in ~valid-from ~valid-to]}
+                   :for-valid-time [:in ~(or valid-from :now) ~valid-to]}
```
FiV0: `nil` is now an expression so I need to be explicit here and also in `xtql/plan.clj` below.

src/test/clojure/xtdb/sql_test.clj
```diff
@@ -2965,3 +2965,22 @@ FROM dates"))))
              (xt/q tu/*node* [(format "FROM (XTQL $$ %s $$) t SELECT *, ? AS x"
                                       (pr-str '#(from :bar [{:xt/id %} *])))
                               2 "x"])))))
+
+(t/deftest temporal-filter-expressions-3306
+  (xt/execute-tx tu/*node* [[:put-docs :docs {:xt/id 1 :xt/valid-from #inst "2000-01-01" :xt/valid-to #inst "2100-01-01"}]])
+
+  (t/is (= [#:xt{:id 1}]
+           (xt/q tu/*node* "SELECT * FROM docs FOR VALID_TIME AS OF (CURRENT_TIMESTAMP - INTERVAL 'PT1M')")))
```
jarohen: can you explicitly set current-timestamp in the opts given you use it in the query?

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -64,7 +66,16 @@
                                  (.getObject 0))
                              (time/sql-temporal->micros expr/*default-tz*))
               :now (-> (expr/current-time)
-                       (time/instant->micros))))
+                       (time/instant->micros))
+              :expr (let [param-types (expr/->param-types args)
+                          projection (expr/->expression-projection-spec "_temporal_expression"
+                                                                        (expr/form->expr arg {:param-types param-types})
+                                                                        {:param-types param-types})]
+                      (util/with-open [dummy-rel (vw/open-args alloc [42])
```
jarohen: rather than dummy-rel, we have (the potentially now misnamed) `vw/empty-args` which doesn't require opening/closing

core/src/main/antlr/xtdb/antlr/Sql.g4
```diff
@@ -595,11 +595,11 @@ tableTimePeriodSpecification
 
 // replace usages of this with dateTimeExpr when possible - #3306
 periodSpecificationExpr
-    : literal #PeriodSpecLiteral
-    | parameterSpecification #PeriodSpecParam
-    | ('NOW' | 'CURRENT_TIMESTAMP') #PeriodSpecNow
+    : ('NOW' | 'CURRENT_TIMESTAMP') #PeriodSpecNow
+    | expr #PeriodSpecExpr
```
jarohen: possible to remove `periodSpecificationExpr` entirely now (i.e. inline it as `expr`)?

src/test/clojure/xtdb/sql_test.clj
```diff
@@ -2965,3 +2965,22 @@ FROM dates"))))
              (xt/q tu/*node* [(format "FROM (XTQL $$ %s $$) t SELECT *, ? AS x"
                                       (pr-str '#(from :bar [{:xt/id %} *])))
                               2 "x"])))))
+
+(t/deftest temporal-filter-expressions-3306
```
jarohen, line 2988: worth some unit tests against `->temporal-bounds` here?

jarohen: ha, in fact, we have a test in scan-test with a note saying to adapt for `scan/->temporal-bounds` - good time to do that, while you're in context?

jarohen (changes requested):
cheers @FiV0 - actually came out a lot simpler than I was expecting in the end, well done for figuring it out :clap:

core/src/main/clojure/xtdb/operator/scan.clj
FiV0: Yes, but then I don't get anything out of the EE, ie. nothing to run.

FiV0: Can potentially use a `NullVector` of size 1 here.

src/test/clojure/xtdb/sql_test.clj
FiV0: Do you want me to add temporal column predicate constraints to `->temporal-bounds` (if possible)?

FiV0: I have added an extra test. The other test is more about pushing temporal predicates into the `temporal-bounds`.

api/src/main/clojure/xtdb/time.clj
```diff
@@ -110,6 +110,9 @@
 (def ^java.time.Instant end-of-time
   (micros->instant Long/MAX_VALUE))
 
+(def ^java.time.Instant start-of-time
+  (micros->instant (+ Long/MIN_VALUE (long 1000000))))
```
FiV0: O/w I get an overflow in `insant->micros`.

core/src/main/antlr/xtdb/antlr/Sql.g4
```diff
@@ -830,9 +823,13 @@ dmlStatementValidTimeExtents
...
   : PATCH INTO tableName
-    ('FOR' ('PORTION' 'OF')? 'VALID_TIME' 'FROM' validFrom=staticExpr ('TO' validTo=staticExpr)?)?
-    patchSource
-  ;
+    patchStatementValidTimeExtents?
+    patchSource;
+
+// could become dmlStatementValidTimeExtents
+patchStatementValidTimeExtents
+: 'FOR' ('PORTION' 'OF')? 'VALID_TIME' 'FROM' from=staticExpr ('TO' to=staticExpr)? # PatchStatementValidTimePortion
```
FiV0: This could become `dmlStatementValidTimeExtents` at some point.

core/src/main/clojure/xtdb/operator/patch.clj
```diff
@@ -1,27 +1,40 @@
...
-                   (.getObject 0)))
+                   (.getObject 0))
+        :expr (let [param-types (expr/->param-types args)
+                    projection (expr/->expression-projection-spec "_temporal_expression"
+                                                                  (expr/form->expr value {:param-types param-types})
+                                                                  {:param-types param-types})]
+                (util/with-open [res (.project projection allocator vw/empty-args {} args)]
+                  (let [inst-like (or (.getObject res 0) default)]
+                    (time/expect-instant inst-like)
+                    inst-like))))
```
FiV0: I am adding an expression in the RA as it make planning in `sql.clj` easier, this is currently not exposed in SQL PATCH.

src/test/clojure/xtdb/operator/scan_test.clj
```diff
@@ -373,66 +375,127 @@
...
+      (t/testing "nil behaviour"
+        (t/is (= [(time/instant->micros clock-time) (time/instant->micros time/end-of-time)]
+                 (-> (->temporal-bounds vw/empty-args
+                                        '{:for-valid-time [:in [:expr nil] [:expr nil]]}
+                                        snapshot-time)
+                     tu/temporal-bounds->data
+                     :valid-time)))))))
+
+;; TODO adapt for scan/->temporal-bounds
+#_(t/deftest can-create-temporal-min-max-range
```
FiV0: I left this test as a TODO as it mainly concerns predicates on temporal columns which get translated into valid-time and system-time constraints if possible.

FiV0:
The approach is now as follows `FROM` and `:scan` behavior for temporal columns is roughly the same. Meaning one can write an RA plan with `nil`s and these get defaulted to `now` and `end-of-time` respectively (in `scan`). Any other behavior needs to be explicitly dealt with in the planner. This is the case for our DML operations. Here we make sure expressions resulting in nil get defaulted to `start-of-time` and `end-of-time` respectively.

There are still some issues for `start-of-time` and `end-of-time` as certain calls will result in a overflow. I left some comments in the appropriate places.

core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -74,12 +85,14 @@
...
+                      (TemporalDimension/in (->time-μs from (expr/current-time))
+                                            (->time-μs to time/end-of-time)))
 
                 :between (let [[from to] args]
-                           (TemporalDimension/between (->time-μs (or from [:now]))
-                                                      (some-> to ->time-μs)))
+                           (TemporalDimension/between (->time-μs from (expr/current-time))
+                                                      (let [res (->time-μs to time/end-of-time)]
+                                                        (cond-> res
+                                                          (= res Long/MAX_VALUE) dec))))
```
FiV0: Getting an overlow in `TemporalDimension` otherwise.

FiV0:
This is a breaking change because `UPDATE ....`, `DELETE....` and `PATCH ... FOR PORTION OF VALID_TIME NULL TO NULL ...`
changes from `now` to `end-of-time` to `start-of-time` to `end-of-time`.

core/src/main/clojure/xtdb/sql.clj
```diff
@@ -600,7 +596,7 @@
                                         ['pg_catalog table-cols]))
 
                                     (add-warning! env (->BaseTableNotFound sn tn)))
-                expr-visitor (->ExprPlanVisitor env scope)]
+                expr-visitor (->ExprPlanVisitor (assoc env :no-column-refs? true) scope)]
```
jarohen: this is what `scope`'s for - you could consider adding a scope that doesn't have any available variables

FiV0:
Breaking change moved to #4454.

src/test/clojure/xtdb/operator/scan_test.clj
jarohen: let's run this one through - probably worth us either updating or deleting it given we're not likely to ever be _more_ in context than now :slightly_smiling_face:

core/src/main/clojure/xtdb/sql.clj
jarohen: double check default to now here?

core/src/main/clojure/xtdb/xtql/plan.clj
```diff
@@ -364,13 +364,13 @@
...
     (required-vars-available? (.getAt this) #{})
-    [:at (plan-expr (.getAt this))])
+    [:at (or (plan-expr (.getAt this)) :now)])
 
   TemporalFilter$In
   (plan-temporal-filter [this]
     (required-vars-available? (.getFrom this) #{})
     (required-vars-available? (.getTo this) #{})
-    [:in (plan-expr (.getFrom this)) (plan-expr (.getTo this))])
+    [:in (or (plan-expr (.getFrom this)) 'xtdb/start-of-time) (plan-expr (.getTo this))])
```
jarohen, line 373: asymmetry?

to be solved later - leave a comment

src/test/clojure/xtdb/operator/scan_test.clj
jarohen: test removed on main :pray:

jarohen (approved):
minor changes, good to merge without further review :clap:

PR #4445, opened 2025-05-13
add xt.metrics_timers and xt.metrics_gauges tables

jarohen:
```sql
SELECT * FROM xt.metrics_gauges;
```
image

```sql
SELECT * FROM xt.metrics_timers;
```
image


jarohen:
not actually merged, I messed up

PR #4446, opened 2025-05-13
add xt.metrics_timers and xt.metrics_gauges tables

jarohen:
reopen #4445


jarohen:
I may return to this at some point - for now, minimising WIP

PR #4458, opened 2025-05-16
Defaulting to `start-of-time` for explicit NULL from in VALID_TIME and SYSTEM_TIME filters

FiV0:
This means a `FROM table FOR VALID_TIME/SYSTEM_TIME FROM NULL to NULL` now means a table scan for all of time.


core/src/main/clojure/xtdb/operator/scan.clj
```diff
@@ -74,11 +77,12 @@
 
                 ;; overlaps [time-from time-to]
                 :in (let [[from to] args]
-                      (TemporalDimension/in (->time-μs (or from [:now]))
-                                            (some-> to ->time-μs)))
+                      (prn args)
```
jarohen: prn

PR #4488, opened 2025-05-28
Add finish block endpoint

### LINKED ISSUE #4338: Add ability to manually cut a block
For example, before/after important migrations/batch jobs, it'd be useful for the user to be able to finish a block manually.

Probably a few options for how to do so:

* Healthz endpoint (my current but weak preference, so that it can be easier locked down)
* SQL statement (available to all, currently? doesn't sound like a Good Thing™)
* In-process API addition (harder to access for those only connecting through pgwire)
* others?

Any of these will need to send the flush-block message to the tx-log, s.t. all nodes agree on when to make the cut.

tggreene:
Adds healthz endpoint under system/finish-block to be able to manually cut a block.

Closes #4338

### DIFF
--- a/core/src/main/clojure/xtdb/healthz.clj
+++ b/core/src/main/clojure/xtdb/healthz.clj
@@ -16,7 +16,9 @@
            (xtdb.api.metrics HealthzConfig)
            xtdb.api.Xtdb$Config
            xtdb.BufferPoolKt
-           (xtdb.indexer LiveIndex LogProcessor)))
+           (xtdb.indexer LiveIndex LogProcessor)
+           xtdb.api.log.Log
+           xtdb.api.log.Log$Message$FlushBlock))
 
 (defn get-ingestion-error [^LogProcessor log-processor]
   (.getIngestionError log-processor))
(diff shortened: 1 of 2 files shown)


core/src/main/clojure/xtdb/healthz.clj
```diff
@@ -59,7 +61,17 @@
...
 
                 ["/healthz/ready" {:name :ready
-                                   :get (fn [_] {:status 200, :body "Ready."})}]]
+                                   :get (fn [_] {:status 200, :body "Ready."})}]
+
+                ["/system/finish-block" {:name :finish-block
+                                         :post (fn [{:keys [^Log log ^LiveIndex live-index]}]
+                                                 (try
+                                                   (let [latest-completed-block-tx (:tx-id (.getLatestCompletedBlockTx live-index))
+                                                         ^Log$Message$FlushBlock flush-msg (Log$Message$FlushBlock. (or latest-completed-block-tx -1))]
```
jarohen: nit: shouldn't need this type hint given it's a constructor call?

src/test/clojure/xtdb/healthz_test.clj
```diff
@@ -108,3 +113,43 @@
...
+                                                     [:put-docs :bar {:xt/id "bar6"}]]))))
+
+          (t/is (= nil (.getLatestCompletedBlockTx live-index)))
+
+          (t/testing "successful block flush"
+            (let [resp (clj-http/post (->system-url port "finish-block") {:throw-exceptions false})]
+              (t/is (= 200 (:status resp)))
+              (t/is (= "Block flush message sent successfully." (:body resp)))))
+
+          (Thread/sleep 2000)
```
jarohen: should be able to drop this Thread/sleep with the extra transaction if you run the execute-tx first, and then the getLatestCompletedBlockTx check

src/test/clojure/xtdb/healthz_test.clj
```diff
@@ -108,3 +113,43 @@
...
+          (t/is (= 5420 (:tx-id (xt/execute-tx node [[:put-docs :bar {:xt/id "bar7"}]
+                                                     [:put-docs :bar {:xt/id "bar8"}]
+                                                     [:put-docs :bar {:xt/id "bar9"}]]))))
+
+          (t/testing "second successful block flush"
+            (let [resp (clj-http/post (->system-url port "finish-block") {:throw-exceptions false})]
+              (t/is (= 200 (:status resp)))
+              (t/is (= "Block flush message sent successfully." (:body resp)))))
+
+          (Thread/sleep 2000)
```
jarohen: Same again here - send through a dummy tx and then check latest-completed-block-tx

Wherever possible, we avoid sleeps in tests - only when we absolutely cannot test it any other way.

src/test/clojure/xtdb/healthz_test.clj
```diff
@@ -108,3 +113,43 @@
...
+
+          (t/is (= 2097 (:tx-id (.getLatestCompletedBlockTx live-index))))
+
+          (t/is (= 5420 (:tx-id (xt/execute-tx node [[:put-docs :bar {:xt/id "bar7"}]
+                                                     [:put-docs :bar {:xt/id "bar8"}]
+                                                     [:put-docs :bar {:xt/id "bar9"}]]))))
+
+          (t/testing "second successful block flush"
```
jarohen: is it additionally worth a test here that sends through a message that _doesn't_ flush the block? i.e. one that doesn't lead to a flush, because either there haven't been any transactions since, or because its tx-id is out of sync?

jarohen (approved):
thanks @tggreene - fine to go in without further review :pray: 

few minor points below - could you address those and then either bundle them into the existing commit, or squash-merge?

cheers!

src/test/clojure/xtdb/healthz_test.clj
tggreene: I'm struggling to do this using execute-tx as checking the latest completed block is causing the latest tx to progress, is there a true dummy execute-tx I can call which doesn't increase the latest tx?

Issue #4512, opened 2025-06-02
GCing L1+ files proposal


FiV0:
### Current state
- No GC
- Table block files show the current set of active files and garbage files (- trie metadata). 

### What we need!
- GC L1+ files 
- Make sure no actively running queries (running for less than 6 hours) are impacted by this change.
- Files that are still active as of 5 block files behind the head can not be garbage collected.
- We want to make sure that the Trie catalog gets updated when files get GCed

### How queries are run 
- Download all the files necessary for the query.
- Run the scan

#### Approach
- Make trie catalog state more amenable to our GCing approach
	- Separate `live + nascent` and `garbage`
- Add a TrieDetails2 message from which the TrieCatalog state can be completely reconstructed without message replaying. This helps us to not keep garbage around for replaying.
- Add Garbage `as-of` to the TrieCatalog, this then allows us to query for the garbage before a particular timestamp.
- Actually do GC. Query for the garbage. Remove it from the OS. Send delete message on the log.
	- Should the delete be retried?
	- So we delete all files from the OS and then send the message. What happens if a delete doesn't go through?

Potential issues at the boundaries (assuming the cutoff for GCing is 5 blocks):
- Node is running ~5 blocks behind but not yet unhealthy. It takes more than 5 blocks indexing time to download all files for a query. The node would receive an object missing while still downloading all files (although S3 might guarantee the delivery of all files if we started the download process for all necessary files for the query)
- Node is the furthest ahead, but it takes more than ~10 blocks of indexing time to download all files. The same issue as above occurs.

The happy path for this approach is: Node `n1` gets a query and downloads all relevant files to local disk(cache) in less than ~5 blocks indexing time. If some other node `n2` is running way ahead and deletes data from object storage relevant for the query on node `n1` the data in the caches should not be affected as the deletion only affects object storage and caches on `n2`. New queries on `n1` will not use the same block file so won't run into this issue.

PR #4515, opened 2025-06-03
Add grace period configuration to garbage collector

FiV0:
- Add gracePeriod parameter to GarbageCollectorConfig with 6-hour default
- Update garbage collection logic to respect grace period when deleting blocks. Meaning we either keep 10 blocks or more if there are more in the gracePeriod.
- Prevent deletion of blocks newer than the grace period cutoff
- Add ClockUtil test utility for time-based testing

Part of #4512.


FiV0:
No longer needed. Updated #4512 with a different approach that no longer requires a grace period for the block catalog. There will likely still be a grace period, but it won't affect block files.

PR #4519, opened 2025-06-05
Add list expressions, use within emit-table-cols.

danmason:
Github action runs: https://github.com/danmason/xtdb/actions/workflows/build.yml?query=branch%3Aimplement-list-expr++

Adds in the `ListExpression` interface, alongside `xtdb.expression.list`, and makes use of this with emit-col-table. Simplifies how we handle lists within there, and acts as an incremental/equivalence change which should make reasoning around batching generate-series simpler.

### DIFF
--- a/core/src/main/clojure/xtdb/expression/list.clj
+++ b/core/src/main/clojure/xtdb/expression/list.clj
@@ -3,7 +3,7 @@
             [xtdb.types :as types]
             [xtdb.util :as util])
   (:import (clojure.lang IPersistentMap)
-           (xtdb.arrow ListExpression RelationReader ListValueReader)))
+           (xtdb.arrow ListExpression RelationReader)))
 
 (def vec-writer-sym (gensym 'vec_writer))
 
@@ -15,13 +15,17 @@
           {:field field
            :->list-expr (eval `(fn [~(-> expr/schema-sym (expr/with-tag IPersistentMap))
                                     ~(-> expr/args-sym (expr/with-tag RelationReader))]
-                                 (let [~@(expr/batch-bindings emitted-expr)
-                                       ^ListValueReader lvr# ~(continue (fn [_t c]
-                                                                          c))]
-                                   (reify ListExpression
-                                     (getSize [_#] (.size lvr#))
-                                     (writeTo [_# ~vec-writer-sym start# len#]
-                                       (dotimes [~expr/idx-sym len#]
-                                         (.writeValue ~vec-writer-sym (.nth lvr# (+ start# ~expr/idx-sym)))))))))}))
+                                 (let [~@(expr/batch-bindings emitted-expr)]
+                                   ~(continue (fn [t c]
+                                                (case (types/col-type-head t)
+                                                  (:list :set :fixed-size-list)
+                                                  `(let [lvr# ~c]
+                                                     (reify ListExpression
+                                                       (getSize [_#] (.size lvr#))
+                                                       (writeTo [_# ~vec-writer-sym start# len#]
+                                                         (dotimes [~expr/idx-sym len#]
+                                                           (.writeValue ~vec-writer-sym (.nth lvr# (+ start# ~expr/idx-sym)))))))
+
+                                                  nil))))))}))
       (util/lru-memoize) ; <<no-commit>>
       expr/wrap-zone-id-cache-buster))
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/xtdb/arrow/ListExpression.kt
```diff
@@ -0,0 +1,10 @@
+package xtdb.arrow
+
+import xtdb.vector.IVectorWriter
+import org.apache.arrow.vector.types.pojo.Field
+
+interface ListExpression {
+    fun field(): Field
```
danmason: Not strictly necessary here, given that it's already returned by `:field-type`. Could probably drop it, unless there's any reason we might need to keep it/use it?

core/src/main/clojure/xtdb/expression/list.clj
```diff
@@ -0,0 +1,26 @@
...
+
+(def vec-writer-sym (gensym 'vec_writer))
+
+(def compile-list-expr
+  (-> (fn [expr opts]
+        (let [expr (expr/prepare-expr expr)
+              {:keys [return-type continue] :as emitted-expr} (expr/codegen-expr expr opts) 
+              el-type (types/unnest-field (types/col-type->field return-type))]
+          {:field-type el-type 
+           :!list-expr-fn (eval `(fn [~(-> expr/args-sym (expr/with-tag RelationReader))]
```
jarohen: naming: I'd call this `:->list-expr` - `!` prefix is for things that need to be deref'd

core/src/main/kotlin/xtdb/arrow/ListExpression.kt
jarohen: in Kotlin this would be `val field: Field` and `val size: Int`, called from Clojure as `getField` and `getSize` respectively

core/src/main/clojure/xtdb/expression/list.clj
```diff
@@ -0,0 +1,26 @@
+(ns xtdb.expression.list
```
jarohen, line 1: let's have some tests for this - it's testable in its own right and there's a number of cases to consider:

- happy case
- expr that returns nil
- expr that doesn't return a list
- expr that _might_ return a list, but can potentially return other things as well

core/src/main/clojure/xtdb/expression/list.clj
```diff
@@ -0,0 +1,26 @@
...
+  (:import (xtdb.arrow ListExpression RelationReader ListValueReader)))
+
+(def vec-writer-sym (gensym 'vec_writer))
+
+(def compile-list-expr
+  (-> (fn [expr opts]
+        (let [expr (expr/prepare-expr expr)
+              {:keys [return-type continue] :as emitted-expr} (expr/codegen-expr expr opts) 
+              el-type (types/unnest-field (types/col-type->field return-type))]
+          {:field-type el-type
```
jarohen: naming: just `:field` here - `:field-type` is a separate Arrow type

core/src/main/clojure/xtdb/operator/table.clj
```diff
@@ -133,33 +134,22 @@
...
-
-                      (util/with-close-on-catch [el-rdr (.copy (or (some-> list-rdr .getListElements (.withName (str out-col)))
-                                                                   (vr/vec->reader (ZeroVector. (str out-col))))
-                                                               allocator)]
-
-                        (vr/rel-reader [el-rdr] (.getValueCount el-rdr))))))}))
+                  (if (instance? ArrowType$Null (.getType named-field))
+                    (vr/rel-reader [(vr/vec->reader (ZeroVector. (str out-col)))])
+                    (let [^ListExpression list-expr (!list-expr-fn args)
+                          out-vec-writer (vw/->writer (.createVector named-field allocator))]
```
jarohen: need some resource handling here - `createVector` creates a resource that might leak if any of the rest of this function throws an exception.

have a look at `util/with-close-on-catch`? (and lmk if it's unclear what to do :slightly_smiling_face:)

core/src/main/clojure/xtdb/expression/list.clj
jarohen: I'd also make `schema` available here, as it is in `emit-projection` - it's _possible_ (although unlikely) that the expression here will require it

jarohen (changes requested):
Thanks @danmason - a good first foray into the EE! :clap: 

Happy case looks good - let's tighten up on the unhappy cases and then this will be a good increment to land

core/src/main/kotlin/xtdb/arrow/ListExpression.kt
jarohen: sure, yep!

jarohen (approved):
sweet! :pray:

Issue #4285, opened 2025-03-19
Memory Node + submit-tx hanging


FiV0:
The following hangs at some point
```clj
(time
   (with-open [node (xtn/start-node)]
     (doseq [batch (partition-all 1000 (range 250000))]
       (xt/submit-tx node [(into [:put-docs :docs] (map (fn [idx] {:xt/id idx})) batch)]))))
```
if I replace `submit-tx` with `execute-tx` all seems to run fine.

PR #4526, opened 2025-06-12
Garbage Collection of trie files

### LINKED ISSUE #3195: Background process to garbage collect old index files
Once obj-store index files have been marked as garbage, they can be manually evicted from cached, and deleted from the object store

See #3191 

* L0 files must be preserved, so that we can use them as the basis of future data migrations
* Cannot remove garbage immediately as some nodes may be running behind (but to a limit - see #behind)
* Should handle the state of existing nodes - i.e. the process should naturally deal with any garbage backlog

### LINKED ISSUE #4512: GCing L1+ files proposal

FiV0:
This PR gradually introduces the garbage collection of trie files.
- The tries for a table trie catalog are now split into `live+nascent` and `garbage` for every `[level recency part]` branch. This facilitates the removal of the trie information from the trie catalog once the trie file is really garbage collected. 
- We add two additional fields to `TrieDetails` (we could also make it a `TrieDetails2` if that is preferred). There were only additions needed to make the trie catalog state reconstructable from `TrieDetails`.
   - TrieState - LIVE | NASCENT | GARBAGE - the state of the trie as of the latest block
   - ~`as_of_system_time` - the last system time the trie state changed (or created if there was no change)~
   - `garbage_as_of`- the system time this trie become garbage (if garbage) otherwise non set
- ~Each trie in the trie catalog now has a `as-of` key which indicates the time the trie changed state or was created. This `as-of` time is a system time of a `TriesAdded` message from the log. 
When a trie `trie-particular` enters the trie catalog it will first have an `as-of` time that corresponds to the system time of the `TriesAdded` message. Later `t-particular` might become `live` (in case it was initially `nascent`) or `garbage` (in case it was superseded).  `as-of` will then be the system time of the `TriesAdded` message that contained the trie `t-other` that made `t-particular` become `live` or `garbage`.~
- The actual GCing then looks at the earliest block to keep (currrentBlock - 9), takes the latest completed tx from that block and subtracts a `garbage-lifetime` (6 hours be default). Any garbage that is older than this timestamp will get collected.
    
I have so far not introduced a `TriesDeleted` message. The idea is that nodes will roughly garbage collect files at the same time. The first node that collects the garbage will delete the files from object storage and from the local caches. Subsequent nodes will garbage collect only the local versions and the deletion in OS will be a noop.

Todo:
- [x] The fast-pathing of tries in the live-index needs to be dealt with differently. 
      - ~Option 1: Don't include the latest blocks in the block files.~
      - ~Option 2: Wait for the acknowledgement of the tries added message, but not on the hot path (see #4825).~
      - Option 3: Live index uses `latest-completed-tx` time as the `as-of` time for the trie-catalog. 

- [x] ~I am using `instant-source` also for TriesAdded messages (mainly to deterministically create `as-of` timestamps) . The downside is that testing in other situations is a lot more confusing (the mock clock loses a tick for `finish-block` for example) and is also responsible for most of the test file changes in this PR. Should we do something different?~
Added an extra option to `LocalLog`


core/src/main/clojure/xtdb/indexer/live_index.clj
```diff
@@ -173,12 +172,12 @@
                                                           (catch Exception _
                                                             (throw (.exception ^StructuredTaskScope$Subtask %)))))))
                                    (util/rethrowing-cause))]
-            (let [added-tries (for [[table-name {:keys [trie-key data-file-size trie-metadata]}] table-metadata]
-                                (trie/->trie-details table-name trie-key data-file-size trie-metadata))]
+            (let [added-tries (for [[table-name {:keys [trie-key data-file-size trie-metadata state]}] table-metadata]
+                                (trie/->trie-details table-name trie-key data-file-size trie-metadata state))
+                  ;; TODO this will halt the indexer again so needs to be changes
+                  ^Log$MessageMetadata message-metadata @(.appendMessage log (Log$Message$TriesAdded. Storage/VERSION added-tries))]
```
FiV0: This `deref` can't stay. See #4285.

core/src/main/clojure/xtdb/trie_catalog.clj
```diff
@@ -244,38 +273,67 @@
...
                         added-tries))))
 
   (getTableNames [_] (set (keys !table-cats)))
 
+  (garbageCollectTries [_ table-name as-of]
+    (doseq [{:keys [trie-key]} (garbage-tries (.get !table-cats table-name) as-of)]
+      (.deleteIfExists buffer-pool (Trie/dataFilePath table-name trie-key)))
+    (.compute !table-cats table-name
+              (fn [_table-name tries]
+                (remove-garbage tries as-of))))
```
FiV0: I am doing this in a two step approach, to make sure stuff is really deleted before removing it from the trie catalog. Might be a bit inefficient.

core/src/main/kotlin/xtdb/api/log/LocalLog.kt
```diff
@@ -116,9 +116,7 @@ class LocalLog(rootPath: Path, private val instantSource: InstantSource, overrid
         try {
             val res = Array(msgs.size) { idx ->
                 val (msg) = msgs[idx]
-                // we only use the instantSource for Tx messages so that the tests
-                // that check files can be deterministic
-                val ts = if (msg is Message.Tx) instantSource.instant() else Instant.now()
+                val ts =  instantSource.instant()
```
FiV0: This is what creates most test file changes and might also confuse other people testing, but it makes my particular test scenario easier.

core/src/main/kotlin/xtdb/api/GarbageCollectorConfig.kt
```diff
@@ -9,9 +9,11 @@ import java.time.Duration
 data class GarbageCollectorConfig(
     var enabled: Boolean = true,
     var blocksToKeep: Int = 10,
+    var garbageLifetime: Duration = Duration.ofHours(6),
```
jarohen: I know it's an arbitrary number, but I'd probably stick this one closer to a day - a day's worth of garbage still won't be the end of the world :slightly_smiling_face:

core/src/main/proto/xtdb/log/proto/log.proto
```diff
@@ -35,9 +35,17 @@ message TrieMetadata {
     bytes iid_bloom = 8;
 }
 
+enum TrieState {
+    LIVE = 0;
```
jarohen: pedantic: but I'd probably go nascent -> live -> garbage here

core/src/main/clojure/xtdb/indexer/live_index.clj
FiV0: Use `latest-completed-tx` time in the live-index.

core/src/main/kotlin/xtdb/api/log/LocalLog.kt
FiV0: Add a testing flag to `LocalLog` to use the `instantSource` for everything.

FiV0:
This is currently blocked by https://github.com/xtdb/xtdb/issues/4545

FiV0:
My feeling is that the `live+nascent` + `garbage` split is quite brittle (in the sense it's quite easy to mess something up without it being caught by tests or being immediately obvious). For example the implicit assumption that everything is sorted by block index descending.

core/src/main/clojure/xtdb/garbage_collector.clj
```diff
@@ -1,26 +1,35 @@
...
   (:import [xtdb.api Xtdb$Config GarbageCollectorConfig]
            [xtdb.garbage_collector GarbageCollector]))
 
-(defmethod xtn/apply-config! :xtdb/garbage-collector [^Xtdb$Config config _ {:keys [enabled? blocks-to-keep approx-run-interval]}]
+(defmethod xtn/apply-config! :xtdb/garbage-collector [^Xtdb$Config config _
+                                                      {:keys [enabled? blocks-to-keep garbage-lifetime approx-run-interval]}]
   (.garbageCollector config
                      (cond-> (GarbageCollectorConfig.)
-                       enabled? (.enabled enabled?)
+                       (not (nil? enabled?)) (.enabled enabled?)
```
jarohen: nit: `some?`

core/src/main/clojure/xtdb/trie_catalog.clj
jarohen: I think there's a race here - you're calculating the garbage tries at time T1, deleting those files from the object-store, then recalculating the garbage-tries at time T2 to remove from the trie-cat? need to re-use the same list.

core/src/main/clojure/xtdb/trie_catalog.clj
```diff
@@ -261,38 +286,68 @@
...
+
+(def ^:dynamic *force-old-trie-details*)
+
 (defmethod ig/init-key :xtdb/trie-catalog [_ {:keys [^BufferPool buffer-pool, ^BlockCatalog block-cat]}]
   (log/debug "starting trie catalog...")
   (let [[_ table->table-block] (table-cat/load-tables-to-metadata buffer-pool block-cat)
-        cat (TrieCatalog. (ConcurrentHashMap.) *file-size-target*)]
-    (doseq [[table-name {:keys [tries]}] table->table-block]
-      (.addTries cat table-name tries))
+        cat (if (and (some-> table->table-block first val :tries first new-trie-details?) *force-old-trie-details*)
```
jarohen: I'm uneasy about having quite a large fork here just for testing purposes - this makes sense when you're just adjusting thresholds, say, but once you're adding different code-paths you start to run the risk of deviations between code-under-test and real code. Let's see what we can do to reduce the gap?

core/src/main/proto/xtdb/log/proto/log.proto
```diff
@@ -35,9 +35,17 @@ message TrieMetadata {
...
+    GARBAGE = 2;
+}
+
 message TrieDetails {
     string table_name = 1;
     string trie_key = 2;
     int64 data_file_size = 3;
     TrieMetadata trie_metadata = 4;
+    TrieState trie_state = 5;
+    int64 garbage_as_of = 6;
```
jarohen, line 50: pbuf has a specific 'timestamp' type - not sure of its exact semantics but may be preferable to int64

src/test/clojure/xtdb/trie_catalog_test.clj
```diff
@@ -273,50 +284,53 @@
...
-      (let [cat (cat/trie-catalog node)]
-        (t/is (= #{"l00-rc-b03"
-                   "l01-rc-b02"
-                   "l02-rc-p0-b01"
-                   "l02-rc-p1-b01"
-                   "l02-rc-p2-b01"
-                   "l02-rc-p3-b01"}
-                 (->> (cat/current-tries (cat/trie-state cat "public/foo"))
-                      (into (sorted-set) (map :trie-key)))))))))
+  (doseq [val [true false]]
```
jarohen: try not to use clj.core fn names as var-names, save the next poor sod from wondering why it no longer works to call `(val map-entry)` within the `doseq` :sweat_smile: 

`force-old-trie-details`?

src/test/clojure/xtdb/trie_catalog_test.clj
```diff
@@ -344,3 +358,109 @@
...
+                    "tables/public$foo/meta/l00-rc-b00.arrow"
+                    "tables/public$foo/meta/l00-rc-b01.arrow"
+                    "tables/public$foo/meta/l00-rc-b02.arrow"
+                    "tables/public$foo/meta/l00-rc-b03.arrow"
+                    "tables/public$foo/meta/l01-rc-b01.arrow"
+                    "tables/public$foo/meta/l01-rc-b02.arrow"
+                    "tables/public$foo/meta/l01-rc-b03.arrow"]
+
+                   (->>
+                    (.listAllObjects bp)
```
jarohen: listAllObjects can optionally take a prefix, save you `filter`ing here

core/src/main/clojure/xtdb/trie_catalog.clj
jarohen: I think I could also use an explanation of what the flag does, I can't immediately spot it from the code :derp:

core/src/main/clojure/xtdb/trie_catalog.clj
jarohen: separately, shall we try to extract the side effects from here? i.e. not have the trie-cat responsible for actually deleting the files? would keep this component pure and be easier to test

aware that the two-step process may not make this as easy, might have to separate it out into 'getGarbageTries' and 'deleteTries'?

core/src/main/kotlin/xtdb/garbage_collector/GarbageCollector.kt
```diff
@@ -2,41 +2,68 @@ package xtdb.garbage_collector
...
-            while (isActive) {
-                try {
-                    LOGGER.debug("Starting block garbage collection")
-                    blockCatalog.garbageCollectBlocks(blocksToKeep)
-                    LOGGER.debug("Block garbage collection completed")
-                } catch (e: Exception) {
-                    LOGGER.warn("Block garbage collection failed", e)
-                } catch (e: Throwable) {
-                    throw RuntimeException("Error encountered during Block garbage collection: ", e)
+        if (enabled) {
```
jarohen: I think I'd be tempted to have a separate `start` method here now that it might not be enabled - more explicit than a flag - can then conditionally call that in the init-key

jarohen (changes requested):
Hey @FiV0, nice one :clap: test distinction is good, drawing a good line between unit tests of the exact business logic and e2e tests of the behaviour.

Few changes below to address, and probably a few to discuss, lmk when's a good time :slightly_smiling_face:

core/src/main/clojure/xtdb/trie_catalog.clj
FiV0: Actually, I think it's currently a bug. It should be `(not *force-old-trie-details*)`.

I can't take out the other code path because I need the `garbage-as-of` and `state` if I want to construct the trie-catalog state without applying the tries. The problem is that I don't know if the last block file is with old or new trie-details.

core/src/main/proto/xtdb/log/proto/log.proto
FiV0: Keep for now. See also `system-time`.

FiV0:
Ok. Good to go from my side for another round of review.

jarohen (approved):
neat, thanks @FiV0 :pray: 

thoughts on following up with a tries-deleted msg - benefits/tradeoffs?

FiV0:
> thoughts on following up with a tries-deleted msg - benefits/tradeoffs?

Upside. Likely arriving quicker at the exact same state of the trie catalog ( a lagging node would delete it's garbage quicker, less likelyhood of the error state where the file reference still exists, but it is not in the object store).

Downside. Another extra message to go through the log (log congestion), albeit the deletion can probably happen async.

Issue #4387, opened 2025-04-25
Memory leak on query with left-joins, generate_series, and window function


egg-juxt:
The query is self-contained, no data needed:

```
WITH systemCTE AS (
     FROM GENERATE_SERIES(1,10000) AS system(_id)
     SELECT
       system._id AS _id,
       ROW_NUMBER() OVER (PARTITION BY system._id) AS rn
   )
   FROM
     systemCTE AS system
       LEFT JOIN systemCTE AS load2 USING (_id)
       LEFT JOIN GENERATE_SERIES(1,10000) AS gen(t) ON TRUE
   SELECT
     system.nmi AS nmi
   LIMIT 10
```

XTDB logs report leak:
```
java.lang.IllegalStateException: Memory was leaked by query. Memory leaked: (134217720)
Allocator(BoundQuery/openCursor) 0/134217720/404160488/9223372036854775807 (res/actual/peak/limit)

	at org.apache.arrow.memory.BaseAllocator.close(BaseAllocator.java:504)
	at xtdb.util$close.invokeStatic(util.clj:52)
	at xtdb.util$close.invoke(util.clj:35)
	at xtdb.query$wrap_cursor$reify__40293.close(query.clj:105)
	at xtdb.util$close.invokeStatic(util.clj:52)
	at xtdb.util$close.invoke(util.clj:35)
	at xtdb.query$open_cursor_as_stream$fn__40406.invoke(query.clj:347)
	at clojure.lang.AFn.run(AFn.java:22)
	at java.base/java.util.stream.Streams$1.run(Streams.java:845)
	at java.base/java.util.stream.AbstractPipeline.close(AbstractPipeline.java:323)
	at xtdb.api$plan_q$reify__20378.reduce(api.clj:54)
	at clojure.core$reduce.invokeStatic(core.clj:6964)
	at clojure.core$into.invokeStatic(core.clj:7038)
	at clojure.core$into.invoke(core.clj:7029)
	at xtdb.api$q.invokeStatic(api.clj:115)
	at xtdb.api$q.invoke(api.clj:69)
	at xtdb.api$q.invokeStatic(api.clj:112)
	at xtdb.api$q.invoke(api.clj:69)
	at scenarios.commons$query_BANG_.invokeStatic(commons.clj:132)
	at scenarios.commons$query_BANG_.doInvoke(commons.clj:129)
	at clojure.lang.RestFn.invoke(RestFn.java:413)
[...]

PR #4539, opened 2025-06-16
Add batched list operator, use for generate_series.

### LINKED ISSUE #4412: Make `GENERATE_SERIES` work in batches
Currently the query
```sql
FROM GENERATE_SERIES(1,200000000) AS system(_id)
LIMIT 10
```
fails with 
```
1. Caused by io.netty.util.internal.OutOfDirectMemoryError
   failed to allocate 536870880 byte(s) of direct memory (used:
   8183422486, max: 8401190912)
```
this makes sense looking at the plan:
```clj
[:top
  {:skip nil, :limit 10}
  [:project
   [{_id system.1/_id}]
   [:table #:system.1{_id (generate_series 1 200000000 1)}]]]
```
as the EE needs generate the whole series. Purely from SQL syntax this is hard to guess, as one would have expected this to work in batches as with a standard table.

danmason:
Resolves #4412 

Github action runs here: https://github.com/danmason/xtdb/actions/workflows/build.yml?query=branch%3Abatched-list-operator

Adds in a batched list operator into XTDB, using it from the SQL planner to handle generate_series. Adds a set of tests around the behavior of this. 

### Note on GENERATE_SERIES for Timestamps

This would still an issue for timestamps, as we evaluate generate_series for those eagerly, for example:

```
;; Would generate 3.6 billion timestamps (3600 seconds × 1,000,000)
  (t/is (= [{:ts #xt/date-time "2020-01-01T00:00:00.000Z"}
            {:ts #xt/date-time "2020-01-01T00:00:00.001Z"}
            {:ts #xt/date-time "2020-01-01T00:00:00.002Z"}
            {:ts #xt/date-time "2020-01-01T00:00:00.003Z"}
            {:ts #xt/date-time "2020-01-01T00:00:00.004Z"}]
           (xt/q tu/*node* "SELECT ts FROM generate_series(TIMESTAMP '2020-01-01 00:00:00', TIMESTAMP '2020-01-01 01:00:00', INTERVAL 'PT0.000001S') AS t(ts) LIMIT 5"))
        "timestamp generate_series with microsecond interval and LIMIT should not blow memory")
```

Will still throw an OoM - within ts_series itself.

### Note on #4387 

I wrote the following test against generate_series on main based on the issue, getting it to reliably throw an out of memory against a 6GB Heap by upping the sizes of GENERATE_SERIES:
```
(t/deftest test-oom-on-left-joined-generate-series-4387
  (t/is (= [{:xt/id 1 :rn 1}]
           (xt/q tu/*node* "WITH systemCTE AS (SELECT system._id AS _id, ROW_NUMBER() OVER (PARTITION BY system._id) AS rn FROM GENERATE_SERIES(1,100000) AS system(_id)) SELECT system.* FROM systemCTE AS system LEFT JOIN systemCTE AS load2 USING (_id) LEFT JOIN GENERATE_SERIES(1,100000) AS gen(t) ON TRUE LIMIT 1;"))))
```

Following this change - this _still_ throws an OoM, so not certain these changes would necessarily fixed the observed memory leak.

### DIFF
--- a/core/src/main/clojure/xtdb/logical_plan.clj
+++ b/core/src/main/clojure/xtdb/logical_plan.clj
@@ -184,6 +184,14 @@
     (mapv symbol (if (map? table)
                    (keys table)
                    (keys (first table))))
+            
+    [:list explicit-column-names _]
+    (vec explicit-column-names)
+    
+    [:list list]
+    (mapv symbol (if (map? list)
+                   (keys list)
+                   (keys (first list))))
 
     [:scan _scan-opts columns]
     (mapv ->projected-column columns)
--- a/core/src/main/clojure/xtdb/operator/list.clj
+++ b/core/src/main/clojure/xtdb/operator/list.clj
@@ -1,0 +1,62 @@
+(ns xtdb.operator.list
+  (:require [clojure.spec.alpha :as s]
+            [xtdb.expression :as expr]
+            [xtdb.expression.list :as expr-list]
+            [xtdb.logical-plan :as lp]
+            [xtdb.types :as types]
+            [xtdb.util :as util]
+            [xtdb.vector.reader :as vr]
+            [xtdb.vector.writer :as vw])
+  (:import (org.apache.arrow.vector.types.pojo Field)
+           (org.apache.arrow.memory BufferAllocator)
+           (xtdb ICursor)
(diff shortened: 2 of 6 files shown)


jarohen (approved):
:100:

PR #4546, opened 2025-06-19
Simplify join construction via emit-join-expr

danmason:
Github action runs: https://github.com/danmason/xtdb/actions/workflows/build.yml?query=branch%3Ajoin-tidy-up++

This PR restructures emit-join-expr to simplify join operator construction.

Previously, emit-join-expr required a nested function to construct the join cursor, leading to a back-and-forth passing around cursors and duplicated logic across join types.

With this change, emit-join-expr now takes a map of options (e.g. :build-side, :join-type, a field merging function, and whether to construct pushdown bloom filters, etc), and each join is represented based on these.  Moves all the logic for cursor construction and field projection within emit-join-expr.

### TPCH run results

Ran using `./gradlew tpch -PscaleFactor=0.1`, summary of times taken:

On `main` ([11bb869](https://github.com/xtdb/xtdb/commit/11bb8697f525f238a9b133bf139b2a98e0136f45)):
```
ingest: 20984 ms
cold-queries: 74765 ms
hot-queries: 67021 ms
```

On this branch (based on `main`):
```
ingest: 19928ms
cold-queries: 68332ms
hot-queries: 65092ms
```

### DIFF
--- a/core/src/main/clojure/xtdb/operator/join.clj
+++ b/core/src/main/clojure/xtdb/operator/join.clj
@@ -392,7 +392,11 @@
        (into {} (map (comp (juxt #(symbol (.getName ^Field %)) identity)
                            #(.getField ^ProjectionSpec %))))))
 
-(defn- emit-join-expr {:style/indent 2} [{:keys [condition left right]} {:keys [param-fields] :as _args}]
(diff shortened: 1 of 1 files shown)


jarohen (approved):
well done @danmason, a significant simplification :clap:

PR #4582, opened 2025-07-09
Pick build side on left outer join based on smaller relation.

danmason:
Github action builds: https://github.com/danmason/xtdb/actions/workflows/build.yml?query=branch%3Aleft-outer-join-flip-sides

### Support for building left outer joins from either side
- Planner now selects the join build/probe sides based on row count statistics:
  - If we do not know the size of a relation based on row-count, we default to Long/MAX_VALUE for it's size.
  - If the **right** side is smaller (or equal), default to building the right-hand side (same as current behavior).
  - If the **left** side is smaller, flip the join sides and build the left-hand side instead.
    - Makes use of `matched-build-idxs` / `unmatched-build-idxs` to ensure all values on the build side (the left side) are returned.
    - A `side-flipped?` flag is passed to:
      - Ensure correct output flipping in `join-rels`
      - Enable correct probing logic in `probe-outer-join-select`
   - We _also_ make use of pushdown-blooms, as these are applicable in the case where we dont NEED to keep all of the probe side.
- In the `unmatched-build-idxs` path:
  - The `nil-row-idx` entry is removed **only if** the sides are **not flipped** (i.e., when using `with-nil-row`, as in `full-outer-join`).

## PATCH performance notes

Doing some testing against this locally - mainly checking what kind of memory usage we see SPECIFICALLY on the build side with yourkit:
```
;; Submit docs
  (def node (xtn/start-node {:log [:local {:path "patch-test"}]
                             :storage [:local {:path "patch-test"}]}))
  (run! (fn [n]
          (xt/submit-tx node (mapv (fn [i]
                                          [:put-docs :foo {:xt/id (+ i n), :a "a", :b "b"}])
                                        (range 0 1000))))
        (range 0 500000 1000))
  (tu/then-await-tx node)

  ;; With YOURKIT running, memory profiling:
  ;; Memory usage - of build-phase.

  ;; MAIN
  (time (xt/execute-tx node ["PATCH INTO foo RECORDS {_id: 50001, c: 'c'}"]))
  ;; Memory usage of build side (approx): 139MiB
  (time (xt/execute-tx node ["PATCH INTO foo RECORDS {_id: 'foo1', c: 'c'}"]))
  ;; Memory usage of build side (approx): 140MiB

  ;; After LOJ changes (with pushdown blooms)
  (time (xt/execute-tx node ["PATCH INTO foo RECORDS {_id: 70000, c: 'c'}"]))
  ;; Memory usage (approx): 3KiB
  (time (xt/execute-tx node ["PATCH INTO foo RECORDS {_id: 'baz', c: 'c'}"]))
  ;; Memory usage (approx): 3KiB
```

### DIFF
--- a/core/src/main/clojure/xtdb/operator/join.clj
+++ b/core/src/main/clojure/xtdb/operator/join.clj
@@ -178,7 +178,7 @@
 
 #_{:clj-kondo/ignore [:unused-binding]}
 (defmulti ^xtdb.arrow.RelationReader probe-phase
-  (fn [join-type probe-rel rel-map matched-build-idxs]
+  (fn [join-type probe-rel rel-map matched-build-idxs side-flipped?]
     join-type))
 
 (defn- probe-inner-join-select
@@ -205,10 +205,14 @@
   "Takes a relation (probe-rel) and its mapped relation (via rel-map) and returns a relation with the columns of both."
   [^RelationReader probe-rel
    ^IRelationMap rel-map
-   [^ints probe-sel, ^ints build-sel :as _selection-pair]]
-  (let [built-rel (.getBuiltRelation rel-map)]
-    (vr/rel-reader (concat (.select built-rel build-sel)
-                           (.select probe-rel probe-sel)))))
+   side-flipped?
+   [^ints probe-sel, ^ints build-sel :as _selection]]
+  (let [built-rel (.getBuiltRelation rel-map)
+        selected-probe-rel (.select probe-rel probe-sel)
+        selected-build-rel (.select built-rel build-sel)]
+    (vr/rel-reader (if side-flipped? 
+                     (concat selected-probe-rel selected-build-rel)
+                     (concat selected-build-rel selected-probe-rel)))))
 
 (defn- probe-semi-join-select
   "Returns a single selection of the probe relation, that represents matches for a semi-join."
@@ -224,18 +228,20 @@
   [_join-type
    ^RelationReader probe-rel
    ^IRelationMap rel-map
-   _matched-build-idxs]
-  (join-rels probe-rel rel-map (probe-inner-join-select probe-rel rel-map)))
+   _matched-build-idxs
+   _side-flipped?]
+  (join-rels probe-rel rel-map false (probe-inner-join-select probe-rel rel-map)))
 
 (defmethod probe-phase ::semi-join
   [_join-type
    ^RelationReader probe-rel
    ^IRelationMap rel-map
-   _matched-build-idxs]
+   _matched-build-idxs
+   _side-flipped?]
   (.select probe-rel (probe-semi-join-select probe-rel rel-map)))
 
 (defmethod probe-phase ::anti-semi-join
-  [_join-type, ^RelationReader probe-rel, ^IRelationMap rel-map, _matched-build-idxs]
+  [_join-type, ^RelationReader probe-rel, ^IRelationMap rel-map, _matched-build-idxs, _side-flipped?]
   (let [rel-map-prober (.probeFromRelation rel-map probe-rel)
         matching-probe-idxs (IntStream/builder)]
     (dotimes [probe-idx (.getRowCount probe-rel)]
@@ -248,7 +254,8 @@
   [_join-type
    ^RelationReader probe-rel
    ^IRelationMap rel-map
-   _matched-build-idxs]
+   _matched-build-idxs
+   _side-flipped?]
   (.select probe-rel (probe-semi-join-select probe-rel rel-map)))
 
 (defn- int-array-concat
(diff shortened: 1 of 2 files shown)


danmason:
I've spent this morning evaluating left-side build behavior in left outer joins for the PATCH case, some findings from that:
- Memory usage on the build side is drastically reduced - for 500k docs, it’s ~140MiB on main, and just ~3KiB on my branch.
- Average timings (patching both an existing and non-existing doc) are currently about the same regardless of which side is built - expected, since no filtering was applied on the probe side yet.
- When building the left side, we can apply pushdown bloom filters to the probe side. Having done this, however, the timings still don’t improve - which was surprising.

```
(t/deftest patch-performance-test
  (util/with-tmp-dirs #{node-tmp-dir}
    (let [local-node (xtn/start-node {:log [:local {:path (.resolve node-tmp-dir "log")}]
                                      :storage [:local {:path (.resolve node-tmp-dir "storage")}]})]
      (log/info "Inserting 100K Docs")
      (run! (fn [n]
              (xt/submit-tx local-node (mapv (fn [i]
                                               [:put-docs :foo {:xt/id (+ i n), :a (str "a" (+ i n))}])
                                             (range 0 1000))))
            (range 0 100000 1000))
      (tu/then-await-tx local-node)
      (log/info "Inserted 100K Docs")

      (log/info "Patching 10 existing docs...")
      (let [existing-times (doall
                            (for [i (range 10)]
                              (let [id (+ 50000 i)
                                    start-time (System/nanoTime)]
                                (xt/execute-tx local-node
                                               [(str "PATCH INTO foo RECORDS {_id: " id ", c: 'c" i "'}")])
                                (/ (- (System/nanoTime) start-time) 1000000.0))))]
        (log/info (str "Average time for existing docs: "
                       (/ (reduce + existing-times) (count existing-times)) "ms")))
      
      (log/info "Patching 10 non-existing docs...")
      (let [non-existing-times (doall
                                (for [i (range 10)]
                                  (let [id (str "foo" i)
                                        start-time (System/nanoTime)]
                                    (xt/execute-tx local-node
                                                   [(str "PATCH INTO foo RECORDS {_id: '" id "', c: 'c" i "'}")])
                                    (/ (- (System/nanoTime) start-time) 1000000.0))))]
        (log/info (str "Average time for non-existing docs: "
                       (/ (reduce + non-existing-times) (count non-existing-times)) "ms")))

      (t/is (= [{:xt/id 50000, :a "a50000", :c "c0"}]
               (xt/q local-node "SELECT * FROM foo WHERE _id = 50000")))

      (t/is (= [{:xt/id "foo0", :c "c0"}]
               (xt/q local-node "SELECT * FROM foo WHERE _id = 'foo0'"))))))
;; =>
;; Average Time taken for PATCH on existing doc before LOJ changes (main): 252ms
;; Average Time taken for PATCH on non existing doc before LOJ changes (main): 272ms
;; Time taken for PATCH on existing doc after LOJ changes (without pushdown blooms): 254ms
;; Time taken for PATCH on non existing doc after LOJ changes (without pushdown blooms): 219ms
;; Time taken for PATCH on existing doc after LOJ changes (with pushdown blooms): 290ms
;; Time taken for PATCH on non existing doc after LOJ changes (with pushdown blooms): 239ms
```

After digging into it, I suspect pushdown blooms aren't working as intended in this case. Specifically:
- The bloom filters are built using the build-key-col-name, which (in the patch case) is a renamed field: new/_iid.
- But in the scan (probe side), we’re filtering on the plain _iid column.
- I THINK the bloom filter likely isn’t being applied correctly in that case - though I'm not entirely sure on their expected behaviour?

### UPDATE
Reading through the code, all of the above seemed to be working as intended and the pushdown blooms WERE being applied - I suspect it may be the case that we're not seeing a huge gain from them in the original performance test that I did due to both being a relatively small amount of data AND not guaranteed that compaction had ran yet, so they may not have been gaining much performance wise.

Given that merging this should help ease some memory pressure, I don't think speed tests are particularly blocking - though can follow up with further performance testing against a larger dataset following some compaction.

jarohen (approved):
This looks great @danmason, well done 👏 

Might be a subjective personal aversion to boolean flags - might this involve less flag-checking/more intentful to introduce another join-type? e.g. `:left-outer-join-flipped`?

danmason:
One last note on PATCH performance following these changes:

Ran a test where I:

- Inserted 2 million docs.
- Synced the node.
- Ran compact-all with a timeout of 5 minutes.
- Ran two PATCH txes - one patching an existing doc in the inserted docs, one patching one that didnt exist. 

```
Times from those:
;; Average Time taken for PATCH on existing doc before LOJ changes (main): 9253ms
;; Average Time taken for PATCH on non existing doc before LOJ changes (main): 9389ms
;; Time taken for PATCH on existing doc after LOJ changes (with pushdown blooms): 6772ms
;; Time taken for PATCH on non existing doc after LOJ changes (with pushdown blooms): 8087ms
```

Not quite the rapid speed up I would have HOPED for, but perhaps I need a better understanding of pushdown blooms/there are more easy wins to be had here. Worth pointing a profiler against it, and perhaps building an edge version we can test on the Synergy simulator data.

As mentioned, going to get my PR merged for the difference in memory usage alone. I have unlinked "efficiency of PATCH" from this PR for now, as I would still like to work on it's performance.

PR #4583, opened 2025-07-09
Helm terraform private networking

### LINKED ISSUE #4566: Helm/Terraform to default to private networking
Modify all xtdb kubernetes service templates from LoadBalancer to ClusterIP - means that the services will no longer be accessible over the open internet but remain accessible via kubectl port-forward or similar.

tggreene:
This PR just changes LoadBalancer service to ClusterIP which means services won't be exposed publicly by default.

Docs updated to reflect change and use of kubectl port-forward

Closes #4566

### DIFF
--- a/docs/src/content/docs/ops/guides/starting-with-aws.adoc
+++ b/docs/src/content/docs/ops/guides/starting-with-aws.adoc
@@ -306,7 +306,7 @@ For more information on the available configuration options and fetching the cha
 
 NOTE: As it will take some time for the XTDB nodes to be marked as ready (as they need to pass their initial startup checks) it may take a few minutes for the XTDB cluster to be accessible.
 
-NOTE: The xtdb service is only available via ClusterIP be default so as to not expose the service publicly
+NOTE: The xtdb service is only available via ClusterIP by default so as to not expose the service publicly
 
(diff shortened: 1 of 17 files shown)


docs/src/content/docs/ops/guides/starting-with-aws.adoc
```diff
@@ -304,28 +304,29 @@ For more information on the available configuration options and fetching the cha
...
-Once the XTDB cluster is up and running, you can access it via the LoadBalancer service that was created.
+NOTE: The xtdb service is only available via ClusterIP be default so as to not expose the service publicly
 
-To get the external IP of the LoadBalancer service, run:
+Once the XTDB cluster is up and running, you can access it via the ClusterIP service that was created.
+
+To port forward the service locally
 ```bash
-kubectl get svc xtdb-service --namespace xtdb-deployment
+kubectl port-forward service/xtdb-service --namespace xtdb-deployment 8080:8080
```
danmason, line 315: Could you update the instructions within the various NOTES.txt files too? See: https://github.com/xtdb/xtdb/blob/main/aws/helm/templates/NOTES.txt

danmason (changes requested):
Added a quick note about another bit of documentation to update here, but once we're happy this works on all of the cloud vendors should be good to go :)

PR #4594, opened 2025-07-14
Add user declaration to docker files

### LINKED ISSUE #4086: Ensure XTDB docker images can run as non-root user.
As per our previous investigations into the trivy security actions (issue #3953), we had a few lingering security actions we could not action:

> The following three actions we will have to leave out:
> 
>     Runs as root user [MEDIUM]
>     Runs with GID <= 10000 [LOW]
>     Runs with UID <= 10000 [LOW]
> 
> I believe to run as a non-root user we'd need to move where some of the files are stored within the dockerfiles, so I've left these off.

We could not set the numerous pieces of recommended `SecurityContext` config as a result of this - and we have now seen this issue again causing issues when attempting to run an unprivileged node in podman - #4084 

* [ ] check the repro there before marking this one resolved

We want to ensure that _all_ of the XTDB images are configured with sensible storage paths/configured volumes and can be ran as a non-root user:
- `xtdb-standalone` will need to setup sensible volume mounts/directories for local directory log/bufferpool. 
- The cloud images will need to setup sensible volume mounts/directories for the local disk cache.

tggreene:
Very simple change, just chown /var/lib/xtdb and switch user. Also makes the uid and gid >= 10000. Sets security context runAsUser, runAsGroup to 10000.

Resolves #4086.

For release notes:

## Docker Running as a Non-Root User

To improve security and align with container best practices, the XTDB container now runs as a non-root user by default (UID 20000).

If you’re mounting a host volume (e.g. `/tmp/xtdb-data-dir:/var/lib/xtdb`), ensure the container has permission to write to it:

Docker users:
```
sudo chown -R 20000:20000 /tmp/xtdb-data-dir
```

Podman users:
Ensure the directory is owned by your user, and run with `--userns=keep-id`:

```
podman run -it --pull=always \
  --userns=keep-id \
  -p 5432:5432 \
  -p 8080:8080 \
  -v /tmp/xtdb-data-dir:/var/lib/xtdb \
  ghcr.io/xtdb/xtdb
```

### DIFF
--- a/aws/helm/templates/statefulset.yaml
+++ b/aws/helm/templates/statefulset.yaml
@@ -24,6 +24,8 @@ spec:
         {{- toYaml .Values.tolerations | nindent 8 }}
       affinity:
         {{- toYaml .Values.affinity | nindent 8 }}
+      securityContext:
+        {{- toYaml .Values.securityContext | nindent 12 }}
       volumes: 
         - name: "tmp"
           emptyDir: {}
@@ -65,8 +67,6 @@ spec:
             - name: {{ $key }}
               value: {{ $value | quote }}
             {{- end }}
-          securityContext:
-            {{- toYaml .Values.securityContext | nindent 12 }}
           startupProbe:
             {{- toYaml .Values.startupProbe | nindent 12 }}
           livenessProbe:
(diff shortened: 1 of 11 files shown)


tggreene:
I've checked the docker image/security context against a running cluster

danmason (approved):
Looking good to me, especially if it's been tested against a running cluster - keen for us to try running trivy/security checks against one of the images/helm charts again, see if that clears most of our security vulns :)

danmason (approved):
Good stuff :)

PR #4621, opened 2025-07-22
Fix a few typos in the SQL overview doc

anderseknert:
Otherwise great docs!


refset:
Thanks @anderseknert :pray:

PR #4622, opened 2025-07-23
Pass _iid pushdown bloom to ScanCursor, use to filter rows.

### LINKED ISSUE #4555: Improve efficiency of PATCH
A Design Partner has identified long-running ingestion statements, probably PATCH statements.

We have been able to reproduce a long-running PATCH operation locally, on a big loca dataset. In Clojure:

```
(time
    (execute-tx! ["PATCH INTO system_active_power_reading_w RECORDS {_id: 'patch_test'}"]))
"Elapsed time: 16734.788253 msecs"
```

We suspect that the :patch operation could be executing an inefficient hash join. Plan for reference:
```
[:patch
 {:table public/system_active_power_reading_w}
 [:project
  [{_iid new/_iid}
   {_valid_from (coalesce old/_valid_from nil (current-timestamp))}
   {_valid_to (coalesce old/_valid_to nil xtdb/end-of-time)}
   {doc (_patch old/doc new/doc)}]
  [:left-outer-join
   [#:new{_iid old/_iid}]
   [:rename
    new
    [:project
     [{_iid (_iid _id)} {doc {:_id _id}}]
     [:project [{_id xt.values.1/_id}] [:rename xt.values.1 [:table [_id] [{:_id "patch_test"}]]]]]]
   [:rename
    old
    [:patch-gaps
     {:valid-from nil, :valid-to nil}
     [:project
      [_iid _valid_from _valid_to {doc {}}]
      [:order-by
       [[_iid] [_valid_from]]
       [:scan
        {:table public/system_active_power_reading_w, :for-valid-time [:in nil nil]}
        [_iid _valid_from _valid_to]]]]]]]]]
```

danmason:
## Summary

Resolves #4555

This PR adds support for pushing down `_iid` info from JOIN into scan, and using that within `ScanCursor` to filter rows more efficiently. This speeds up operations like `PATCH` by avoiding unnecessary row processing when `_iid` values are known ahead of time.

### Implementation

- If a single `_iid` is present (ie, single patch join), we construct an `IidSelector` and apply `maybeSelect` for fast binary search filtering.
- If multiple `_iid`s are present, we filtering by the _iid bloom filters.
- This hybrid approach gives us a fast path for single-record PATCH operations, while also speeding up multi document patches significantly.

---

## PATCH Performance

The above comparisons are captured using the PATCH benchmark - we take averages for each of these from ten runs, also using execute-tx to ensure it has been fully indexed (which adds it's own overhead):

### PATCHING against 500k docs on main:
```
patching a single existing document: 3636.05ms
patching a single non-existing document: 3742.57ms
patching multiple (4) existing documents: 4000.65ms
```

### PATCHING against 500k docs on this branch:
```
patching a single existing document: 27.27 ms
patching a single non-existing document: 15.66 ms
patching multiple (4) existing documents: 144.33 ms
```

### PATCHING against 2million docs on this branch:
```
patching a single existing document: 37.88 ms
patching a single non-existing document: 23.06 ms
patching multiple (4) existing documents: 1357.25 ms
```

This doesn't even run on `main` due to `/tmp` exhaustion during sort. Additionally, it shows that the table size doesn't have a massive performance impact on this branch.

### DIFF
--- a/core/src/main/clojure/xtdb/operator/join.clj
+++ b/core/src/main/clojure/xtdb/operator/join.clj
@@ -174,7 +174,7 @@
                                    build-col (.vectorForOrNull build-rel (str build-col-name))
                                    ^MutableRoaringBitmap pushdown-bloom (nth pushdown-blooms col-idx)]
                                (dotimes [build-idx (.getRowCount build-rel)]
-                                 (when (and iid-set (string/includes? (str build-col-name) "iid"))
+                                 (when (and iid-set (= (name build-col-name) "_iid"))
                                    (.add iid-set (.getBytes build-col build-idx)))
                                  (.add pushdown-bloom ^ints (BloomUtils/bloomHashes build-col build-idx))))))))))
 
@@ -342,10 +342,12 @@
     (build-phase build-cursor rel-map pushdown-blooms iid-set)
 
     (boolean
-     (or (let [advanced? (boolean-array 1)]
+     (or (let [advanced? (boolean-array 1)
+               probe-iid-keys (filter #(= (name %) "_iid") (.probeKeyColumnNames rel-map))]
            (binding [scan/*column->pushdown-bloom* (cond-> scan/*column->pushdown-bloom*
                                                      (some? pushdown-blooms) (conj (zipmap (.probeKeyColumnNames rel-map) pushdown-blooms)))
-                     scan/*iid-set* iid-set]
+                     scan/*column->iid-set* (cond-> scan/*column->iid-set*
+                                              (some? iid-set) (conj (zipmap probe-iid-keys (repeat iid-set))))] 
              (when-not probe-cursor
                (util/with-close-on-catch [probe-cursor (->probe-cursor)]
                  (set! (.probe-cursor this) probe-cursor)))
--- a/core/src/main/clojure/xtdb/operator/join.clj
+++ b/core/src/main/clojure/xtdb/operator/join.clj
@@ -14,7 +14,7 @@
             [xtdb.vector.reader :as vr]
             [xtdb.vector.writer :as vw])
   (:import [clojure.lang IFn]
-           (java.util ArrayList Iterator List HashSet)
+           (java.util ArrayList HashSet Iterator List Set)
            (java.util.function IntConsumer)
            (java.util.stream IntStream)
            (org.apache.arrow.memory BufferAllocator)
@@ -160,7 +160,7 @@
 (defmethod lp/emit-expr :cross-join [join-expr args]
   (emit-cross-join (emit-join-children join-expr args)))
 
-(defn- build-phase [^ICursor build-cursor, ^IRelationMap rel-map, pushdown-blooms, ^HashSet iid-set]
+(defn- build-phase [^ICursor build-cursor, ^IRelationMap rel-map, pushdown-blooms, ^Set iid-set]
   (.forEachRemaining build-cursor
                      (fn [^RelationReader build-rel]
                        (let [rel-map-builder (.buildFromRelation rel-map build-rel)
@@ -335,7 +335,7 @@
                      ^IRelationMap rel-map
                      ^RoaringBitmap matched-build-idxs
                      pushdown-blooms
-                     ^HashSet iid-set
+                     ^Set iid-set
                      join-type]
   ICursor
   (tryAdvance [this c]
--- a/core/src/main/clojure/xtdb/operator/join.clj
+++ b/core/src/main/clojure/xtdb/operator/join.clj
@@ -1,7 +1,7 @@
 (ns xtdb.operator.join
   (:require [clojure.set :as set]
             [clojure.spec.alpha :as s]
-            [clojure.string]
+            [clojure.string :as string]
             [clojure.walk :as walk]
             [xtdb.error :as err]
             [xtdb.expression :as expr]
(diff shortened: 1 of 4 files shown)


danmason:
### Testing 1

Tried a quick and relatively hacky approach to avoid our specific hashing/roaring bitmap checking:
- Collected all the individual build-side IIDs as ByteBuffers into a Set.
- Passed that Set down through scan and into the ScanCursor.
- Then just checked for Set membership per row in the cursor.

Performance ended up being pretty much the same as when we did this using hashing/intersection:
```
patch-existing-average-ms: 146.16
patch-non-existing-average-ms: 92.99
```

I guess it’s not too surprising - Set.contains() still does hashing internally, so we’re not really bypassing much. 

### Testing 2

Decided to do a quick test against this by changing approach somewhat:
- We already have IidSelector. which we use to do binary search when we have an iid select in hand.
- Using the iid set I passed through to scan - took the first value off of it, and used it to construct an IidSelector.
- Remove any of the above set based testing or bloom hashing - just using the maybeSelect with the IidSelector.

That brings times for PATCH for 500k docs down from 140ms to around 27ms on existing docs, and 90ms down to 18ms on non existing docs (obviously, thats only against a single iid so not ideal and also incorrect if you have MORE than one iid, such as a patch of multiple records, but was useful to check)

### Testing 3

For the sake of comparison, decided to try a sort of mixed approach of both, so:
- In the case where I have a single iid in my pushdowns - I transform it into an IIDSelector, and use that to filter results with maybeSelect.
- Otherwise, I use a row by row pushdownbloom based filtering type approach. 

In a sense, that first step is sort of a "quick path" when I'm patching a single record, otherwise it falls back on pushdown bloom row level filtering. Perhaps unsurprisingly, see a big difference in performance dependent on whether or not I am patching multiple documents.

Patching against a table with 2million documents:
```
patching a single existing document: 37.88ms
patching a single non existing document: 23.06ms
patching multiple (4) existing documents: 1357.25ms
```

This will in large part be due to the iidselector performing a binary search when finding a single iid, so it has a lot less work to do.

core/src/main/clojure/xtdb/operator/join.clj
```diff
@@ -174,6 +174,8 @@
                                    build-col (.vectorForOrNull build-rel (str build-col-name))
                                    ^MutableRoaringBitmap pushdown-bloom (nth pushdown-blooms col-idx)]
                                (dotimes [build-idx (.getRowCount build-rel)]
+                                 (when (and iid-set (string/includes? (str build-col-name) "iid"))
```
jarohen: I'd probably go a little tighter here - assuming this is a symbol, could you try `(= (name build-col-name) "_iid")`?

danmason:
Spent some time trying to tidy this up a little further, though seemed to incur a performance regression in doing so.

Was attempting to replace some of the work in scan.clj, where currently:
- We pass down the iid-set from join.clj.
- We also grab the _iid pushdown-bloom, which is already set.
- If (count iid-set) = 1, we create an IIDSelector which gets passed as col-preds.
- If (count iid-set) > 1, we pass down the iid pushdown blooms to scan cursor.
- In ScanCursor, we either use the IIDSelector if present (fast path) to do a binary scan or filter row by row with the iid pushdown bloom.

Change/cleanup was to take pushdown blooms out of the loop and do some of this checking within Kotlin, so:
- Pass the iid-set, whatever it may be, directly into the scancursor.
- Call to maybeSelect with it - if it is not null and size == 1, take the first (only) entry, create an IidSelector and call select on it.
- Otherwise, do selects as before and when we go to filter row by row, if iid set is not null and size of the set >1, use it to filter rows by checking set.contains(iid)

For some reason, this made the fast path MUCH slower, which confused me since it was supposed to be an equivalence change. Not entirely sure what would cause that, so I've left it be for now.

core/src/main/clojure/xtdb/operator/join.clj
```diff
@@ -160,7 +160,7 @@
 (defmethod lp/emit-expr :cross-join [join-expr args]
   (emit-cross-join (emit-join-children join-expr args)))
 
-(defn- build-phase [^ICursor build-cursor, ^IRelationMap rel-map, pushdown-blooms]
+(defn- build-phase [^ICursor build-cursor, ^IRelationMap rel-map, pushdown-blooms, ^HashSet iid-set]
```
jarohen: nit: we usually type-hint these as `^Set`, i.e. use the interface rather than the concrete impl

core/src/main/clojure/xtdb/operator/join.clj
```diff
@@ -327,25 +329,29 @@
...
 
 (deftype JoinCursor [^BufferAllocator allocator, ^ICursor build-cursor,
-                     ^:unsynchronized-mutable ^ICursor probe-cursor 
+                     ^:unsynchronized-mutable ^ICursor probe-cursor
                      ^:unsynchronized-mutable ^RelationWriter nil-rel-writer
                      ^IFn ->probe-cursor
                      ^IRelationMap rel-map
                      ^RoaringBitmap matched-build-idxs
                      pushdown-blooms
+                     ^HashSet iid-set
```
jarohen: `^Set`

jarohen (approved):
sweet, cheers @danmason 🙏

danmason:
Ernesto saw some issues against PATCH:

Table that PATCH is applied against has a number of entries for each id, bounded by valid time (ie, each one 5 minutes between valid from and valid to, all entries had this bounded valid time). There were also GAPS between data (ie, one day's last entry was at 12pm, then next entry was at 11am the next day).

When PATCH was applied (with no temporal info set within the PATCH, ie, from now) it would fill in the gaps within the data and then add a final entry with an unbounded valid to. When then applied, PATCH worked as expected. 

Checking this against main - we see that the SAME problem occurs, which would suggest that the issue lies within the temporal gaps part of PATCH. Shall go ahead with merging this given that the PATCH went from taking 27seconds to 40ms, but will write up a minimal repro/test and an issue.

PR #4641, opened 2025-07-28
Introduced pg_type columns typcategory and typarray

### LINKED ISSUE #4629: Support `pg_type`'s `typcategory` and `typarray` for fully supporting PostgresJS init request
Until now, we have configured PostgresJS with the option `fetch_types: false`. But our current support of `pg_type` allows for enabling that option; the client will just receive some warnings on not-found columns.

From a test with `fetch_types: true`:
```
  console.log
    {
      severity: 'WARNING',
      severity_local: 'WARNING',
      code: '01000',
      message: 'Column not found: a.typcategory'
    }

      at onnotice (src/lib/xtdb-sql.service.ts:151:14)

  console.log
    {
      severity: 'WARNING',
      severity_local: 'WARNING',
      code: '01000',
      message: 'Column not found: b.typarray'
    }

      at onnotice (src/lib/xtdb-sql.service.ts:151:14)
```

The ability to enable `fetch_types` is more important than it seems, as it prevents an indefinite blocking of PostgresJS' `sql.reserve()` function (see https://github.com/porsager/postgres/issues/751), for which we are currently applying a workaround. However, the issue doesn't seem to be fully solved when the connection pool starts to grow, as `sql.reserve()` seems to be blocking indefinitely when it is the first operation on a newly opened connection.

egg-juxt:
For fully supporting PostgresJS init query.

### DIFF
--- a/core/src/main/clojure/xtdb/pgwire/types.clj
+++ b/core/src/main/clojure/xtdb/pgwire/types.clj
@@ -262,6 +262,8 @@
                 :col-type :i64
                 :oid 20
                 :typlen typlen
+                :typcategory :numeric
+                :typarray :_int8
                 :typsend "int8send"
                 :typreceive "int8recv"
                 :read-binary (fn [_env ba] (-> ba ByteBuffer/wrap .getLong))
@@ -277,6 +279,8 @@
                 :col-type :i32
                 :oid 23
                 :typlen typlen
+                :typcategory :numeric
+                :typarray :_int4
                 :typsend "int4send"
                 :typreceive "int4recv"
                 :read-binary (fn [_env ba] (-> ba ByteBuffer/wrap .getInt))
@@ -292,6 +296,7 @@
                 :col-type :i16
                 :oid 21
                 :typlen typlen
+                :typcategory :numeric
                 :typsend "int2send"
                 :typreceive "int2recv"
                 :read-binary (fn [_env ba] (-> ba ByteBuffer/wrap .getShort))
@@ -323,6 +328,7 @@
                   :col-type :f32
                   :oid 700
                   :typlen typlen
+                  :typcategory :numeric
                   :typsend "float4send"
                   :typreceive "float4recv"
                   :read-binary (fn [_env ba] (-> ba ByteBuffer/wrap .getFloat))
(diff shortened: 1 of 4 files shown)


lang/js/test/test.js
```diff
@@ -14,7 +14,7 @@ beforeEach (async () => {
     host: process.env.PG_HOST || "localhost",
     port: process.env.PG_PORT || 5439,
     database: uuid.v4().toString(),
-    fetch_types: false, // currently required https://github.com/xtdb/xtdb/issues/3607
+    fetch_types: true, // beware setting to false: https://github.com/porsager/postgres/issues/751
```
jarohen: if this is the default, let's drop it entirely

core/src/main/clojure/xtdb/pgwire/types.clj
```diff
@@ -254,6 +254,7 @@
        :default {:typname "default"
                  :col-type :utf8
                  :oid 0
+                 :typcategory :z-internal
```
jarohen: let's exclude `:default` from the pg_type list, then we shouldn't need to invent `:z-internal`

core/src/main/clojure/xtdb/pgwire/types.clj
```diff
@@ -367,6 +376,7 @@
                 :col-type :uuid
                 :oid 2950
                 :typlen typlen
+                :typcategory :user-defined
```
jarohen, line 378: is this user-defined in PG? I'd be surprised if so, given it's built-in

jarohen (changes requested):
thanks @egg-juxt - minor changes below

core/src/main/clojure/xtdb/pgwire/types.clj
egg-juxt: Yes, I copied the values from a real postgres.

image

core/src/main/clojure/xtdb/pgwire/types.clj
egg-juxt: It's actually excluded already, and `:null` too. (Note: the "Z" category = internal exists). Removed line.

PR #4655, opened 2025-08-01
[WIP] Memory hash trie `RelationMap`

FiV0:
Some commits that try to use our `MemoryHashTrie` implementation for `join`, `set` and `group by` operators.

- The first two commits move things from Clojure -> Kotlin and should be purely non-functional changes.
- I then do all kinds of changes to the existing `MemoryHashTrie` implementation so that it can be used in the `RelationMap`. 
- The last two commits mostly copy the `MemoryHashTrie` at that point and try to create a transient version `MutableMemoryHashTrie` because the Build side of joins (and group bys), for which the hash trie is used here, doesn't need other persistent versions.
- The last commit tries to use binary search on the `MutableMemoryHashTrie` but is non functional. 

As of second to last commit, TPCH 0.1 is 15% off from `main`.

Things I have not tried:
- Using a hash trie per block on the probe side and merging the build and probe side hash tries.


FiV0:
Closing in favour of #4666. This is more a trace of all the different things that were tried.

PR #4657, opened 2025-08-04
Move RelationMap from Clojure -> Kotlin

FiV0:
This is a non-functional change. Moving the RelationMap from Clojure to Kotlin.

- [x] ~I think the pattern that needs to be sorted out is how to deal with keyword option maps if Clojure calls Kotlin and Kotlin calls Clojure. This is especially the case for the EE. Currently I map to strings and then back to keywords which feels a bit awkward.~ Pattern ok for now.

### DIFF
--- a/core/src/main/kotlin/xtdb/Util.kt
+++ b/core/src/main/kotlin/xtdb/Util.kt
@@ -1,5 +1,7 @@
 package xtdb
 
+import clojure.lang.Keyword
+import clojure.lang.Symbol
 import java.nio.ByteBuffer
 import java.util.*
 
@@ -11,3 +13,10 @@ val UUID.asByteBuffer: ByteBuffer
     }
 
 val UUID.asBytes: ByteArray get() = asByteBuffer.array()
+
+val String.kw: Keyword get() = Keyword.intern(this)
+
+val String.symbol: Symbol get() = Symbol.intern(this)
+
+fun Map<*, *>.toClojureMap(): clojure.lang.IPersistentMap =
+    clojure.lang.PersistentHashMap.create(this)
--- a/core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
+++ b/core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
@@ -1,14 +1,213 @@
 package xtdb.expression.map
 
+import clojure.lang.IFn
+import com.carrotsearch.hppc.IntObjectHashMap
+import java.util.function.IntBinaryOperator
+import java.util.function.IntConsumer
+import org.apache.arrow.memory.util.hash.MurmurHasher
 import org.apache.arrow.vector.types.pojo.Field
-import xtdb.arrow.RelationReader
+import org.roaringbitmap.RoaringBitmap
+import xtdb.*
 
-interface RelationMap {
-    fun buildFields(): Map<String, Field>
-    fun buildKeyColumnNames(): List<String>
-    fun probeFields(): Map<String, Field>
-    fun probeKeyColumnNames(): List<String>
-    fun buildFromRelation(inRelation: RelationReader): RelationMapBuilder
-    fun probeFromRelation(inRelation: RelationReader): RelationMapProber
-    fun getBuiltRelation(): RelationReader
+import xtdb.arrow.RelationReader
+import xtdb.arrow.RelationWriter
+import xtdb.arrow.VectorReader
+import xtdb.util.Hasher
+import xtdb.util.requiringResolve
+
+class RelationMap(
+    val buildFields : Map<String, Field>,
+    val buildKeyColumnNames: List<String>,
+    val probeFields : Map<String, Field>,
+    val probeKeyColumnNames:  List<String>,
+    private val storeFullBuildRel : Boolean,
+    private val relWriter: RelationWriter,
+    private val buildKeyCols: List<VectorReader>,
+    private val hashToBitmap: IntObjectHashMap<RoaringBitmap>,
+    private val nilKeysEqual: Boolean = false,
+    private val thetaExpr: Any? = null,
+    private val paramTypes: Map<String, Any> = emptyMap(),
+    private val args: RelationReader? = null
+) : AutoCloseable {
+
+    private val equiComparatorFn = requiringResolve("xtdb.expression.map/->equi-comparator") as IFn
+    private val thetaComparatorFn = requiringResolve("xtdb.expression.map/->theta-comparator") as IFn
+
+    @JvmOverloads
+    fun andIBO(p1: IntBinaryOperator? = null, p2: IntBinaryOperator? = null): IntBinaryOperator {
+        return if (p1 == null && p2 == null) {
+            IntBinaryOperator { _, _ -> 1 }
+        } else if (p1 != null && p2 != null) {
+            IntBinaryOperator { l, r ->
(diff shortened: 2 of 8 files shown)


core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
```diff
@@ -1,14 +1,236 @@
...
+        val probeKeyCols = probeKeyColumnNames.map { probeRel.vectorForOrNull(it) as VectorReader }
+        
+        // Create equi-comparators for key columns
+        val equiComparators = buildKeyCols.zip(probeKeyCols).map { (buildCol, probeCol) ->
+            equiComparatorFn.invoke(
+                probeCol, buildCol, args,
+                mapOf(
+                    "nil-keys-equal?" to nilKeysEqual,
+                    "param-types" to paramTypes
+                ).mapKeys { it.key.toKeyword() } .toClojureMap()
```
FiV0: Find a better pattern for this scenario.

core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
```diff
@@ -1,14 +1,236 @@
...
+        
+        // Add theta comparator if needed
+        val comparator = if (thetaExpr != null) {
+            val thetaComparator = thetaComparatorFn.invoke(
+                probeRel, buildRel, thetaExpr, args,
+                mapOf(
+                    "build-fields" to buildFields.mapKeys { it.key.toSymbol() } .toClojureMap() ,
+                    "probe-fields" to probeFields.mapKeys { it.key.toSymbol() } .toClojureMap(),
+                    "param-types" to paramTypes.mapKeys { it.key.toSymbol() } .toClojureMap()
+                ).mapKeys { it.key.toKeyword() } .toClojureMap()
```
FiV0: Same issue.

core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
```diff
@@ -1,14 +1,236 @@
...
+            }
+        }
+        return -1
+    }
+
+    fun createHasher(cols: List<VectorReader>): IndexHasher {
+        val hasher = Hasher.Xx()
+        return if (cols.size == 1) {
+            val col = cols[0]
+            object : IndexHasher {
```
jarohen: make IndexHasher a `fun interface` (if it's not already) and then you can slim this down significantly

core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
```diff
@@ -1,14 +1,236 @@
...
+            IntBinaryOperator { l, r ->
+                val lRes = p1.applyAsInt(l, r)
+                if (lRes == -1) -1 else minOf(lRes, p2.applyAsInt(l, r))
+            }
+        } else {
+            p1 ?: p2!!
+        }
+    }
+
+    fun findInHashBitmap(hashBitmap: RoaringBitmap?, comparator: IntBinaryOperator, idx: Int, removeOnMatch: Boolean): Int {
```
jarohen: Kotlin idiom - I'd probably make hashBitmap the receiver here because you're essentially applying the method on that object

(receivers can be nullable in Kotlin)

core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
```diff
@@ -1,14 +1,236 @@
...
+    private val paramTypes: Map<String, Any> = emptyMap(),
+    private val args: RelationReader? = null
+) : AutoCloseable {
+
+    private val equiComparatorFn = requiringResolve("xtdb.expression.map/->equi-comparator") as IFn
+    private val thetaComparatorFn = requiringResolve("xtdb.expression.map/->theta-comparator") as IFn
+
+    fun andIBO() = andIBO(null, null)
+
+    fun andIBO(p1: IntBinaryOperator? = null, p2: IntBinaryOperator? = null): IntBinaryOperator {
```
jarohen, line 37: add `@JvmOverloads` to this one and you don't need the 0-arg version

core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
```diff
@@ -1,14 +1,236 @@
...
+    }
+
+    companion object {
+        @JvmStatic
+        fun returnedIdx(insertedIdx: Int): Int = -insertedIdx - 1
+        @JvmStatic
+        fun insertedIdx(returnedIdx: Int): Int = if (returnedIdx < 0) -returnedIdx - 1 else returnedIdx
+    }
+
+    fun computeHashBitmap(rowHash: Int): RoaringBitmap {
```
jarohen: nit: expr body

core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
```diff
@@ -1,14 +1,236 @@
...
+
+    @Suppress("NAME_SHADOWING")
+    fun buildFromRelation(inRel: RelationReader): RelationMapBuilder {
+        val inRel = if (storeFullBuildRel) {
+            inRel
+        } else {
+            RelationReader.from(buildKeyColumnNames.map { inRel.vectorFor(it) })
+        }
+        
+        val inKeyCols = buildKeyColumnNames.map { inRel.vectorForOrNull(it) as VectorReader }
```
jarohen, line 93: if you know the vector's there you can use `vectorFor`, or even `inRel[it]`)

core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
```diff
@@ -1,14 +1,236 @@
...
+
+        return object : RelationMapBuilder {
+            override fun add(inIdx: Int) {
+                add(computeHashBitmap(hasher.hashCode(inIdx)), inIdx);
+            }
+            
+            override fun addIfNotPresent(inIdx: Int): Int {
+                val hashBitmap = computeHashBitmap(hasher.hashCode(inIdx))
+                val outIdx = findInHashBitmap(hashBitmap, comparatorLazy, inIdx, false)
+                return if (outIdx >= 0) {
```
jarohen: one line, perhaps?

core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
jarohen: it's not terrible, I'd go with this for now and wait for some inspiration

jarohen (approved):

core/src/main/kotlin/xtdb/Util.kt
```diff
@@ -10,4 +10,12 @@ val UUID.asByteBuffer: ByteBuffer
         flip()
     }
 
-val UUID.asBytes: ByteArray get() = asByteBuffer.array()
\ No newline at end of file
+val UUID.asBytes: ByteArray get() = asByteBuffer.array()
+
+fun String.toKeyword(): clojure.lang.Keyword = clojure.lang.Keyword.intern(this)
```
jarohen: these exist - I think `String.kw`?

core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
FiV0: Yes. I think I had that at some point, but it blew up. I suspect join columns don't need to be there.

PR #4663, opened 2025-08-05
Add `USE {dbname}` and `SET DATABASE = {dbname}`

### LINKED ISSUE #4662: SQL statement to switch database (mid-session/connection)
Postgres forces users to close a connection and re-open a new connection in order to 'switch' the current database (`psql` does this on the users behalf). However many other databases offer `USE {dbname}` statements which dynamically change the current database without breaking the session/connection. XTDB could supporting this behaviour also. This may improve performance and stability in multi-tenant applications which routinely switch between databases based on in-flight requests.

From the [CockroachDB docs](https://www.cockroachlabs.com/docs/stable/schema-design-database):
> Here are some best practices to follow when creating and using databases:
Do not use the preloaded defaultdb database. Instead, create your own database with a CREATE DATABASE statement, and change it to the SQL session's [current database](https://www.cockroachlabs.com/docs/v25.3/sql-name-resolution#current-database) by executing a USE {databasename}; statement, by passing the --database={databasename} flag to the [cockroach sql command](https://www.cockroachlabs.com/docs/v25.3/cockroach-sql#general), or by specifying the database parameter in the [connection string](https://www.cockroachlabs.com/docs/v25.3/connection-parameters#connect-using-a-url) passed to your database schema migration tool.

Allegedly you can also do `SET DATABASE = {dbname}` in CockroachDB also, according to https://stackoverflow.com/questions/43331605/how-do-i-change-cockroachdbs-default-database

Per DuckDB, this should also be able to set the schema: https://duckdb.org/docs/stable/sql/statements/use

refset:
...how did Claude do @jarohen? :slightly_smiling_face:

It never did figure out how to run a single Clojure test / test-ns using `gradlew`, but otherwise it was pretty flawless in getting this working without any serious input from me.

```
Bash(./gradlew :test --tests "xtdb.sql.multi-db-test" --no-daemon)
  ⎿  Error: FAILURE: Build failed with an exception.

     * What went wrong:
     Execution failed for task ':test'.
     > No tests found for given includes: [xtdb.sql.multi-db-test](--tests filter)
```

(...is it possible to run a single Clojure test like this? if so we should add an example to `dev/README.adoc`)


refset:
This specific design/approach is not enough - requires also preventing switching DBs within a transaction, at least.

jarohen:
I did try to land this one - fell down with prepared statements, which the pgjdbc client re-uses within (what it thinks is) a connection. Best to park for now.

Issue #4592, opened 2025-07-11
Disk Based Join


FiV0:
One of our DP's is seeing OOM when doing large joins. This is due to us not spilling to disk for large joins. 
The ticket tries to address this issue by spilling to disk. We are going for a hash based approach for now (Grace Hash Join).
Say we have relations $S$ and $R$.
- Partition $S$ and $R$ via a hash function and spill the partitions to disk
- For equi-joins the hash partitions of the same hash then need to be brought into memory and joined.
- For theta-joins we need to do a nested loop join on the different partitions (doing a special sort based approach for certain predicates is out of scope for now)
- If the initial bucketing (4) is not enough (a bucket becomes to large again) we split the problematic bucket again (dividing it into 4 sub-buckets). This is better for data skew as global bucket splitting approach might do a lot of unnecessary work.

PR #4666, opened 2025-08-08
`MutableMemoryHashTrie` for `RelationMap`.

FiV0:
(for #4592)

This adds a `MutableMemoryHashTrie` class that has a similar interface to the existing `MemoryHashTrie`. We are using the new HashTrie implementation on the build side of RelationMap. It replaces the `IntObjectHashMap` + `RoaringBitmap` values that we are currently using. The idea is that this will make it easier to use existing tooling (currently mainly used in the compactor) for the `HashTrie` interface for off disk joining, grouping and set operations. Currently this implementation runs a bit slower (the comparison is TPCH 0.1) then `main`.

The following table shows TPCH 0.1 run times (all queries except Q21) for different branching factors and page-sizes
```
"Page Size: " 32 " branch-factor: " 16
"Elapsed time: 29540.182785 msecs"
"Page Size: " 64 " branch-factor: " 16
"Elapsed time: 28970.345773 msecs"
"Page Size: " 96 " branch-factor: " 16
"Elapsed time: 29521.322433 msecs"
"Page Size: " 128 " branch-factor: " 16
"Elapsed time: 29487.788614 msecs"
"Page Size: " 256 " branch-factor: " 16
"Elapsed time: 31607.618253 msecs"
"Page Size: " 32 " branch-factor: " 32
"Elapsed time: 36701.340508 msecs"
"Page Size: " 64 " branch-factor: " 32
"Elapsed time: 29450.644377 msecs"
"Page Size: " 96 " branch-factor: " 32
"Elapsed time: 29932.627244 msecs"
"Page Size: " 128 " branch-factor: " 32
"Elapsed time: 30089.853236 msecs"
"Page Size: " 256 " branch-factor: " 32
"Elapsed time: 29792.72954 msecs"
"Page Size: " 32 " branch-factor: " 64
"Elapsed time: 29262.566839 msecs"
"Page Size: " 64 " branch-factor: " 64
"Elapsed time: 28986.29487 msecs"
"Page Size: " 96 " branch-factor: " 64
"Elapsed time: 29258.650799 msecs"
"Page Size: " 128 " branch-factor: " 64
"Elapsed time: 29044.098769 msecs"
"Page Size: " 256 " branch-factor: " 64
"Elapsed time: 29795.626605 msecs"
```

### DIFF
--- a/core/src/main/clojure/xtdb/expression/map.clj
+++ b/core/src/main/clojure/xtdb/expression/map.clj
@@ -5,8 +5,7 @@
             [xtdb.util :as util]
             [xtdb.vector.reader :as vr]
             [xtdb.vector.writer :as vw])
-  (:import (com.carrotsearch.hppc IntObjectHashMap)
-           (java.util Map)
+  (:import (java.util Map)
            java.util.function.IntBinaryOperator
            (org.apache.arrow.memory BufferAllocator)
            (org.apache.arrow.vector NullVector VectorSchemaRoot)
@@ -112,7 +111,6 @@
         build-key-col-names (get opts :build-key-col-names key-col-names)
         probe-key-col-names (get opts :probe-key-col-names key-col-names)
 
-        hash->bitmap (IntObjectHashMap.)
         schema (Schema. (-> build-fields
                             (cond-> (not store-full-build-rel?) (select-keys build-key-col-names))
                             (->> (mapv (fn [[field-name field]]
@@ -126,15 +124,18 @@
             (.copyRow 0)))
 
         (let [build-key-cols (mapv #(vw/vec-wtr->rdr (.vectorFor rel-writer (str %))) build-key-col-names)]
-          (RelationMap. (update-keys build-fields str)
+          (RelationMap. allocator
+                        (update-keys build-fields str)
                         (map str build-key-col-names)
                         (update-keys probe-fields str)
                         (map str probe-key-col-names)
                         (boolean store-full-build-rel?)
                         rel-writer
                         build-key-cols
-                        hash->bitmap
                         (boolean nil-keys-equal?)
                         theta-expr
                         (update-keys param-types str)
-                        args))))))
+                        args
+                        (boolean with-nil-row?)
+                        64
+                        4))))))
--- a/core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
+++ b/core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
@@ -1,21 +1,23 @@
 package xtdb.expression.map
 
 import clojure.lang.IFn
-import com.carrotsearch.hppc.IntObjectHashMap
+import org.apache.arrow.memory.BufferAllocator
 import java.util.function.IntBinaryOperator
 import java.util.function.IntConsumer
 import org.apache.arrow.memory.util.hash.MurmurHasher
 import org.apache.arrow.vector.types.pojo.Field
-import org.roaringbitmap.RoaringBitmap
 import xtdb.*
+import xtdb.arrow.IntVector
 
 import xtdb.arrow.RelationReader
 import xtdb.arrow.RelationWriter
 import xtdb.arrow.VectorReader
+import xtdb.trie.MutableMemoryHashTrie
 import xtdb.util.Hasher
 import xtdb.util.requiringResolve
 
 class RelationMap(
+    val allocator: BufferAllocator,
     val buildFields : Map<String, Field>,
     val buildKeyColumnNames: List<String>,
     val probeFields : Map<String, Field>,
@@ -23,12 +25,23 @@ class RelationMap(
     private val storeFullBuildRel : Boolean,
     private val relWriter: RelationWriter,
     private val buildKeyCols: List<VectorReader>,
-    private val hashToBitmap: IntObjectHashMap<RoaringBitmap>,
     private val nilKeysEqual: Boolean = false,
     private val thetaExpr: Any? = null,
     private val paramTypes: Map<String, Any> = emptyMap(),
-    private val args: RelationReader? = null
+    private val args: RelationReader? = null,
+    withNilRow: Boolean = false,
+    pageLimit: Int = 16,
+    levelBits: Int = 2
 ) : AutoCloseable {
+
+    private val hashColumn: IntVector = IntVector(allocator, "xt/join-hash", false)
+    private var buildHashTrie : MutableMemoryHashTrie
+
+    init {
+        if (withNilRow) hashColumn.writeInt(0)
+        buildHashTrie = MutableMemoryHashTrie.builder(hashColumn.asReader).setPageLimit(pageLimit).setLevelBits(levelBits).build()
+    }
+
 
     private val equiComparatorFn = requiringResolve("xtdb.expression.map/->equi-comparator") as IFn
     private val thetaComparatorFn = requiringResolve("xtdb.expression.map/->theta-comparator") as IFn
@@ -47,22 +60,6 @@ class RelationMap(
         }
     }
 
-    fun RoaringBitmap?.findInHashBitmap(comparator: IntBinaryOperator, idx: Int, removeOnMatch: Boolean): Int {
-        if (this == null) return -1
-
-        val iterator = this.intIterator
-        while (iterator.hasNext()) {
-            val testIdx = iterator.next()
-            if (comparator.applyAsInt(idx, testIdx) == 1) {
-                if (removeOnMatch) {
-                    this.remove(testIdx)
-                }
-                return testIdx
-            }
-        }
-        return -1
-    }
-
     private fun createHasher(cols: List<VectorReader>): IndexHasher {
         val hasher = Hasher.Xx()
         return IndexHasher { index -> cols.foldRight(0) { col, acc -> MurmurHasher.combineHashCode(acc, col.hashCode(index, hasher)) } }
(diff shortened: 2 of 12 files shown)


core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
```diff
@@ -105,34 +95,58 @@ class RelationMap(
...
+                "Expected outIdx $outIdx to match hashColumn valueCount $outIdxHashColumn"
+            }
             return returnedIdx(outIdx)
         }
 
-
         return object : RelationMapBuilder {
-            override fun add(inIdx: Int) {
-                add(computeHashBitmap(hasher.hashCode(inIdx)), inIdx);
+            override fun add(inIdx: Int)  {
```
jarohen: at first glance this just looks like a recursive call back to the same function? I'm guessing it's not, otherwise more would probably break!

core/src/main/kotlin/xtdb/trie/MutableMemoryHashTrie.kt
```diff
@@ -0,0 +1,281 @@
...
+import org.roaringbitmap.RoaringBitmap
+import xtdb.arrow.VectorReader
+import xtdb.trie.HashTrie.Companion.DEFAULT_LEVEL_BITS
+import java.nio.ByteBuffer
+import java.nio.ByteOrder
+import java.util.function.IntUnaryOperator
+
+private const val PAGE_LIMIT = 1024
+
+// hashReader is assumed to be a FixedSizeBinary vector
```
jarohen: hashReader is assumed to be a fixed-width vector?

core/src/main/kotlin/xtdb/trie/MutableMemoryHashTrie.kt
```diff
@@ -0,0 +1,281 @@
...
+            if (!sorted) sortData(trie)
+            val idx = binarySearch(trie, hash)
+            if (idx < 0) return IntArray(0)
+            return data.take(dataCount).drop(idx).takeWhile{ trie.compare(it, trie.reusePtr1, hash) == 0 }.toIntArray()
+        }
+
+        override fun findValue(trie: MutableMemoryHashTrie, hash: ByteArray, comparator: IntUnaryOperator, removeOnMatch: Boolean): Int {
+            val data = if (dataCount <= 16) {
+                when (deletions) {
+                    null -> data.take(dataCount).filter { trie.compare(it, trie.reusePtr1, hash) == 0 }
```
jarohen: these take/drop/filter seem a little out of place here when everything else is low-level array ops - probably an easy isolated perf win?

core/src/test/kotlin/xtdb/trie/MutableMemoryHashTrieTest.kt
```diff
@@ -0,0 +1,78 @@
...
+        val hashes = List(testSize) {r.nextInt() }
+        hashes.forEachIndexed( {index, hash ->
+            hashReader.writeInt(hash)
+            trie += index
+        })
+
+
+        hashes.forEachIndexed( { index, hash ->
+            val newIndex = hashes.size + index
+            assertTrue(trie.addIfNotPresent(hash, newIndex, {_ -> 1}, {throw Exception("Should not happen!")}).first != newIndex, "Hash $hash should already be present")
```
jarohen: `throw Exception("Should not happen!")` -> `error("Should not happen")`

core/src/main/kotlin/xtdb/expression/map/RelationMap.kt
```diff
@@ -163,38 +177,31 @@ class RelationMap(
...
-                
-                hashBitmap?.let { bitmap ->
-                    val iterator = bitmap.intIterator
-                    while (iterator.hasNext()) {
-                        val outIdx = iterator.next()
-                        if (comparator.applyAsInt(inIdx.toInt(), outIdx) == 1) {
-                            c.accept(outIdx)
-                        }
+                for (outIdx in buildHashTrie.findCandidates(hashCode)) {
+                    if (comparator.applyAsInt(inIdx, outIdx) == 1) {
```
jarohen: `comparator` naming is misleading here, especially with `applyAsInt` - I spent most of the PR review assuming this was a `java.util.Comparator`, or at least a function that returned positive for >, 0 for =, negative for < 😅 

it seems you're effectively providing an IntPredicate instead?

core/src/main/kotlin/xtdb/trie/HashTrie.kt
```diff
@@ -23,42 +23,61 @@ interface HashTrie<N : Node<N>, L : N> {
 
     val leaves get() = rootNode?.leaves ?: emptyList()
 
+    val LEVEL_BITS: Int
```
jarohen: given that these now seem to be 'travelling together' everywhere, it's probably time to make them a struct. can then give the struct the 'bucketFor' methods for more idiomatic OO.

(comment maybe a bit vague, lmk and it's probably a quick pairing session)

core/src/main/kotlin/xtdb/trie/HashTrie.kt
```diff
@@ -23,42 +23,61 @@ interface HashTrie<N : Node<N>, L : N> {
...
-        const val LEVEL_MASK: Int = LEVEL_WIDTH - 1
+        const val DEFAULT_LEVEL_BITS: Int = 2
+        const val DEFAULT_LEVEL_WIDTH: Int = 1 shl DEFAULT_LEVEL_BITS
+        const val DEFAULT_LEVEL_MASK: Int = DEFAULT_LEVEL_WIDTH - 1
 
         @JvmStatic
-        fun bucketFor(pointer: ArrowBufPointer, level: Int): Byte {
-            assert(level * LEVEL_BITS < pointer.length * java.lang.Byte.SIZE)
-            val bitIdx = level * LEVEL_BITS
+        fun bucketFor(byteArray: ByteArray, level: Int, levelBits: Int = DEFAULT_LEVEL_BITS, levelMask: Int = DEFAULT_LEVEL_MASK): Byte {
```
jarohen: yeah, let's get this into a class now 🙂

core/src/main/kotlin/xtdb/trie/HashTrie.kt
```diff
@@ -67,6 +86,14 @@ interface HashTrie<N : Node<N>, L : N> {
     }
 }
 
+abstract class BaseHashTrie<N : Node<N>, L : N>(
```
jarohen: this probably goes if you have hash-bucketing as its own class - 'composition rather than inheritance', as they say

core/src/main/clojure/xtdb/expression/map.clj
```diff
@@ -126,15 +124,18 @@
...
                         rel-writer
                         build-key-cols
-                        hash->bitmap
                         (boolean nil-keys-equal?)
                         theta-expr
                         (update-keys param-types str)
-                        args))))))
+                        args
+                        (boolean with-nil-row?)
+                        96
```
jarohen: 64/4 seemed fastest? also powers of 2, so may well be a situation where those are easier to work with

jarohen (changes requested):
Thanks @FiV0 🙏 comments below/above/wherever they appear 🙂

core/src/main/kotlin/xtdb/trie/HashTrie.kt
```diff
@@ -18,6 +18,51 @@ fun conjPath(path: ByteArray, idx: Byte): ByteArray {
...
+    fun compareToPath(pointer: ArrowBufPointer, path: ByteArray): Int {
+        for (level in path.indices) {
+            val cmp = bucketFor(pointer, level).toInt() compareTo (path[level].toInt())
+            if (cmp != 0) return cmp
+        }
+        return 0
+    }
+}
+
+val DEFAULT_BUCKETER = Bucketer()
```
FiV0: There are multiple places (`EventRowPointer`, `TrieWriter` ..)  that call `bucketFor` but don't have a hashTrie at hand, so I added this one.

core/src/main/kotlin/xtdb/trie/MutableMemoryHashTrie.kt
```diff
@@ -0,0 +1,288 @@
...
+            if (dataCount == 0) return
+            val tmp = data.take(dataCount).toTypedArray()
+            tmp.sortWith { leftIdx, rightIdx -> trie.compare(leftIdx, rightIdx, trie.reusePtr1, trie.reusePtr2) }
+            tmp.forEachIndexed { index, i -> data[index] = i }
+            sorted = true
+        }
+
+
+        // Beware that mixing this method with `removeOnMatch` set to true, will result in non correct behaviour
+        override fun findCandidates(trie: MutableMemoryHashTrie, hash: ByteArray): IntArrayList {
```
FiV0, line 225: One thing to be aware of when using this structure is that it is optimized for the `RelationMap` use case. For example mixing `findCandidates` with `findValue` (and `removeOnMatch` set to true) will result in bad behaviour.

FiV0:
Good for another round of review. I left two comments.

core/src/main/kotlin/xtdb/trie/HashTrie.kt
```diff
@@ -18,6 +18,51 @@ fun conjPath(path: ByteArray, idx: Byte): ByteArray {
...
+const val DEFAULT_LEVEL_BITS = 2
+const val DEFAULT_LEVEL_WIDTH = 1 shl DEFAULT_LEVEL_BITS
+
+data class Bucketer(val levelBits: Int = DEFAULT_LEVEL_BITS) {
+
+    val levelWidth: Int = 1 shl levelBits
+    val levelMask: Int = levelWidth - 1
+
+    init {
+        if (levelBits !in setOf(2,4,8)) {
```
jarohen: nit: `require(levelBits !in ...) { "levelBits must be ..." }`

core/src/main/kotlin/xtdb/trie/HashTrie.kt
jarohen: fair - companion object for this one, though, within Bucketer:

```
companion object {
  val DEFAULT = Bucketer()
}
```

then `Bucketer.DEFAULT`

core/src/main/kotlin/xtdb/trie/MutableMemoryHashTrie.kt
jarohen: comment as to why, rather than just 'will behave badly'?

I'd guess something to do with deletions?

jarohen (approved):
minor nits now, good to go without further review 👏

PR #4672, opened 2025-08-11
Implement openid connect based authentication.

### LINKED ISSUE #3999: OpenID Connect (OIDC) authenticator
We want to support authentication on pgwire via OpenID Connect, across common grant types.

Want to implement OIDC authentication covering the following flows:
- [x] Password grant: for end users (e.g., data scientists using psql, metabase users).
- [x] Client credentials flow: for ingestion services, custom apps 
  - requiring proper handling of HTTP basic auth
  - methodFor mappings for this - ie, assume client credentials passed as password if user: oid-client
- [x] Device grant (since this is already implemented)

We must handle token expiry/refresh:
- [x] Tokens must be stored and checked for expiry on every access (within pgwire).
- [x] Automatically refresh expired access tokens using refresh tokens.
- [x] If refresh fails (e.g., token revoked), terminate the session and return a clear error.

danmason:
Resolves #3999 
Github action runs: https://github.com/danmason/xtdb/actions/workflows/build.yml?query=branch%3Aoidc-authn

Implements OpenID Connect (OIDC) authentication for pgwire connections, supporting multiple OAuth2/OIDC grant flows for different use cases.

- Added OpenIdConnect authenticator with support for password, client credentials, and device authorization flows
  - We take in an "issuer-url", from which we fetch the "/.well-known/openid-configuration" for the sake of route discovery.
- Updated Kotlin Authenticator interface, adding in new methods which return different data class types depending on the type of auth.
  - All of these implement `AuthResult` which at it's minimum contains a `user-id`.
  - Existing user table password auth returns a `SimpleResult`.
  - All OpenIdConnect auth methods return some implemenation of `OAuthResult`, which contains additional fields (used currently for the sake of checking token expiry/handling refresh - in the future we can return authz claims and such) 
- Supported flows:
    - Uses `methodFor` to route authentication based on user/address rules
    - **Password flow**: Standard username/password authentication for end users
    - **Client credentials**:`client_id:client_secret` passed as password, mapped via user rules (e.g., `oid-client` user)
    - **Device auth**: Interactive flow with verification URL, polls for completion
- Implemented OAuth token management with automatic refresh and expiry handling
  - OAuth results are stored within pgwire connection state after the initial msg-auth call.
  - We implement and use the "revalidate" method on the authenticator to:
    - Check if a token is expired.
    - Attempt to refresh the token if it is expired. 
  - We wrap most XTDB api calls within pgwire in a `with-auth-check` macro that calls down to this method.
- Testing and documentation:
  - Have a set of unit and integration levels tests of the various oauth flows.
  - Docker Compose setup with Keycloak service for development.

### Updates since last review

- Client credential password handling - no longer base64, client-id:client-secret 
  - Parsing these out in pgwire and passing separately into verification.
- authn now uses and passes around Kotlin data objects for auth responses.
- Using anomalies for authn errors, and then send-ex  handling for these errors on pgwire side.
- Adding and passing around InstantSource to OpenIdConnect - using this within numerous tests also, so I don't have to rely on thread sleeps.
- Making revalidate a method on the authenticator implementation, testing revalidate logic within authn_test.
  - We use this in pgwire with a macro that wraps individual api XTDB calls, rather than checking that at the handle-msg level.
- oidc_integration_test now uses  createConnectionBuilder on the node and is generally much simpler.
- Fetch expires-in from initial device_auth and using this to check if the device auth has timed out.
  - I think that the token endpoint actually does something similar when handling device auth - it will return an auth_expired error code, so this may not be entirely necessary - still, serves as an explicit loop break for us, though, so can't hurt.
- We now transform the results of OAuth HTTP calls fairly specifically, grabbing and renaming specific keys which we are using. 
- Added a set of OIDC auth pgwire protocol tests.

### DIFF
--- a/core/src/main/clojure/xtdb/authn.clj
+++ b/core/src/main/clojure/xtdb/authn.clj
@@ -1,22 +1,32 @@
 (ns xtdb.authn
   (:require [buddy.hashers :as hashers]
             [integrant.core :as ig]
+            [hato.client :as http]
             [xtdb.api :as xt]
             [xtdb.node :as xtn]
-            [xtdb.query :as q])
+            [xtdb.query :as q]
+            [xtdb.error :as err])
   (:import [java.io Writer]
-           (xtdb.api Authenticator Authenticator$Factory Authenticator$Factory$UserTable Authenticator$Method Authenticator$MethodRule Xtdb$Config)
-           (xtdb.query IQuerySource)))
+           (xtdb.api Authenticator Authenticator$DeviceAuthResponse Authenticator$Factory Authenticator$Factory$OpenIdConnect
+                     Authenticator$Factory$UserTable Authenticator$Method Authenticator$MethodRule Xtdb$Config
+                     SimpleResult OAuthPasswordResult OAuthClientCredentialsResult OAuthResult)
+           (xtdb.query IQuerySource)
+           [java.net URI]
+           [java.time Duration Instant InstantSource]))
 
-(defn verify-pw [^IQuerySource q-src, db-cat, user password]
-  (when password
-    (with-open [res (-> (.prepareQuery q-src "SELECT passwd AS encrypted FROM pg_user WHERE username = ?"
-                                       db-cat {:default-db "xtdb"})
-                        (.openQuery {:args [user]}))]
+(defn verify-pw
+  [^IQuerySource q-src db-cat user password]
+  (when-not password
+    (throw (err/incorrect :xtdb/authn-failed (format "password authentication failed for user: %s" user))))
 
-      (when-let [{:keys [encrypted]} (first (.toList (q/cursor->stream res {:key-fn #xt/key-fn :kebab-case-keyword})))]
-        (when (:valid (hashers/verify password encrypted))
-          user)))))
+  (with-open [res (-> (.prepareQuery q-src
+                                     "SELECT passwd AS encrypted FROM pg_user WHERE username = ?"
+                                     db-cat {:default-db "xtdb"})
+                      (.openQuery {:args [user]}))]
+    (let [{:keys [encrypted]} (first (.toList (q/cursor->stream res {:key-fn #xt/key-fn :kebab-case-keyword})))]
+      (if (and encrypted (:valid (hashers/verify password encrypted)))
+        user
+        (throw (err/incorrect :xtdb/authn-failed (format "password authentication failed for user: %s" user)))))))
 
 (defn- method-for [rules {:keys [remote-addr user]}]
   (some (fn [{rule-user :user, rule-address :address, :keys [method]}]
@@ -28,11 +38,13 @@
 (defn read-authn-method [method]
   (case method
     :trust Authenticator$Method/TRUST
-    :password Authenticator$Method/PASSWORD))
+    :password Authenticator$Method/PASSWORD
+    :device-auth Authenticator$Method/DEVICE_AUTH
+    :client-credentials Authenticator$Method/CLIENT_CREDENTIALS))
 
 (defmethod print-dup Authenticator$Method [^Authenticator$Method m, ^Writer w]
   (.write w "#xt.authn/method ")
-  (print-method (case (str m) "TRUST" :trust, "PASSWORD" :password) w))
+  (print-method (case (str m) "TRUST" :trust, "PASSWORD" :password, "DEVICE_AUTH" :device-auth, "CLIENT_CREDENTIALS" :client-credentials) w))
 
 (defmethod print-method Authenticator$Method [^Authenticator$Method m, ^Writer w]
   (print-dup m w))
@@ -53,7 +65,9 @@
 (defmethod xtn/apply-config! :xtdb/authn [^Xtdb$Config config, _, [tag opts]]
   (xtn/apply-config! config
                      (case tag
-                       :user-table ::user-table-authn)
+                       :user-table ::user-table-authn
+                       :openid-connect ::openid-connect-authn
+                       tag)
                      opts))
 
 (defmethod xtn/apply-config! ::user-table-authn [^Xtdb$Config config, _, {:keys [rules]}]
@@ -65,7 +79,8 @@
     (method-for rules {:user user, :remote-addr remote-addr}))
 
   (verifyPassword [_ user password]
-    (verify-pw q-src db-cat user password)))
+    (let [user-id (verify-pw q-src db-cat user password)]
+      (SimpleResult. user-id))))
 
 #_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
 (defn ->user-table-authn [^Authenticator$Factory$UserTable cfg, q-src, db-cat]
@@ -75,6 +90,211 @@
   (into {:q-src (ig/ref :xtdb.query/query-source)
          :db-cat (ig/ref :xtdb/db-catalog)}
         opts))
+
+(defn- validate-oidc-config [config discovery-url]
+  (let [{:keys [token_endpoint userinfo_endpoint]} config]
+    (cond
+      (not token_endpoint)
+      (throw (err/incorrect :xtdb/oidc-config-discovery-error
+                            (format "OIDC configuration missing required token_endpoint from %s" discovery-url)
+                            {:body config}))
+
+      (not userinfo_endpoint)
+      (throw (err/incorrect :xtdb/oidc-config-discovery-error
+                            (format "OIDC configuration missing required userinfo_endpoint from %s" discovery-url)
+                            {:body config}))
+
+      :else config)))
+
+(defn discover-oidc-config [issuer-url]
+  (let [discovery-url (str issuer-url "/.well-known/openid-configuration")]
+    (try
+      (let [{:keys [status body]} (http/get discovery-url
+                                            {:throw-exceptions false
+                                             :as :json})]
+        (if (= 200 status)
+          (validate-oidc-config body discovery-url)
+          (throw (err/incorrect :xtdb/oidc-config-discovery-error
+                                (format "Failed to discover OIDC configuration from %s: %s" discovery-url (:error body))
+                                {:status status, :body body}))))
+      (catch Exception e
+        (throw (err/incorrect :xtdb/oidc-config-discovery-error
+                              (format "Error thrown when fetching OIDC configuration from %s: %s" discovery-url (.getMessage e))
+                              {:exception e}))))))
+
+(defn calculate-expires-at [{:keys [expires_in]} ^Instant now]
+  (some->> expires_in (.plusSeconds now)))
+
+(defn oauth-token [{:keys [oidc-config client-id client-secret ^InstantSource instant-src]} opts]
+  (let [token-endpoint (:token_endpoint oidc-config)
+        {:keys [status body]} (http/post token-endpoint
+                                         {:form-params (into {:client_id client-id
+                                                              :client_secret client-secret}
+                                                             opts)
+                                          :throw-exceptions false
+                                          :coerce :always
+                                          :as :json})]
+    {:status status
+     :body {:access-token (:access_token body)
+            :refresh-token (:refresh_token body)
+            :expires-at (calculate-expires-at body (.instant instant-src))}}))
+
+(defn oauth-userinfo [{:keys [oidc-config]} token]
+  (let [userinfo-endpoint (:userinfo_endpoint oidc-config)
+        {:keys [status body]} (http/get userinfo-endpoint
+                                        {:oauth-token token
+                                         :throw-exceptions false
+                                         :as :json})]
+    (if (= 200 status)
+      {:user-id (:sub body)}
+      (throw (err/incorrect :xtdb/authn-failed
+                            (format "Failed to obtain user info from %s: %s" userinfo-endpoint (:error body))
+                            {:status status, :body body})))))
+
+(defn token-expired? [^OAuthResult auth-result ^Instant now]
+  (when-let [expires-at (.getExpiresAt ^OAuthResult auth-result)]
+    (.isAfter now expires-at)))
+
+;; TODO: Could live in Kotlin
+(defn refresh-token [authn ^OAuthResult auth-result]
+  (cond
+    ;; For password/device auth flows - use refresh token from OAuthPasswordResult
+    (instance? OAuthPasswordResult auth-result)
+    (let [oauth-result ^OAuthPasswordResult auth-result
+          refresh-token (.getRefreshToken oauth-result)
+          {:keys [status body]} (oauth-token authn {:grant_type "refresh_token"
+                                                    :refresh_token refresh-token})]
+      (if (= 200 status)
+        (.withExpiry oauth-result ^Instant (:expires-at body))
+        (throw (err/incorrect :xtdb/authn-failed
+                              (format "Failed to refresh OAuth token: %s" (:error body))
+                              {:status status, :body body}))))
+
+    ;; For client credentials flow - get a fresh token using stored client credentials
+    (instance? OAuthClientCredentialsResult auth-result)
+    (let [creds-result ^OAuthClientCredentialsResult auth-result
+          client-id (.getClientId creds-result)
+          client-secret (.getClientSecret creds-result)
+          {:keys [status body]} (oauth-token authn {:grant_type "client_credentials"
+                                                    :scope "openid"
+                                                    :client_id client-id
+                                                    :client_secret client-secret})]
+      (if (= 200 status)
+        (.withExpiry creds-result ^Instant (:expires-at body))
+        (throw (err/incorrect :xtdb/authn-failed
+                              (format "Failed to refresh client credentials token: %s" (:error body))
+                              {:status status, :body body}))))
+
+    :else
+    (throw (err/incorrect :xtdb/authn-failed "No valid token or refresh token available for refresh"))))
+
+(defn oauth-device-info [{:keys [oidc-config client-id client-secret]}]
+  (let [device-endpoint (:device_authorization_endpoint oidc-config)]
+    (when-not device-endpoint
+      (throw (err/incorrect :xtdb/authn-failed
+                            "OIDC provider does not support device authorization flow - missing device_authorization_endpoint")))
+    (let [{:keys [status body]} (http/post device-endpoint
+                                           {:form-params {:client_id client-id
+                                                          :client_secret client-secret
+                                                          :scope "openid"}
+                                            :throw-exceptions false
+                                            :as :json})]
+      (if (= 200 status)
+        {:device-code (:device_code body)
+         :verification-uri-complete (:verification_uri_complete body)
+         :interval (:interval body)
+         :expires-at (calculate-expires-at body (.instant (InstantSource/system)))}
+        (throw (err/incorrect :xtdb/authn-failed
+                              (format "Failed to obtain device info from %s: %s" device-endpoint (:error body))
+                              {:status status, :body body}))))))
+
+
+(defrecord DeviceAuthResponse [authn url device-code ^Duration interval ^Instant expires-at]
+  Authenticator$DeviceAuthResponse
+  (getUrl [_] url)
+
+  (await [_]
+    (let [{:keys [oidc-config ^InstantSource instant-src]} authn]
+      (loop []
+        (when (.isAfter (.instant instant-src) expires-at) (throw (err/incorrect :xtdb/authn-failed "Device authorization has expired")))
+
+        (let [token-endpoint (:token_endpoint oidc-config)
+              {:keys [status body]} (http/post token-endpoint
+                                               {:form-params {:client_id (:client-id authn)
+                                                              :client_secret (:client-secret authn)
+                                                              :grant_type "urn:ietf:params:oauth:grant-type:device_code"
+                                                              :device_code device-code}
+                                                :throw-exceptions false
+                                                :coerce :always
+                                                :as :json})]
+          (case (long status)
+            200 (let [access-token (:access_token body)
+                      refresh-token (:refresh_token body)
+                      expires-at (calculate-expires-at body (.instant instant-src))
+                      {:keys [user-id]} (oauth-userinfo authn access-token)]
+                  (OAuthPasswordResult. user-id expires-at access-token refresh-token))
+            400 (if (= "authorization_pending" (:error body))
+                  (do (Thread/sleep (.toMillis interval)) (recur))
+                  (throw (err/incorrect :xtdb/authn-failed
+                                        (format "Device authentication failed: %s" (:error body))
+                                        {:status status, :body body})))
+            (throw (err/incorrect :xtdb/authn-failed
+                                  (format "Device authentication failed with status %d" status)
+                                  {:status status, :body body}))))))))
+
+(defrecord OpenIdConnect [oidc-config client-id client-secret rules ^InstantSource instant-src]
+  Authenticator
+  (methodFor [_ user remote-addr]
+    (method-for rules {:user user, :remote-addr remote-addr}))
+
+  (verifyPassword [this user password]
+    (let [{:keys [status body]} (oauth-token this {:grant_type "password", :scope "openid"
+                                                   :username user, :password password})]
+      (if (= 200 status)
+        (let [{:keys [access-token refresh-token expires-at]} body
+              {:keys [user-id]} (oauth-userinfo this access-token)]
+          (OAuthPasswordResult. user-id expires-at access-token refresh-token))
+        (throw (err/incorrect :xtdb/authn-failed
+                              (format "Password authentication failed for user: %s" user)
+                              {:status status, :body body})))))
+
+  (startDeviceAuth [this _user]
+    (let [{:keys [device-code verification-uri-complete interval expires-at]} (oauth-device-info this)]
+      (->DeviceAuthResponse this (.toURL (URI. verification-uri-complete)) device-code (Duration/ofSeconds interval) expires-at)))
+
+  (verifyClientCredentials [this client-id client-secret]
+    (let [{:keys [status body]} (oauth-token this {:grant_type "client_credentials",
+                                                   :scope "openid",
+                                                   :client_id client-id
+                                                   :client_secret client-secret})]
+      (if (= 200 status)
+        (let [{:keys [access-token expires-at]} body
+              {:keys [user-id]} (oauth-userinfo this access-token)]
+          (OAuthClientCredentialsResult. user-id expires-at access-token client-id client-secret))
+        (throw (err/incorrect :xtdb/authn-failed
+                              (format "Client credentials authentication failed for client: %s" client-id)
+                              {:status status, :body body})))))
+
+  (revalidate [this auth-result]
+    (when (token-expired? auth-result (.instant instant-src))
+      (refresh-token this auth-result))))
+
+#_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
+(defn ->oidc-authn [^Authenticator$Factory$OpenIdConnect cfg]
+  (let [oidc-config (discover-oidc-config (.getIssuerUrl cfg))]
+    (->OpenIdConnect oidc-config
+                     (.getClientId cfg)
+                     (.getClientSecret cfg)
+                     (<-rules-cfg (.getRules cfg))
+                     (.getInstantSource cfg))))
+
+(defmethod xtn/apply-config! ::openid-connect-authn [^Xtdb$Config config, _, {:keys [issuer-url client-id client-secret rules ^InstantSource instant-src]}]
+  (.authn config
+          (cond-> (Authenticator$Factory$OpenIdConnect. (.toURL (URI. issuer-url))
+                                                        client-id
+                                                        client-secret
+                                                        (->rules-cfg rules))
+            instant-src (.instantSource instant-src))))
 
 (defmethod ig/init-key :xtdb/authn [_ {:keys [^Authenticator$Factory authn-factory, q-src, db-cat]}]
   (.open authn-factory q-src db-cat))
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -31,15 +31,14 @@
            [java.nio.channels FileChannel]
            [java.nio.file Path]
            [java.security KeyStore]
-           [java.time Clock Duration ZoneId]
-           [java.util Map]
+           [java.time Clock Duration Instant ZoneId]
            [java.util.concurrent ExecutorService Executors TimeUnit]
            [javax.net.ssl KeyManagerFactory SSLContext]
            (org.antlr.v4.runtime ParserRuleContext)
-           (org.apache.arrow.memory BufferAllocator RootAllocator)
+           (org.apache.arrow.memory BufferAllocator)
            org.apache.arrow.vector.types.pojo.Field
            (xtdb.antlr Sql$DirectlyExecutableStatementContext SqlVisitor)
-           (xtdb.api DataSource DataSource$ConnectionBuilder ServerConfig Xtdb$Config)
+           (xtdb.api Authenticator DataSource DataSource$ConnectionBuilder OAuthResult ServerConfig Xtdb$Config)
            xtdb.api.module.XtdbModule
            xtdb.arrow.RelationReader
            (xtdb.error Anomaly Incorrect Interrupted)
@@ -62,7 +61,7 @@
 
                    server-state
 
-                   authn-rules]
+                   ^Authenticator authn]
   DataSource
   (createConnectionBuilder [_]
     (DataSource$ConnectionBuilder. "localhost" port))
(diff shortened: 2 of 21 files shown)


core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1409,14 +1459,18 @@
...
   (try
     (log/trace "Read client msg" {:cid cid, :msg msg})
 
     (err/wrap-anomaly {}
       (if (and (:skip-until-sync? @conn-state) (not= :msg-sync msg-name))
         (log/trace "Skipping msg until next sync due to error in extended protocol" {:cid cid, :msg msg})
-        (handle-msg* conn msg)))
+        (do
+          ;; Validate and refresh tokens for messages that execute queries or handle data
+          (when (and authn #{:msg-simple-query :msg-execute :msg-copy-data :msg-copy-done} msg-name)
```
danmason: I'm only checking for expired tokens/refreshing for particular message types that handle database read/write - happy to change this.

jarohen: shall we wrap the calls to XT rather than the messages? e.g. submit-tx, execute-tx, openQuery

core/src/main/kotlin/xtdb/api/Authenticator.kt
```diff
@@ -1,34 +1,54 @@
...
+import java.net.URL
+import clojure.lang.IPersistentMap
 
 val DEFAULT_RULES = listOf(MethodRule(TRUST))
 
 interface Authenticator : AutoCloseable {
     fun methodFor(user: String?, remoteAddress: String?): Method
 
-    fun verifyPassword(user: String, password: String): String =
+    fun verifyPassword(user: String, password: String): IPersistentMap? =
```
jarohen: given this is kt-land let's return a bean here. may need to be an interface with user-id as the return-type here, with specific impls for each authn implementation

core/src/main/kotlin/xtdb/api/Authenticator.kt
```diff
@@ -1,34 +1,54 @@
...
 
+    interface DeviceAuthResponse {
+        val url: URL
+        fun await(): IPersistentMap?
+    }
+
+    fun startDeviceAuth(user: String): DeviceAuthResponse =
+        throw UnsupportedOperationException("device auth not supported")
+
+    @Throws(IllegalArgumentException::class)
```
jarohen: dw about `@Throws` unless you need it to be present in the Java bytecode

core/src/main/kotlin/xtdb/api/Authenticator.kt
```diff
@@ -1,34 +1,54 @@
...
+    fun verifyClientCredentials(clientCredentials: String): IPersistentMap? =
+        throw UnsupportedOperationException("client credentials auth not supported")
```
jarohen: `throw Unsupported(...)` - we have an unsupported class that you can attach an error key to

core/src/main/kotlin/xtdb/api/Authenticator.kt
```diff
@@ -45,5 +65,18 @@ interface Authenticator : AutoCloseable {
...
                     .invoke(this, querySource, dbCatalog) as Authenticator
         }
+
+        @Serializable
+        @SerialName("!OpenIdConnect")
+        data class OpenIdConnect(
+            val issuerUrl: URL,
+            @Serializable(with = StringWithEnvVarSerde::class) val clientId: String,
+            @Serializable(with = StringWithEnvVarSerde::class) val clientSecret: String,
+            override var rules: List<MethodRule> = DEFAULT_RULES
```
jarohen: (not for this PR) we may move rules up, if we later allow multiple authn providers

core/src/main/clojure/xtdb/authn.clj
```diff
@@ -76,6 +87,171 @@
...
+(defn discover-oidc-config [issuer-url]
+  (try
+    (let [discovery-url (str issuer-url "/.well-known/openid-configuration")
+          {:keys [status body]} (http/get discovery-url
+                                          {:throw-exceptions false
+                                           :as :json})]
+      (when (= 200 status)
+        body))
+    (catch Exception e
+      (throw (IllegalArgumentException. (format "Failed to discover OIDC configuration from %s: %s" issuer-url (.getMessage e)))))))
```
jarohen: `throw (err/incorrect ...)`

core/src/main/clojure/xtdb/authn.clj
```diff
@@ -76,6 +87,171 @@
          :db-cat (ig/ref :xtdb/db-catalog)}
         opts))
 
+(defn discover-oidc-config [issuer-url]
+  (try
+    (let [discovery-url (str issuer-url "/.well-known/openid-configuration")
+          {:keys [status body]} (http/get discovery-url
+                                          {:throw-exceptions false
+                                           :as :json})]
+      (when (= 200 status)
```
jarohen: probably not for discover, but any other status codes that we need to handle specifically here?

core/src/main/clojure/xtdb/authn.clj
```diff
@@ -76,6 +87,171 @@
...
+        body))
+    (catch Exception e
+      (throw (IllegalArgumentException. (format "Failed to discover OIDC configuration from %s: %s" issuer-url (.getMessage e)))))))
+
+(defn oauth-token [{:keys [oidc-config client-id client-secret]} opts] 
+  (let [token-endpoint (:token_endpoint oidc-config)]
+    (when token-endpoint
+      (http/post token-endpoint
```
jarohen: other status codes to throw an exception, perhaps?

jarohen: more generally, let's use anomalies here and catch them at the outermost level in pgwire?

core/src/main/clojure/xtdb/authn.clj
```diff
@@ -76,6 +87,171 @@
...
+                 {:form-params (into {:client_id client-id
+                                      :client_secret client-secret}
+                                     opts)
+                  :throw-exceptions false
+                  :coerce :always
+                  :as :json}))))
+
+(defn calulcate-expires-at [oauth-token]
```
jarohen: thai po

core/src/main/clojure/xtdb/authn.clj
```diff
@@ -76,6 +87,171 @@
...
+  (when-let [expires-in (:expires_in oauth-token)]
+    (+ (System/currentTimeMillis) (* expires-in 1000))))
+
+(defn oauth-userinfo [{:keys [oidc-config]} token]
+  (let [userinfo-endpoint (:userinfo_endpoint oidc-config)]
```
jarohen: nit: when-let

core/src/main/clojure/xtdb/authn.clj
```diff
@@ -76,6 +87,171 @@
...
+
+(defn refresh-token [authn {:keys [refresh-token client-id client-secret] :as oauth}] 
+  (cond
+    ;; For password/device auth flows - use refresh token
+    refresh-token
+    (let [{:keys [status body]} (oauth-token authn {:grant_type "refresh_token"
+                                                    :refresh_token (:refresh_token oauth)})]
+      (when (= 200 status)
+        {:expires-at (calulcate-expires-at body)
+         :refresh_token (:refresh_token body)}))
```
jarohen: `:refresh-token`?

core/src/main/clojure/xtdb/authn.clj
```diff
@@ -76,6 +87,171 @@
...
+                                                            :scope "openid"}
+                                              :throw-exceptions false
+                                              :as :json})]
+        (when (= 200 status)
+          body)))))
+
+(defn parse-client-creds [^String credentials]
+  (cond
+    (nil? credentials)
+    {:error "Missing client credentials"}
```
jarohen: `throw (err/incorrect ...)`

core/src/main/clojure/xtdb/authn.clj
```diff
@@ -76,6 +87,171 @@
...
+    
+    (string/blank? credentials)
+    {:error "Empty client credentials"}
+    
+    :else
+    (try
+      (let [decoded (String. (.decode (Base64/getDecoder) credentials) "UTF-8")
```
jarohen: as discussed, drop b64

core/src/main/clojure/xtdb/authn.clj
```diff
@@ -76,6 +87,171 @@
...
+      (catch Exception e
+        {:error (format "Invalid base64 encoding: %s" (.getMessage e))}))))
+
+(defrecord DeviceAuthResponse [authn url device-code ^Duration interval]
+  Authenticator$DeviceAuthResponse
+  (getUrl [_] url)
+
+  (await [_]
+    (loop []
+      (let [{:keys [status body] :as token-response} (oauth-token authn
```
jarohen: nit: indentation

core/src/main/clojure/xtdb/authn.clj
```diff
@@ -76,6 +87,171 @@
...
+                                                                 {:grant_type "urn:ietf:params:oauth:grant-type:device_code"
+                                                                  :device_code device-code})]
+        (case (long status)
+          200 (let [user-info (oauth-userinfo authn (:access_token body))]
+                {:user-id (:sub user-info)
+                 :oauth-session {:user-info user-info
+                                :expires-at (calulcate-expires-at token-response)
+                                :refresh-token (:refresh_token body)}})
+          400 (when (= "authorization_pending" (:error body))
+                (Thread/sleep interval)
```
jarohen: timeout?

core/src/main/clojure/xtdb/authn.clj
```diff
@@ -76,6 +87,171 @@
...
+
+  (startDeviceAuth [this _user]
+    (let [{:keys [device_code verification_uri_complete interval]} (oauth-device-info this)]
+      (->DeviceAuthResponse this (.toURL (URI. verification_uri_complete))
+                            device_code (Duration/ofSeconds interval))))
+
+  (verifyClientCredentials [this client-credentials]
+    (let [parsed (parse-client-creds client-credentials)] 
+      (if (:error parsed)
+        (throw (IllegalArgumentException. (:error parsed)))
```
jarohen: anomaly

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -242,6 +248,13 @@
           (log/error ex "Uncaught exception processing message")
           {::severity :error, ::error-code "XX000"})))))
 
+(defn validate-and-refresh-token [{:keys [conn-state ^Authenticator authn]}]
```
jarohen, line 251: - implement as an Authenticator method - `revalidate`?
- `(:oauth @conn-state)` -> `(:authn-state @conn-state)`
- have Authenticator be given an `InstantSource`

core/src/main/kotlin/xtdb/api/Authenticator.kt
```diff
@@ -1,34 +1,54 @@
...
+    interface DeviceAuthResponse {
+        val url: URL
+        fun await(): IPersistentMap?
+    }
+
+    fun startDeviceAuth(user: String): DeviceAuthResponse =
+        throw UnsupportedOperationException("device auth not supported")
+
+    @Throws(IllegalArgumentException::class)
+    fun verifyClientCredentials(clientCredentials: String): IPersistentMap? =
```
jarohen: take client-id and client-secret here - it's a pgw impl detail that they're colon-separated

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -293,45 +306,80 @@
...
+          ;; asking for client-id/client-secret
+          (pgio/cmd-write-msg conn pgio/msg-auth {:result 3})
+
+          ;; we go idle until we receive a message
+          (when-let [{:keys [msg-name] :as msg} (pgio/read-client-msg! frontend)]
+            (if (not= :msg-password msg-name)
+              (throw (err-invalid-auth-spec (str "client credentials authentication failed for user: " user)))
+
+              (try
+                (if-let [auth-result (.verifyClientCredentials authn (:password msg))]
```
jarohen: split client-creds string at this point

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -293,45 +306,80 @@
...
+                  (let [user-id (:user-id auth-result)
```
jarohen: nit: you can include these as destructuring in the auth-result if-let (although changing to beans anyway)

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -293,45 +306,80 @@
...
+                        oauth-session (:oauth-session auth-result)]
+                    (pgio/cmd-write-msg conn pgio/msg-auth {:result 0})
+                    (when oauth-session
+                      (swap! (:conn-state conn) assoc :oauth oauth-session))
+                    (startup-ok conn (assoc startup-opts "user" user-id)))
+                  (throw (err-invalid-passwd (str "client credentials authentication failed for user: " user))))
+                (catch IllegalArgumentException e
```
jarohen: once you're throwing anomalies, you may not even need catches here necessarily. even if you do, you might 'just' be able to `send-ex` here

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1409,14 +1457,18 @@
...
   (try
     (log/trace "Read client msg" {:cid cid, :msg msg})
 
     (err/wrap-anomaly {}
       (if (and (:skip-until-sync? @conn-state) (not= :msg-sync msg-name))
         (log/trace "Skipping msg until next sync due to error in extended protocol" {:cid cid, :msg msg})
-        (handle-msg* conn msg)))
+        (do
+          ;; Validate and refresh tokens for messages that execute queries or handle data
+          (when (and authn #{:msg-simple-query :msg-execute :msg-copy-data :msg-copy-done} msg-name)
```
jarohen: we talked about wrapping the submit-tx/execute-tx/openQuery calls instead so that we ensure this check happens when the request passes through the bottleneck

src/test/clojure/xtdb/authn_test.clj
```diff
@@ -0,0 +1,222 @@
...
+  (:import [dasniko.testcontainers.keycloak KeycloakContainer]
+           [java.net URI]
+           [java.util Base64]
+           [org.keycloak.representations.idm ClientRepresentation CredentialRepresentation UserRepresentation RealmRepresentation]
+           [xtdb.api Authenticator]))
+
+(defonce ^KeycloakContainer container
+  (KeycloakContainer. "quay.io/keycloak/keycloak:26.0"))
+
+(defn seed! [^KeycloakContainer c & {:keys [access-token-lifespan]}]
```
jarohen: I usually go for multi-arity here

```clojure
(defn seed! 
  ([c] (seed! c {}))
  ([c {:keys [...]}]
   ...))
```

src/test/clojure/xtdb/authn_test.clj
```diff
@@ -0,0 +1,222 @@
...
+  (fn [f]
+    (tu/with-container container
+      (fn [_]
+        (f)))))
+
+(defn encode-client-creds [^String client-id ^String client-secret]
+  (let [client-cred-str (format "%s:%s" client-id client-secret)]
+    (.encodeToString (Base64/getEncoder) (.getBytes client-cred-str "UTF-8"))))
+
+(t/deftest test-parse-client-cred
```
jarohen: -> pgw-test

src/test/clojure/xtdb/authn_test.clj
```diff
@@ -0,0 +1,222 @@
...
+  (t/testing "test correct encoding"
+    (let [encoded (encode-client-creds "test-client" "test-secret")
+          result (authn/parse-client-creds encoded)]
+      (t/is (= {:client-id "test-client" :client-secret "test-secret"} result))))
+
+  (t/testing "test invalid inputs"
+    (t/is (= {:error "Missing client credentials"}
```
jarohen: `(t/is (anomalous? [...] (authn/parse-client-creds ...)))`

src/test/clojure/xtdb/authn_test.clj
```diff
@@ -0,0 +1,222 @@
...
+           [java.net URI]
+           [java.util Base64]
+           [org.keycloak.representations.idm ClientRepresentation CredentialRepresentation UserRepresentation RealmRepresentation]
+           [xtdb.api Authenticator]))
+
+(defonce ^KeycloakContainer container
+  (KeycloakContainer. "quay.io/keycloak/keycloak:26.0"))
+
+(defn seed! [^KeycloakContainer c & {:keys [access-token-lifespan]}]
+  (with-open [admin-client (.getKeycloakAdminClient c)]
```
jarohen: I would probably have written this in Kotlin, because it's all interop - np that it's in Clojure, i.e. dw about changing it now :)

src/test/clojure/xtdb/authn_test.clj
```diff
@@ -0,0 +1,222 @@
...
+      (t/is (= {:error "Credentials must contain exactly 2 parts separated by ':'"} 
+               (authn/parse-client-creds no-colon))))
+    
+    (let [too-many-parts (.encodeToString (Base64/getEncoder) (.getBytes "client:secret:extra" "UTF-8"))]
+      (t/is (= {:error "Credentials must contain exactly 2 parts separated by ':'"} 
+               (authn/parse-client-creds too-many-parts))))))
+
+(t/deftest test-token-expiry
+  (t/testing "calculates expiry correctly"
+    (let [current-time (System/currentTimeMillis)
```
jarohen: some form of mocking out the current-time here - InstantSource, and various `tu` utils

src/test/clojure/xtdb/authn_test.clj
```diff
@@ -0,0 +1,222 @@
...
+(t/deftest test-invalid-issuer-url
+  (t/testing "error caught when issuer URL is invalid"
+    (let [invalid-url (.toURL (URI. "http://invalid-url"))]
+      (t/is (thrown-with-msg? IllegalArgumentException #"Failed to discover OIDC configuration from" (authn/discover-oidc-config invalid-url)))))
+  
+  (t/testing "discovery fails with unknown realm"
+    (let [unknown-realm-url (.toURL (URI. (str (.getAuthServerUrl container) "/realms/other-realm")))]
+      (t/is (nil? (authn/discover-oidc-config unknown-realm-url))))))
+
+(t/deftest test-refresh-token-errors
```
jarohen, line 95: will also need test-revalidate in here

src/test/clojure/xtdb/authn_test.clj
```diff
@@ -0,0 +1,222 @@
...
+            (t/is (some? auth-result))
+            (t/is (some? oauth-session))
+            (t/is (= client-id (:client-id oauth-session)))
+            (t/is (not (authn/token-expired? oauth-session)))
+            
+            ;; Wait for token to expire
+            (Thread/sleep 5000)
+            
+            ;; Token should now be expired
+            (t/is (authn/token-expired? oauth-session))
```
jarohen: have `token-expired?` take a current-time

src/test/clojure/xtdb/oidc_integration_test.clj
```diff
@@ -0,0 +1,166 @@
...
+(t/deftest ^:integration test-oidc-password-flow
+  (authn-test/seed! authn-test/container)
+  (let [port (tu/free-port)
+        oidc-config [:openid-connect {:issuer-url (str (.getAuthServerUrl authn-test/container) "/realms/master")
+                                      :client-id "xtdb"
+                                      :client-secret "xtdb-secret"
+                                      :rules [{:user "test-user" :method :password :address "127.0.0.1"}]}]]
+    (with-open [_node (xtn/start-node {:authn oidc-config
+                                       :server {:port port}})]
+      (let [ds-spec {:dbtype "postgresql"
```
jarohen: use xt/client for this

src/test/clojure/xtdb/oidc_integration_test.clj
```diff
@@ -0,0 +1,166 @@
...
+
+(t/use-fixtures :once
+  (fn [f]
+    (tu/with-container authn-test/container
+      (fn [_]
+        (f)))))
+
+(t/deftest ^:integration test-oidc-password-flow
+  (authn-test/seed! authn-test/container)
+  (let [port (tu/free-port)
```
jarohen: dw about `free-port` here, start-node will create one

src/test/clojure/xtdb/oidc_integration_test.clj
jarohen: and then you can use `createConnectionBuilder` to supply username and password

src/test/clojure/xtdb/oidc_integration_test.clj
```diff
@@ -0,0 +1,166 @@
...
+                     :password (authn-test/encode-client-creds client-id client-secret)}]
+
+        (t/testing "successful client credentials authentication"
+          (with-open [conn (jdbc/get-connection ds-spec)]
+            (let [result (jdbc/execute-one! conn ["SELECT 'authenticated' as status"])]
+              (t/is (= "authenticated" (:status result))))))
+
+        
+        (t/testing "failed client credentials authentication with badly encoded credentials"
+          (t/is (thrown-with-msg? PSQLException #"ERROR: client credentials error: Invalid base64 encoding: Illegal base64 character 2d"
```
jarohen: there's a pgw-test util here which unwraps returned anomalies, so you can then test with `(t/is (anomalous? ...))` again

src/test/clojure/xtdb/oidc_integration_test.clj
```diff
@@ -0,0 +1,166 @@
...
+        
+        (t/testing "session persists across multiple queries"
+          (with-open [conn (jdbc/get-connection ds-spec)]
+            (let [result1 (jdbc/execute-one! conn ["SELECT 1 as first_query"])
+                  result2 (jdbc/execute-one! conn ["SELECT 2 as second_query"])]
+              (t/is (= 1 (:first_query result1)))
+              (t/is (= 2 (:second_query result2))))))
+        
+        (t/testing "multiple concurrent connections work"
+          (let [conn1 (jdbc/get-connection ds-spec)
```
jarohen: `with-open`

src/test/clojure/xtdb/oidc_integration_test.clj
```diff
@@ -0,0 +1,166 @@
...
+                     :password (authn-test/encode-client-creds client-id client-secret)}]
+        
+        (t/testing "token expiry and automatic refresh works"
+          ;; Initial connection should work
+          (with-open [conn (jdbc/get-connection ds-spec)]
+            (let [result (jdbc/execute-one! conn ["SELECT 'initial' as status"])]
+              (t/is (= "initial" (:status result)))))
+          
+          ;; Wait for token to expire
+          (Thread/sleep 5000)
```
jarohen: smell 🙂

src/test/clojure/xtdb/pgwire_protocol_test.clj
```diff
@@ -64,14 +64,16 @@
   (^java.lang.AutoCloseable [frontend startup-opts]
    (->conn frontend startup-opts [{:user nil, :method #xt.authn/method :trust, :address nil}]))
   (^java.lang.AutoCloseable [frontend startup-opts authn-rules]
-   (let [conn (pgwire/map->Connection {:server {:server-state (atom {:parameters {"server_encoding" "UTF8"
+   (let [authn (authn/->UserTableAuthn authn-rules
```
jarohen, line 68: could add some OIDC tests here, check the protocol messages are correct

docs/src/content/docs/ops/config/authentication.adoc
```diff
@@ -2,28 +2,23 @@
 title: Authentication
 ---
 
-== Server-level settings
+XTDB provides authentication to control database access and secure connections. Authentication rules determine which users can connect and what credentials they must provide.
```
jarohen: sentence per line

docs/src/content/docs/ops/config/authentication.adoc
```diff
@@ -2,28 +2,23 @@
...
-The above authentication rule means that the `admin` user requires a password connecting from `localhost`.
-The authentication rules are passed to the server at startup and can be configured as below.
-The rules are considered in order until the first one matches.
-If none matches the connection is rejected.
+== Authentication Providers
+
+XTDB supports two authentication providers:
+
+- **User Table** (`!UserTable`): Uses an internal user table with password-based authentication
+- **OpenID Connect** (`!OpenIdConnect`): Integrates with enterprise identity providers like Keycloak, Auth0, or Azure AD
```
jarohen: s/enterprise/external/

Keycloak, Auth0, AWS Cognito or Azure Entra

docs/src/content/docs/ops/config/authentication.adoc
```diff
@@ -32,23 +27,78 @@ authn: !UserTable
...
 ----
 ALTER USER ada WITH PASSWORD 'LOVELACE'
 ----
+
+**Password Validation**::
+When `PASSWORD` method is specified, credentials are validated against the `pg_user` table entries.
+
+== OpenID Connect (OIDC)
+
+The `!OpenIdConnect` authentication method integrates with enterprise identity providers like Keycloak, Auth0, or Azure AD.
```
jarohen: same as above

docs/src/content/docs/ops/config/authentication/oidc.adoc
```diff
@@ -0,0 +1,114 @@
...
+== Troubleshooting
+
+**Authentication Failed**::
+Verify client ID/secret configuration and issuer URL accessibility.
+
+**Token Expired**::
+Check access token lifespan settings in identity provider.
+
+**Invalid Credentials (Client Credentials)**::
+Verify base64 encoding: `echo -n "client-id:client-secret" | base64`
```
jarohen: b64

docs/src/content/docs/ops/config/authentication/oidc.adoc
```diff
@@ -0,0 +1,114 @@
+---
+title: OpenID Connect Authentication
+---
+
+XTDB integrates with OpenID Connect identity providers (Keycloak, Auth0, Azure AD) for enterprise authentication. OIDC authentication supports multiple OAuth 2.0 flows for different client types.
+
+== Identity Provider Requirements
+
+**Required Capabilities**::
+- OpenID Connect Discovery (/.well-known/openid_configuration)
```
jarohen: adoc nit: consistency about your bullet points

`*` for top-level, `**` for second

docs/src/content/docs/ops/config/authentication/oidc.adoc
```diff
@@ -0,0 +1,114 @@
...
+  # Should be stored as an environment variable for security.
+  # (Can be set as an !Env value)
+  clientSecret: !Env OIDC_CLIENT_SECRET
+
+  # Authentication rules determine which OAuth flow applies to each connection.
+  # Rules are evaluated in order until the first match.
+  # See the main Authentication page for complete rule syntax.
+  rules:
+    # Service accounts use client credentials flow
+    - user: oidc-client
```
jarohen, line 47: specific note to use `oidc-client` as your username, and the colon-delimited password

docs/src/content/docs/ops/config/authentication/oidc.adoc
```diff
@@ -0,0 +1,114 @@
...
+      method: CLIENT_CREDENTIALS
+
+    # Specific users can use device authorization flow
+    - user: contractor
```
jarohen: TODO (out of scope of PR) ignore username given in connection in device flow - same as client-creds (i.e. use oidc-device)

docs/src/content/docs/ops/config/authentication/oidc.adoc
```diff
@@ -0,0 +1,114 @@
...
+== Authentication Flows
+
+NOTE: Not all OIDC providers support all OAuth 2.0 flows. PASSWORD and DEVICE_AUTH flows require specific provider support and configuration. Check your provider's documentation for supported grant types.
+
+=== CLIENT_CREDENTIALS Method
+
+Uses OAuth Client Credentials flow for service accounts and automated processes.
+
+**Authentication Format**::
+Base64-encoded `client-id:client-secret` used as password.
```
jarohen: b64

docs/src/content/docs/ops/config/authentication/oidc.adoc
```diff
@@ -0,0 +1,114 @@
...
+    - method: PASSWORD
+----
+
+== Authentication Flows
+
+NOTE: Not all OIDC providers support all OAuth 2.0 flows. PASSWORD and DEVICE_AUTH flows require specific provider support and configuration. Check your provider's documentation for supported grant types.
+
+=== CLIENT_CREDENTIALS Method
+
+Uses OAuth Client Credentials flow for service accounts and automated processes.
```
jarohen, line 85: here's where you can put the note around what to specify as username - that the provided username in the connection is completely ignored except for deciding on the auth method.

if the user is _only_ using client-creds/only using device auth, they don't need to filter by user - this is only if they want to support multiple flows and differentiate by the conn's username

docs/src/content/docs/ops/config/authentication/oidc.adoc
```diff
@@ -0,0 +1,114 @@
...
+----
+
+=== DEVICE_AUTH Method
+
+Uses OAuth Device Authorization flow for applications that cannot securely store secrets.
+
+**Flow Process**::
+1. Client requests device code from XTDB
+2. User visits verification URL and enters user code
+3. Client polls XTDB until user completes authentication
```
jarohen: XTDB polls identity provider

jarohen (changes requested):

core/src/main/kotlin/xtdb/api/Authenticator.kt
danmason: Going for a kotlin data class here - varied per authn implementation.

src/test/clojure/xtdb/oidc_integration_test.clj
danmason: Can't seem to find what you're referencing here?

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -242,6 +248,16 @@
           (log/error ex "Uncaught exception processing message")
           {::severity :error, ::error-code "XX000"})))))
 
+(defn validate-and-refresh-token [{:keys [conn-state ^Authenticator authn]}]
+  (when-let [^OAuthResult authn-state (:authn-state @conn-state)]
+    (some->> (.revalidate authn authn-state) (swap! conn-state assoc :authn-state))))
```
jarohen: not a fan of using threading macros when `swap!`'s the target, the reader has to reason about what's happening - try to limit threading macros to cases when you're transforming a value.

`when-let` is likely much clearer here

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -972,7 +1028,7 @@
 
     (letfn [(->cursor ^xtdb.IResultCursor [xt-args]
               (util/with-close-on-catch [args-rel (vw/open-args allocator xt-args)]
-                (.openQuery prepared-query (assoc query-opts :args args-rel))))
+                (with-auth-check conn (.openQuery prepared-query (assoc query-opts :args args-rel)))))
```
jarohen: I'd probably stick `with-auth-check` around the with-close-on-catch here, feels like you're doing the auth check halfway through 🙂

core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1032,7 +1088,7 @@
 
                      (let [^RelationReader args-rel (aget !args 0)]
                        (case (:statement-type inner)
-                         :query (util/with-close-on-catch [inner-cursor (.openQuery inner-pq (assoc query-opts :args args-rel))]
+                         :query (util/with-close-on-catch [inner-cursor (with-auth-check conn (.openQuery inner-pq (assoc query-opts :args args-rel)))]
```
jarohen: same here (although I know that's the first thing) - this one is more about making it easier to spot rather than hiding it under a binding

src/test/clojure/xtdb/pgwire_protocol_test.clj
```diff
@@ -349,3 +350,108 @@
...
+    (try
+      (pgwire/cmd-startup-pg30 conn startup-opts)
+      (catch Exception e 
+        (if (#{:xtdb/authn-failed :xtdb/invalid-client-credentials} (:xtdb.error/code (ex-data e)))
+          (doto conn
+            (pgwire/send-ex e)
+            (pgwire/handle-msg* {:msg-name :msg-terminate}))
+          (throw e))))))
+
+(deftest ^:integration test-oidc-password-auth-successs
```
jarohen, line 390: niice ☺️ liking these tests, very clear to see

jarohen (approved):
nice one @danmason ☺️

PR #4691, opened 2025-08-20
Baby's first property testing

Akeboshiwind:
Definitely needs more love, but already @danmason and I have discovered some issue:
- Nullable DUV issue on an earlier version of main
- https://github.com/xtdb/xtdb/issues/4690
- https://github.com/xtdb/xtdb/issues/4692
- https://github.com/xtdb/xtdb/issues/4693
- https://github.com/xtdb/xtdb/issues/4694

~Probably should use `clojure.test.check` or something~

### DIFF
--- a/src/test/clojure/xtdb/baby_deterministic_test.clj
+++ b/src/test/clojure/xtdb/baby_deterministic_test.clj
@@ -1,0 +1,218 @@
+(ns xtdb.baby-deterministic-test
+  (:require [clojure.test :as t]
+            [xtdb.api :as xt]
+            [xtdb.compactor :as c]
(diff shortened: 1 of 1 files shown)


Akeboshiwind:
Hmm, the `test.check` stuff finds errors but I find it hard to actually reproduce them with the examples.

FiV0:
FYI https://github.com/xtdb/xtdb/blob/main/src/test/clojure/xtdb/vector/arrow_generative_test.clj. I added this at some point. They were failing all the time, so we turned them off.

jarohen:
superseded - thanks @Akeboshiwind :pray:

PR #4696, opened 2025-08-21
add vector functions (cosine_distance, l2_distance, dot_product)

refset:


refset:
It was a fun to spike this, but I don't intend to push this through to completion for the moment, so closing for now.

Semi-related, interesting piece on why you'd probably want to bring these things in alongside BM25 or similar instead of just relying on vector searching: https://techcommunity.microsoft.com/blog/azuredevcommunityblog/doing-rag-vector-search-is-not-enough/4161073

PR #4713, opened 2025-08-27
Autoscaling in azure benchmark cluster

### LINKED ISSUE #4706: Benchmark application nodes should auto-scale using k8s horizontal scaling.
Currently, when setting up benchmark infrastructure, we configure a **fixed count of nodes** in the application pool.  
- These nodes run continuously - regardless of whether a benchmark is running, or how many nodes are actually needed.  
- This leads to unnecessary cost and idle resource usage - this would especially be the case if we leave the infrastructure deployed between runs.

We want to enable autoscaling for the application nodes in the Azure benchmark cluster so that:
- Nodes scale **up** when benchmark workloads run.
- Nodes scale **down** once workloads complete (ie, helm chart has been uninstalled)

## TODO
- [x] Update shared Terraform modules to accept min_count and max_count.
  - We can default these to be equal to the node count, so there is a fixed number of nodes in the sample terraform. 
- [x] Update benchmark infra to use these:
  - Initial node count of zero. 
  - Min count of zero nodes on the application pool. 
  - Some reasonable max count, say 3.
- [x] Verify autoscaling ad-hoc on at least one benchmark (e.g. Auctionmark with multiple nodes).
  - Should start with zero nodes in the node pool.
  - Should scale to the number of nodes it requires.
  - Following uninstall/clearup of benchmark, should scale back down to zero nodes.

tggreene:
I've checked the autoscaling behaviour on AKS and looks to be what we expect i.e. scale to max when in high use (auctionmark multiple nodes) and scale down following although not immediately to 0, I haven't seen it fall below 1 after an hour.

Closes #4706.

### DIFF
--- a/azure/terraform/variables.tf
+++ b/azure/terraform/variables.tf
@@ -46,8 +46,14 @@ variable "aks_system_pool_vm_size" {
   default     = "Standard_D2pds_v6"
 }
 
-variable "aks_system_pool_node_count" {
-  description = "Number of nodes in the XTDB AKS system node pool"
+variable "aks_system_pool_node_max_count" {
+  description = "Maximum number of nodes in the XTDB AKS system node pool"
+  type        = number
+  default     = 2
+}
+
+variable "aks_system_pool_node_min_count" {
+  description = "Minimum number of nodes in the XTDB AKS system node pool"
   type        = number
   default     = 2
 }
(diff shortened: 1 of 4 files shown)


azure/terraform/variables.tf
```diff
@@ -70,6 +76,18 @@ variable "aks_application_pool_node_count" {
   default     = 3
 }
 
+variable "aks_application_pool_node_max_count" {
```
danmason, line 79: All looks sensible to me assuming our sample terraform deploy worked as expected, would suggest adding this under terraform.tfvars also.

tggreene: Whoops, completely spaced that, thanks Dan

PR #4716, opened 2025-08-27
Add XTQL set operators: union, distinct, except etc.

### LINKED ISSUE #2980: XTQL set operators
Should work like `->` in terms of scoping (implicit passing of params)

- [ ] union/union-all
- [ ] intersect/intersect-all (SQL naming)
- [ ] except/except-all (likewise)
- [ ] distinct

refset:
Resolves #2980

Given my non-familiarity with the XTQL compiler I asked Claude to add a few larger test examples (which we could happily drop if the implementation looks sound). This PR was largely implemented by Claude.

(context: required for manually translating Datalog or-joins into XTQL)

### DIFF
--- a/api/src/main/clojure/xtdb/xtql.clj
+++ b/api/src/main/clojure/xtdb/xtql.clj
@@ -624,6 +624,66 @@
                                                    :message "Union must contain a least one sub query"})))
   (->UnionAll (mapv #(parse-query % env) queries)))
 
+(defrecord Union [union-queries]
+  XtqlQuery
+  UnparseQuery (unparse-query [_] (list* 'union (mapv unparse-query union-queries))))
+
+(defmethod parse-query 'union [[_ & queries :as this] env]
+  (when (> 1 (count queries))
+    (throw (err/illegal-arg :xtql/malformed-union {:union this
+                                                   :message "Union must contain a least one sub query"})))
+  (->Union (mapv #(parse-query % env) queries)))
+
+(defrecord Intersect [intersect-queries]
+  XtqlQuery
+  UnparseQuery (unparse-query [_] (list* 'intersect (mapv unparse-query intersect-queries))))
+
+(defmethod parse-query 'intersect [[_ & queries :as this] env]
+  (when (> 1 (count queries))
+    (throw (err/illegal-arg :xtql/malformed-intersect {:intersect this
+                                                       :message "Intersect must contain a least one sub query"})))
+  (->Intersect (mapv #(parse-query % env) queries)))
+
+(defrecord IntersectAll [intersect-all-queries]
+  XtqlQuery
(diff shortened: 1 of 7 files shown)


jarohen (changes requested):
thanks @refset/Claude 🙏 

- intersect and except need some more test cases that check out their multiplicity handling - see set_test.clj for examples
- comments above `t/is`: these can either be elided if it's obvious, or they can become strings as the second argument to `t/is` - e.g. `(t/is (= expected actual) "comment describing test")`
- don't need `_` as the final binding in a let - just move the RHS into the body of the let

api/src/main/clojure/xtdb/xtql.clj
```diff
@@ -624,6 +624,66 @@
                                                    :message "Union must contain a least one sub query"})))
   (->UnionAll (mapv #(parse-query % env) queries)))
 
+(defrecord Union [union-queries]
+  XtqlQuery
+  UnparseQuery (unparse-query [_] (list* 'union (mapv unparse-query union-queries))))
+
+(defmethod parse-query 'union [[_ & queries :as this] env]
+  (when (> 1 (count queries))
+    (throw (err/illegal-arg :xtql/malformed-union {:union this
```
jarohen: `err/illegal-arg` -> `err/incorrect` (and also check the args, Claude has misunderstood this one 😅)

core/src/main/clojure/xtdb/xtql/plan.clj
```diff
@@ -474,7 +474,67 @@
...
+                {:ra-plan [:distinct [:union-all (:ra-plan left-plan) (:ra-plan right-plan)]]
+                 :provided-vars (set/union (:provided-vars left-plan) (:provided-vars right-plan))}))
+            (plan-query (first union-queries))
+            (rest union-queries)))
+
+  Intersect
+  (plan-query [{:keys [intersect-queries]}]
+    (reduce (fn [left-plan query]
+              (let [right-plan (plan-query query)]
+                {:ra-plan [:distinct [:intersect (:ra-plan left-plan) (:ra-plan right-plan)]]
```
jarohen: can `distinct` be applied at the outside here? otherwise we're distincting every pair.

(might be what's required - main aim is that it's consistent with SQL (which I know doesn't have variadic intersect 🙂))

api/src/main/clojure/xtdb/xtql.clj
refset: Noting that I won't make this change across the rest of the ns in this PR (so breaking consistency a little)

core/src/main/clojure/xtdb/xtql/plan.clj
refset: good spot :bulb:

refset:
Not required currently. We'll revisit and reimplement as needed.

PR #4719, opened 2025-08-28
Initial property testing work - generators, gradle task and some block boundary tests

### LINKED ISSUE #4722: Setup property-based testing in the test suite
Want to add some initial support for setting up and running property-based tests in XTDB using clojure.test.check. This includes a Gradle task for running them with configurable iterations, some shared generators/utilities, and at least one example test in the suite.

- [x] Add shared generators/utilities for our existing simple/composite Arrow types, using test.check generators.
- [x] Add an initial set of property tests (using the generators and test.check).
  - Ensure these are tagged such that they do not get ran with the other unit tests.
  - Want to have a configurable number of iterations and easy ability to re-run a particular seed 
- [x] Add a Gradle task for running property tests with a configurable number of iterations.

danmason:
Resolves #4722 

(Based on some initial pairing work within #4691, and the contents of `arrow_generative_test`)

This PR introduces an early implementation of property-based testing infrastructure for XTDB, with a focus on block boundary behavior and type handling across database operations.

Added within this PR:
- Type Generators
  - Generators for XTDB’s supported "simple" data types (numerical, temporal, etc.)
  - Generators for collections, structs, and onions
  - Recursive generators for nested values
  - Document/record generators with configurable ID sets, ensuring overlapping keys across test runs
  - Safe keyword generation to avoid case-sensitivity conflicts in struct fields
- Some initial plumbing for running property tests: 
  - `run-property-test` utility with configurable iteration counts and reproducible seeds
    - As a note, did attempt to use `defspec` from clojure.test though ran into issues surfacing errors.  
  - Property tests tagged using ^:property metadata
  - New Gradle task for running property tests, supporting configurable iteration counts
- Initial Block Boundary Tests
  - Covering ingestion and query behavior of documents with differing types across block boundaries

Property tests are isolated from unit tests and run via a dedicated Gradle task (similar to `integration-test`):
```
# Default 100 iterations
./gradlew property-test

# Custom iteration count
./gradlew property-test -Piterations=10000
```

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -160,6 +160,7 @@ allprojects {
             testImplementation(libs.testcontainers.kafka)
             testImplementation(libs.testcontainers.keycloak)
             testImplementation(libs.testcontainers.minio)
+            testImplementation(libs.clojure.test.check)
         }
 
         if (plugins.hasPlugin("dev.clojurephant.clojure")) {
@@ -419,7 +420,6 @@ dependencies {
     testImplementation("metosin", "jsonista", "0.3.3")
     testImplementation("clj-commons", "clj-yaml", "1.0.27")
     testImplementation("org.xerial", "sqlite-jdbc", "3.39.3.0")
-    testImplementation("org.clojure", "test.check", "1.1.1")
     testImplementation("clj-kondo", "clj-kondo", "2023.12.15")
     testImplementation("com.github.igrishaev", "pg2-core", "0.1.33")
     testImplementation(libs.hato)
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -121,7 +121,7 @@ allprojects {
             // To stub an AWS region
             environment("AWS_REGION", "eu-west-1")
             useJUnitPlatform {
-                excludeTags("integration", "jdbc", "timescale", "s3", "minio", "slt", "docker", "azure", "google-cloud")
+                excludeTags("integration", "property", "jdbc", "timescale", "s3", "minio", "slt", "docker", "azure", "google-cloud")
             }
 
             /*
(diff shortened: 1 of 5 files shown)


danmason:
Already encountered a fail for this at 100 iterations from `block-boundaries-all-live` - after failing, test.check returns a "shrunk" failure which should be relatively minimal to repro with:

```clojure
{:total-nodes-visited 198,
 :depth 97,
 :pass? false,
 :result
 #xt/error [:fault :xtdb.error/unknown "Cannot invoke \"org.apache.arrow.vector.types.pojo.Field.getType()\" because \"field\" is null" {}],
 :result-data
 {:clojure.test.check.properties/error
  #xt/error [:fault :xtdb.error/unknown "Cannot invoke \"org.apache.arrow.vector.types.pojo.Field.getType()\" because \"field\" is null" {}]},
 :time-shrinking-ms 23160,
 :smallest
 [[{:a nil, :b nil, :h {:b #{}}, :c nil, :xt/id 1}]
  [{:h {}, :a nil, :xt/id 1} {:a nil, :b nil, :xt/id 1}]]}
```

This also returns a seed, which can be passed as an option into `run-property-test` (in this case, `1756393002180`).

Sure enough, can write a failing test using the records from "smallest":
```
(t/deftest block-boundaries-all-live-fail
  (with-open [node (xtn/start-node {:databases {:xtdb {:log [:in-memory {:instant-src (tu/->mock-clock)}]}}
                                    :compactor {:threads 0}})]
    (xt/execute-tx node [[:put-docs :docs {:a nil, :b nil, :h {:b #{}}, :c nil, :xt/id 1}]])
    (xt/execute-tx node [[:put-docs :docs {:h {}, :a nil, :xt/id 1} {:a nil, :b nil, :xt/id 1}]])
  
    (t/is (= :bang (xt/q node "SELECT * FROM docs ORDER BY _id")))))
```

jarohen (approved):
nice one @danmason @Akeboshiwind - good to merge as-is 👏 

I'd probably look at adding some between-block tests where all of the documents on either side of the block boundary have the same types - i.e. all ints on one side, all strings on the other, or not present on the other, etc - these are the ones that we've seen issues with recently.

PR #4733, opened 2025-09-02
Add github action to bench

### LINKED ISSUE #4705: Add nightly job to trigger Azure benchmarks
We want to setup github actions for running benchmarks on Azure.

## Notes  
- CI should **not** create/destroy infra.
- CI should only:  
  - Authenticate against AKS and verify it is available.  
  - Trigger the benchmark workloads (e.g. via Helm).  
- To avoid having application nodes running 24/7:  
  - Consider enabling Kubernetes autoscaling in the Terraform templates.  
  - Nodes should scale up when benchmarks run and scale down once jobs are deleted.  
- We may also want a **separate job/action** that:  
  - Cleans up completed benchmark runs.  
  - Collects any generated artifacts (logs, results) for storage/analysis.  

## TODO:
- [x] Cluster scales application nodes up/down automatically based on workload (autoscaling). 
  - #4706
- [ ] Add a github action that triggers benchmark workloads on AKS.  
  - Can likely start with TPCH here.
  - Action authenticates against AKS and verifies cluster availability.  
  - Benchmarks run independently of CI (Action does not need to monitor logs).  
  - Action surfaces basic run metadata.
  - Should be ran **nightly**.   
- [ ] Add a separate action to clean up old runs and collect any artifacts.  
  - Should run some amount of time after the initial benchmark run.
  - Should gather & upload logs and any other generated artifacts.

tggreene:
Very simply adds a call to GH to invoke the nightly benchmark cleanup job.

Also adds a try catch around the main benchmark run so we can report failures.

### DIFF
--- a/modules/bench/build.gradle.kts
+++ b/modules/bench/build.gradle.kts
@@ -26,8 +26,10 @@ dependencies {
     // bench
     api("com.github.oshi", "oshi-core", "6.3.0")
     api("pro.juxt.clojars-mirrors.hiccup", "hiccup", "2.0.0-alpha2")
+    api("org.testcontainers", "testcontainers", "1.20.1")
 
     api("software.amazon.awssdk", "s3", "2.25.24")
+    api("clj-http", "clj-http", "3.13.1")
 }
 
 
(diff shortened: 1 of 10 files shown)


danmason (approved):
LGTM :)

PR #4734, opened 2025-09-02
Add nightly benchmark jobs

### LINKED ISSUE #4705: Add nightly job to trigger Azure benchmarks

tggreene:
This adds two jobs:
 - Nightly Benchmark TPC-H (which runs nightly at midnight)
   - Gets k8s credentials
   - Runs benchmark with configuration
   - Posts to slack to notify job start with parameters
 - Nightly Benchmark Cleanup (which is invoked via a callback)
   - Gets k8s credentials
   - Checks whether a helm deployment is present exits early if not
   - Captures logs and uploads to azure
   - Messages slack with status of job, time taken ms and a link to the log

### DIFF
--- a/.github/workflows/nightly-benchmark-cleanup.yml
+++ b/.github/workflows/nightly-benchmark-cleanup.yml
@@ -1,0 +1,129 @@
+name: Nightly Benchmark Cleanup
+
+on:
+  workflow_dispatch:
+    inputs:
+      benchType:
+        description: 'Type of benchmark to clean up'
+        required: true
+        type: choice
+        options:
+          - tpch
+          - auctionmark
+          - readings
+      status:
+        description: 'Status of the benchmark'
+        required: true
+        type: choice
+        options:
+          - success
+          - failure
+      nodeId:
+        description: 'Node ID of the benchmark'
+        required: true
+        type: string
+
+permissions:
+  contents: write
+
+concurrency:
+  group: cleanup-benchmarks
+  cancel-in-progress: false
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Azure CLI Login
+        uses: azure/login@v2
+        with:
+          client-id: ${{ secrets.BENCHMARK_AZURE_CLIENT_ID }}
+          tenant-id: ${{ secrets.BENCHMARK_AZURE_TENANT_ID }}
+          subscription-id: ${{ secrets.BENCHMARK_AZURE_SUBSCRIPTION_ID }}
+
+      - name: Acquire Kubernetes Configuration
+        run: |
+          az aks get-credentials --resource-group cloud-benchmark-resources --name xtdb-bench-cluster
+
+      - name: Check Deployment
+        id: check-deployment
+        continue-on-error: true
+        run: |
+          # exit 0 if exists, non-zero if not
+          helm status xtdb-benchmark -n cloud-benchmark >/dev/null 2>&1
+
+      - name: Capture Benchmark Logs
+        if: steps.check-deployment.outcome == 'success'
+        id: logs
+        run: |
+          # Wait for job logs with retries
+          max_attempts=12
+          attempt=1
+          while [ $attempt -le $max_attempts ]; do
+            if kubectl logs ${{ inputs.nodeId }} --namespace cloud-benchmark > benchmark.log 2>/dev/null; then
+              if grep -q '"time-taken-ms":' benchmark.log; then
+                time_taken=$(grep -o '"time-taken-ms":[0-9]\+' benchmark.log | tail -1 | cut -d: -f2)
+                echo "time_taken=$time_taken" >> "$GITHUB_OUTPUT"
+                break
+              fi
+            fi
+            echo "Attempt $attempt: Waiting for benchmark logs..."
+            sleep 5
+            attempt=$((attempt + 1))
+          done
+
+      - name: Set Timestamp
+        id: timestamp
+        run: |
+          echo "value=$(date -u +"%Y-%m-%d-%H-%M-%S")" >> "$GITHUB_OUTPUT"
+          echo "date=$(date -u +"%Y/%m/%d")" >> "$GITHUB_OUTPUT"
+
+      - name: Upload Benchmark Logs
+        id: upload
+        uses: actions/upload-artifact@v4
+        with:
+          name: benchmark-logs-${{ inputs.benchType }}-${{ steps.timestamp.outputs.value }}-${{ inputs.nodeId }}-${{ inputs.status }}
+          path: benchmark.log
+          retention-days: 7
+
+      - name: Run Cleanup Script
+        if: ${{ always() && steps.check-deployment.outcome == 'success' }}
+        run: |
+          ./modules/bench/cloud/clear-bench.sh azure
+
+      - name: Compose Slack Message
+        id: compose
+        run: |
+          type="${{ inputs.benchType }}"
+          [ "$type" = "tpch" ] && type="TPC-H"
+
+          if [ -n "${{ steps.logs.outputs.time_taken }}" ]; then
+            base="$type Benchmark completed in ${{ steps.logs.outputs.time_taken }}ms"
+          else
+            base="$type Benchmark cleanup"
+          fi
+
+          {
+            echo "msg<<EOF"
+            echo "$base — {status_message}"
+            if [ "${{ steps.upload.outcome }}" = "success" ]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly-benchmark-cleanup.yml
```diff
@@ -0,0 +1,174 @@
...
+            az storage blob upload \
+              --connection-string "${{ secrets.BENCHMARK_AZURE_STORAGE_CONNECTION_STRING }}" \
+              --container-name benchmark-logs \
+              --file benchmark.log \
+              --name "${{ inputs.benchType }}/${{ steps.timestamp.outputs.date }}/${{ steps.timestamp.outputs.value }}-${{ inputs.nodeId }}-${{ inputs.status }}.log" \
+              --content-type "text/plain; charset=utf-8" \
+              --overwrite true
+
+      - name: Generate read-only SAS URL for the log
+        id: sas
```
danmason: What's the SAS URL generation doing for us, out of interest?

.github/workflows/nightly-benchmark-cleanup.yml
```diff
@@ -0,0 +1,174 @@
...
+            base="$type Benchmark completed in ${{ steps.logs.outputs.time_taken }}ms"
+          else
+            base="$type Benchmark cleanup"
+          fi
+
+          {
+            echo "msg<<EOF"
+            echo "$base — {status_message}"
+            if [ "${{ steps.upload.outcome }}" = "success" ]; then
+              echo ":link: <${{ steps.sas.outputs.url }}|benchmark log>"
```
danmason: Could we perhaps upload the logs as an artifact on the cleanup job itself, and link to that? (ie, upload them as a github artifact?). I realize we need them stored longer term for the sake of check regressions, just curious.

.github/workflows/nightly-benchmark-tpch.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Run Benchmark
+        if: steps.check-deployment.outcome != 'success'
+        run: |
+          helm dependency update ./modules/bench/cloud/helm
+          helm upgrade --install "xtdb-benchmark" ./modules/bench/cloud/helm \
+            --namespace "cloud-benchmark" \
+            --create-namespace \
+            -f ./modules/bench/cloud/azure/values.yaml \
+            --set "benchType=tpch" \
+            --set "tpch.scaleFactor=${{ inputs.scale_factor }}" \
```
danmason: Would this/the inputs be unset on the nightly run version of that? Might need to handle for that here, somehow, or pass inputs on the nightly one.

.github/workflows/nightly-benchmark-cleanup.yml
```diff
@@ -0,0 +1,168 @@
...
+            attempt=$((attempt + 1))
+          done
+
+      - name: Set Timestamp
+        id: timestamp
+        run: |
+          echo "value=$(date -u +"%Y-%m-%d-%H-%M-%S")" >> "$GITHUB_OUTPUT"
+          echo "date=$(date -u +"%Y/%m/%d")" >> "$GITHUB_OUTPUT"
+
+      - name: Upload Benchmark Logs
```
danmason, line 84: What storage account are we using for the persistent log uploads? Is it separate to the infra deployment (ie, so delete-infra wouldnt get rid of it)?

.github/workflows/nightly-benchmark-cleanup.yml
tggreene: Allows us to post a link in slack which is authed so you can immediately view it

.github/workflows/nightly-benchmark-cleanup.yml
tggreene: I thought about that, I just naturally assumed github files would be pricier but definitely worth thinking about

.github/workflows/nightly-benchmark-tpch.yml
tggreene: It would take the default value of 0.01 which is probably a bit low. What do you think it should be for the nightly run?

.github/workflows/nightly-benchmark-cleanup.yml
```diff
@@ -0,0 +1,168 @@
...
+          - auctionmark
+          - readings
+      status:
+        description: 'Status of the benchmark'
+        required: true
+        type: choice
+        options:
+          - success
+          - failure
+      nodeId:
```
danmason, line 21: Realize nodeId is passed along with the callback and we use it to identify which pod to scrape - what sets XTDB_NODE_ID?

.github/workflows/nightly-benchmark-cleanup.yml
```diff
@@ -0,0 +1,168 @@
...
+                  --https-only \
+                  --full-uri \
+                  -o tsv)
+            echo "url=$URL" >> "$GITHUB_OUTPUT"
+
+      - name: Run Cleanup Script
+        if: steps.check-deployment.outcome == 'success'
+        run: |
+          # Since we're using Azure for benchmarks
+          ./modules/bench/cloud/clear-bench.sh azure
```
danmason, line 95: If the rest of this workflow fails - are we left with a hanging helm deployment that doesn't get cleared up?

danmason (changes requested):
Have left a few comments :)

.github/workflows/nightly-benchmark-tpch.yml
danmason: Would it? Is that not the default for workflow_call only? 

(For nightly, perhaps something like 0.5/1 for now?)

.github/workflows/nightly-benchmark-cleanup.yml
danmason: Gotcha - so if I wanted to read this file on the benchmark job in k8s itself, I would be able to just grab it with my credentials on that bucket?

.github/workflows/nightly-benchmark-cleanup.yml
danmason: I don't think github stores them forever, admittedly - I think for the sake of "having the last few runs logs available to us to read" it should be fine, I'd probably upload any other artifacts to github also.

.github/workflows/nightly-benchmark-tpch.yml
tggreene: ☝️ You're absolutely right it wouldn't

.github/workflows/nightly-benchmark-cleanup.yml
tggreene: It's set in the benchmark-x.yaml

          env:
            - name: XTDB_NODE_ID
              valueFrom:
                fieldRef:
                  fieldPath: metadata.name

danmason: Ah gotcha, cool!

.github/workflows/nightly-benchmark-cleanup.yml
tggreene: Yup, but I think I'll switch to github storage anyway

.github/workflows/nightly-benchmark-cleanup.yml
tggreene: I'll fix this!!

.github/workflows/nightly-benchmark-cleanup.yml
danmason: Fair! I wonder if it's still worthwhile storing things on Azure for the sake of pulling them/checking regressions, depends how hard pulling previous days logs would be on github.

PR #4735, opened 2025-09-02
Move BuildSide to Disk above IN_MEMORY_THRESHOLD

FiV0:
This moves the build side to disk in batches of ~100k rows and reads everything back into one relation on the way out.

### DIFF
--- a/core/src/main/kotlin/xtdb/operator/join/BuildSide.kt
+++ b/core/src/main/kotlin/xtdb/operator/join/BuildSide.kt
@@ -1,16 +1,22 @@
 package xtdb.operator.join
 
 import org.apache.arrow.memory.BufferAllocator
+import org.apache.arrow.vector.ipc.ArrowFileReader
+import org.apache.arrow.vector.types.Types
+import org.apache.arrow.vector.types.pojo.Field
 import org.apache.arrow.vector.types.pojo.Schema
 import org.roaringbitmap.RoaringBitmap
-import xtdb.arrow.IntVector
-import xtdb.arrow.RelationReader
-import xtdb.arrow.VectorReader
+import xtdb.arrow.*
 import xtdb.expression.map.IndexHasher
-import xtdb.trie.MutableMemoryHashTrie
+import xtdb.util.openReadableChannel
+import xtdb.util.openWritableChannel
 import xtdb.vector.OldRelationWriter
+import java.nio.file.Files
+import java.nio.file.Path
 import java.util.function.IntConsumer
 import java.util.function.IntUnaryOperator
+import kotlin.io.path.createDirectories
+import kotlin.io.path.deleteIfExists
 
 internal const val NULL_ROW_IDX = 0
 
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/xtdb/operator/join/BuildSide.kt
```diff
@@ -1,35 +1,83 @@
...
+    private val diskSchema: Schema =  Schema(schema.fields.plus(HASH_COLUMN_FIELD))
+    private val tmpFile: Path =
+        Files.createTempDirectory("xtdb-build-side-tmp").let { rootPath ->
+            Files.createTempFile(rootPath.createDirectories(), "build-side", ".arrow") }
+
+    private fun internalRelToExternalRel(internalRel: OldRelationWriter) =
+        OldRelationWriter(al, internalRel.vectors.filter { it.name != HASH_COLUMN_NAME }).also { it.rowCount = internalRel.rowCount }
+
+    private fun internalRelToHashColumn(internalRel: OldRelationWriter): VectorWriter =
+        internalRel.vectorForOrNull(HASH_COLUMN_NAME)!!
```
jarohen: nit: `vectorFor` if you're certain it's there - or, `internalRel[HASH_COLUMN_NAME]`

core/src/main/kotlin/xtdb/operator/join/BuildSide.kt
```diff
@@ -1,35 +1,83 @@
...
+
+    // The reference counting should happen on the internalRel
+    private val internalRel = OldRelationWriter(al, diskSchema)
+    private val externalRel = internalRelToExternalRel(internalRel)
+    private val hashColumn: VectorWriter = internalRelToHashColumn(internalRel)
+    private val unloadChannel = tmpFile.openWritableChannel()
+    private val unloader = ArrowUnloader.open(unloadChannel, diskSchema, ArrowUnloader.Mode.FILE)
```
jarohen, line 48: ArrowUnloader is lower-level than I thought you'd go with here - Relation.startUnload means you don't have to worry about record-batches. that said, if you think this one is easier to use, that's probably a signal in itself 🙂

src/test/clojure/xtdb/disk_join_test.clj
```diff
@@ -0,0 +1,29 @@
...
+  (:require [clojure.test :as t :refer [deftest]]
+            [xtdb.api :as xt]
+            [xtdb.test-util :as tu])
+  (:import (xtdb.operator.join BuildSideConfig)))
+
+(defn with-disk-threshold
+  ([threshold] (partial with-disk-threshold threshold))
+  ([threshold f]
+   (let [old-threshold BuildSideConfig/IN_MEMORY_THRESHOLD]
+     (set! BuildSideConfig/IN_MEMORY_THRESHOLD threshold)
```
jarohen: you probably want `(try (f) (finally (set! ... old-threshold)))`, just in case `(f)` throws. 

if you're looking to replicate dynamic bindings, I'd probably go with a ThreadLocal? otherwise this may affect tests that run in parallel (which I don't think we have, but Good Practice™ not to prevent it)

core/src/test/kotlin/xtdb/operator/join/BuildSideTest.kt
```diff
@@ -0,0 +1,174 @@
...
+    private fun BuildSide.getMatches(hash: Int): List<Int> {
+        val matches = mutableListOf<Int>()
+        forEachMatch(hash) { matches.add(it) }
+        return matches
+    }
+
+    @Test
+    fun testBuildSideWithDiskSpill(al: BufferAllocator) {
+        val schema = Schema(listOf(
+            Field.nullable ("id", Types.MinorType.INT.type),
```
jarohen: we have a new(-ish) DSL for this - try `Fields.I32.nullable.toArrowField("id")` etc?

core/src/main/kotlin/xtdb/operator/join/BuildSide.kt
```diff
@@ -1,35 +1,83 @@
...
     private val withNilRow: Boolean
 ) : AutoCloseable {
-    private val relWriter = OldRelationWriter(al, schema)
 
-    private val hashColumn: IntVector = IntVector(al, "xt/join-hash", false)
+    internal var toDisk: Boolean = false
+    private val diskSchema: Schema =  Schema(schema.fields.plus(HASH_COLUMN_FIELD))
+    private val tmpFile: Path =
+        Files.createTempDirectory("xtdb-build-side-tmp").let { rootPath ->
+            Files.createTempFile(rootPath.createDirectories(), "build-side", ".arrow") }
```
jarohen, line 36: do you need to create both the temp-dir and temp-file here?

core/src/test/kotlin/xtdb/operator/join/BuildSideTest.kt
```diff
@@ -0,0 +1,174 @@
...
+
+            buildSide.build()
+
+            assertEquals(buildSide.toDisk, true)
+
+            val builtRelation = buildSide.builtRel
+
+            assertEquals(10  , builtRelation.rowCount)
+
+            val idVector = builtRelation.vectorFor("id")
```
jarohen: can do `builtRelation.toMaps()` here and save yourself pulling each vector out individually?

core/src/test/kotlin/xtdb/operator/join/BuildSideTest.kt
```diff
@@ -0,0 +1,174 @@
...
+            val expectedValues = listOf(null, 100, 200, 300, 100, 200, 300, 100, 200, 300)
+            assertEquals(expectedValues, valueVector.toList())
+
+            val hasher = IndexHasher.fromCols(listOf(idVector))
+            val val2Hash = hasher.hashCode(2) // hash for id=2
+            val expectedMatches = listOf(2, 5, 8)
+            assertEquals(expectedMatches, buildSide.getMatches(val2Hash).sorted())
+        }
+
+        relation.close()
```
jarohen: `use`, if you need to close a relation

core/src/main/kotlin/xtdb/operator/join/BuildSide.kt
```diff
@@ -1,35 +1,83 @@
...
+
+    private fun internalRelToExternalRel(internalRel: OldRelationWriter) =
```
jarohen, line 38: let's run through the distinction between internal and external here - it seems that the shared mutable state is causing you some complexity here. let's see if we can combine/split these at the boundaries instead, save you maintaining both throughout?

core/src/main/kotlin/xtdb/operator/join/BuildSide.kt
```diff
@@ -1,35 +1,83 @@
...
 import java.util.function.IntConsumer
 import java.util.function.IntUnaryOperator
+import kotlin.io.path.createDirectories
+import kotlin.io.path.deleteIfExists
 
 internal const val NULL_ROW_IDX = 0
 
+object BuildSideConfig {
+    @JvmField
+   var IN_MEMORY_THRESHOLD = 100_000
```
jarohen: let's pop this as a defaulted parameter to `BuildSide`?

jarohen (changes requested):
thanks @FiV0 🙏 let's have a pair up on this, would like to better understand the internal/external rel split?

core/src/main/kotlin/xtdb/operator/join/BuildSide.kt
FiV0: I am getting missing `ARROW_MAGIC` when trying to read the batches back. Does this need get set explicitly somehow with `Relation.startUnload`.

core/src/main/kotlin/xtdb/operator/join/BuildSide.kt
FiV0: I wasn't sure where the root folder should go. Should I add a rootPath similar to how the `DiskCache` does it?

PR #4742, opened 2025-09-03
Helm template fixes

tggreene:
A couple of fixes/improvements for the bench helm templates:
 - Fix auctionmark flags, we were being tripped up by the presence of --no-load even though we were passing false: 10:48:56.558 [main] INFO  xtdb.bench.auctionmark - {:scale-factor 0.1, :no-load? true, :only-load? true, :duration #xt/duration "PT10M", :seed 0}
   - Now we just pass the flag if noLoad or loadOnly is true, throwing if we see both as true
 - Use the xtdbConfig.kafkaBootstrapServers to configure the wait-for-kafka containers, just removing a potential foot gun that I've fired myself


danmason (approved):
LGTM :+1:

PR #4749, opened 2025-09-04
Add further property tests for block boundaries, document history and PATCH.

### LINKED ISSUE #4730: Add property tests for Block boundary behaviors
We have a number of tests already merged as part of the initial work done within #4719, though there are still a few other tests we would like to add:

- [x] Split these out into their own testing namespace, out of node_test.
- [x] From @jarohen - "some between-block tests where all of the documents on either side of the block boundary have the same types - i.e. all ints on one side, all strings on the other, or not present on the other, etc - these are the ones that we've seen issues with recently."

danmason:
Resolves #4730 

- Adds a property test to ensure that writing a block of a single field type followed by a block of another type remains intact and queryable across block boundaries.
- Adds a property test to verify that inserting a block of single-typed values followed by a block of nil fields remains intact and queryable across block boundaries.
- Renames existing block boundary tests.

Also adds a few other tests against record insertion:
- Adds a property test writing to the same document many times across multiple transcations - verifying history of document and final state. 
- Adds a property test to verify that multiple writes to the same document within a single transaction (each document with distinct valid times) - verifying history of document and final state. 
- Adds a property test against patching the same doc multiple times, verifying history is present and resulting doc is as expected.

Have annotated all currently failing property tests with associated issues.

### DIFF
--- a/src/main/clojure/xtdb/test_generators.clj
+++ b/src/main/clojure/xtdb/test_generators.clj
@@ -136,16 +136,14 @@
      (-> (zipmap field-keys field-values)
          (assoc :xt/id id)))))
 
-(defn typed-vector-vs-gen
-  [element-gen min-length max-length]
+(defn typed-vector-vs-gen [element-gen min-length max-length]
   (gen/let [vector-name non-blank-string-gen
             vs (gen/vector element-gen min-length max-length)]
     {:vec-name vector-name
      :vs vs
      :type element-gen}))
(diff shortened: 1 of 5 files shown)


danmason:
Have raised new issues from the PATCH test: https://github.com/xtdb/xtdb/issues/4751&&  https://github.com/xtdb/xtdb/issues/4749

Issue #4758, opened 2025-09-09
'market data' use case with handful of infrequently updated rows causes query degradation


jarohen:
See also: #4097, #4111 

'market data' use case from the [bitemp taxonomy blog](https://xtdb.com/blog/building-a-bitemp-index-1-taxonomy) - although we'd recommend giving this type of data a TTL now as its valid-to, which transforms it into 'readings' data.

* The L1 'current' ('l01-rc') file ends up with a large VT range (let's say ~P3M) because there's one or two entities that haven't been updated in that time.
* Counting that table, for example, then needs to pull in the l01-rc. However, because rows in the recent history could all temporally affect rows in the l01-rc (mainly their valid-from/valid-to) we also have to scan all the pages that overlap in valid time.
* Because there's such a large valid time, this additionally pulls in the last P3M of history too.
* With this history, at the client, we're seeing ~30s for a query that, with this history manually removed, takes of the order of 10-15ms.

jarohen:
@Akeboshiwind I'm'a optimistically close this one, let's re-open if need be

PR #4759, opened 2025-09-09
Adding actual max recency to compacted files

jarohen:
(accidentally auto-closed #4746)

first part of #4758

We've been seeing some performance issues related to a [market-data-like dataset](https://xtdb.com/blog/building-a-bitemp-index-1-taxonomy) - in this case, entities where the user has repeatedly re-inserted the same/similar data with unbounded valid-to, to the extent that they now have quite a lot of history for each entity.

One improvement that we can make in this area to help them out is to reduce the number of files that the scan operator was including in the scan. When we partition by recency (weekly buckets) we then use the following Monday as the partition key - then, though, when deciding what files to read for an as-of-now query, we assume that there are events in that file up to that Monday, and hence include it.

This PR gets the compactor to add the actual maximum recency of the file to the trie-metadata, and then uses this to filter out files.

We store the recency of each event in the output data files, because when we come to re-compact that file, the 'current' file that originally superseded the event is not included in the same compaction job.

e.g. let's say an event is superseded in L1 - the historical event goes to L1H; the superseding event to L1C. when it comes time to compact L1H -> L2H, the L1C file isn't included, and so the L1H event would look 'current' again.

This will need a compaction reset to take advantage of the new metadata, but will cope fine without. I'll separately update the compaction reset tool to be able to reset a single table. (https://github.com/xtdb/xtdb/issues/4745)

related: https://github.com/xtdb/xtdb/pull/4111


jarohen:
not doing this any more, tactical fix applied and longer term fix requires more thought

PR #4775, opened 2025-09-12
pgwire thread interrupt handling for cmd-cancel messages

### LINKED ISSUE #4757: Investigate memory leaking via pgwire cmd-cancel'd queries
For context, it has been observed that Netty memory is breaching memory limits and causing `OOMKilled` events (in Kubernetes).

A possible culprit which warrants investigation is the lack of termination handling for the pgwire execution thread - as noted in this long-standing TODO: https://github.com/xtdb/xtdb/blob/25d0f2e01f50f5e7f959a510e3ea120f44550ca5/core/src/main/clojure/xtdb/pgwire.clj#L178

Original commit https://github.com/xtdb/xtdb/commit/806cf47e33a026279c1395b6d2c583642fd72ac0#diff-fb16a6a947e20d7e43ed9084cdad0672f4ebc2bcca5033fac0dce2826c3314e5R1177

refset:
Testing all the behaviour here is a little complicated (could possibly do with more tests), but via psql it now looks like:

(actual long-running query)
```sql
xtdb=> FROM generate_series(1, 100) x, generate_series(1, 100) y, generate_series(1, 220) z;
^CCancel request sent
ERROR:  query cancelled during execution
DETAIL:  
```

(long enough to be interesting, but the actual query executes quite fast and the  human-speed cancelling kicks in during pgwire batch processing so we have to chunk the processing)
```sql
xtdb=> FROM generate_series(1, 100) x, generate_series(1, 100) y, generate_series(1, 22) z;
^CCancel request sent
ERROR:  query cancelled during execution
DETAIL:  
```

(also works fine in a very simple case)
```
xtdb=> select pg_sleep('PT10S');
^CCancel request sent
ERROR:  query cancelled during execution
DETAIL:  
```

(cancel flag state on the connection gets reset properly)
```sql
xtdb=> select 1;
 _column_1 
-----------
         1
(1 row)

xtdb=> 
```

Yet-to-be-proven hypothesis (WIP): Reduced consumption of Netty memory can be observed vs. main with `FROM generate_series(1, 1000000) x (x) LEFT JOIN generate_series(1, 1000000) y (y) ON x = y;`

Despite these changes, clients still won't be able to cancel queries containing long-running, non-batched operators, e.g. `from generate_series(TIMESTAMP '2020-01-01Z', TIMESTAMP '2025-01-01Z', INTERVAL 'PT1S') x;`

### DIFF
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -156,7 +156,7 @@
            {::severity :fatal, ::error-code "3D000"}))
 
 (defn- err-invalid-auth-spec [msg] (ex-info msg {::severity :error, ::error-code "28000"}))
-(defn- err-query-cancelled [msg] (ex-info msg {::severity :error, :error-code "57014"}))
+(defn- err-query-cancelled [msg] (ex-info msg {::severity :error, ::error-code "57014"}))
 
 (defn- notice [msg]
   {:severity "NOTICE"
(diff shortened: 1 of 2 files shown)


refset:
Closing for now.

PR #4829, opened 2025-09-29
Benchmark tracking and alerting

### LINKED ISSUE #4709: Track and Report Performance Regressions on Cloud Benchmark Runs
We want to automatically collect results of various benchmark steps from the nightly benchmark runs, and then compare them against historical baselines, reporting any regressions.  

## TODO
- [x] Store/export nightly benchmark results in a structured format.  
- [x] Persist results somewhere accessible across runs.
- [ ] Fetch prior run results to define a historical baseline to compare against.
- [ ] Compute deltas between the latest run and the baseline.  
- [ ] Flag regressions based on defined thresholds (e.g. % slowdowns).  
- [ ] Integrate regression reporting with Slack notifications in the benchmark process.  
- [ ] Provide a simple report or dashboard to visualize historical nightly trends.

tggreene:
This PR adds:
 - A request to Azure monitor logs in the benchmark cleanup github action
 - Azure resources in terraform for storing logs in dev/cloud/azure/benchmark-metrics
 - Azure resources for alerting based on those logs

I will squash commits when I merge.


danmason:
@tggreene Any reason this lives under dev/cloud specifically, rather than under the bench directory?

tggreene:
I think it was your idea https://juxt.slack.com/archives/C09BMP7SD40/p1758116354583909. I'm happy either way.

danmason:
> I think it was your idea https://juxt.slack.com/archives/C09BMP7SD40/p1758116354583909. I'm happy either way.

Ah gotcha - yeah, in hindsight I'd probably move it somewhere under `bench` given that it's all benchmark related (just not somewhere it gets automatically spun up/tore down by CI).

danmason (approved):
As discussed - while we may be able to simplify a lot of this using log aggregation over k8s/docker, I think this is a good step in the right direction so it's worth merging! Only thing I would like particularly is for us to get pinged on slack when an alert fires :)

tggreene:
Okay I think that's good! Thanks Dan!

PR #4835, opened 2025-09-30
[bench] Raise errors from benchmark threads

tggreene:
This PR raises exceptions from pool and concurrent threads to the main thread and rethrows exceptions in the tpch query function. This means overall that if an exception is throw in the benchmark the process should exit non-zero and come back as failed in nightly benchmark runs rather than getting swallowed.

### DIFF
--- a/modules/bench/src/main/clojure/xtdb/bench.clj
+++ b/modules/bench/src/main/clojure/xtdb/bench.clj
@@ -19,7 +19,7 @@
            (java.time Clock Duration InstantSource)
            java.time.Duration
            (java.util Comparator Random)
-           (java.util.concurrent Executors TimeUnit)
+           (java.util.concurrent ExecutionException Executors TimeUnit)
            (java.util.concurrent.atomic AtomicLong)
            (oshi SystemInfo)))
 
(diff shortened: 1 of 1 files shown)


modules/bench/src/main/clojure/xtdb/bench.clj
```diff
@@ -245,13 +257,25 @@
...
                             (when-not (.awaitTermination executor (.toMillis join-wait) TimeUnit/MILLISECONDS)
-                              (throw (ex-info "Task threads did not stop within join-wait" {:task task, :executor executor}))))))
+                              (.shutdownNow executor)
+                              (when-not (.awaitTermination executor (.toMillis join-wait) TimeUnit/MILLISECONDS)
+                                (throw (ex-info "Task threads did not stop within join-wait" {:task task, :executor executor}))))
+                            ;; Propagate any worker exceptions to the main thread
+                            (doseq [f futures]
+                              (try
+                                (.get f) ; will throw ExecutionException if the Runnable raised
+                                (catch java.util.concurrent.ExecutionException e
```
jarohen: nit: import ExecutionException

modules/bench/src/main/clojure/xtdb/bench.clj
```diff
@@ -245,13 +257,25 @@
...
+                                  (let [cause (.getCause e)]
+                                    (log/error cause "Benchmark worker failed in :concurrently" {:task task})
+                                    (throw (ex-info "Benchmark worker failed" {:task task} cause))))
+                                (catch java.lang.InterruptedException e
```
jarohen: nit: don't need to fully-qualified InterruptedException - java.lang imported by default

modules/bench/src/main/clojure/xtdb/bench.clj
```diff
@@ -224,13 +224,25 @@
                                                                        thread-loop)))))]
 
                 (fn run-pool [worker]
-                  (run! #(start-thread worker %) (range thread-count))
+                  ;; Launch threads, collect Futures so we can collect any exceptions after termination
+                  (let [futures (mapv #(start-thread worker %) (range thread-count))]
```
jarohen, line 228: nit: indent `let` body

jarohen (approved):
minor nits but tbh, as discussed, we probably want to implement this with some safer structured concurrency these days

feel free to merge without further review 👌

PR #4861, opened 2025-10-07
Query cancellation through interruption

egg-juxt:
Several cancellation mechanisms have been considered:
- Using FutureTask in-thread:
  - FutureTask with blocking `get`
  - **FutureTask with non-blocking checks: CHOSEN - easy, reliable, most appropriate for in-thread usage, preserves current threading architecture with one dedicated thread per connection**
- explicit thread interruption
- virtual thread:
  - spawn through executor, explicit interruption
  - spawn through new thread, explicit interruption
  - spawn through executor, task cancellation
- Kotlin's structured concurrency: discarded - doesn't use thread interruption for cancellation
- Java's structured concurrency: discarded - preview API in Java 21. XTDB currently deviates from standard InterruptedException (sometimes throws Interrupted anomaly)

For code (sometimes untested, or incorrect), see branches
https://github.com/egg-juxt/xtdb/tree/backup/query-cancellation-through-interruption-research
https://github.com/egg-juxt/xtdb/tree/query-cancellation-through-interruption-vthread-draft

### DIFF
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -33,7 +33,7 @@
            [java.nio.file Path]
            [java.security KeyStore]
            [java.time Clock Duration ZoneId]
-           [java.util.concurrent ExecutorService Executors TimeUnit]
+           [java.util.concurrent ExecutorService Executors Future$State FutureTask TimeUnit]
            [javax.net.ssl KeyManagerFactory SSLContext]
            (org.antlr.v4.runtime ParserRuleContext)
            (org.apache.arrow.memory BufferAllocator)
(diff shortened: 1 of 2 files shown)


core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1260,47 +1258,56 @@
...
+    (swap! conn-state assoc :cancel-query! #(when (.cancel task true)
+                                              (log/debug "Query cancelled")))
+    (try
+      (.run task) ; in-thread execution
+      (finally
+        (swap! conn-state dissoc :cancel-query!)))
+    (condp = (.state task)
+      Future$State/SUCCESS (do)
+      Future$State/CANCELLED (do
+                               (Thread/interrupted) ; FutureTask may leak interruption status
```
jarohen, line 1272: yeah...?

jarohen (approved):
thanks @egg-juxt - yep, seems reasonable, good analysis 🙏

core/src/main/clojure/xtdb/pgwire.clj
egg-juxt: Yes, didn't know that, and should be infrequent, but [it can happen](https://bugs.openjdk.org/browse/JDK-8269467).

PR #4892, opened 2025-10-21
Move project wide test utils to gradle testFixtures source set, using java-test-fixtures plugin

mbutlerw:
Removes the need for src/main, making it clear that only test code should live under src/testFixtures as they will not be included in the main jar/artifact.


mbutlerw:
Contents merged manually in https://github.com/xtdb/xtdb/commit/bfa45178f224e15bb5584b571d83df9e32cc3cc4

PR #4901, opened 2025-10-22
[bench] yakbench

### LINKED ISSUE #4893: Add Yakbench Benchmark
- [x] Add generators for yakbench
- [x] Look into :get-items, :rss-items, performance should be comparable
- [x] Make user-items, user-subs follow a poisson distribution (particularly interested in the tail)
- [x] Compare to existing benchmark results

tggreene:
This PR implements the yakbench benchmark based on Jacob O'Bryants dataset for his product yakread.

It comprises generator functions for generating data as per the original dataset as closely as possible, and queries from Jacob.

Initially the benchmark generates data in memory forming joins between key records like user items, user subs and feed items. The dataset is quite large so in order to fit into memory the benchmark generates the data in chunks (0.1 scale factor chunks to be precise). It's an assumption that this doesn't have an impact on query performance.

In order to generate data suitably similar to the original dataset we distributed counts for key relationships like user items with a lognormal poisson distribution, this allows us to model the heavy tailed data as closely as possible.

You can see some stats/graphs on how closely the two datasets match [here](https://github.com/xtdb/xtdb/issues/4893#issuecomment-3432960182).

Closes #4893

### DIFF
--- a/modules/bench/src/main/clojure/xtdb/bench/yakbench.clj
+++ b/modules/bench/src/main/clojure/xtdb/bench/yakbench.clj
@@ -23,52 +23,8 @@
 (def recent-ad-inst (.toInstant #inst "2025-01-01T00:00:00Z"))
 
 (defn benchmarks-queries
-  [{{:keys [user-id user-email]} :picked-user
-    :keys [active-user-ids biggest-feed]}]
-  (cond-> [{:id :get-user-by-email
-            :f #(xt/q % ["select * from users where user$email = ?" user-email])}
-           {:id :get-user-by-id
-            :f #(xt/q % ["select * from users where _id = ?" user-id])}
-           {:id :get-user-id-by-email
-            :f #(xt/q % ["select _id from users where user$email = ?" user-email])}
-           {:id :get-user-email-by-id
-            :f #(xt/q % ["select user$email from users where _id = ?" user-id])}
-           {:id :get-feeds
-            :f #(xt/q % [(str "select count(sub$feed$feed) from subs "
-                              "where sub$user = ? and sub$feed$feed is not null")
-                         user-id])}
-           {:id :get-items
-            :f #(xt/q % [(str "select count(i._id) "
-                              "from subs s "
-                              "join items i on i.item$feed$feed = s.sub$feed$feed "
-                              "where s.sub$user = ? "
-                              "and s.sub$feed$feed is not null")
-                         user-id])}
-           {:id :read-urls
-            :f #(xt/q % [(str "select distinct i.item$author_url "
-                              "from user_items ui "
-                              "join items i on i._id = ui.user_item$item "
-                              "where ui.user_item$user = ? "
-                              "and coalesce(ui.user_item$viewed_at, ui.user_item$favorited_at, ui.user_item$skipped_at) is not null "
-                              "and i.item$author_url is not null")
-                         user-id])}
-           {:id :email-items
-            :f #(xt/q % [(str "select s._id as sub_id, i._id as item_id "
-                              "from subs s "
-                              "join items i on i.item$email$sub = s._id "
-                              "where s.sub$user = ?")
-                         user-id])}
-           {:id :rss-items
-            :f #(xt/q % [(str "select s._id as sub_id, i._id as item_id "
-                              "from subs s "
-                              "join items i on i.item$feed$feed = s.sub$feed$feed "
-                              "where s.sub$user = ? and s.sub$feed$feed is not null")
-                         user-id])}
-           {:id :user-items
-            :f #(xt/q % [(str "select user_item$item, user_item$viewed_at, user_item$favorited_at, user_item$skipped_at "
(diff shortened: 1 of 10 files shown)


modules/datasets/src/main/clojure/xtdb/datasets/yakbench.clj
```diff
@@ -0,0 +1,424 @@
...
+        rest-groups (rest (str/split rest-str #"-"))]
+    (UUID/fromString (str/join "-" (cons prefix rest-groups)))))
+
+(defn gen-users
+  "For users we add an extra :user/k field that is a poisson-distributed long.
+  This is used to determine the number of items and subs a user has."
+  [random n]
+  (vec
+   (for [_ (range n)
+         :let [id (next-uuid random)
```
jarohen: don't let-bind when you're only going to use it once - at its usage site it's given a name, so you're not gaining anything by giving it another one here

modules/datasets/src/main/clojure/xtdb/datasets/yakbench.clj
```diff
@@ -0,0 +1,424 @@
...
+
+(defn gen-subs
+  [random users feeds]
+  (let [feed-ids (mapv :xt/id feeds)
+        n-feeds (count feed-ids)]
+    (->> users
+         (mapcat (fn [{user :xt/id
+                       user-k :user/k}]
+                   (let [k (min (long (* user-k 0.25)) n-feeds)
+                         chosen (->> (repeatedly #(next-int random n-feeds))
```
jarohen: similarly here - `(for [feed-id (->> (repeatedly ...) ...)] ...)`

modules/datasets/src/main/clojure/xtdb/datasets/yakbench.clj
```diff
@@ -0,0 +1,424 @@
...
+     {:xt/id (next-uuid random)
+      :skip/user (:xt/id (uniform-nth random users))
+      :skip/created-at (next-instant random)})))
+
+(defn generate-data
+  "Generate synthetic data according to a scale factor and return a map of
+   table keyword -> vector of docs"
+  [random scale]
+  (let [{:keys [users feeds items ads ad-credits ad-clicks digests bulk-sends skips]} baseline-counts
+        n-users (round-down (* users scale))
```
jarohen: inlining again here, little value in the names and it means the code's ~3x as many lines ... which may sound like it's golfing but it genuinely does make a difference when you're reading it - helps one's eyes focus on what's important

modules/datasets/src/main/clojure/xtdb/datasets/yakbench.clj
```diff
@@ -0,0 +1,424 @@
...
+        (let [p (* p (.nextDouble random))]
+          (if (> p L)
+            (recur (inc k) p)
+            k))))
+    :else
+    (let [k (long (Math/round (+ lambda (* (Math/sqrt lambda)
+                                           (.nextGaussian random)))))]
+      (max 0 k))))
+
+(defn lognormal-lambda
```
jarohen: as discussed, let's see if we can delegate to kixi.stats or similar here

modules/datasets/src/main/clojure/xtdb/datasets/yakbench.clj
```diff
@@ -0,0 +1,424 @@
...
+
+(def feed-items-stats
+  {:mean 736.407677681279
+   :median 25
+   :p90 589
+   :p99 12004
+   :min 1
+   :max 363444})
+
+(defn round-down [x]
```
jarohen: can we get these into a bench util namespace? something like `bench.stats`?

modules/bench/src/main/clojure/xtdb/bench/yakbench.clj
```diff
@@ -0,0 +1,531 @@
...
+
+    (println "User subs stats")
+    (pp/print-table [(user-subs-stats conn)])
+    (println)
+
+    (println "Feed items stats")
+    (pp/print-table [(feed-items-stats conn)])
+    (println)))
+
+(defn load-data!
```
jarohen: if this is required to make the dataset loadable at arbitrary scale, let's get this into the dataset ns?

modules/bench/src/main/clojure/xtdb/bench/yakbench.clj
```diff
@@ -0,0 +1,531 @@
...
+                       (f conn))
+          [_ pstats] (tufte/profiled
+                      {}
+                      (doseq [{:keys [id f]} suite
+                              _ (range iterations)]
+                        (p id (f conn))))]
+      {:profile (profile-data @pstats)
+       :formatted (tufte/format-pstats pstats)})))
+
+(defn distribution-stats
```
jarohen: -> bench.stats?

jarohen (approved):
thanks @tggreene - happy to merge this without further review once you've taken a look through the comments :)

PR #4912, opened 2025-10-27
Add MockDriver for Compactor and first simulation test

FiV0:
This tries to improve our testing approach on the Deterministic Simulation Testing side. We add add a MockDriver for the Driver interface of the compactor. The driver of the compactor deals with all the actual compaction job execution and the signalling to the main compactor thread when jobs are done. The MockDriver tries to make the ordering of these jobs and their respective signalling deterministic. We also add a first simple L00 -> L01 compaction test.  The main part this machinery is currently testing is the TrieCatalog.

Things still to do in later iterations:
- Deterministic splitting of files if a compaction job has multiple outputs (historical vs current).
- Deterministic simulation of multiple node (really just multiple trie catalogs).
- Deterministic simulation of the Garbage Collector

### DIFF
--- a/core/src/main/clojure/xtdb/trie_catalog.clj
+++ b/core/src/main/clojure/xtdb/trie_catalog.clj
@@ -344,6 +344,9 @@
               (fn [_table tries]
                 (remove-garbage tries garbage-trie-keys))))
 
+  (listAllTrieKeys [this table]
+    (mapv :trie-key (all-tries (trie-state this table))))
+
   PTrieCatalog
   (trie-state [_ table] (.get !table-cats table))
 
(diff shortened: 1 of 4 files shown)


core/src/main/clojure/xtdb/trie_catalog.clj
```diff
@@ -344,6 +344,9 @@
               (fn [_table tries]
                 (remove-garbage tries garbage-trie-keys))))
 
+  (listAllTrieKeys [this table]
+    (mapv :trie-key (all-tries (trie-state this table))))
+
```
FiV0, line 349: I wasn't sure if we want to keep this bit, but as I am not quite sure yet which properties we are actually looking for and what we will be asserting against I kept it for now.

danmason (approved):
Seems like it'd be good to get this onto main and running with the property tests :)

PR #4922, opened 2025-10-31
add `export-snapshot` CLI tool to export a snapshot of an XT object-store

jarohen:
Quick CLI tool to export a snapshot of an XTDB database out to a sibling directory in its own object store - this directory can then be moved elsewhere and new nodes started.

Similarly to compactor-reset, set the CMD to `export-snapshot -f node-config.yaml <db-name>`, then you should get an `exports/b00_20251031T170000Z` directory in your object store (exact name given in the logs).

Then, start a new log topic on your destination node, and increase the log epoch:

```yaml
log: !Kafka
  ...
  topic: new-topic
  epoch: 1
storage: !Remote
  objectStore: !S3
    bucket: my-bucket
    prefix: exports/...
```

### DIFF
--- a/core/src/main/clojure/xtdb/export.clj
+++ b/core/src/main/clojure/xtdb/export.clj
@@ -1,0 +1,111 @@
+(ns xtdb.export
+  (:require [clojure.tools.logging :as log]
+            [xtdb.db-catalog :as db]
+            [xtdb.error :as err]
+            [xtdb.node :as xtn]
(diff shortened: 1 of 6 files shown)


refset:
Caveats: 
- all copyObjects must finish before GC deletes some of files 
- the snapshot will not contain anything from the "live index", i.e. transactions from the log that do not yet appear in the object store, to mitigate this you can use the finish chunk API

PR #4935, opened 2025-11-03
Flattening multiple indirections

jarohen:
Again, guided by the profile, we're spending a lot of time resolving indirections on a row-by-row basis.

This PR:
- removes the initial DUV indirection that comes from pulling the 'put' vector - as soon as we select on that leg, we know that it's only put values, and so we can jump out of the DUV indirection
- ensures that when we call 'select' on an indirect vector, we eagerly resolve the two indirections s.t. we've only got one.


jarohen:
merged 079be31ab0e70c6d0026108cb514e0b10a2d49ba

Issue #4926, opened 2025-11-03
Reduce peak memory usage of table metadata


jarohen:
In addition to #4920, we also needn't open up all the table metadata at once. The merge-planner traverses the IID key space, and we only need the metadata for a file open when we're testing that specific part of the IID space - at the moment, MergePlanner.plan eagerly opens metadata for all of the segments for the duration of the method.

PR #4937, opened 2025-11-04
not opening metadata for segments that don't match the path pred (#4926)

jarohen:
related to #4926 - simple fix to not open metadata for segments that cannot possibly match the path predicate.


jarohen:
not actually merged, I messed up

Issue #4946, opened 2025-11-05
Keep latest live block-idx per `level` `recency` `part`.


FiV0:
We currently always assume that we can figure out if a trie is stale from the existing tries in the trie-catalog. In theory this true, but it requires a complex calculation of looking at the the level of the incoming trie and the levels above it. We are currently only checking at the current level. The following test exhibits the problem
```clj
(t/deftest keep-latest-live-block-idx-around
  (let [cat (apply-msgs ["l00-rc-b00" 10] ["l00-rc-b01" 10] ["l00-rc-b02" 10] ["l00-rc-b03" 10]
                        ["l01-rc-b00" 20] ["l01-rc-b01" 20] ["l01-rc-b02" 20] ["l01-rc-b03" 20]
                        ["l02-rc-p0-b03" 20] ["l02-rc-p1-b03" 20] ["l02-rc-p2-b03" 20] ["l02-rc-p3-b03" 20])]

    (t/is (true? (cat/stale-msg? cat (trie/parse-trie-key "l01-rc-b00"))))

    (let [new-cat (cat/remove-garbage cat (map :trie-key (cat/garbage-tries cat as-of-time)))]

      (t/is (true? (cat/stale-msg? new-cat (trie/parse-trie-key "l01-rc-b00")))))))
```

The solution is to keep a block index per `[level recency part]` key of the trie catalog data structure. This will also simplify some calculation in the trie-catalog logic.

Issue #4947, opened 2025-11-05
Dealing with races in `supersede-partial-tries`


FiV0:
When superseding files in the trie-catalog we just supersede everything up to block-idx of the new file . This can cause issues if their are races between two part groups that become complete for the same [`level`, `recency`] where the part group with a higher block-idx completes before the part group with a lower block-idx. I am calling something like `l03-rc-p{0,1,2,3}-bX` a part group where X is fixed. 
The following test shows that this then results in the rejection of tries which are then no longer covered as we are rejecting tries up to the block index of the "newer" part group. 

```clj
(require '[xtdb.compactor-test :as ct])

(t/deftest race-in-supersede-partial-tries
  
  (t/is (= #{"l03-rc-p00-b0f"
             "l03-rc-p01-b0f"
             "l03-rc-p02-b0f"
             "l03-rc-p03-b0f"          
             
             "l03-rc-p00-b11f"
             "l03-rc-p01-b11f"
             "l03-rc-p02-b11f"
             "l03-rc-p03-b11f"}
           
           (->> (apply curr-tries
                       (concat (ct/level-seq 2 32) 
                               [["l03-rc-p00-b0f"]
                                ["l03-rc-p01-b0f"]
                                ["l03-rc-p02-b0f"]

                                ["l03-rc-p00-b11f"]
                                ["l03-rc-p01-b11f"]
                                ["l03-rc-p02-b11f"]

                                ;; "correct" order
                                ["l03-rc-p03-b0f"]
                                ["l03-rc-p03-b11f"]]))
                (filter (comp #{3} :level trie/parse-trie-key))
                set)

           (->> (apply curr-tries
                       (concat (ct/level-seq 2 32) 
                               [["l03-rc-p00-b0f"]
                                ["l03-rc-p01-b0f"]
                                ["l03-rc-p02-b0f"]

                                ["l03-rc-p00-b11f"]
                                ["l03-rc-p01-b11f"]
                                ["l03-rc-p02-b11f"]
                                ;; higer part group completes earlier
                                ["l03-rc-p03-b11f"]
                                
                                ["l03-rc-p03-b0f"]
                                ]))
                (filter (comp #{3} :level trie/parse-trie-key))
                set))))
```

FiV0:
See the comment at https://github.com/xtdb/xtdb/issues/5001#issuecomment-3527861590 for why this is a non issue.

PR #4948, opened 2025-11-05
Split live and nascent in the TrieCatalog

### LINKED ISSUE #4930: Inconsistent Trie Catalog behaviour
The following is a trace I have obtained from a DST test. It contains the tries in order as they are being applied to the trie-catalog. It likely is not the smallest example of this issue, but I wanted to start somewhere. I will update the ticket with smaller examples once I find them.

```clj
(def trace-seq
  ;; I generate L1 explicitly in the test
  [["l02-rc-p0-b03" 20]
   ["l02-rc-p0-b07" 20]
   ["l02-rc-p0-b0b" 20]
   ["l02-rc-p0-b0f" 20]


   ["l03-rc-p00-b0f" 20]
   ["l03-rc-p01-b0f" 20]
   ["l03-rc-p02-b0f" 20]
   ["l03-rc-p03-b0f" 20]])
```
When applying the last trie `"l03-rc-p03-b0f"`, the trie `l02-rc-p0-b03` becomes a job that is available again, even though it is an input to `"l03-rc-p03-b0f"`. It seems the state of the trie-catalog is then no longer in a correct state.

```clj
(t/deftest trace-test
  (binding [cat/*file-size-target* 16]
    (t/is (= #{{:trie-keys ["l01-rc-b00" "l01-rc-b01" "l01-rc-b02" "l01-rc-b03"],
                :part [3],
                :out-trie-key "l02-rc-p3-b03"}
               {:trie-keys ["l01-rc-b00" "l01-rc-b01" "l01-rc-b02" "l01-rc-b03"],
                :part [2],
                :out-trie-key "l02-rc-p2-b03"}
               {:trie-keys
                ["l02-rc-p0-b03" "l02-rc-p0-b07" "l02-rc-p0-b0b" "l02-rc-p0-b0f"],
                :part [0 3],
                :out-trie-key "l03-rc-p03-b0f"}
               {:trie-keys ["l01-rc-b00" "l01-rc-b01" "l01-rc-b02" "l01-rc-b03"],
                :part [1],
                :out-trie-key "l02-rc-p1-b03"}}
             (apply calc-jobs (concat (level-seq 1 16) (butlast trace-seq)))))

    (t/is (= #{{:trie-keys ["l01-rc-b00" "l01-rc-b01" "l01-rc-b02" "l01-rc-b03"],
                :part [3],
                :out-trie-key "l02-rc-p3-b03"}
               {:trie-keys ["l01-rc-b00" "l01-rc-b01" "l01-rc-b02" "l01-rc-b03"],
                :part [2],
                :out-trie-key "l02-rc-p2-b03"}
               {:trie-keys ["l01-rc-b00" "l01-rc-b01" "l01-rc-b02" "l01-rc-b03"],
                :part [1],
                :out-trie-key "l02-rc-p1-b03"}
               {:trie-keys ["l01-rc-b00" "l01-rc-b01" "l01-rc-b02" "l01-rc-b03"],
                :part [0],
                :out-trie-key "l02-rc-p0-b03"}}
             (apply calc-jobs (concat (level-seq 1 16) trace-seq))))))
```

FiV0:
Currently we maintain a list of `live+nascent` files for every `[level recency part]` key in the trie-catalog. The consequence is that for certain calculations (for example compactor job calculation) live and nascent are not differentiated. This means together with GCing certain files can be partially superseded and other files recompacted anew (see also #4946). By splitting `live` and `nascent` we assure that only live files are taken into account for compactor job calculation. This partially addresses the following scenario (which is also written down as a test in #4930):
- Assuming all L1 files up to block 0x0f (16) are present
- Plus the following files `"l02-rc-p0-b03", "l02-rc-p0-b07" "l02-rc-p0-b0b", "l02-rc-p0-b0f"` (just called `l02`s further down)
- Also assume that all (it's not a necessary condition that all of them are nascent just easier to explain) the `l02`'s are nascent meaning some other part of the part group for the same block is missing.
- As all input files for the compaction jobs `l03-rc-p{00,01,02,03}-b0f` are present we would previously compact those and hence these `l03`'s could potentially be live.
- Non of the part groups of the `l02` are live.
- This means for query we would for example  take `l1-rc-b03` and `l03-rc-p00-b0f` as both of them are marked live. There is a partial overlap in data with `l03-rc-p00-b0f`  $\subseteq$  `l1-rc-b03`.

With this change we only consider compaction of `l03-rc-p{00,01,02,03}-b0f` when all the part groups of the `l02`'s (`l02-rc-p{0,1,2,3}-b03` and so forth) are live. The compactor will currently also recompact the `l02`s if they  garbage collected and the relevant `l1` files are still live. In theory this change should also get rid of this scenario as one can only compact if all part groups of the `l02`'s have become live and hence all `l01` of these part groups have been superseded before we consider any `l03` job. With a fix for #4946 we will also further make sure not to recompact superseded files.

One impact this change has is that in certain scenarios compaction will be delayed. All the input files for the `l03` part group might exist (some of them nascent), but we will only consider compaction once they become live.

### DIFF
--- a/core/src/main/clojure/xtdb/compactor.clj
+++ b/core/src/main/clojure/xtdb/compactor.clj
@@ -20,17 +20,16 @@
   (getOutputTrieKey [_] out-trie-key)
   (getPartitionedByRecency [_] partitioned-by-recency?))
 
-(defn- l0->l1-compaction-job [table {{l0 :live+nascent} [0 nil []], {l1c :live+nascent} [1 nil []]} {:keys [^long file-size-target]}]
-  (when-let [live-l0 (seq (->> l0
-                               (take-while #(= :live (:state %)))))]
+(defn- l0->l1-compaction-job [table {{l0 :live} [0 nil []], {l1c :live} [1 nil []]} {:keys [^long file-size-target]}]
+  (when-let [live-l0 (seq l0)]
 
     (let [{l0-trie-key :trie-key, :keys [block-idx]} (last live-l0)
           {l1-trie-key :trie-key} (->> l1c
                                        (take-while #(< (:data-file-size %) file-size-target))
                                        first)]
       (->Job table (-> []
-                            (cond-> l1-trie-key (conj l1-trie-key))
-                            (conj l0-trie-key))
+                       (cond-> l1-trie-key (conj l1-trie-key))
+                       (conj l0-trie-key))
              nil ; part
              (Trie$Key. 1 nil nil block-idx)
              true ; partitioned-by-recency?
@@ -41,7 +40,6 @@
   [l2h-tries l1h-tries {:keys [^long file-size-target]}]
 
   (->> l1h-tries
-       (take-while #(= :live (:state %)))
        reverse
 
        (reductions (fn [[acc-tries acc-size] {:keys [^long data-file-size] :as trie}]
@@ -57,9 +55,9 @@
                  tries)))))
 
 (defn- l2h-compaction-jobs [table table-tries opts]
-  (for [[[level recency _part] {l1h-tries :live+nascent}] table-tries
+  (for [[[level recency _part] {l1h-tries :live}] table-tries
         :when (and recency (= level 1))
-        :let [input-tries (l2h-input-files (get-in table-tries [[2 recency []] :live+nascent]) l1h-tries opts)]
+        :let [input-tries (l2h-input-files (get-in table-tries [[2 recency []] :live]) l1h-tries opts)]
         :when input-tries]
     (->Job table (mapv :trie-key input-tries) (byte-array 0)
            (Trie$Key. 2 recency nil (:block-idx (last input-tries)))
@@ -67,7 +65,7 @@
            )))
 
 (defn- tiering-compaction-jobs [table table-tries {:keys [^long file-size-target]}]
-  (for [[[level recency part] {files :live+nascent}] table-tries
+  (for [[[level recency part] {files :live}] table-tries
         :when (or (and (nil? recency) (= level 1))
                   (> level 1))
         :let [live-files (-> files
@@ -78,7 +76,10 @@
         p (range cat/branch-factor)
         :let [out-level (inc level)
               out-part (conj part p)
-              {lnp1-block-idx :block-idx} (first (get-in table-tries [[out-level recency out-part] :live+nascent]))
+              {level-above-live :live level-above-nascent :nascent} (get table-tries [out-level recency out-part])
+              {lnp1-live-block-idx :block-idx} (first level-above-live)
+              {lnp1-nascent-block-idx :block-idx} (first level-above-nascent)
+              lnp1-block-idx (or lnp1-nascent-block-idx lnp1-live-block-idx)
 
               in-files (-> live-files
                            (cond->> lnp1-block-idx (drop-while #(<= (:block-idx %) lnp1-block-idx)))
--- a/core/src/main/clojure/xtdb/trie_catalog.clj
+++ b/core/src/main/clojure/xtdb/trie_catalog.clj
@@ -17,28 +17,29 @@
            (xtdb.util TemporalBounds)))
 
 ;; table-tries data structure
-;; values is a map of live+nascent and garbage tries
+;; values is a map of live, nascent and garbage tries
 ;; tries :: {:keys [level recency part block-idx state]}
 ;;   sorted by block-idx desc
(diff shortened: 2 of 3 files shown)


core/src/main/clojure/xtdb/compactor.clj
```diff
@@ -78,7 +76,10 @@
         p (range cat/branch-factor)
         :let [out-level (inc level)
               out-part (conj part p)
-              {lnp1-block-idx :block-idx} (first (get-in table-tries [[out-level recency out-part] :live+nascent]))
+              {level-above-live :live level-above-nascent :nascent} (get table-tries [out-level recency out-part])
+              {lnp1-live-block-idx :block-idx} (first level-above-live)
+              {lnp1-nascent-block-idx :block-idx} (first level-above-nascent)
+              lnp1-block-idx (or lnp1-nascent-block-idx lnp1-live-block-idx)
```
FiV0, line 82: This will change once more with a fix for #4946.

core/src/main/clojure/xtdb/trie_catalog.clj
```diff
@@ -133,19 +132,18 @@
...
-                         (cond-> trie
-                           (and (= trie-state :nascent)
-                                (= trie-block-idx block-idx))
-                           (assoc :state :live)))
-                       %))))
+       (every? (fn [{:keys [live nascent] :as _ln}]
+                 (let [{^long ln-live-block-idx :block-idx :or {ln-live-block-idx -1}} (first live)
+                       {^long ln-nascent-block-idx :block-idx :or {ln-nascent-block-idx -1}} (first nascent)
+                       ln-block-idx (max ln-live-block-idx ln-nascent-block-idx)]
+                   (>= ln-block-idx block-idx))))))
```
FiV0, line 140: This is another location where we need to be careful with races, see #4947.

src/test/clojure/xtdb/compactor_test.clj
```diff
@@ -242,13 +256,16 @@
...
+                  (and (= level required-level) (= part-prefix (take (count part-prefix) (vec part)))))
+                trie/parse-trie-key
+                :out-trie-key)
+          res))
+
 (t/deftest test-l2+-compaction-jobs
   (binding [cat/*file-size-target* 16]
-    (t/is (= #{ ;; L2 [0] is full, compact L3 [0 0] and [0 1]
-               (job "l03-rc-p00-b0f" ["l02-rc-p0-b03" "l02-rc-p0-b07" "l02-rc-p0-b0b" "l02-rc-p0-b0f"] [0 0])
-               (job "l03-rc-p01-b0f" ["l02-rc-p0-b03" "l02-rc-p0-b07" "l02-rc-p0-b0b" "l02-rc-p0-b0f"] [0 1])
```
FiV0, line 249: Added another case for this scenario as `l03` will only happen if the part groups are full.

core/src/main/clojure/xtdb/trie_catalog.clj
jarohen: TODO update this comment

jarohen (approved):
As discussed, all good 👌 

- update comment below
- update issue or PR with whatever notes you need to be confident that you'll understand this change when you come back to it
- prep a demo for showcase with enough context that the team understand it 🙂

core/src/main/clojure/xtdb/trie_catalog.clj
```diff
@@ -133,19 +132,18 @@
...
-(defn- mark-block-idx-live [tries ^long block-idx]
-  (update tries :live+nascent
-          #(doall (map (fn [{trie-state :state, trie-block-idx :block-idx, :as trie}]
-                         (cond-> trie
-                           (and (= trie-state :nascent)
-                                (= trie-block-idx block-idx))
-                           (assoc :state :live)))
-                       %))))
+       (every? (fn [{:keys [live nascent] :as _ln}]
+                 (let [{^long ln-live-block-idx :block-idx :or {ln-live-block-idx -1}} (first live)
```
jarohen, line 137: @FiV0 just to confirm - do these `^long` type hints throw when `(first live)` or `(first nascent)` are empty? was a `when-let` previously

I'm assuming not because I'd be stunned if everything passed if it were the case, but good to check 🙂

FiV0: Not even this seems to throw
```clj
(let [{^long foo :foo} {:foo "str"}]
  foo)
;; => "str"
```

PR #4972, opened 2025-11-08
add LPAD and RPAD

refset:
e.g. often used for displaying data:
```
SELECT relname table_name,
       lpad(to_char(reltuples, 'FM9,999,999,999'), 13) row_count
FROM pg_class
LEFT JOIN pg_namespace
    ON (pg_namespace.oid = pg_class.relnamespace)
WHERE nspname NOT IN ('pg_catalog', 'information_schema')
AND relkind = 'r'
ORDER BY reltuples DESC;
```
=>
```
              table_name               |   row_count
---------------------------------------+---------------
 trips                                 | 1,120,559,744
 uber_trips_2015                       |    14,270,497
 spatial_ref_sys                       |         3,911
 central_park_weather_observations_raw |         2,372
 nyct2010                              |         2,167
 uber_taxi_zone_lookups                |           265
 yellow_tripdata_staging               |             0
 cab_types                             |             0
 uber_trips_staging                    |             0
 green_tripdata_staging                |             0
```
(overview of approximate counts, from https://tech.marksblogg.com/billion-nyc-taxi-rides-redshift.html)

### DIFF
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -1228,6 +1228,56 @@
    :->call-code (fn [[code]]
                   `(str->buf (.toLowerCase (resolve-string ~code))))})
 
+(letfn [(pad-impl [left? str-code len-code pad-code]
+          (let [s (gensym 's)
+                len (gensym 'len)
+                slen (gensym 'slen)
+                pad (gensym 'pad)
+                pad-len (gensym 'pad-len)
+                needed (gensym 'needed)
+                full-pads (gensym 'full-pads)
+                partial (gensym 'partial)
+                padding (gensym 'padding)]
+            `(let [~s (resolve-string ~str-code)
+                   ~len (long ~len-code)]
+               (if (<= ~len 0)
+                 (str->buf "")
+                 (let [~slen (.length ~s)]
+                   (if (>= ~slen ~len)
+                     (str->buf (.substring ~s 0 ~len))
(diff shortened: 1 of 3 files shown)


jarohen:
@refset what's happening with this one? ready to review?

refset:
I'm happy with the logic itself, that the result is Postgres-compliant (maybe with the exception of NULL handling looking again now :thinking: probably need to test that also), and that the tests pass, but I wasn't 100% sure if the codegen pattern is idiomatic (or if there's a better option?) - a quick review would be great :pray: 

Possibly also needs unicode handling (or testing at least)

refset:
Was thinking further about unicode handling, and I just spotted that Datafusion has extensive SLT-format tests for their standard lib: https://github.com/apache/datafusion/blob/36ec9f1de0aeabca60b8f7ebe07d650b8ef03506/datafusion/sqllogictest/test_files/string/string_query.slt.part#L1385

And for comparison, their vectorized(!) implementation: https://github.com/apache/datafusion/blob/36ec9f1de0aeabca60b8f7ebe07d650b8ef03506/datafusion/functions/src/unicode/lpad.rs

jarohen (changes requested):
@refset try to get the pad function in Clojure rather than in the EE - i.e. should be able to `(defn lpad [s len pad] ...)` and then just emit `(lpad ~str-code ~len-code ~pad-code)`

subjectively, I probably wouldn't use a `left?` flag here - extract out the common functionality if it's worthwhile (can create a `pad` function, for example, which would yield the pad string)

PR #4986, opened 2025-11-11
TxSink Phase 1

### LINKED ISSUE #4983: TxSink Phase 1: Replicate log
Once this is complete a user will be able to:
- Configure an xtdb node in `tx-sink` mode (possibly like how the compactor reset is used)
- Configure an output log + format for the `tx-sink`
- Read messages off the configured output log in the format specified

OOS (but to follow soon after):
- Ability to start from an arbitrary offset
- Robust handing of restarts

Akeboshiwind:
The goal here is to have a way of replicating the transaction log as a first step towards CDC.

To that end this PR adds a `tx-sink` command which starts up a node which indexes transactions into a configured output log in a format readable by a downstream consumer.

To run it set the CMD of the docker container to `tx-sink -f node-config.yaml` and add the following to the config (in addition to the rest of your config):
```yaml
TxSink:
  dbName: xtdb         # default
  format: transit+json # default (and only supported format currently)
  outputLog: !Kafka    # required
    cluster: kafkaCluster
    topic: output-topic
```

Transactions are output in the following format:
```clj
{:transaction {:id <the tx key>}
 :system-time #xt/instant "2024-01-01T00:00Z"
 :source {:db "xtdb"}
 :tables [; An entry per table affected in the transaction
          {:db "xtdb"
           :schema "public"
           :table "docs"
           :ops [; An entry per operation on a row in the table
                 {:op :put ; Can be :put, :delete, or :erase
                  :iid "some-iid-uuid"
                  :valid-from #xt/zdt "2024-01-01T00:00Z"
                  :valid-to nil
                  ; :doc is only present of :op is :put
                  :doc {:_id 1, :a "a"}}
                 ...]}
          ...]}
```

Additional notes:
- TxSink won't be started unless you use the `tx-sink` startup command in main. This is intentional because we only want a single node running this process (at least while our upstream log has a single partition).
- On restart this first increment may re-output recent messages to the output log, the next increment will fix that

Closes #4983


core/src/main/clojure/xtdb/tx_sink/main.clj
```diff
@@ -0,0 +1,11 @@
+(ns xtdb.tx-sink.main
```
jarohen: assuming there may well be a reason for separating this, but if not, I'd stick it at the bottom of xt.tx-sink¸ s.t. it's all co-located

core/src/main/clojure/xtdb/tx_sink.clj
```diff
@@ -0,0 +1,86 @@
...
+      :output-log (ig/ref ::output-log)
+      :db-name db-name}
+   ::output-log {:tx-sink-conf tx-sink-conf
+                 :base base}})
+
+(defrecord TxSink [^Log output-log encode-as-bytes db-name]
+  Indexer$TxSink
+  (onCommit [_ tx-key live-idx-tx]
+    (util/with-open [live-idx-snap (.openSnapshot live-idx-tx)]
+      (->> {:transaction {:id tx-key}
```
jarohen: nit: we tend to avoid using `->>` when it's not a data-structure you're transforming - i.e. avoid including side-effects in a threading macro - threading macros do make reading easier when you know all the functions you're passing to take the value as the first/last arg, but otherwise the reader has to stop and visually parse what's happening.

so in this case, I'd probably:

```clojure
(.appendMessage output-log
                (-> {...}
                    encode-as-bytes
                    Log$Message$Tx.))
```

src/test/clojure/xtdb/tx_sink_kafka_test.clj
```diff
@@ -0,0 +1,117 @@
...
+        (binding [*bootstrap-servers* (.getBootstrapServers c)]
+          (f))))))
+
+(defn consume-messages [bootstrap-servers topic]
+  (let [props {"bootstrap.servers" bootstrap-servers
+               "group.id" (str "test-consumer-" (random-uuid))
+               "key.deserializer" "org.apache.kafka.common.serialization.ByteArrayDeserializer"
+               "value.deserializer" "org.apache.kafka.common.serialization.ByteArrayDeserializer"
+               "auto.offset.reset" "earliest"}
+        ^KafkaConsumer consumer (KafkaConsumer. ^java.util.Map props)
```
jarohen: I know it's 'only a test', but best to use `with-open` here, make it 'obviously no bugs' 🙂

core/src/main/clojure/xtdb/tx_sink.clj
```diff
@@ -0,0 +1,86 @@
...
+        system-from-vec (.vectorFor live-relation "_system_from")
+        valid-from-vec (.vectorFor live-relation "_valid_from")
+        valid-to-vec (.vectorFor live-relation "_valid_to")
+        op-vec (.vectorFor live-relation "op")
+        put-vec (.vectorFor op-vec "put")]
+    (doall
+      (for [i (range start-pos pos)]
+        (let [leg (.getLeg op-vec i)
+              put? (= leg "put")
+              data (when put? (.getObject put-vec i))]
```
jarohen: I'd inline these, little value in the names and it means the code is further away from the place it takes effect (I'm big on co-locality of related code 🙂)

i.e. `(cond-> {...} (= (.getLeg op-vec i) "put") (.getObject put-vec i)))`

jarohen (changes requested):
nice one @Akeboshiwind, well done and congratulations on your first significant XT feature 👏 

On reflection, I'd tweak the payload slightly (on the assumption it's a relatively small change in practice):

```clojure
{:transaction {:id <the tx key>}
 :system-time #xt/instant "..." ; use instants here
 :source {:db "xtdb"}
 :tables [{:db "xtdb" ; not strictly necessary I guess?
           :schema "public"
           :table "docs"
           :ops [{:op :put
                  :iid #uuid "..."
                  ;; system-time -> top-level
                  :valid-from #xt/instant "..."
                  :valid-to #xt/instant "..."
                  :doc {...} ; iff put
                  }]}
           ...]}
```

reason to group by tables is that removes any implication that the ops are ordered between tables (which they're not, at this point)

PR #4994, opened 2025-11-12
Add per-cursor level query tracing.

### LINKED ISSUE #4985: Add nested per-cursor OpenTelemetry spans
We should extend OpenTelemetry tracing to include nested per-cursor spans under the main pgwire.query span - These cursor spans should capture high-level timing to support improved query introspection and performance analysis, working in a similar manner to our currently implemented EXPLAIN ANALYZE.

danmason:
Resolves #4985 
Github actions: https://github.com/danmason/xtdb/actions?query=branch%3Atracing-scan-cursor

- Passes the query tracer and `pgwire.query` span down as query-opts from pgwire.
  - This allows us to create spans with pgwire.query as the parent, so they're all nested under there.
- Add TracingCursor/wrapTracing to ICursor.
  - This is where the bulk of the tracing implementation lives. 
  - TracingCursor wraps another ICursor and creates a child span under the provided parent (e.g. `pgwire.query`). 
  - The span is started when the cursor is first advanced and records metrics such as total execution time, time to first block, number of blocks, and total rows processed. These metrics are attached as span tags on close. 
- Calling TracingCursor/wrapTracing from all of our operators - essentially, similar to how we call wrapExplainAnalyze.
- Updated the tracer tests to verify full span hierarchy rather than just top-level spans - building a tree of nested spans, ensuring children appear in start-time order, and filtering out time-related attributes for deterministic comparison.

image

### DIFF
--- a/core/src/main/kotlin/xtdb/ICursor.kt
+++ b/core/src/main/kotlin/xtdb/ICursor.kt
@@ -91,6 +91,7 @@ interface ICursor : Spliterator<RelationReader>, AutoCloseable {
             private var rowCount: Long = 0;
             private var started = false
             private var startTime = clock.instant()
+            private var endTime = clock.instant()
             private var timeToFirstPage: Duration? = null
             private var totalTime: Duration = Duration.ZERO
             private var span: Span? = null
@@ -115,7 +116,9 @@ interface ICursor : Spliterator<RelationReader>, AutoCloseable {
                     pageCount++
                     c.accept(rel)
                 }.also {
-                    totalTime += Duration.between(pageStart, clock.instant())
+                    val pageEnd = clock.instant()
+                    totalTime += Duration.between(pageStart, pageEnd)
+                    endTime = pageEnd
                 }
             }
 
@@ -128,7 +131,6 @@ interface ICursor : Spliterator<RelationReader>, AutoCloseable {
                     }
                     s.tag("cursor.page_count", pageCount.toString())
                     s.tag("cursor.row_count", rowCount.toString())
-                    val endTime = startTime.plus(totalTime)
                     s.end(endTime.asMicros, java.util.concurrent.TimeUnit.MICROSECONDS)
                 }
             }
--- a/core/src/main/kotlin/xtdb/ICursor.kt
+++ b/core/src/main/kotlin/xtdb/ICursor.kt
@@ -1,11 +1,15 @@
 package xtdb
 
+import io.micrometer.tracing.Tracer
+import io.micrometer.tracing.Span
 import org.apache.arrow.memory.BufferAllocator
 import xtdb.api.query.IKeyFn
 import xtdb.api.query.IKeyFn.KeyFn.SNAKE_CASE_STRING
 import xtdb.arrow.RelationReader
+import xtdb.time.InstantUtil.asMicros
 import java.lang.AutoCloseable
 import java.time.Duration
+import java.time.Instant
 import java.time.InstantSource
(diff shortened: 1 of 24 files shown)


core/src/main/kotlin/xtdb/ICursor.kt
```diff
@@ -77,7 +81,65 @@ interface ICursor : Spliterator<RelationReader>, AutoCloseable {
...
+                    rowCount += rel.rowCount
+                    pageCount++
+                    c.accept(rel)
+                }.also {
+                    totalTime += Duration.between(pageStart, clock.instant())
+                }
+            }
+
+            override fun close() {
+                span!!.tag("cursor.total_time_ms", totalTime.toMillis().toString())
```
jarohen: start with `val span = this.span!!` if this is the case, save yourself the `!!` throughout.

if you don't end up merging this with the EA cursor (i.e. if span genuinely isn't nullable at this point) you might want `lateinit var span` as the field instead - see [lateinit docs](https://kotlinlang.org/docs/properties.html#late-initialized-properties-and-variables).

core/src/main/kotlin/xtdb/ICursor.kt
```diff
@@ -77,7 +81,65 @@ interface ICursor : Spliterator<RelationReader>, AutoCloseable {
...
+                timeToFirstPage?.let { ttfp ->
+                    span!!.tag("cursor.time_to_first_page_ms", ttfp.toMillis().toString())
+                }
+                span!!.tag("cursor.page_count", pageCount.toString())
+                span!!.tag("cursor.row_count", rowCount.toString())
+                inner.close()
+                val endTime = startTime.plus(totalTime)
```
jarohen: I might be tempted to keep endTime as a var on the class too, given that you can manually supply it here - i.e. bump it at the end of each call to tryAdvance, then use that time here?

jarohen (changes requested):
nice one @danmason - very valuable addition to the observability of XT 👌 

potential idea:
- have the tracing cursor able to accept nullable tracer
- wrap with tracing cursor if explain-analyse is set or if there's a tracer
- remove the explain-analyse cursor.

as always, your call, you know how similar they actually are.

more generally (out of scope of this PR) could you consider moving this trace out of pgwire, if you haven't already? we're currently looking at modernising the flight-sql driver and would be good to have the tracing there too. somewhere like xtdb.query? I guess you may have already looked at this and concluded that there's meaningful work going on in pgwire too... 🤔

danmason:
> more generally (out of scope of this PR) could you consider moving this trace out of pgwire, if you haven't already? we're currently looking at modernising the flight-sql driver and would be good to have the tracing there too. somewhere like xtdb.query? I guess you may have already looked at this and concluded that there's meaningful work going on in pgwire too... 🤔

One of the main reasons I have it within pgwire is that the query timer is within there - now, I believe I moved that within there as previously our query timer was entirely completely inaccurate & still acting as if we were using the api calls rather than going through pgwire, so it was ONLY timing execute-tx calls. 

Given that we DO go through `openQuery` in pgwire, we could likely use that as our span starting point? We need to pass the tracer/span down in there anyway, so it's one I can follow up with. We could always have separate pgwire level tracing if we want, I don't know how valuable having the time taken for msg.bind even is, really.

> Remove the explain-analyse cursor.

I can certainly look at doing that, yeah - I didn't plan to dig too far into how the explain-analyze cursor was outputting its results, but could see how hard it is to do from the tracingCursor.

PR #4997, opened 2025-11-12
Add max-block-idx to partitions, fixes #4946

### LINKED ISSUE #4946: Keep latest live block-idx per `level` `recency` `part`.

FiV0:
We are adding a max-block-idx to the partition proto. This is the largest block idx the partition has ever seen (including live, nascent and garbage). We use this twofold:
- Decide on stale messages (#4946)
- Decide on what to compact next (making sure we don't compact a file twice)

### DIFF
--- a/src/test/clojure/xtdb/trie_catalog_test.clj
+++ b/src/test/clojure/xtdb/trie_catalog_test.clj
@@ -175,15 +175,18 @@
   (t/is (= #{{:level 1,
               :part [],
               :recency #xt/date "2020-01-01",
-              :tries ["l01-r20200101-b01"]}
+              :tries ["l01-r20200101-b01"]
+              :max-block-idx 1}
              {:level 1,
               :part [],
               :recency #xt/date "2020-01-02",
-              :tries ["l01-r20200102-b01"]}
+              :tries ["l01-r20200102-b01"]
+              :max-block-idx 1}
              {:level 0,
               :part [],
               :recency nil,
-              :tries ["l00-rc-b02" "l00-rc-b01" "l00-rc-b00"]}}
+              :tries ["l00-rc-b02" "l00-rc-b01" "l00-rc-b00"]
+              :max-block-idx 2}}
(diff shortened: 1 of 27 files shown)


src/test/clojure/xtdb/trie_catalog_test.clj
```diff
@@ -405,7 +407,7 @@
                                            #xt/instant "2025-01-01T00:00:00Z")
                         (into (sorted-set) (map :trie-key))))))
 
-    ;; new, with trie-state
+    ;; new, with trie-state, without max-block-idx
```
jarohen, line 410: nice, like it 👌

jarohen (approved):
solid 🪨 

thanks mate 🙏

PR #5032, opened 2025-11-20
Make simulation test iterations configurable by system property.

danmason:
Github actions: https://github.com/danmason/xtdb/actions?query=branch%3Aconfigurable-sim-test-iterations++ 

Required a replacement of RepeatedTest in most places as this cannot have non-compile time values - making use of ParameterizedTest and a MethodSource for the iterations instead.


jarohen (approved):
as discussed:
- try drop `@DisplayName`, rename the tests back
- keep `@Timeout` on the tests themselves (try passing timeout to `runTest` if that's easier)

no worries if not

Issue #5060, opened 2025-11-28
Multi-db support in GC


jarohen:
## Progress

**Status**: completed

- [x] Restructure `GarbageCollector.kt` — extract `ForDatabase`, reshape `Impl` as node-level coordinator
- [x] Update Clojure init-key to pass `DatabaseCatalog` to `Impl`
- [x] Update `NodeSimulationTest` to use `ForDatabase`
- [x] Verify all tests pass

jarohen:
### Chalk — Multi-db GC implementation

- [x] Extract `ForDatabase` from current `GarbageCollector.Impl` (per-database GC logic)
- [x] Reshape `Impl` as node-level coordinator that iterates `DatabaseCatalog`
- [x] Update Clojure integrant init-key
- [x] Update `NodeSimulationTest` for new structure
- [x] Run tests and verify (all pass — 28 targeted + full suite)

<details><summary>Approach</summary>

Mirroring the Compactor's structure: `ForDatabase` holds per-database GC logic (block + trie collection), `Impl` is the node-level coordinator with the periodic loop that iterates all databases in the catalog each cycle.

ForDatabase instances are lazily created and cached in the Impl (keyed by database name), rather than stored on Database objects — GC doesn't need the tight log-processor coupling that Compactor has.

</details>

<details><summary>Implementation details</summary>

`GarbageCollector.ForDatabaseImpl` is the extracted per-database class (was `Impl`).
`GarbageCollector.Impl` is now the node-level coordinator that holds a `ConcurrentHashMap<DatabaseName, ForDatabase>` cache.

Each GC cycle: read `dbCatalog.databaseNames`, prune cache for detached dbs, create/reuse `ForDatabase` per db, run `collectGarbage()` on each sequentially.

No changes to `BlockGarbageCollector`, `DatabaseCatalog`, `Database`, or `GarbageCollectorConfig`.

</details>

PR #5068, opened 2025-12-01
Split out impl and driver on garbage collector.

danmason:
Github actions: https://github.com/danmason/xtdb/actions?query=branch%3Arework-gc

- Reworks garbage collector to act a bit more like the compactor, with deletion of tries as a suspending function within the driver.
  - Switching this to an interface/Impl will ALSO help us with #5060  
- Also adds a MockDriver implementation.

### DIFF
--- a/core/src/main/kotlin/xtdb/garbage_collector/GarbageCollector.kt
+++ b/core/src/main/kotlin/xtdb/garbage_collector/GarbageCollector.kt
@@ -2,11 +2,14 @@ package xtdb.garbage_collector
 
 import kotlinx.coroutines.*
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.plus
 import org.slf4j.LoggerFactory
 import xtdb.database.IDatabase
+import xtdb.table.TableRef
 import xtdb.time.microsAsInstant
 import xtdb.trie.Trie.dataFilePath
 import xtdb.trie.Trie.metaFilePath
+import xtdb.trie.TrieKey
 import java.io.Closeable
 import java.time.Duration
 import java.time.Instant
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/xtdb/garbage_collector/GarbageCollector.kt
```diff
@@ -16,77 +19,108 @@ import kotlin.time.Duration.Companion.seconds
...
             }
         }
-    }
 
-    override fun close() {
-        runBlocking { withTimeout(5.seconds) { parentJob.cancelAndJoin() } }
-        LOGGER.debug("GarbageCollector shut down")
+        override fun collectGarbage() {
+            LOGGER.debug("Starting block garbage collection")
+            blockGc.garbageCollectBlocks()
```
danmason, line 105: Have left blockGc for now since it's primarily deletion of tries we are interested in testing various interleavings against - we may wish to bring that back into this class and move the side effect (ie, deleting a list of blocks) into the driver.

core/src/test/kotlin/xtdb/garbage_collector/GarbageCollectorMockDriver.kt
```diff
@@ -0,0 +1,37 @@
...
+            for (garbageTrie in garbageTries) {
+                LOGGER.debug("Deleting data file for garbage trie {}", garbageTrie)
+                bufferPool.deleteIfExists(tableName.metaFilePath(garbageTrie))
+                yield()
+                LOGGER.debug("Deleting meta file for garbage trie {}", garbageTrie)
+                bufferPool.deleteIfExists(tableName.dataFilePath(garbageTrie))
+                yield()
+            }
+            LOGGER.debug("Removing {} garbage tries from catalog for table {}", garbageTries.size, tableName)
+            trieCatalog.deleteTries(tableName, garbageTries)
```
danmason: I have some yield() calls above for us to yield our suspending function between deletions of tries from the buffer pool - I'm wondering if we might want to introduce interleavings of trieCatalog deletions here also (ie, once deleted from the buffer pool, also delete one-by-one from the trie catalog and add yields between those calls).

jarohen: I'd probably shift these up into the code-under-test - i.e. have the driver only handle the fine-grained side-effects, particularly as each can be observed individually.

so maybe `deletePath(...)` and `deleteTries` as methods on the driver?

jarohen (approved):
cheers @danmason 🙏 

one comment regarding the granularity of the methods but otherwise good to go

PR #5088, opened 2025-12-09
Add tracing on transactions

danmason:
GH actions: https://github.com/danmason/xtdb/actions?query=branch%3Atrace-transactions++

Passes a tracer through the indexer and uses it to gather timings/info about individual transactions.

Adds top level transaction batch tracing which wraps all operations in the batch.
image

Adds tracing to SQL indexer and ensures any queries ran by the SQL indexer get captured and traced as a child span. 
- We pass the tracer + query text down within the openQuery calls made within the indexer, and these all get captured accordingly.

image

Also adds slightly less detailed tracing to put-docs, delete-docs, erase-docs and patch-docs indexers - these aren't really getting used at the moment (aside from put-docs) and the level of detail is much less, so arguable if it's worth keeping them - I have separated these to a separate commit.
- We're only capturing what table they are being applied to, rather than the underlying operation itself. 

image


mbutlerw (approved):
Looks good to me.

Issue #5055, opened 2025-11-26
TxSink can "catch up" when starting behind the indexer


Akeboshiwind:
i.e. The indexer's start offset is *after* the offset the last message posted by TxSink

PR #5115, opened 2025-12-19
TxSink - Add initialScan

### LINKED ISSUE #5054: TxSink can optionally start processing from the beginning of time
Adds a config option similar to:
```yaml
TxSink:
  initialScan: true
  ...
```

Only takes effect if no message is found on the output topic to get a last posted offset from.
Scans the L0 files for events to output.

Akeboshiwind:
Currently TxSink will skip messages in two scenarios:
1. When the `latestSubmittedTx` from the object store is later than the TxId from the last message on the output-log
2. When there is no last message on the output log (i.e. we're starting up for the first time)

<details><summary>Visual description</summary>

image

</details>

This PR aims to fix that by in both scenarios reading events from the object store.
In case 1 we read the last message from the output-log and see what blocks we need to read in order to catch up.
In case 2 we default to skipping reading from the beginning of time unless you specify `initialScan: true` in the txSink config:

```yaml
txSink:
  ...
  initialScan: true ; defaults to false
```

We read from the object store instead of from the log because:
- The log may have been truncated so we can't rely on it
- It might be quicker to read from L0 files instead of processing the whole transaction (not confirmed this though)

---

> [!WARNING]
> There is a breaking change related to the output format of `:doc`s in the output. They now output as "proper" transit, the biggest  change is that keys are now keywords and `"_id"` is now `:xt/id`
> Here's an example of how a transaction:
> ```sql
> BEGIN TRANSACTION;
>   INSERT INTO docs RECORDS {_id: 1, a: 1, b: {c: [1, 2, 3], d: 'test'}}, {_id: 2, a: 2}, {_id: 3, a: 3};
>   UPDATE docs SET a = 4 WHERE _id = 1;
>   DELETE FROM docs WHERE _id = 1;
>   ERASE FROM docs WHERE _id = 1;
>   INSERT INTO other RECORDS {_id: 1};
> COMMIT;
> ```
> Would output:
> ```clj
> {:transaction {:id #xt/tx-key {:tx-id 1, :system-time #xt/instant "2020-01-02T00:00:00Z"}},
>  :system-time #xt/instant "2020-01-02T00:00:00Z",
>  :source {:db xtdb, :block-idx 0},
>  :tables [{:db xtdb, :schema public, :table docs,
>            :ops [{:iid <iid>, :valid-from #xt/instant "2020-01-02T00:00:00Z", :valid-to nil, :op :put, :doc {:xt/id 1, :a 1, :b {:c [1 2 3], :d test}}}
>                  {:iid <iid>, :valid-from #xt/instant "2020-01-02T00:00:00Z", :valid-to nil, :op :put, :doc {:xt/id 2, :a 2}}
>                  {:iid <iid>, :valid-from #xt/instant "2020-01-02T00:00:00Z", :valid-to nil, :op :put, :doc {:xt/id 3, :a 3}}
>                  {:iid <iid>, :valid-from #xt/instant "2020-01-02T00:00:00Z", :valid-to nil, :op :put, :doc {:xt/id 1, :a 4, :b {:c [1 2 3], :d test}}}
>                  {:iid <iid>, :valid-from #xt/instant "2020-01-02T00:00:00Z", :valid-to nil, :op :delete}]}
>           {:db xtdb, :schema public, :table other,
>            :ops [{:iid <iid>, :valid-from #xt/instant "2020-01-02T00:00:00Z", :valid-to nil, :op :put, :doc #:xt{:id 1}}]}
>           {:db xtdb, :schema xt, :table txs,
>            :ops [{:iid <iid>, :valid-from #xt/instant "2020-01-02T00:00:00Z", :valid-to nil, :op :put, :doc {:xt/id 1, :system-time #xt/zdt "2020-01-02T00:00Z[UTC]", :committed true}}]}]}
> ```
> Particularly note the different format of `:doc` and the addition of `:block-idx` to the `:source`

> [!NOTE]
> The order of output `:opt` is only guaranteed between *iid*s, even within the same transaction.
> For example:
> ```sql
> BEGIN TRANSACTION;
>   INSERT INTO docs RECORDS {_id: 'a', v: 0};
>   INSERT INTO docs RECORDS {_id: 'a', v: 1};
>   INSERT INTO docs RECORDS {_id: 'b'};
> COMMIT
> ```
> Could output:
> ```clj
> {...
>  :tables [{...
>            :table "docs"
>            :opt [{... :doc {:xt/id "a" :v 0}}
>                   {... :doc {:xt/id "a" :v 1}}
>                   {... :doc {:xt/id "b"}}]}]}
> ; or
> {...
>  :tables [{...
>            :table "docs"
>            :opt [{... :doc {:xt/id "b"}}
>                   {... :doc {:xt/id "a" :v 0}}
>                   {... :doc {:xt/id "a" :v 1}}]}]}
> ```
> `:opt` for the same `iid` will always be in the same order as in the transaction.

---

Changes:
- When configured with `initial-scan` TxSink now scans `L0` files from the beginning of time out to the output-log
- When TxSink starts up and the previous message is from *before* the end of the previous block, reads events from `L0` files to catch up with the log
- Refactors `read-relation-rows` which also changes the output format of `:doc`
  - Used to output string columns `"_id"`, now outputs keyword columns `:xt/id`
- Make the TxSink node a read-only node

Fixes: #5054, #5055

### DIFF
--- a/core/src/main/clojure/xtdb/tx_sink.clj
+++ b/core/src/main/clojure/xtdb/tx_sink.clj
@@ -115,13 +115,18 @@
 
 (def ^:dynamic *after-block-hook* nil)
 
+(defn gt [a b] (pos? (compare a b)))
+
 (defn- output-block-txs!
-  [^Log output-log grouped-txs db-name block-idx encode-fn]
-  (reduce (fn [_ [system-time events]]
-            (let [{:keys [tx-key message]} (events->tx-output system-time events db-name block-idx encode-fn)]
-              @(.appendMessage output-log message)
-              tx-key))
-          nil
+  [^Log output-log grouped-txs db-name block-idx encode-fn last-key]
+  (reduce (fn [last-key [system-time events]]
+            (let [{tx-key' :tx-key :keys [message]} (events->tx-output system-time events db-name block-idx encode-fn)]
+              (if (or (nil? last-key) (gt tx-key' last-key))
+                (do
+                  @(.appendMessage output-log message)
+                  tx-key')
+                last-key)))
+          last-key
           grouped-txs))
 
 (defn backfill-from-l0!
@@ -131,12 +136,14 @@
             (->> (trie-cat/l0-blocks trie-cat)
                  (drop-while #(>= (or last-block-idx -1) (:block-idx %)))))]
     (loop [last-key (some-> last-message :transaction :id)
-           last-block-idx (some-> last-message :source :block-idx)]
+           ;; Decrement because the last block may not have been fully processed
+           ;; (e.g. if the last message was from the live index before a flush)
+           last-block-idx (some-> last-message :source :block-idx dec)]
       (if-let [blocks (seq (fetch-new-blocks last-block-idx))]
         (let [[new-last-key new-last-block-idx]
-              (reduce (fn [_ {:keys [block-idx tables]}]
+              (reduce (fn [[last-key _] {:keys [block-idx tables]}]
                         (let [grouped-txs (->> tables (read-l0-events al bp) group-events-by-system-time)
-                              new-last-key (output-block-txs! output-log grouped-txs db-name block-idx encode-fn)]
+                              new-last-key (output-block-txs! output-log grouped-txs db-name block-idx encode-fn last-key)]
                           (when *after-block-hook* (*after-block-hook* block-idx))
                           [new-last-key block-idx]))
                       [last-key last-block-idx]
@@ -165,8 +172,6 @@
    ::output-log {:tx-sink-conf tx-sink-conf
                  :base base
                  :db-name db-name}})
-
-(defn gt [a b] (pos? (compare a b)))
 
 (defrecord TxSink [^Log output-log encode-fn ^BlockCatalog block-cat db-name last-tx-key]
   Indexer$TxSink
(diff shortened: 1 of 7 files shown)


core/src/main/clojure/xtdb/tx_sink.clj
```diff
@@ -52,10 +56,94 @@
...
+        event (read-l0-data-file allocator buffer-pool (Trie/dataFilePath table trie-key))]
+    {:table table :event event}))
+
+(defn group-events-by-system-time
+  [events]
+  (-> (group-by #(get-in % [:event :system-from]) events)
+      (update-vals (fn [entries]
+                     (-> (group-by :table entries)
+                         (update-vals #(mapv :event %)))))
+      (->> (into (sorted-map)))))
```
Akeboshiwind: Use `sort-by` instead of `sorted-map`

core/src/main/clojure/xtdb/tx_sink.clj
```diff
@@ -52,10 +56,94 @@
...
+                                        :schema (.getSchemaName table)
+                                        :table (.getTableName table)
+                                        :ops events}))
+                                (into []))}
+                  encode-fn
+                  Log$Message$Tx.)}))
+
+(def ^:dynamic *after-block-hook* nil)
+
+(defn backfill-from-l0!
```
Akeboshiwind, line 132: - Inner loop -> reduce
- Inner loop separate function
- Outer loop -> something like loop + reduce
- Move refresh! + get new blocks to letfn or something

PR #5134, opened 2026-01-08
Fix: EXPLAIN queries were erroring when tracer enabled.

### LINKED ISSUE #5091: EXPLAIN always returns an error on Metabase
Snapshot:
Image

However, when run in my local env, there doesn't seem to be an issue: `(xt/q *node* ["EXPLAIN FROM system"])` responds adequately.

egg-juxt:
Small fix.

@danmason, can you review the test in case you'd want anything better?


danmason (approved):
Looks good to me! :)

PR #5136, opened 2026-01-09
Enshrine trie catalog state in a spec

FiV0:
This tries to enshrine the possible trie catalog states in a spec and asserts against that spec on every change.


FiV0:
Superseded by #5152 as this was from the xtdb repo itself.

PR #5149, opened 2026-01-12
feat: Handling cursoring within cmd-exec-query.

### LINKED ISSUE #5131: Metabase CSV export only returns 500 rows instead of full result set
### Expected Behavior

CSV exports should contain all query results (or respect Metabase's export limit settings) for a table.

### Actual Behavior

When exporting query results from Metabase to CSV (or other formats), only the first 500 rows are being exported, even for tables with far more than 500 rows.
- Running the query via the metabase UI displays the expected 2000 rows
- This appears to have been introduced between versions 0.55.7 and 0.55.x of metabase - we first encountered it when upgrading from 0.55.7 -> 0.57.7.

### Steps to Reproduce

1. Run an XTDB node with a server on port 5432.
2. Set up the XTDB node with a table with > 500 rows.
```clojure
(doseq [batch (partition-all 1000 (range 10000))]
  (xt/submit-tx node 
    (into [] (mapv (fn [i] [:put-docs :foo {:xt/id i :val i}]) batch))))
```
3. Run `metabase/metabase:<version>` via Docker Compose on one of the specified versions.
4. Connect Metabase to the XTDB server as a PostgresSQL connection.
5. Run a query that returns >500 rows (e.g., `SELECT * FROM foo`)
6. Observe that Metabase UI shows first 2000 results correctly
7. Export results to CSV
8. Observe that CSV only contains 500 rows

### Version Testing

Ran through the above steps on a few different versions of metabase:
| XTDB Version | CSV Export Works |
|--------------|------------------|
| 0.55.7       | ✅ Full table exported |
| 0.55.x       | ❌ Only 500 rows |
| 0.56.7       | ❌ Only 500 rows |
| 0.57.7       | ❌ Only 500 rows |

### Hypothesis

Something changed between 0.55.7 and 0.55.x that affects how Metabase retrieves data during export operations. This likely relates to our pgwire implementation, potential areas of interest:

- **Fetch size parameter**: May not be properly handled
- **Cursor support**: Portal/cursor functionality may not be fully implemented

danmason:
Resolves #5131
Github actions: https://github.com/danmason/xtdb/actions/workflows/build.yml?query=branch%3Apgwire-cursoring-5131

### Background

Metabase uses cursors when exporting query results. In this mode, the client repeatedly issues Execute calls with a row limit until the server signals completion. Prior to this change, XTDB would stop sending rows once limit was reached, but failed to:
- correctly signal that the portal was suspended
- preserve any remaining rows already read from relations.

As a result, clients such as Metabase would silently miss rows during export, typically observing only the first 500 rows even though the full result set was available.

### What this changes

- Implements the correct pgwire message flow when an `Execute` reaches its limit but more rows remain:
  - Sends `PortalSuspended` instead of `CommandComplete`
- Buffers any leftover rows from partially-consumed relations when a limit is applied
- Persists buffered rows on the portal so they can be returned on subsequent Execute calls

This brings XTDB’s cursor / portal behavior in line with pgwire expectations and fixes incremental fetch semantics. Metabase exports now return all rows (or correctly respect metabase export limits):
- Tested this by exporting a table with 100,000 rows - previously we'd only return the first 500, now we return everything. 

### Tests

Adds a regression test covering cursor-based fetching with setFetchSize on JDBC to pgwire_test, ensuring that all rows are returned across multiple fetches.

### DIFF
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -1327,7 +1327,7 @@
       Future$State/FAILED (throw (.exceptionNow task)))))
 
 (defn cmd-exec-query [{:keys [conn-state !closing? query-error-counter] :as conn}
-                      {:keys [limit statement-type query ^IResultCursor cursor pg-cols portal-name pending-rows]
+                      {:keys [limit statement-type query ^IResultCursor cursor pg-cols portal-name pending-rows total-rows-sent]
                        :as _portal}]
   ;; Create an implicit transaction if one hasn't already been started
   (let [transaction (get-in @conn-state [:transaction])]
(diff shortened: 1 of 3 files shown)


core/src/main/clojure/xtdb/pgwire.clj
```diff
@@ -1364,11 +1374,23 @@
...
+                                              ;; Buffer any remaining rows for next Execute 
+                                              (dotimes [idx (- row-count num-to-send)] 
+                                                (vswap! !pending conj (serialize-row rel (+ num-to-send idx))))))))))))
 
-      (pgio/cmd-write-msg conn pgio/msg-command-complete {:command (str (statement-head query) " " @!n-rows-out)}))
+      ;; Save any pending rows back to portal
+      (when portal-name
+        (swap! conn-state assoc-in [:portals portal-name :pending-rows] @!pending))
+
+      (if (= @!n-rows-out limit)
```
jarohen: this one just raises my alarm of 'what happens if the result-set size happens to be a multiple of the limit'?

jarohen (changes requested):
nice one @danmason 👏 

one comment below which'd be good to see a test for. I'd also like to see some pgwire protocol tests for this one, so that we can confirm the exact messages going over the wire for the different cases 🙏

core/src/main/clojure/xtdb/pgwire.clj
danmason: Have added some an additional test of cursoring with a few different fetch sizes - does that help cover this? :)

PR #5150, opened 2026-01-12
Fusion Benchmark

### LINKED ISSUE #5125: Benchmark: Add improved readings benchmark
Build on the readings benchmark with observations from one of our DPs.

Discussion [here](https://juxt.slack.com/archives/C09BMP7SD40/p1765797713519819).

tggreene:
There's a docstring in the fusion.clj describing the benchmark, we may want to drop it but I found it helpful whilst working on it.

Link to original discussion [here](https://juxt.slack.com/archives/C09BMP7SD40/p1765797713519819)

Closes #5125

### DIFF
--- a/modules/bench/src/main/clojure/xtdb/bench/fusion.clj
+++ b/modules/bench/src/main/clojure/xtdb/bench/fusion.clj
@@ -11,73 +11,9 @@
   (:import (java.time Duration Instant)
            (java.util UUID)))
 
-(hugsql/def-sqlvec-fns "xtdb/bench/fusion.sql")
+;; See fusion.md for benchmark documentation
 
-;; ============================================================================
-;; FUSION BENCHMARK
-;; ============================================================================
-;;
-;; A benchmark based on actual design partner usage patterns to catch
-;; regressions and enable optimization.
-;;
-;; ## Data Model (11 tables)
-;;
-;; Core tables:
-;;   1. organisation    - Manufacturers (5 rows)
-;;   2. device_series   - Device series per org (25 rows, 5 per org)
-;;   3. device_model    - Reference data (50 rows, 2 per series)
-;;   4. site            - Location data (~= system count)
-;;   5. system          - Main table, constantly updated (configurable count, default 10k)
-;;   6. device          - Links systems to device models (2× system count)
-;;   7. readings        - High volume time-series (system_id, value, valid_time)
-;;
-;; Registration test tables (for cumulative registration query):
-;;   8. test_suite      - Test suite definitions (1 suite)
-;;   9. test_case       - Test cases within suites (5 cases per suite)
-;;   10. test_suite_run - Suite executions per system (1 per system, 80% pass rate)
-;;   11. test_case_run  - Case execution results (5 per suite run)
-;;
-;; ## Workload Characteristics (matching production patterns)
-;;
-;; Readings ingestion:
-;;   - Batch size: ~1000
-;;   - Insert frequency: every 5min of valid_time
-;;   - System-time lag:
-;;     * 80%: 0-2ms delay (near real-time)
-;;     * 20%: 0-50ms delay (delayed batch processing)
-;;
-;; System updates:
-;;   - Small-batch UPDATEs: batch size 30
-;;   - Simulates constant trickle of system state changes (~every 5min)
-;;   - Update frequency: configurable rounds (default 10)
-;;
-;; ## Query Suite (5 queries)
-;;
-;;   1. system-settings           - Point-in-time system lookup
-;;   2. readings-for-system       - Time-series scan with temporal join
-;;   3. system-count-over-time   - Complex 4-table temporal aggregation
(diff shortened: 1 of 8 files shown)


modules/bench/src/main/clojure/xtdb/bench/fusion.clj
```diff
@@ -0,0 +1,625 @@
...
+            [clojure.tools.logging :as log]
+            [xtdb.api :as xt]
+            [xtdb.bench :as b]
+            [xtdb.bench.random :as random]
+            [xtdb.test-util :as tu]
+            [xtdb.util :as util])
+  (:import (java.time Duration Instant)
+           (java.util UUID)))
+
+;; ============================================================================
```
jarohen: extract to an md readme file?

modules/bench/src/main/clojure/xtdb/bench/fusion.clj
```diff
@@ -0,0 +1,625 @@
...
+;; - Large-batch INSERT workload (readings)
+;; - Temporal scatter: data arriving with variable system-time lag
+;; - Complex multi-CTE queries with window functions
+;; - range_bins temporal aggregation patterns
+;; - Multi-table temporal joins with CONTAINS predicates
+;;
+;; ## Usage
+;;
+;; Default (10k systems × 1k readings):
+;;   ./gradlew fusion
```
jarohen: nit: `fusion-bench`...? 🤔 

only because I wasn't sure it was clearly a benchmark task, but then again maybe that's only because I recognise (e.g.) 'tpch' to be a benchmark much quicker 🙂 

I'm talking myself out of it...

modules/bench/src/main/clojure/xtdb/bench/fusion.clj
```diff
@@ -0,0 +1,625 @@
...
+   :name model-name
+   :capacity-kw (random-float rng 5 15)})
+
+(defn generate-site [rng nmi]
+  {:xt/id nmi
+   :address (str (random-int rng 1 999) " Solar Street")
+   :postcode (str (random-int rng 1000 9999))
+   :state (random/uniform-nth rng ["NSW" "VIC" "QLD" "SA" "WA"])})
+
+(defn generate-system-record [rng system-id nmi base-time]
```
jarohen: as discussed, slimming this one down

jarohen (commented):
cheers @tggreene, nice one 🙏 

couple of higher-level considerations:
- shall we bring in [HugSQL](https://hugsql.org/) for the SQL queries, save them being massive inline strings?
- shall we make this more of an Auctionmark-style benchmark rather than distinct stages? some of the issues we've seen have been related to the interleaving of different types of queries/transactions.

modules/bench/src/main/clojure/xtdb/bench/fusion.clj
```diff
@@ -0,0 +1,625 @@
...
+;;   - Batch size: ~1000
+;;   - Insert frequency: every 5min of valid_time
+;;   - System-time lag:
+;;     * 80%: 0-2ms delay (near real-time)
+;;     * 20%: 0-50ms delay (delayed batch processing)
+;;
+;; System updates:
+;;   - Small-batch UPDATEs: batch size 30
+;;   - Simulates constant trickle of system state changes (~every 5min)
+;;   - Update frequency: configurable rounds (default 10)
```
Akeboshiwind: An important property for performance is that some small number of systems stop getting updates over time.

This leads to a trail of systems in valid time across blocks which can cause performance problems.
Basically this if it makes sense:
```clj
(loop [system-ids system-ids
       updates-per-system updates-per-system]
  (if (or (empty? system-ids) (zero? updates-per-system))
    nil
    (do
      (do-update system-ids)
      (recur (->> system-ids shuffle (take (int (* 0.9 (count system-ids)))))
             (dec updates-per-system)))))
```

modules/bench/src/main/clojure/xtdb/bench/fusion.clj
```diff
@@ -0,0 +1,625 @@
...
+(defn generate-test-case-run [rng run-id suite-run-id case-id suite-passed? base-time]
+  (let [case-passed? (or suite-passed? (random/chance? rng 0.7))
+        test-start (.plusSeconds base-time 43200)] ; Same as suite run start
+    {:xt/id run-id
+     :test-suite-run-id suite-run-id
+     :test-case-id case-id
+     :status (if case-passed? "OK" "FAILED")
+     :executed-at test-start}))
+
+(defn ->init-tables-stage
```
Akeboshiwind, line 155: In a real production system these (and the readings and system updates) are all mixed.

I'm not *certain* if this matters or not for the queries.
I suspect for the queries w/ `generate_series` it won't but I'm less sure for the ones where we join two temporal tables together.

Akeboshiwind (commented):
Apologies for the late review!
The system laggards comment is important I think, the other comment I'm less sure of.

Issue #5140, opened 2026-01-09
`block-idx` not sorted in descending order for certain partitions in the trie catalog of the `test-l2+-compaction` test


FiV0:
I have been trying to use #5136 to identify issues in the current tests. `test-l2+compaction` seems to be one of them.
The `garbage` list doesn't seem to be sorted in descending order for some partition in that test after some compaction has happened.

FiV0:
Failing test for this issue
```clojure
(t/deftest issues-with-garbage-being-unordered-5140
  (t/is (= [2 1 0] 
           (->> (-> (apply-msgs ["l00-rc-b00" 10] ["l00-rc-b01" 10] ["l00-rc-b02" 10] ["l00-rc-b03" 10] 
                                ["l01-rc-b00" 10] ["l01-rc-b01" 20] ["l01-rc-b02" 10] ["l01-rc-b03" 10]
                                ["l02-rc-p0-b01" 10] ["l02-rc-p1-b01" 10] ["l02-rc-p2-b01" 10] ["l02-rc-p3-b01" 10])
                    :tries
                    (get-in [[1 nil []] :garbage]))
                (map :block-idx)))))
```

Issue #5139, opened 2026-01-09
`max-block-idx` should always be set in trie-catalog


FiV0:
```clj
(t/deftest max-block-index-always-set-test
  (let [{:keys [tries] :as _table-trie-cat} (apply-msgs ["l02-r20200203-b00" 10])]
    (t/is (true? (every? (comp int? :max-block-idx) (vals tries))))))
```
I ran into this while testing #5136 on `test-protobuf-addtions`. This should only be an issue with the old format.

PR #5152, opened 2026-01-12
Enshrine trie catalog state in a spec

FiV0:
This adds a `::catalog-tries` spec which checks some invariants of the trie-catalog that should always be met. This check is compiled in as an assertion but not enabled by default. There are two aspects that enable spec assertions
- `s/*compile-asserts*` - true by default - can be changed with `-Dclojure.spec.compile-asserts=false`. If this is false asserts are completely elided when the code gets compiled.
- `(s/check-asserts?)` - false be default, can be changed with `(s/check-asserts true)` or `-Dclojure.spec.check-asserts=true`.
So without further explicit changes to the images, the spec assertions are complied in, but not checked as `(s/check-asserts?)` is false by default. 

Issues so far found with this spec:
- #5139
- #5140

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -163,6 +163,10 @@ allprojects {
             val simulationIterations = project.findProperty("simulationIterations")?.toString() ?: "100"
             systemProperty("xtdb.simulation-test-iterations", simulationIterations)
 
+            if (project.hasProperty("specAsserts")) {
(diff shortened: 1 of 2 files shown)


FiV0:
Just realized that for L0 and L1 it's not strictly necessary that the block indices increase by a value of 1. If a table doesn't receive data for some block there will be no L0 or L1 for this block idx.

PR #5178, opened 2026-01-23
feat: add tries-deleted message s.t. we can run GC on an independent machine

jarohen:
This PR is the first step in adding a one-shot GC task and/or a separate garbage collecting XT instance.

Previously, deleteTries was called directly on the local trieCatalog, on the assumption that every node will try to delete the same garbage. With this change, we instead put a deleted-tries message on the log so that files GC'd by any machine on the cluster are removed from all trie-catalogs - this is so that we needn't run GC on every node.

Key design decisions:
- GC does NOT eagerly update the catalog - only LogProcessor does, avoiding the eager update issues seen with the Compactor
- If message emission fails, log warning but continue (deletion succeeded, message is to update the trie catalogs)


jarohen:
closing this one because of #5188 - may well revisit

PR #5216, opened 2026-02-04
feat(sql): support multiple window functions with different specs

### LINKED ISSUE #5211: feat: support for multiple window functions in one operator
grep: `(throw (UnsupportedOperationException. "TODO: only one window function supported at the moment!"))`

jarohen:
N.B. thus far, entirely vibed as a nothing-to-lose background task - will review myself prior to inflicting it on anyone else.

## Summary
- Support queries with multiple window functions that have different partition/order specs (#5211)
- Previously threw `UnsupportedOperationException`

## Changes
- Each window function now gets its own group-mapper and sort-mapping based on its window spec
- Added `invert-sort-mapping` and `compose-mappings` helpers to reorder window columns when specs have different sort orders
- Moved window tests to dedicated `xtdb.sql.window-test` namespace

## Test plan
- [x] Existing window function tests pass
- [x] New tests for multiple window functions with different partition specs
- [x] Tests for mixed functions (ROW_NUMBER + LAG) with independent partitions


jarohen:
`claude --resume 5cbc056c-686a-459d-b8ba-3fe01ab4d91a`

PR #5242, opened 2026-02-12
tidy(allium): verify specs against code, fix discrepancies

henrygarner:
## Summary

- Reviewed all six Allium specs against their implementations to identify where spec and code diverge
- Fixed the specs where the code was correct, fixed the code where the spec revealed a real issue
- Formalised the `addTries` algorithm in trie-cat.allium (previously deferred as a black box)

### Spec corrections

- **live-index**: `GetTableTx` is idempotent (`computeIfAbsent`), not create-every-time. Row op modelled as Arrow union, not flat fields. Single-writer constraint noted as structurally enforced by LogProcessor.
- **memory-hash-trie**: `SplitIntoBranch` shows explicit defaults for new leaves. `sorted_data_cache` documented as volatile.
- **db**: `LeaderFinishesBlock` now includes `table_catalog.finishBlock` and `block_catalog.refresh`. Timestamp semantics documented.
- **gc**: L0 exclusion explained (compaction timing unpredictable, storage cost negligible). Null `block_time` guard added.
- **trie-cat**: `addTries` fully formalised — insertion strategy per level, levelled vs tiered supersession, nascent→live transitions, partition group completion. Catalog entity restructured around `Shard`/`ShardState`.

### Code fix

- **MemoryHashTrie**: `sortedData` now `@Volatile` to ensure visibility across concurrent readers on weak memory models.

## Test plan

- [x] `MutableMemoryHashTrieTest` passes with the `@Volatile` change

### DIFF
--- a/allium/live-index.allium
+++ b/allium/live-index.allium
@@ -1,14 +1,15 @@
 -- live-index.allium
 --
--- NOTE: generated, unverified. may contain errors.
---
 -- Scope: In-memory staging for transactions before block compaction
 -- Includes: LiveIndex, LiveTable, row operations, transaction lifecycle, snapshot visibility
 -- Excludes: Block storage, trie serialization, Arrow memory management
 --
 -- Concurrency model:
 --   - Single-writer thread: all transaction operations (StartTransaction, LogPut,
---     CommitTransaction, AbortTransaction, FinishBlock, NextBlock) execute serially
+--     CommitTransaction, AbortTransaction, FinishBlock, NextBlock) execute serially.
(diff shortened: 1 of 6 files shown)


jarohen:
> L0 exclusion explained (compaction timing unpredictable, storage cost negligible)

reason for this is that L0 forms a complete set of the data - so if anything does go pear-shaped, we can completely reconstruct the deeper levels from the set of L0s.

couple of times, in the earlier days of the compactor, this was _very_ useful :slightly_smiling_face:

henrygarner:
I've never hammered ESC so quickly as when Claude started proposing making L0 eligible for deletion 😳

PR #5249, opened 2026-02-16
feat(group-by): add min/max support for variable-width types

### LINKED ISSUE #2128: min/max agg for variable-width values

tggreene:
The existing reducing-agg-factory approach can't work for variable-width
types (utf8, varbinary, keyword, uuid, uri) because VariableWidthVector
is append-only with no random-access writes (no set-value-code).

Instead, a new VarWidthMinMax Kotlin class tracks per-group winners as
ByteArray values. Comparison is done directly against the incoming
ByteBuffer (avoiding an allocation on every losing row) and only copies
to a ByteArray when the value actually wins.

Note: winners are stored on the Java heap, outside Arrow's allocator
accounting.

### DIFF
--- a/core/src/main/clojure/xtdb/operator/group_by.clj
+++ b/core/src/main/clojure/xtdb/operator/group_by.clj
@@ -15,7 +15,7 @@
            (org.apache.arrow.memory BufferAllocator)
            (xtdb ICursor)
            (xtdb.arrow MonoVector RelationReader Vector VectorReader VectorType VectorWriter)
-           (xtdb.arrow.agg AggregateSpec AggregateSpec$Factory Average Count GroupMapper GroupMapper$Mapper GroupMapper$Null RowCount StdDevPop StdDevSamp Sum VariancePop VarianceSamp)
+           (xtdb.arrow.agg AggregateSpec AggregateSpec$Factory Average Count GroupMapper GroupMapper$Mapper GroupMapper$Null RowCount StdDevPop StdDevSamp Sum VarWidthMinMax VariancePop VarianceSamp)
            (xtdb.expression.map RelationMapBuilder)
            xtdb.operator.distinct.DistinctRelationMap
            xtdb.types.LeastUpperBound))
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/xtdb/arrow/agg/VarWidthMinMax.kt
```diff
@@ -0,0 +1,79 @@
...
+            if (hasZeroRow && winners.isEmpty())
+                outVec.writeNull()
+
+            return outVec
+        }
+
+        override fun close() {}
+    }
+
+    companion object {
```
jarohen: check these two don't exist as util elsewhere - pretty sure we have buf.toByteArray(), compareUnsigned probably better belongs in a util file somewhere (likely next to toByteArray)

core/src/main/clojure/xtdb/operator/group_by.clj
```diff
@@ -370,42 +379,42 @@
...
-                       (cond-> (util/with-close-on-catch [agg-specs (LinkedList.)]
-                                 (doseq [^AggregateSpec$Factory factory agg-factories]
-                                   (.add agg-specs (.build factory allocator)))
-
-                                 (GroupByCursor. allocator in-cursor
-                                                 (->group-mapper allocator (select-keys vec-types group-cols))
-                                                 (vec agg-specs)
-                                                 false))
-                         (or explain-analyze? (and tracer query-span)) (ICursor/wrapTracing tracer query-span)))})))))
+                   (fn [{:keys [vec-types], :as inner-rel}]
```
jarohen: in future, unless you're changing code, can you exclude these kinds of whitespace change from your PR? I'm playing spot the difference otherwise 😅

src/test/clojure/xtdb/operator/group_by_test.clj
```diff
@@ -479,15 +479,53 @@
...
   (t/is (thrown-with-msg? RuntimeException #"Incomparable types in min/max aggregate"
                           (tu/query-ra '[:group-by {:columns [{min (min a)}]}
-                                         [:table {:rows [{:a 32} {:a "foo"}]}]])))
-
-  (t/is (thrown-with-msg? RuntimeException #"Unsupported type in min/max aggregate"
-                          (tu/query-ra '[:group-by {:columns [{min (min a)}]}
-                                         [:table {:rows [{:a "32"} {:a "foo"}]}]]))))
+                                         [:table {:rows [{:a 32} {:a "foo"}]}]]))))
+
+(t/deftest test-min-max-strings
```
jarohen, line 487: just to double-check - null behaviour in SQL works differently if there are grouping columns vs if there aren't. the other tests go through these cases - I'll assume you've cross-checked?

jarohen (approved):
hey @tggreene, nice one 👏 once you've taken a look through the comments, fine to merge without further review

src/test/clojure/xtdb/operator/group_by_test.clj
tggreene: Added empty relation, no grouping, empty relation, with grouping. I think the all nulls cases are covered by 499 and 531.

I took the chance to be a bit more thorough, added tests for the different var width types too, do you think that's unnecessary?

PR #5265, opened 2026-02-20
Exhaustive allocator metrics

egg-juxt:
From commit message:

Added memory metrics (allocated, limit) for all interesting arrow allocators:
- guarantees covering all child allocators (actually adds some missing allocators)
- single metric for all child allocators, tagged by allocator-name and db-name, making it directly configurable by users
- no need to explicitly set up metrics for each allocator

image


jarohen:
will defer to @danmason and @tggreene on this one

tggreene (approved):
This looks great, ship it!

PR #5280, opened 2026-02-24
fix(sql): support Postgres-style backreferences (\N) in REGEXP_REPLACE

### LINKED ISSUE #5278: Support Postgres-style backreferences (\1) in REGEXP_REPLACE replacement strings
REGEXP_REPLACE currently only supports Java-style `$1` backreferences in replacement strings.

Postgres (and most SQL databases) use `\1`, which was silently treated as the literal character `1`.

tggreene:
REGEXP_REPLACE replacement strings use \1, \2 etc for backreferences in Postgres (and most SQL databases per POSIX convention), but Java's Matcher.replaceAll() uses $1, $2. We were passing the replacement string through as-is, so \1 was silently treated as the literal character '1'.

Implemented in the expression engine (as `postgres-backrefs->java`) means it runs at query time on the actual resolved replacement string, which is the correct layer for bridging Postgres replacement semantics to Java's Matcher.

Discovered while comparing clickbench query results against Postgres - Q29 (REGEXP_REPLACE for domain extraction) was returning literal 1.

Closes #5278

### DIFF
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -1112,8 +1112,17 @@
                     `(boolean (re-find (compile-regex ~(or re-literal `(resolve-string ~needle-code)) ~flags)
                                        (resolve-string ~haystack-code))))}))
 
+(defn- postgres-backrefs->java
+  "Converts \\N backreferences to Java's $N in a single pass.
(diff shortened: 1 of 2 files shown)


tggreene:
I think more thought it needed before proceeding with this change

PR #5283, opened 2026-02-24
feat(bench/fusion): add multi-reading aggregation query

tggreene:
Adds a new benchmark query that exercises multi-table range_bins aggregation, MATERIALIZED CTEs, ROW_NUMBER delta calculations, AGE/EXTRACT time arithmetic, and UNION of heterogeneous result sets.

New tables: reading_a (system active power, 5min), reading_b (site active power, 5min), reading_c (state of charge, 15min), control_plan (sparse temporal events).

The query aggregates readings from three typed tables, computes charge rate from SoC deltas, and unions with control plan target boundaries. Added to both staged-only and OLTP workload (5% weight).

feat(bench/fusion): add OLTP writes for reading_b and reading_c

tidy(bench/fusion): add type hints to suppress reflection warnings

### DIFF
--- a/modules/bench/src/main/clojure/xtdb/bench/fusion.clj
+++ b/modules/bench/src/main/clojure/xtdb/bench/fusion.clj
@@ -1,5 +1,6 @@
 (ns xtdb.bench.fusion
-  (:require [clojure.string :as str]
+  (:require [clojure.set :as set]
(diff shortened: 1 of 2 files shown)


tggreene:
I think this is a show/ship, but if anyone's interested in the new query it's here!

PR #5296, opened 2026-02-26
feat(bench): track scan-perf stats over time via github-action-benchmark

mbutlerw:
Since moving scan-perf from K8s to GHA (29bf3be0f), we lost historical tracking — only a one-shot Slack message remained.

Uses github-action-benchmark to store results on a gh-pages branch and auto-generate timeseries charts at dev/bench/scan-perf/. Extracts total time and 4 profile scan stages from the bench log.

Uses -PbenchLogFile to produce a clean JSON-only log, avoiding jq parse errors from Gradle output mixed into stdout.

Results are only stored when running on main, so ad-hoc workflow_dispatch runs on feature branches won't pollute the historical data. Different parameter configs (n-items, id-type) appear as separate series.

Also removes the dead scan-perf entry from K8s chart configs.

Requires one-time setup: create orphan gh-pages branch and enable GitHub Pages in repo settings.

### DIFF
--- a/core/src/main/antlr/xtdb/antlr/Sql.g4
+++ b/core/src/main/antlr/xtdb/antlr/Sql.g4
@@ -584,7 +584,7 @@ whereClause : 'WHERE' searchCondition ;
 groupByClause : 'GROUP' 'BY' (setQuantifier)? groupingElement (',' groupingElement)* ;
 
 groupingElement
-    : columnReference # OrdinaryGroupingSet
+    : expr # ExpressionGroupingSet
     | '(' ')' # EmptyGroupingSet
     ;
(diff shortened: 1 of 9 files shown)


.github/workflows/scan-perf.yml
```diff
@@ -17,6 +17,9 @@ on:
   schedule:
     - cron: '0 19 * * 1-5'
 
+permissions:
+  contents: write
```
mbutlerw: Requires giving write permission to this action (so it can push to gh-pages branch). Does not appear to be easily scoped to a specific branch. Action is tagged to specific commit.

PR #5301, opened 2026-02-26
fix: pass real logOffset into BlockBoundary to fix latestProcessedMsgId metric flipflopping

egg-juxt:
See https://juxt.slack.com/archives/G018EA25K71/p1772125999666639

Alternative: just not update lastProcessedMsgId when offset < 0.

Passing a real offset was preferred because:
- FlushBlock is a real message, so it should update lastProcessedMsgId as well
- No special cases, no -1 surprises
- Would have needed to make msgId optional throughout ReplicaLogProcessor.processRecords, which would have been more of a hassle


egg-juxt:
This one is not applicable anymore since latest Log changes. (Source and ReplicaLogProcessor have been recombined, so the message passed between the two doesn't exist any more. That message was the one that was carrying a -1 and confounded the completedMsgId meter. This meter is now based on liveIndex.latestCompletedTx anyway.)

PR #5312, opened 2026-03-02
fix(pgwire): disallow !Env in ATTACH DATABASE config

### LINKED ISSUE #5311: Disallow !Env within `ATTACH DATABASE`
Currently users can use `!Env` within the yaml config supplied to `ATTACH DATABASE` like so:
```sql
ATTACH DATABASE my_secondary WITH $$
  log: !Kafka
    cluster: kafkaCluster
    topic: !Env MY_TOPIC
$$
```

The `!Env` value is resolved at submission time and then baked into the configuration meaning all other nodes will use the values from the node it was submitted to.
This is misleading `!Env` looks dynamic but is actually a one-shot snapshot.

Akeboshiwind:
!Env tags are resolved at submission time on the submitting node, then the resolved value is baked into the source log via protobuf. Other nodes (or the same node after restart) use whatever was resolved at attach time, not their own environment — making !Env misleading.

Reject !Env in the YAML string before parsing, with a clear error message directing users to use literal values instead.

resolves #5311

### DIFF
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -903,7 +903,11 @@
                                            :db-config (try
                                                         (or (some-> (.configYaml ctx)
                                                                     (sql/plan-expr env)
-                                                                    (Database$Config/fromYaml))
+                                                                    (as-> yaml
+                                                                      (do (when (re-find #"!Env\b" yaml)
+                                                                            (throw (IllegalArgumentException.
+                                                                                     "!Env is not supported in ATTACH DATABASE - environment variables are resolved at submission time and baked into the log, use literal values instead")))
+                                                                          (Database$Config/fromYaml yaml))))
                                                             (Database$Config.))
                                                         (catch Exception e
                                                           (throw (err/incorrect :xtdb/invalid-database-config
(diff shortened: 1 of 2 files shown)


Akeboshiwind:
### Chalk — Explored and backed out @Contextual serializer refactor

- [x] Explored regex-based !Env rejection in pgwire
- [x] Explored @Contextual refactor approach
- [x] Implemented tidy commit with @Contextual across 17 files
- [x] Discovered kaml YAML tree navigation limitations
- [x] Backed out all changes

<details><summary>Explored regex-based !Env rejection in pgwire</summary>

Implemented pattern matching to detect and reject !Env tags in incoming YAML before parsing. Initial implementation worked but required careful handling of multiline values and edge cases. Eventually reverted in favour of exploring architectural alternatives.

</details>

<details><summary>Explored @Contextual refactor approach</summary>

Investigated switching from @file:UseSerializers annotations to @Contextual serializers, allowing runtime selection between env-resolving and env-rejecting serializers based on context (e.g. ATTACH vs initial config). This approach decoupled serializers from their usage context.

</details>

<details><summary>Implemented tidy commit with @Contextual across 17 files</summary>

Created a tidy refactor commit converting serializer annotations across core, aws, azure, google-cloud, and kafka modules. The change itself was clean and architecturally sound. However, the downstream complexity made it not worthwhile.

</details>

<details><summary>Discovered kaml YAML tree navigation limitations</summary>

kaml (kotlinx.serialization YAML) uses an internal tree representation for tagged nodes (!Env, etc). Built-in kotlinx.serialization serializers fail when applied to these nodes with "Must call beginStructure()" errors. Custom logic is required to navigate and filter the YAML tree before serialization. This added significant complexity — effectively requiring a second YAML parser layer.

</details>

<details><summary>Backed out all changes</summary>

The @Contextual refactor approach was architecturally sound but the implementation complexity (custom kaml tree navigation + maintaining parallel serializer sets) was not justified by the benefit. Simpler alternatives (regex rejection, or targeted serializer overrides) would be more maintainable.

</details>

PR #5315, opened 2026-03-04
Add CDC demo docker-compose setup

Akeboshiwind:
Self-contained docker-compose stack for demoing end-to-end Debezium CDC into XTDB: source Postgres → Debezium Connect → Kafka → xtdb-cdc node → shared Kafka log → main XTDB node (queryable via pgwire).

See the README for instructions


refset:
Hey @Akeboshiwind if there's no testing associated with this, it would probably be best to live in https://github.com/xtdb/driver-examples (ultimately replacing one of the existing dirs (see `/debezium`, mostly likely, but getting in there in the first instance would be good)

Akeboshiwind:
Gah, saw your comment too late!

Issue #3129, opened 2024-01-31
Add support for grouping by expressions in SQL


mbutlerw:
Spec only allows grouping by columns present on the table resulting from the FROM clause. (among other related features)

However this is a common requirement and extension and something we support in XTQL.

`SELECT ... FROM docs GROUP BY (docs.x + 1)`

refset:
A real example someone hit trying to use CAST [Play](https://play.xtdb.com/?version=2.0.0-b1&type=sql&txs=W3sic3lzdGVtLXRpbWUiOm51bGwsInR4cyI6IklOU0VSVCBJTlRPIGRvY3MgKF9pZCwgZm9vKSBWQUxVRVMgKDEwLCAnYmFyMScpO1xuSU5TRVJUIElOVE8gZG9jcyAoX2lkLCBmb28pIFZBTFVFUyAoMTIsICdiYXIyJyk7XG5JTlNFUlQgSU5UTyBkb2NzIChfaWQsIGZvbykgVkFMVUVTICgxNCwgJ2JhcjMnKTtcbklOU0VSVCBJTlRPIGRvY3MgKF9pZCwgZm9vKSBWQUxVRVMgKDEzLCAnYmFyNCcpOyJ9XQ%3D%3D&query=U0VMRUNUIGNvdW50KDEpIGFzIGZmLCBDQVNUKF92YWxpZF9mcm9tIEFTIERBVEUpIEZST00gZG9jcwpncm91cCBieSBDQVNUKF92YWxpZF9mcm9tIEFTIERBVEUpOw%3D%3D)
`SELECT count(1) as ff, CAST(_valid_from AS DATE) FROM docs
GROUP BY CAST(_valid_from AS DATE);`

refset:
[...]

Issue #3130, opened 2024-01-31
Add support for derived column references in SQL group by


mbutlerw:
Not supported in SQL spec, however a fairly common feature present in postgres and mysql (although by no means all dbms).

Consideration here that it breaks the typical SQL order of operations, whereby the group by is always executed before the select list.

`SELECT 2 AS b FROM docs GROUP BY b`

Assuming calculating the derived column is deterministic, this functionality can usually be replaced by repeating the expression in both the select list and group by clause.

FiV0:
As this is not SQL spec, we decided to put it on the backburner for now.

refset:
Needed for Metabase (assuming we want nested field access):
```
14:28:51.289 [pgwire-connection--pool-149-thread-2] DEBUG xtdb.pgwire - Interpreting SQL:  -- Metabase
SELECT ("xt"."txs"."error"#>> array[$1]::text[])::text AS "error → message" FROM "xt"."txs" GROUP BY "error → message" ORDER BY "error → message" ASC LIMIT 1000
```
#4235

jarohen:
see Jeremy's commits ^^

PR #5316, opened 2026-03-04
feat(sql): support expressions and SELECT aliases in GROUP BY

mbutlerw:
GROUP BY previously only accepted column references. This adds support for arbitrary expressions (`GROUP BY UPPER(category)`) and SELECT alias references (`GROUP BY cat`).

GROUP BY resolves names through its own scope rather than the GroupInvariantColsTracker scope used by SELECT/HAVING. This prevents GROUP BY's internal column references from polluting !implied-gicrs — routing resolution through the tracker was a category error since GROUP BY defines what's grouped, it doesn't need observing.

Alias precedence follows PostgreSQL/DuckDB: FROM columns first, SELECT aliases as fallback for unqualified names.

When a GROUP BY expression matches a SELECT expression, the planner reuses the SELECT col-sym and pre-projects via a :map node below :group-by, rewriting the SELECT projection to a pass-through.

The !implied-gicrs validation uses a permissive gb-referenced-cols check — columns appearing inside GROUP BY expressions are considered covered.
This over-accepts some edge cases that PostgreSQL would reject (e.g. SELECT b ... GROUP BY UPPER(b) where b is not functionally determined). Strict expression-level validation is left as future work.

### DIFF
--- a/core/src/main/antlr/xtdb/antlr/Sql.g4
+++ b/core/src/main/antlr/xtdb/antlr/Sql.g4
@@ -584,7 +584,7 @@ whereClause : 'WHERE' searchCondition ;
 groupByClause : 'GROUP' 'BY' (setQuantifier)? groupingElement (',' groupingElement)* ;
 
 groupingElement
-    : columnReference # OrdinaryGroupingSet
+    : expr # ExpressionGroupingSet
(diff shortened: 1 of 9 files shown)


mbutlerw:
Closes #3129 #3130.

PR #5318, opened 2026-03-04
fix: inherit parent allocator limit in Kotlin child allocators

tggreene:
The Kotlin `openChildAllocator` was using `Long.MAX_VALUE` as the limit, allowing child allocators (buffer pool, compactor, metadata manager, all storage impls) to allocate without bound.
The Clojure equivalent (`->child-allocator`) already correctly inherits the parent's limit via `.getLimit`.

Spotted whilst looking into fusion oom kill.


jarohen:
as discussed, closing this one - the allocator limits are checked all the way up the tree

PR #5331, opened 2026-03-12
feat!(sql): add RANGE function, make GENERATE_SERIES end-inclusive

mbutlerw:
Align with PostgreSQL/DuckDB semantics: GENERATE_SERIES is now end-inclusive and RANGE is end-exclusive, both backed by a single `:generate_series` IR expression with a 4th `inclusive?` boolean arg.

Grammar accepts both GENERATE_SERIES and RANGE keywords via explicit lexer tokens; the SQL visitor checks the token type to set inclusivity.

3-arg codegen-call methods delegate to 4-arg (defaulting exclusive), avoiding body duplication. `recall-with-cast3` preserves trailing arg-types through cast-and-redispatch so the bool flows through date→timestamp cast paths.

Temporal series functions (`date-series`, `ts-series`, `tstz-series`) lift the inclusive/exclusive predicate out of `take-while` for clarity.

Existing test assertions are preserved verbatim under RANGE; new GENERATE_SERIES assertions verify inclusive behaviour.

### DIFF
--- a/core/src/main/antlr/xtdb/antlr/Sql.g4
+++ b/core/src/main/antlr/xtdb/antlr/Sql.g4
@@ -396,7 +396,7 @@ columnReference : identifierChain ;
 
 /// generate_series function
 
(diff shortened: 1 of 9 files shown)


tggreene (approved):
This looks good to me

PR #5341, opened 2026-03-13
fix(log-proc): follower would start processing messages after the pen…

### LINKED ISSUE #5332: Follower buffer overflow caused by orphaned BlockBoundary after leader revocation mid-block-finish
## Summary

Under chaos monkey testing (FHIR deployment, ~25 tables), followers hit `follower-buffer-overflow` errors and enter a crash loop. The root cause is a race condition during leader revocation that leaves an orphaned `BlockBoundary` message on the replica log.

## Symptoms

- Followers report: `Follower buffer overflow: buffered 1024 records while waiting for BlockUploaded(b3174d)`
- The `IngestionStoppedException` propagates through `replicaWatchers`, causing subsequent leader transitions to fail inside the Kafka rebalance callback
- This triggers a `KafkaException: User rebalance callback throws an error`, making the node unhealthy
- The node restarts, causing more rebalances → feedback loop

## Root cause

When a leader is revoked **between writing `BlockBoundary` and completing `BlockUploaded`**, the boundary message is orphaned on the replica log. On revocation, the new follower starts from `latestSubmittedMsgId` (at or after the boundary offset), so it never sees the orphaned boundary and `pendingBlock` remains null.

When the node is subsequently re-elected leader, the transition code sees `pendingBlock=false` and opens a fresh leader. The live index is still full, so it writes a **second** `BlockBoundary` for the same block, followed by the actual `BlockUploaded`.

Other followers (who were tailing from before the incident) see **both** boundaries but only **one** upload. The first `BlockUploaded` matches the first (orphaned) boundary. During the buffered-record replay, the second boundary sets `pendingBlock` again — but there's no second `BlockUploaded`, so the follower buffers indefinitely until overflow.

## Evidence from replica log dump (offsets 1008900–1011000)

Parsed Kafka messages around the incident:

```
offset  timestamp      type             size
1008932 15:43:12.267   ResolvedTx       9769      ← last tx from leader A
1008934 15:43:12.300   BlockBoundary    16        ← ORPHANED: leader A writes boundary
                       ─── leader A revoked at 15:43:23 (11s into block upload) ───
                       ─── leader A re-assigned at 15:43:28 ───
1008936 15:43:28.478   NoOp             4         ← leader B's transition catchup
1008938 15:43:29.523   ResolvedTx       12747     ← leader B processes one more source tx
1008940 15:43:29.550   BlockBoundary    16        ← leader B writes SECOND boundary (same block)
                       ─── leader B uploads block (14.5s) ───
1008942 15:43:44.090   BlockUploaded    1022376   ← leader B's upload (only one!)
1008944 15:43:44.127   ResolvedTx       12035     ← normal processing resumes
```

Both `BlockBoundary` messages decoded to `block_index=5765 (0x174d)`. The first has `latest_processed_msg_id` source offset 57393443, the second 57393444 (one more tx processed by leader B before the index was still full).

Message type distribution across 2100 records in the dump:
- ResolvedTx: 2077
- TriesAdded: 15
- NoOp: 5
- BlockBoundary: 2 (the duplicate)
- BlockUploaded: 1

## Application logs showing the cascade

```
15:43:12.304 statefulset-1 finishing block b3174d        ← leader A starts block finish
15:43:23.158 statefulset-1 revoked, closing leader       ← revoked mid-finish
15:43:28.303 statefulset-1 assigned, transitioning       ← re-elected
15:43:28.491 statefulset-1 transition: pendingBlock=false ← MISSED the orphaned boundary
15:43:29.553 statefulset-1 finishing block b3174d        ← second attempt
15:43:44.098 statefulset-1 finished block b3174d         ← success (writes second boundary+upload)
15:44:23.278 statefulset-2 follower overflow at msgId 562949954432304 (offset 1010992)
15:44:31.173 statefulset-0 follower overflow at msgId 562949954432304 (offset 1010992)
```

Both followers overflow at exactly offset 1010992 = 1008942 (BlockUploaded) + ~1024 buffered records × 2 (transactional offset stride).

## Fix

Track `pendingBlock` state in the leader processor. On revocation, pass the leader's pending block to the new follower, so that when the node is subsequently re-elected, the transition code correctly detects the in-flight block and finishes it rather than writing a duplicate boundary.

egg-juxt:
…dingBlock's BlockUpload, eventually throwing "follower buffer overflow"

Under test in FHIR...

### DIFF
--- a/core/src/main/kotlin/xtdb/indexer/LogProcessor.kt
+++ b/core/src/main/kotlin/xtdb/indexer/LogProcessor.kt
@@ -151,9 +151,9 @@ class LogProcessor(
                 val pendingBlock = oldSys.pendingBlock
                 LOG.debug("revocation: closing leader system (pendingBlock=${pendingBlock != null})")
                 oldSys.close()
-                val latestSubmitted = replicaLog.latestSubmittedMsgId
-                LOG.debug("revocation: opening follower system from $latestSubmitted")
-                openFollowerSystem(latestSubmitted, pendingBlock)
+                val startMsgId = pendingBlock?.boundaryMsgId ?: replicaLog.latestSubmittedMsgId
+                LOG.debug("revocation: opening follower system from $startMsgId (pendingBlock=${pendingBlock != null})")
+                openFollowerSystem(startMsgId, pendingBlock)
             }
 
             is FollowerSystem -> {


egg-juxt:
Superseded by commit a9156a60

Issue #5335, opened 2026-03-13
IllegalStateException on `onPartitionsAssigned`: Cannot finish block 6074 when current block is 6074


egg-juxt:
Seen in the soak env. 2 consecutive log events:

```
2026-03-13T07:44:21.258355711Z ERROR o.a.k.c.c.i.ConsumerRebalanceListenerInvoker
[Consumer clientId=consumer-xtdb-xtdb-log-4, groupId=xtdb-xtdb-log] User provided listener xtdb.api.log.KafkaCluster$KafkaLog$openGroupConsumer$1$1 failed on invocation of onPartitionsAssigned for partitions [xtdb-log-0]
java.lang.IllegalStateException: Cannot finish block 6074 when current block is 6074
	at xtdb.catalog.BlockCatalog.buildBlock(BlockCatalog.kt:68)
	at xtdb.indexer.BlockFinisher.finishBlock(BlockFinisher.kt:70)
	at xtdb.indexer.LogProcessor.onPartitionsAssigned(LogProcessor.kt:123)
	at xtdb.api.log.KafkaCluster$KafkaLog$openGroupConsumer$1$1.onPartitionsAssigned(KafkaCluster.kt:362)
	at org.apache.kafka.clients.consumer.internals.ConsumerRebalanceListenerInvoker.invokePartitionsAssigned(ConsumerRebalanceListenerInvoker.java:64)
	at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.onJoinComplete(ConsumerCoordinator.java:464)
	at org.apache.kafka.clients.consumer.internals.AbstractCoordinator.joinGroupIfNeeded(AbstractCoordinator.java:510)
	at org.apache.kafka.clients.consumer.internals.AbstractCoordinator.ensureActiveGroup(AbstractCoordinator.java:421)
	at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.poll(ConsumerCoordinator.java:550)
	at org.apache.kafka.clients.consumer.internals.ClassicKafkaConsumer.updateAssignmentMetadataIfNeeded(ClassicKafkaConsumer.java:680)
	at org.apache.kafka.clients.consumer.internals.ClassicKafkaConsumer.poll(ClassicKafkaConsumer.java:644)
	at org.apache.kafka.clients.consumer.internals.ClassicKafkaConsumer.poll(ClassicKafkaConsumer.java:625)
[...]

Issue #5330, opened 2026-03-12
Multi-DB Stage 3: external source logs (single partition)


jarohen:
Part of #4492 (multiple-database support).

## Overview

At this stage, we support external (non-XT format) source logs - CDC sources (e.g. Debezium) and Kafka Connect. This allows users to use their existing Kafka data without first migrating it into XTDB, removing a key barrier to adoption.

## Requirements

* External source logs are required to be single-partition at this stage.
* When creating a database against an external source log, the user specifies a 'projection' - how to convert external events into XT documents. This could be done through either:
  * SQL expressions (using the existing fully-fledged SQL expression engine), or
  * Existing Kafka Connect transforms

jarohen:
### Chalk — External source log architecture (design discussion)

- [ ] External source config in ATTACH DATABASE
- [ ] Wrapper around LeaderLogProcessor
- [ ] Serialised consumption
- [ ] External log offset tracking
- [ ] Persist external offsets in block file

<details><summary>External source config in ATTACH DATABASE</summary>

User specifies an optional `external-source` (or similar) param in the ATTACH DATABASE YAML config.
Leaning towards top-level rather than nested under the existing log impl, because nesting would require adding it to every log implementation.
Polymorphic config (like log config) — start with `!Debezium`, expect to add Kafka Connect etc later.

</details>

<details><summary>Wrapper around LeaderLogProcessor</summary>

External log consumption likely wraps the leader log-proc.
LeaderLogProcessor currently has `processRecords`; envisage adding something like `processResolvedTx` or similar.
The wrapper subscribes to the external log, linearises via its own channel, then either forwards source messages or debezium-resolved transactions.

</details>

<details><summary>Serialised consumption</summary>

The wrapper handles subscription and linearisation.
It needs its own channel to serialise external events alongside normal source log events before forwarding to the leader processor.

</details>

<details><summary>External log offset tracking</summary>

Need to store offsets of the external format log so we can resubscribe on restart.
Possibly a change to Watchers?
The transaction being awaited could be either an external transaction or a source message.

**Sub-question: refactor Watchers first?** Users of source watchers may only need to await transactions (not other message types).
Changing source watchers to track transaction IDs rather than message IDs could simplify external log integration.
**First step: audit who uses Watchers and what they need.**

</details>

<details><summary>Persist external offsets in block file</summary>

External log offsets need to be persisted in the block file.
Likely as an arbitrary protobuf (`Any`), decoded on the way back in.
[...]

PR #5365, opened 2026-03-20
feat(debezium): ExternalLog abstraction, block token, CDC indexing (#5330)

jarohen:
## Summary

Phases 4–5b of the external source (CDC) integration work tracked in #5330.

- **`ExternalLog<M>` interface** — generic subscribable log abstraction in core, decoupling the database from Debezium-specific types.
  `ExternalLog.Factory` follows the same extensibility pattern as `Log.Factory` (proto via `google.protobuf.Any`, YAML via kotlinx.serialization, `ServiceLoader` for registration).
  Opened as an integrant component (`:xtdb/external-source-log`) and stored on `DatabaseStorage` for the leader processor to consume.

- **`external_source_token`** — opaque `google.protobuf.Any` threaded through the block pipeline (Block, BlockBoundary, BlockUploaded protos).
  On restart, the leader reads this from the latest block to resume CDC from the correct offset.
  All fields default to null/absent — purely additive, no behaviour change until a later phase sets the token.

- **`LiveIndex.Tx.indexCdcEvent`** — extension that applies a `CdcEvent` directly to the live index, bypassing the source log and Clojure indexer dispatch.
  CDC events are limited to puts and deletes, so the full indexer is unnecessary.

Also includes:
- **Debezium module restructure**: `DebeziumConsumer` → `KafkaDebeziumLog` (runtime consumer implementing `DebeziumLog`), config becomes nested `KafkaDebeziumLog.Factory`.
  `DebeziumLog` extends `ExternalLog<DebeziumMessage>` so the type hierarchy is explicit.
- **Kafka subscription fix**: `KafkaConsumer` leak prevention and paused-partition bug fix in cooperative rebalancing (#5335).
- **`LeaderProcessor` interface restoration**: reverted to interface for the extension point needed in Phase 6.

## Remaining phases (separate PRs)

- Phase 5c: `afterToken` on `ExternalLog.tailAll` for resume semantics
- Phase 6: External log processor + leader integration
- Phase 7: End-to-end integration test with restart/resume

## Test plan

- [x] Full `./gradlew test` passes
- [x] Debezium module tests (proto/YAML round-trips, consumer tests)
- [x] Multi-db tests (ATTACH DATABASE error messages with renamed properties)
- [x] ExternalSourceToken round-trip tests
- [ ] Integration tests with Kafka (CI)

### DIFF
--- a/core/src/main/clojure/xtdb/db_catalog.clj
+++ b/core/src/main/clojure/xtdb/db_catalog.clj
@@ -41,11 +41,12 @@
 (defmethod ig/expand-key ::storage [k _opts]
   {k {:source-log (ig/ref :xtdb/source-log)
       :replica-log (ig/ref :xtdb/replica-log)
+      :external-source (ig/ref :xtdb/external-source-log)
       :buffer-pool (ig/ref :xtdb/buffer-pool)
       :metadata-manager (ig/ref :xtdb.metadata/metadata-manager)}})
 
-(defmethod ig/init-key ::storage [_ {:keys [source-log replica-log buffer-pool metadata-manager]}]
-  (DatabaseStorage. source-log replica-log buffer-pool metadata-manager))
+(defmethod ig/init-key ::storage [_ {:keys [source-log replica-log external-source buffer-pool metadata-manager]}]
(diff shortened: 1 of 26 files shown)


jarohen:
Wrong diff — reopening with correct scope.

PR #5368, opened 2026-03-23
fix(log): make Watchers the single source of truth for message positions

jarohen:
## Summary

During leader→follower→leader transitions, the new leader could resume from a stale source log position, re-processing source messages and writing duplicates to the replica log. Other followers then crash with `srcMsgId < latestSourceMsgId` when they see the non-monotonic source IDs.

## Root cause

When a leader is revoked and becomes a follower, the new follower was initialized with `afterSourceMsgId` from a `val` computed once at `LogProcessor` construction time (from the block catalog). If the leader had processed far beyond that position, the follower started with a stale value.

The follower would then catch up on the replica log, but every `ResolvedTx` it replayed was deduplicated (`msg.txId <= latestTxId` in the shared `liveIndex`) — the deduplication path notified watchers with `null` for the source ID and returned early without advancing the follower's `latestSourceMsgId` field. So the field stayed at the stale block-catalog value.

When the node immediately became leader again (common during Kafka rebalances), the transition read this stale field and started the new leader from the wrong source position.

## Fix

Rather than patching the field-passing (which is fragile — every new message type or transition path needs to correctly propagate both IDs), we make Watchers the single source of truth for source and replica positions.

- **Remove all message ID fields** from the processor interfaces and implementations. Processors still notify Watchers as they process records (they already did), but they no longer maintain their own position state.
- **Add `Watchers.sync()`** — sends a `SyncWatcher` through the event channel (same pattern as the existing await methods) that flushes all prior notifications before returning. Called at the two transition points in `LogProcessor`.
- **Remove `afterSourceMsgId`/`afterReplicaMsgId`** from all factory methods — processors don't need initial positions since Watchers already holds the correct values.

This fixes the bug by construction: the leader's source ID notifications persist in Watchers across the leader→follower transition, unaffected by the follower's deduplicated messages (which send `null` source IDs and don't regress the watermark).

### Secondary fixes

- Leader's `finishBlock` now notifies Watchers of the replica advancement from block uploads — previously `appendToReplica` updated a local field but the block upload's final replica ID was never sent to Watchers.
- Transition's `BlockBoundary` handler now notifies with the block upload's final replica ID, not the boundary's.
- `pendingBlock` assignment in `finishBlock` is now a single expression with the `appendToReplica` call, making the coupling explicit.

## Test plan

- [ ] Existing LogProcessorTest, LeaderLogProcessorTest, FollowerLogProcessorTest, WatchersTest all pass (1229 tests)
- [ ] Verify on benchmark cluster that the `srcMsgId < latestSourceMsgId` error no longer occurs during leader transitions
- [ ] Verify on production-like deployment with rapid rebalances (the scenario from the auctionmark benchmarks)

### DIFF
--- a/core/src/main/clojure/xtdb/log/processor.clj
+++ b/core/src/main/clojure/xtdb/log/processor.clj
@@ -74,29 +74,27 @@
                                                {:keys [meter-registry]} :base}]
   (when (.getEnabled indexer-conf)
     (let [proc-factory (reify LogProcessor$ProcessorFactory
-                         (openFollower [_ pending-block after-source-msg-id]
+                         (openFollower [_ pending-block]
                            (FollowerLogProcessor. allocator buffer-pool db-state compactor-for-db
                                                   source-watchers
-                                                  db-catalog pending-block after-source-msg-id))
+                                                  db-catalog pending-block))
 
-                         (openLeaderProcessor [_ replica-producer after-replica-msg-id]
+                         (openLeaderProcessor [_ replica-producer]
                            (LeaderLogProcessor. allocator
                                                 db-storage replica-producer db-state
                                                 indexer-for-db source-watchers
                                                 (set (.getSkipTxs indexer-conf))
                                                 db-catalog block-uploader
-                                                after-replica-msg-id
                                                 (.getFlushDuration indexer-conf)
                                                 meter-registry))
 
-                         (openTransition [_ replica-producer after-source-msg-id]
+                         (openTransition [_ replica-producer]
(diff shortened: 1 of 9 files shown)


jarohen:
Replaced by #5369 — the Watchers-as-source-of-truth approach had issues found during review (missing replica notifications in `finishBlock`/`uploadBlock`, ordering problems with buffered records, and the `sync()` mechanism adding fragility at transition points).

#5369 reverts this and re-does the fix with explicit position tracking on each processor, plus moves the replica await from Watchers to FLP using `MutableStateFlow`.

PR #5375, opened 2026-03-24
fix(log): exception on Kafka Producer tx `abort` was hiding `commit` exception

egg-juxt:
Solution is to remove `abort` in case of a `commit` exception, as a Kakfa transaction can't be aborted anyway once started committing.

Relates to #5376 where we don't see the exception causing the Kafka `commit` error:

```
level	
ERROR
log	
leader: failed to process log record with msgId 562950016061053 (Tx)
java.lang.IllegalStateException: Transaction already closed
	at xtdb.api.log.KafkaCluster$KafkaLog$openAtomicProducer$1$openTx$1.abort(KafkaCluster.kt:283)
	at xtdb.indexer.LeaderLogProcessor.appendToReplica(LeaderLogProcessor.kt:273)
	at xtdb.indexer.LeaderLogProcessor.handleResolvedTx(LeaderLogProcessor.kt:154)
	at xtdb.indexer.LeaderLogProcessor.processRecords(LeaderLogProcessor.kt:170)
	at xtdb.api.log.KafkaCluster$KafkaLog$openGroupSubscription$pollingJob$1.invokeSuspend(KafkaCluster.kt:363)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
	at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:124)
	at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:89)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:586)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:820)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:717)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:704)
logger	
xtdb.indexer.LeaderLogProcessor
stream	
stdout
time	
2026-03-23T21:45:12.278452311Z
```

### DIFF
--- a/core/src/main/kotlin/xtdb/api/log/Log.kt
+++ b/core/src/main/kotlin/xtdb/api/log/Log.kt
@@ -149,11 +149,15 @@ interface Log<M> : AutoCloseable {
             inline fun <M, R> AtomicProducer<M>.withTx(block: (Tx<M>) -> R): R =
                 openTx().use { tx ->
                     try {
-                        block(tx).also { tx.commit() }
+                        block(tx)
                     } catch (e: Throwable) {
-                        tx.abort()
+                        try {
+                            tx.abort()
+                        } catch (abortEx: Throwable) {
+                            e.addSuppressed(abortEx)
+                        }
                         throw e
-                    }
+                    }.also { tx.commit() }
                 }
         }
     }
--- a/modules/kafka/src/main/kotlin/xtdb/api/log/KafkaCluster.kt
+++ b/modules/kafka/src/main/kotlin/xtdb/api/log/KafkaCluster.kt
@@ -169,11 +169,15 @@ class KafkaCluster(
             inline fun <M, R> AtomicProducer<M>.withTx(block: (Tx<M>) -> R): R =
                 openTx().use { tx ->
                     try {
-                        block(tx).also { tx.commit() }
+                        block(tx)
                     } catch (e: Throwable) {
-                        tx.abort()
+                        try {
+                            tx.abort()
+                        } catch (abortEx: Throwable) {
+                            e.addSuppressed(abortEx)
+                        }
                         throw e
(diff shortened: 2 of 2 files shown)


egg-juxt:
As pointed during showcase today, needs fixing, as tx must be aborted on an exception thrown by the tx block.

egg-juxt:
### Chalk — Fix withTx abort/commit exception handling

- [x] Fix `withTx` so `abort` doesn't run after `commit` has executed
- [x] Ensure abort exceptions don't hide the original caught exception (addSuppressed)
- [x] Apply fix to both `Log.kt` and `KafkaCluster.kt`
- [ ] Commit and cherry-pick to both branches

<details><summary>Fix withTx abort/commit exception handling</summary>

The original `withTx` used `block(tx).also { tx.commit() }` inside a single try — meaning if `commit()` threw, the catch block would attempt `abort()`, which fails with "Transaction already closed" (because `commit()` already set `isOpen = false`). This masked the real commit exception.

Restructured so `commit()` runs outside the try via `.also {}` on the try expression result. Only block exceptions trigger `abort()`, and abort failures are added as suppressed exceptions rather than replacing the original.

Both `Log.AtomicProducer.withTx` and `KafkaCluster.AtomicProducer.withTx` had the same bug — the Kafka version exists as a separate extension to expose `sendOffsetsToTransaction` to the block.

</details>

PR #5379, opened 2026-03-24
feat(sql)!: USER keyword and reserved word column aliases

### LINKED ISSUE #5381: Grafana Compatibility: Add USER as CURRENT_USER
Details in this PR #5379, needed for Grafana table discovery queries

tggreene:
## Summary

- Treat bare `USER` as `CURRENT_USER` per SQL spec (breaking: bare `user` as table/column name now requires quoting)
- Allow reserved keywords as column aliases via an expanded `columnLabel` grammar rule

## Breaking

- `USER` is now a reserved keyword (synonym for `CURRENT_USER`).
  Bare `user` as a table or column name must be quoted as `"user"`.
  Use as an alias (`SELECT x AS user`) is unaffected.

## Design

The `columnLabel` rule now accepts nearly every keyword token in the grammar, so any keyword can be used as a column alias (with or without `AS`).
Data type keywords (`TIME`, `DATE`, `INTEGER`, etc.) are also added to the `identifier` rule so they're usable everywhere — table names, column refs, function names, bare aliases.

This mirrors PostgreSQL's approach: the lexer stamps keywords with token types, and the parser contextually reintroduces them where safe as identifiers or aliases.

One caveat: alias keywords used in ORDER BY / GROUP BY can conflict with clause-starters.
e.g. `SELECT 1 AS HAVING GROUP BY HAVING` throws a parse error.
This matches DuckDB behaviour.

## Tests

- Systematic coverage of keywords tested in alias positions (extracted from grammar at test time)
- `identifier` keywords verified referenceable in ORDER BY and GROUP BY
- Data type keywords still work in type roles (CAST, DATE/TIME/TIMESTAMP literals)

Builds on https://github.com/xtdb/xtdb/pull/5374

### DIFF
--- a/core/src/main/antlr/xtdb/antlr/Sql.g4
+++ b/core/src/main/antlr/xtdb/antlr/Sql.g4
@@ -143,7 +143,7 @@ identifier
         | 'SELECT' | 'INSERT' | 'UPDATE' | 'DELETE' | 'ERASE'
         | 'SETTING'
         | 'ROLE'
-        | 'USER' | 'PASSWORD'
+        | 'PASSWORD'
         | 'VARBINARY' | 'BYTEA'
         | 'URI' | 'OID'
         | 'COPY' | 'FORMAT'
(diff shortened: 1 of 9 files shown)


tggreene:
Breaking this into smaller PRs

Issue #5333, opened 2026-03-13
Remove WARN message when re-registering "follower-log-processor" allocator metric on leader->follower switchover


egg-juxt:
When a node changes from leader to follower, this WARN message will be seen:

```
2026-03-13T08:20:09.05400938Z WARN i.m.core.instrument.MeterRegistry
This Gauge has been already registered (MeterId{name='xtdb.allocator.memory.allocated', tags=[tag(allocator=follower-log-processor),tag(database=xtdb),tag(node-id=xtdb-statefulset-1)]}), the registration will be ignored. Note that subsequent logs will be logged at debug level.
```

PR #5384, opened 2026-03-24
ops: preventing WARN "This Gauge has been already registered"

egg-juxt:
solves #5333


egg-juxt:
These changes apply to the Clojure impl of allocator-meters. But those have been moved to Kotlin, so the same changes have been done in #5393

PR #5385, opened 2026-03-24
tidy: prevent meters holding refs that would prevent GC of monitored objects

egg-juxt:


tggreene (approved):
This looks great, good spot!

PR #5386, opened 2026-03-25
tidy: downgrade cancellation logging from ERROR/WARN to DEBUG

### LINKED ISSUE #5401: ERROR log messages on job cancellation
As seen in FHIR env, when terminating an XTDB node:

```
2026-03-20T09:13:55.399018149Z ERROR xtdb.compactor.Compactor 
error running compaction job: public/condition/l01-rc-b3192c, files in job: '["l01-rc-b3192a" "l00-rc-b3192c"]'
kotlinx.coroutines.JobCancellationException: Job was cancelled
```

## Progress

**Status**: completed

- [x] Compactor: use `debugBlock` to downgrade cancellation from ERROR to DEBUG
- [x] GarbageCollector: same pattern — `debugBlock`, System.Logger, `runInterruptible` for blocking calls
- [x] Verified in FHIR env

egg-juxt:
I have looked for introducing a general pattern here.

solves #5401

### DIFF
--- a/core/src/main/kotlin/xtdb/Util.kt
+++ b/core/src/main/kotlin/xtdb/Util.kt
@@ -2,8 +2,13 @@ package xtdb
 
 import clojure.lang.Keyword
 import clojure.lang.Symbol
+import xtdb.util.warn
 import java.nio.ByteBuffer
+import java.nio.channels.ClosedByInterruptException
 import java.util.*
+import kotlin.coroutines.cancellation.CancellationException
+
+private val LOG: System.Logger = System.getLogger("xtdb.util")
 
 val UUID.asByteBuffer: ByteBuffer
     get() = ByteBuffer.wrap(ByteArray(16)).apply {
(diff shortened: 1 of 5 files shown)


jarohen:
Closing as superseded — both halves of this landed on `main` independently, via different code.

- **The compactor half shipped as 321d2a2c8**, `fix(compactor): don't ERROR-log cancelled compaction jobs`.
  `Compactor.kt` now catches `CancellationException` and rethrows it ahead of the general `Throwable` branch that logs.

- **`GarbageCollector.kt` no longer exists.**
  The GC rewrite in #5511 split it into `TrieGarbageCollector.kt` and `BlockGarbageCollector.kt`, and both catch `CancellationException` before the `Exception` branch and rethrow without logging — so the swallow-into-WARN this PR was fixing is structurally gone.

Rebasing would mean re-deriving the patch against a deleted file, for a fix that is already in place.

One piece is not on `main`: `SegmentMerge.kt:227` still uses `Thread.interrupted()` where this PR proposed `ensureActive()`. That is orthogonal to the logging behaviour and is not what #5401 described, so it is being dropped rather than carried — it wants its own card if anyone still wants the change.

PR #5412, opened 2026-04-01
feat(sql): add string_to_array function

tggreene:
## Summary
- Adds PostgreSQL-compatible `string_to_array(string, delimiter)` function
- Supports NULL delimiter (per-character split) and empty delimiter (whole string as single element)
- Matches PG semantics for empty input, consecutive delimiters, and regex metacharacters in delimiters
- Introduces `strings->list-reader` helper for constructing ListValueReader from string arrays

### DIFF
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -1585,15 +1585,13 @@
 
 (defmethod codegen-call [:string_to_array :utf8 :utf8] [_]
   {:return-type #xt/type [:list :utf8]
-   :continue-call (fn [f [s delim]]
-                    (f #xt/type [:list :utf8]
-                       `(string-to-array (resolve-string ~s) (resolve-string ~delim))))})
+   :->call-code (fn [[s delim]]
+                  `(string-to-array (resolve-string ~s) (resolve-string ~delim)))})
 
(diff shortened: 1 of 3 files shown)


core/src/main/clojure/xtdb/expression.clj
```diff
@@ -1558,6 +1559,48 @@
...
+
+(defn string-to-chars ^xtdb.arrow.ListValueReader [^String s]
+  (if (.isEmpty s)
+    (strings->list-reader (into-array String []))
+    (let [parts (into-array String (mapv str (.toCharArray s)))]
+      (strings->list-reader parts))))
+
+(defmethod codegen-call [:string_to_array :utf8 :utf8] [_]
+  {:return-type #xt/type [:list :utf8]
+   :continue-call (fn [f [s delim]]
```
mbutlerw: NIT: Preferable to use `:->call-code` if the return type is static/known ahead of time. 

```
(defmethod codegen-call [:string_to_array :utf8 :utf8] [_]                                                                                                                                                                                                                                                              
    {:return-type #xt/type [:list :utf8]                    
     :->call-code (fn [[s delim]]                                                                                                                                                                                                                                                                                         
                    `(string-to-array (resolve-string ~s) (resolve-string ~delim)))})

mbutlerw (approved):

PR #5413, opened 2026-04-01
feat(sql): add quote_ident function

tggreene:
## Summary
- Adds PostgreSQL-compatible `quote_ident(name)` for identifier quoting
- Returns the identifier unmodified if safe (lowercase, no reserved words, no special chars)
- Otherwise wraps in double quotes with proper `""` escaping
- Uses the SqlLexer vocabulary at load time to detect reserved words

~Flag on this PR is I'm not completely happy with how we build sql-keywords but also can't think of a way around it. If you do please say so!~

I think I've reached the most reasonable solution I can think of for sql-keywords, matching a-zA-Z_+ on literal names of sql lexer vocabulary, there's only one exception on analyze/analyse and should expand with future modifications of vocabulary without issue.

### DIFF
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -1568,7 +1568,10 @@
   Multi-alternative rules (e.g. ANALYZE : 'ANALYZE' | 'ANALYSE') have no single
   literal name — these are added explicitly below."
   (let [vocab (xtdb.antlr.SqlLexer/VOCABULARY)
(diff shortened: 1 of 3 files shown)


mbutlerw (approved):
I think this isn't quite the right approach. 

Ideally we'd want to define our keywords (reserved, unreserved etc.) and then generate the grammar from that.

However given any edge cases/behaviour is limited to this function alone, happy to approve. 

Suggest we maybe leave this as undocumented/experimental however.

PR #5415, opened 2026-04-01
feat(sql): add current_setting('search_path') support

tggreene:
## Summary
- Returns `'"$user", public'` for `current_setting('search_path')` to match PostgreSQL defaults
- Grafana's PostgreSQL plugin queries this during table discovery

### DIFF
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -1394,6 +1394,7 @@
 (defn current-setting [setting-name]
   (case setting-name
     "server_version_num" "160000"
+    "search_path" (str/join ", " explicit-search-path)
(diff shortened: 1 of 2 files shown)


core/src/main/clojure/xtdb/expression.clj
```diff
@@ -1394,6 +1394,7 @@
 (defn current-setting [setting-name]
   (case setting-name
     "server_version_num" "160000"
+    "search_path" "\"$user\", public"
```
mbutlerw: search_path is already modelled in XT, albeit as a static value, current-setting should return this value.

https://github.com/xtdb/xtdb/blob/fed14978aac773263ae6cdcc3cc31ca778cc652b/core/src/main/clojure/xtdb/expression.clj#L1359

Is it required to return $user here for would public alone suffice?

mbutlerw (approved):

PR #5428, opened 2026-04-07
fix(indexer): follower block staleness check skips isFull-triggered blocks

mbutlerw:
When the leader finishes a block via isFull(), the BlockBoundary's latestProcessedMsgId equals the preceding ResolvedTx's txId. The follower's stale check used `latestProcessedMsgId <= latestSourceMsgId`, which evaluated to `T <= T` — true — silently dropping both the BlockBoundary and BlockUploaded.

This meant the follower never called blockCatalog.refresh, liveIndex.nextBlock, or compactor.signalBlock for these blocks. The block catalog froze at its startup value, causing block-lag to grow steadily (~1 per block) until the healthcheck threshold was crossed.

Followers weren't totally broken because L0 tries are delivered separately via TriesAdded messages on the replica log, so queries still found data. The visible symptom was periodic healthcheck 503s as the block-lag grew past the threshold of 5.

Block messages are block-level events — their staleness should be defined by blockIndex against the block catalog, not by comparing source message watermarks. Same fix applied to TransitionLogProcessor.

### DIFF
--- a/core/src/test/kotlin/xtdb/indexer/FollowerLogProcessorTest.kt
+++ b/core/src/test/kotlin/xtdb/indexer/FollowerLogProcessorTest.kt
@@ -7,10 +7,14 @@ import org.junit.jupiter.api.BeforeEach
 import kotlinx.coroutines.test.runTest
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertThrows
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNull
 import xtdb.api.TransactionKey
 import xtdb.api.log.Log
 import xtdb.api.log.ReplicaMessage
 import xtdb.api.log.Watchers
+import xtdb.api.storage.Storage
+import xtdb.block.proto.block
 import xtdb.catalog.BlockCatalog
 import xtdb.catalog.TableCatalog
 import xtdb.compactor.Compactor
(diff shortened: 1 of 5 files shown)


core/src/test/kotlin/xtdb/indexer/TransitionLogProcessorTest.kt
```diff
@@ -0,0 +1,77 @@
...
+        Log.Record(0, offset, Instant.now(), message)
+
+    @Test
+    fun `block boundary not skipped when isFull triggers on same txId`() = runTest {
+        val proc = makeProcessor()
+
+        val txId = 100L
+        proc.processRecords(listOf(
+            record(0, ReplicaMessage.ResolvedTx(txId, Instant.now(), true, null, emptyMap())),
+            record(1, ReplicaMessage.BlockBoundary(0, txId)),
```
mbutlerw, line 70: I'm not entirely sure a transition processor should ever see a BlockBoundary message, as the follower blocks until its caught up to the latest block before transitioning.

However the code is there https://github.com/xtdb/xtdb/blob/63fa1a69e51990594a2e3e50f9fdbfec4da3fe5e/core/src/main/kotlin/xtdb/indexer/TransitionLogProcessor.kt#L91 which suggests it thinks it might, so wanted to add the test for now.

Can circle back with a tidy commit if the code turns out to be dead.

mbutlerw: This is probably a good thing, as the code suggests that the transitionLogProcessor should be uploading blocks, which wouldn't be happening in this case, meaning missing block files.

core/src/test/kotlin/xtdb/indexer/FollowerLogProcessorTest.kt
```diff
@@ -118,6 +126,31 @@ class FollowerLogProcessorTest {
         proc.close()
     }
 
+    @Test
+    fun `block boundary not skipped when isFull triggers on same txId`() = runTest {
+        // Simulates the replica log sequence when isFull() triggers on the leader:
+        // the BlockBoundary's latestProcessedMsgId equals the preceding ResolvedTx's txId.
+        // The follower must still process the block transition.
+        val proc = makeProcessor()
```
jarohen, line 134: nit: `use`

jarohen (approved):

Issue #5170, opened 2026-01-21
Grafana query-builder compatibility


jarohen:
From Claude:

# Grafana Query-Builder Compatibility

[#5170](https://github.com/xtdb/xtdb/issues/5170)

## Overview

Support Grafana's built-in PostgreSQL data source plugin against XTDB's pgwire server.
Users should be able to add XTDB as a PostgreSQL data source in Grafana, use the visual query builder to browse tables/columns, and run time-series and table queries in dashboards.

We target the **existing** Grafana PostgreSQL plugin — no custom plugin needed.

## Strategy

Grafana's PostgreSQL plugin issues a small, well-defined set of introspection queries against `information_schema` and a handful of PostgreSQL functions.
XTDB already has substantial `pg_catalog` and `information_schema` support (Metabase compatibility work).
The delta is primarily a few missing SQL functions and one `current_setting` value.

### Connection & Health Check

On connection, the Grafana backend (`pkg/tsdb/grafana-postgresql-datasource/sqleng`) calls `pool.Ping(ctx)` — a simple pgwire protocol-level ping.
No special SQL is executed at connection time.

### Version Detection

```sql
SELECT current_setting('server_version_num')::int/100 AS version
```

This determines PostgreSQL major version for feature gating.
XTDB supports `current_setting('server_version_num')` but returns the *XTDB* version parsed as major x 1,000,000 + minor x 1,000 + patch (e.g. XTDB `2.0.1` -> `2000001`).
Grafana divides by 100, so it would see version `20000` rather than a recognisable PostgreSQL version like `1600`.
This may need fixing to return `160000` (matching the advertised `server_version = "16"` pgwire parameter).
The `::int` cast syntax works (via `PostgresCastExpr` grammar rule).
Integer division also works.

### TimescaleDB Detection

```sql
SELECT extversion FROM pg_extension WHERE extname = 'timescaledb'
```

Queries `pg_extension` which XTDB does not implement.
This should return *empty* (XTDB is not TimescaleDB), which the plugin handles gracefully — it falls back to standard PostgreSQL time-grouping.

**Requirement**: `pg_extension` table stub returning empty results.
Currently, querying a non-existent `pg_catalog` table throws `UnsupportedOperationException` ("Information Schema table does not exist: ...") rather than returning empty.
Either add a `pg_extension` stub to the information schema, or handle this gracefully at the pgwire layer.

### Table Discovery (Visual Query Builder)

```sql
SELECT
[...]

PR #5430, opened 2026-04-07
feat(sql): allow subqueries in ORDER BY expressions

tggreene:
Grafana's table discovery query uses `ORDER BY CASE WHEN ... IN (SELECT ...) THEN 0 ELSE 1 END` to sort priority schemas first.
This was rejected with "Subqueries are not allowed in this context" because `plan-sort-specification-list` didn't pass a `!subqs` HashMap to the expression visitor.

Follows the same pattern as `wrap-where`: collect subqueries during expression planning, apply them via `apply-sqs` before the order-by node.
Subqueries in window ORDER BY (`OVER (ORDER BY ...)`) remain rejected — only query-level ORDER BY gains subquery support.

Note: correlated scalar subqueries like `ORDER BY (SELECT -t.x)` require a grammar change (subquery before WrappedExpr in exprPrimary) — deferred to a follow-up.

Relates to #5170

### DIFF
--- a/src/test/clojure/xtdb/sql_test.clj
+++ b/src/test/clojure/xtdb/sql_test.clj
@@ -1013,9 +1013,23 @@
     (t/is (= [{:y 11} {:y 21} {:y 31}]
              (xt/q tu/*node* "SELECT x + 1 AS y FROM t ORDER BY y * 2 ASC"))))
 
-  (t/testing "subquery in ORDER BY - not yet supported"
-    (t/is (thrown-with-msg? Exception #"Subqueries are not allowed in this context"
-             (xt/q tu/*node* "SELECT x + 1 AS y FROM t ORDER BY (SELECT 1)")))))
+  (t/testing "subquery in ORDER BY — constant"
+    (t/is (= #{{:y 11} {:y 21} {:y 31}}
+             (set (xt/q tu/*node* "SELECT x + 1 AS y FROM t ORDER BY (SELECT 1)")))))
+
+  (t/testing "subquery in ORDER BY — CASE WHEN ... IN (SELECT ...) pattern"
+    (xt/execute-tx tu/*node* [[:sql "INSERT INTO items (_id, category) VALUES (1, 'a'), (2, 'b'), (3, 'c')"]])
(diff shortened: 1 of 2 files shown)


src/test/clojure/xtdb/sql_test.clj
```diff
@@ -1013,9 +1015,47 @@
...
+             (xt/q tu/*node* "SELECT x, (SELECT 1) AS sq FROM t ORDER BY (SELECT 1), x"))))
+
+  (t/testing "multiple subqueries in ORDER BY"
+    (t/is (= [{:category "b"} {:category "a"} {:category "c"}]
+             (xt/q tu/*node* "SELECT category FROM items
+                              ORDER BY CASE WHEN category IN (SELECT cat FROM priority)
+                                THEN 0 ELSE 1 END,
+                              (SELECT 1)"))))
+
+  ;; correlated scalar subquery in ORDER BY (e.g. ORDER BY (SELECT -t.x))
```
mbutlerw: Would be good to get a failing test case for this, partially as documentation, but also to see that breaks gracefully and preferably with a relevant error message (if possible).

tggreene: The current behaviour in WrapperExpr is a bit surprising, it seems to parse (SELECT -t.x) as a collection of column names (the order is essentially the same as NULL col):

```
xtdb=> INSERT INTO t (_id, x) VALUES (1, 10), (2, 20), (3, 30);
INSERT 0 0
xtdb=> SELECT x + 1 AS y FROM t ORDER BY (SELECT -t.x) ASC;
WARNING:  Column not found: select
 y
----
 21
 11
 31
(3 rows)
```

It seems as much work or even simpler to just implement the correlated subquery than convert this to an error. Do you think I can bundle that together in this PR? I have a ! on that change because it would potentially affect any existing (SELECT ...) not being treated as an expression first but as a subquery, but I think that's probably reasonable behaviour?

tggreene: Thinking about it the grammar change would be "more correct" than the previous state.

mbutlerw (changes requested):
Overall LGTM.

Could we land the grammar change separately with an associated test, as I think it applies to scalar subqueries more broadly.

👍

PR #5431, opened 2026-04-07
feat(sql): support generic function calls in FROM clauses

tggreene:
Grafana's schema discovery queries use `string_to_array(...)` directly in FROM position.
XTDB only supported UNNEST and GENERATE_SERIES as table-valued functions.

Adds a generic `FunctionDerivedTable` grammar rule that wraps scalar function results as single-row derived tables — equivalent to `(VALUES (fn_call(...))) AS t(col)`.
Default column name uses the function name, WITH ORDINALITY is supported, and aggregates are rejected with a clear error.

Relates to #5170


mbutlerw:
Superseded by https://github.com/xtdb/xtdb/pull/5442

PR #5436, opened 2026-04-08
fix!(sql): disambiguate TRIM grammar to prevent greedy expr matching

### LINKED ISSUE #5409: TRIM(expr) fails when argument is a complex expression like array subscript
`TRIM(s[1])` fails to parse because the grammar greedily matches `s` as `trimCharacter`, leaving `[1]` as an unparseable `trimSource`.

```sql
-- works
TRIM(' hello ')
TRIM('x' FROM 'xhellox')

-- fails
TRIM(s[1])
```

The root cause is `trimCharacter=expr? 'FROM'? trimSource=expr` — ANTLR greedily consumes the first `expr`, so any complex single-arg expression gets split incorrectly.

Fix: split into explicit alternatives requiring `FROM` as a delimiter between trim char and source.

tggreene:
Relates to #5170

`trim(s[1])` fails with `trim not applicable to types list and list`.
The old grammar `trimCharacter=expr? 'FROM'? trimSource=expr` is ambiguous — ANTLR greedily matches `s` as `trimCharacter`, leaving `[1]` (parsed as an array literal) as `trimSource`.
Both resolve to list types instead of strings.

Simple change in the end: split into explicit alternatives (`trimChar FROM trimSource | FROM trimSource | trimSource`) so the single-arg path is unambiguous.
Grafana's table discovery query uses `trim(s[i])` to strip whitespace from search_path elements.

### DIFF
--- a/core/src/main/antlr/xtdb/antlr/Sql.g4
+++ b/core/src/main/antlr/xtdb/antlr/Sql.g4
@@ -324,7 +324,7 @@ exprPrimary
          | ',' startPosition=expr (',' stringLength=expr)? )
       ')' # CharacterSubstringFunction
 
-    | 'TRIM' '(' trimSpecification? trimCharacter=expr? 'FROM'? trimSource=expr ')' # TrimFunction
+    | 'TRIM' '(' trimSpecification? (trimCharacter=expr 'FROM' trimSource=expr | 'FROM' trimSource=expr | trimSource=expr) ')' # TrimFunction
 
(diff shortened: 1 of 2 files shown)


mbutlerw (approved):
Technically a breaking change, although not tested/documented, `trim(expr expr)` will no longer be valid. e.g. `"SELECT TRIM('$' '$hello$')"` 

I think this is fine, as its not SQL Standard or PSQL compliant anyway, but we should note it just in case.

PR #5437, opened 2026-04-08
feat(sql): implicit utf8-to-temporal coercion in comparisons

tggreene:
Relates to #5170

Grafana's \`\$__timeFilter\` macro expands to bare ISO strings in temporal comparisons:
\`recorded_at BETWEEN '2026-01-01T00:00:00Z' AND '2026-01-02T00:00:00Z'\`

This is NOT strictly required for grafana compatibility (i.e. table/column discovery).

PG implicitly casts these to the temporal type of the other operand.
Without this, XTDB rejects the comparison as a type mismatch.

Registers codegen-call methods for each comparison operator and temporal type pair (timestamp-tz, timestamp-local, date, time-local) in both operand orderings, casting the utf8 side to match via \`recall-with-cast2\`.
Malformed strings produce clear parse errors at runtime ("could not be parsed" / "invalid format for type").


mbutlerw:
Closing in favour of https://github.com/xtdb/xtdb/pull/5444

PR #5444, opened 2026-04-09
feat(sql): coerce string literals to temporal types in comparisons

mbutlerw:
Bare string literals like '2026-01-01' are now resolved to the temporal type of the other operand in comparison expressions. This matches how DuckDB (STRING_LITERAL) and CockroachDB (StrVal) treat untyped string constants — the literal adapts to context rather than being committed to VARCHAR/utf8.

The coercion lives in codegen-call* rather than as operator-specific codegen-call multimethod methods — multimethod dispatch is committed once matched on types, so non-literal utf8 values would also match with no way to fall through to default dispatch (which produces the type error). codegen-call* lets us try 
  the coercion and fall back naturally.

=== is excluded because UPDATE uses it for change detection on polymorphic columns where utf8 and temporal can coexist — coercing there would mask real type differences.

Gated to user-facing comparison ops only (<, <=, >, >=, =, <>) — predicates return booleans so the coercion can't change what gets stored, unlike COALESCE or GREATEST where it would silently alter the result type. Internal operators (compare, compare_nulls_first, compare_nulls_last, null_eq) are excluded —     
  their operands are always column refs from sorting/grouping, never literals.

Primary motivation is Grafana compatibility: its $__timeFilter macro expands to bare ISO strings in temporal comparisons.                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                          
  Relates to #5170

### DIFF
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -680,6 +680,38 @@
                           ~(f #xt/type :null nil)
                           ~(f #xt/type :bool `(not (== -1 ~l-sym ~r-sym))))))))}))
 
+(def ^:private comparison-ops #{:< :<= :> :>= :== :<>})
+
+(defn- try-coerce-utf8-literal-to-temporal
+  "When a utf8 string literal is compared against a date-time type, coerces the
+  literal to the temporal type before dispatch — so operators only see temporal+temporal."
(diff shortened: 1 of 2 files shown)


tggreene (approved):
Looks good, I hadn't considered the case of columns matching in my version this is much better.

jarohen (approved):

PR #5456, opened 2026-04-13
feat(debezium): add DebeziumEngineSource for direct Postgres CDC

jarohen:
Adds `!DebeziumEngine` as a sibling to `!DebeziumKafka` on `ExternalSource.Factory`.
Reads Postgres CDC directly via the embedded Debezium engine and a logical replication slot — no Kafka between XTDB and the source DB.

## Why

XTDB is bitemporal — ordering is load-bearing for the sink, not a nice-to-have.
The previous `!DebeziumKafka` path consumes CDC that's already been fanned out across Kafka topics and partitions, and that architecture is fundamentally at odds with meaningful replay semantics in two distinct ways.

**Snapshot replay has no natural ordering.**
Initial-snapshot events all share a single LSN (the `pg_export_snapshot` position) and fan out across table topics, with each topic partitioned by PK.
Across partitions Kafka gives no ordering guarantee; across topics there's none at all.
Per-key ordering holds by construction — Debezium keys by PK — but that's the ceiling.
The best a consumer can do is either (a) buffer the entire snapshot before applying and sort by some canonical key, infeasible at production scale, or (b) apply events as they arrive in arbitrary order and rely on per-key convergence.
`!DebeziumKafka` does (b).
For a sink that stamps each incoming event with a `system_time`, the result is arbitrary valid-time ordering for facts that are conceptually "existing state as of LSN X".
Streaming events after the snapshot then add deltas on top of that arbitrary baseline.

**Source-txn atomicity is expensive to reassemble.**
A single source txn spreads across partitions and tables on the Kafka side.
Reassembling it requires correlating data events with `BEGIN`/`END` markers from the transaction-metadata topic, counting events per table from the `END` envelope's `data_collections`.
The mechanism is internally consistent but the costs compound: throughput collapses to "slowest partition × txn size", atomicity-across-tables requires consuming every topic that could participate, and any partition lag stalls every open txn touching that partition.
For the many-small-transactions workload we care about, the cost dominates the benefit.

**Embedded sidesteps both.**
The Debezium engine delivers events in source commit order, in a single in-process stream, with snapshot events arriving as a contiguous block followed by streaming.
One source txn maps cleanly to one XTDB txn, boundary-detected by `source.txId` change.
Snapshot events are batched into fixed-size XTDB txns — still arbitrary within a chunk, inherent to snapshot scan order, but the chunks themselves are serial with a clean switchover to streaming.

## Shape

Config is polymorphic so future connectors (MySQL, etc.) slot in without reshaping the outer type:

```yaml
externalSource: !DebeziumEngine
  backend: !Postgres
    hostname: ...
    publicationName: xtdb_pub       # created first on source
    slotName: xtdb_slot             # our consumer's handle
    tableIncludeList: public.orders,public.customers
  snapshotChunkSize: 1000
  offsetFlushIntervalMs: 1000
```

Engine-level config (`snapshotChunkSize`, `offsetFlushIntervalMs`) lives on the outer `Factory`.
Connector-specific config lives on `backend` — a sealed interface with `Postgres` as its one variant.
Each variant knows how to stamp its own properties onto Debezium's `Properties`.

## Exactly-once-at-sink

Two mechanisms combined:

1. `handleTx(openTx, resumeToken)` commits the XTDB tx and the resume token atomically via the existing `ExternalSource` framework.
2. Per-event `source.lsn` compared against `last_applied_lsn` in the stored token filters replays on restart.

The `offset.flush.interval.ms=1000` setting keeps Debezium's slot acknowledgement near-real-time, bounding the replay window without making it strictly zero — dedup is the safety net.

## Snapshot handling

`snapshot.mode=initial`: snapshot once, stream forever, fail loudly on slot loss.
This deliberately rejects `when_needed` — silently re-snapshotting would produce a valid-time gap in XTDB's bitemporal history for the outage window, which we'd rather surface as an operator-visible failure than absorb.

Initial snapshot events share an LSN so they can't be LSN-deduped.
Debezium doesn't resume interrupted initial snapshots anyway — mid-snapshot crash restarts the scan from the beginning and idempotent re-application handles the overlap.

## The OffsetBackingStore dance

Debezium 3.2.2's `OffsetBackingStore` is instantiated reflectively (`offset.storage=<classname>` + no-arg ctor); no `using(OffsetBackingStore)` builder overload exists.
But we need to inject state from the resumeToken at startup and read Debezium's writes back for the next resumeToken.

Bridge: a static registry keyed by a generated UUID passed through Debezium's config map.
Engine instantiates our class, `configure()` pulls seed state and registers the live handle, callers reach it via `handleFor(id)`.
In-memory only; durability comes from the framework's atomic token commit.

## Behaviour-preserving tidy on the way in

The first commit (`tidy(debezium): extract CDC payload helpers`) moves `writeCdcPayload`, `parseValidTimeMicros`, `parseCdcEnvelope`, and `jsonToJvmValue` out of `DebeziumKafkaSource` and `MessageFormat` into a shared `CdcPayload.kt`.
Pure refactor — both sources receive the same Debezium envelope shape and the same apply path, better to have one copy than three.
Reviewable independently.

## Debezium version bump

3.2.2.Final (was 3.0.2.Final for the Kafka-based integration test).
Earlier 3.0.x throws `NoSuchMethodError` on `SourceTask.commitRecord(SourceRecord)` against Kafka 4.1.x — that overload was removed.
3.2 is the first Debezium line compiled against Kafka 4.

## Known follow-ups

- **MySQL generalisation not done.**
  `ChangeConsumer.classify` extracts `source.lsn` directly (MySQL has no LSN — it's `(file, pos)` or GTID), `writeCdcPayload` routes via `source.schema` (MySQL uses `db`), and `last_applied_lsn` in the proto is a `uint64`.
  The connector-agnostic parts (engine lifecycle, offset store, batch-processing skeleton, envelope parsing) are already ~60–70% of the code and will factor out cleanly when a second concrete backend lands.
  YAGNI until then.
- **No user-facing docs yet** (`dev/doc/` or equivalent).
  Should add before this is pitched as production-ready.
- **Password is a plain String on the Factory**, not routed through `StringWithEnvVarSerde`.
  Existing `!DebeziumKafka` doesn't face this because credentials flow through the referenced Kafka cluster config; here they're direct.
  Worth wiring before exposing this to users.
- **Single-partition only** per `ExternalSource.onPartitionAssigned`, matching `DebeziumKafkaSource`.
  Not a real limitation for Postgres — one replication slot is a singleton source stream anyway.

## Test plan

- [x] Unit tests: offset store bootstrap/snapshot/overwrite/isolation + envelope parser (31 passing)
- [x] Integration tests against TestContainers Postgres 17: snapshot+streaming flow, resume-from-token dedup (2 passing in 16s)
- [x] No reflection / boxed math warnings
- [ ] Manual smoke test against a non-trivial production-like schema
- [ ] Longer-soak test to validate snapshot chunk boundaries and slot advancement under load

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -10,6 +10,7 @@ exposed = "0.61.0"
 google-cloud-storage = "2.60.0"
 guava = "33.4.7-jre"
 confluent = "7.8.0"
+debezium = "3.2.2.Final"
 kafka = "4.1.1"
 kotest = "6.1.11"
 kotlin = "2.3.0"
@@ -152,7 +153,13 @@ guava = { group = "com.google.guava", name = "guava", version.ref = "guava" }
 # Kafka
 kafka-clients = { group = "org.apache.kafka", name = "kafka-clients", version.ref = "kafka" }
 kafka-connect-api = { group = "org.apache.kafka", name = "connect-api", version.ref = "kafka" }
+kafka-connect-runtime = { group = "org.apache.kafka", name = "connect-runtime", version.ref = "kafka" }
 kafka-avro-serializer = { group = "io.confluent", name = "kafka-avro-serializer", version.ref = "confluent" }
+
+# Debezium
+debezium-api = { group = "io.debezium", name = "debezium-api", version.ref = "debezium" }
+debezium-embedded = { group = "io.debezium", name = "debezium-embedded", version.ref = "debezium" }
+debezium-connector-postgres = { group = "io.debezium", name = "debezium-connector-postgres", version.ref = "debezium" }
 
 # Dev/test dependencies
 jovial = { group = "dev.clojurephant", name = "jovial", version = "0.4.3" }
--- a/modules/debezium/build.gradle.kts
+++ b/modules/debezium/build.gradle.kts
@@ -19,6 +19,12 @@ dependencies {
 
     implementation(libs.kafka.avro.serializer)
 
+    implementation(libs.debezium.api)
+    implementation(libs.debezium.embedded)
+    implementation(libs.debezium.connector.postgres)
+    implementation(libs.kafka.connect.api)
+    implementation(libs.kafka.connect.runtime)
+
     testImplementation(kotlin("test"))
     testImplementation(libs.kotlinx.coroutines.test)
     testImplementation(libs.testcontainers)
--- a/modules/debezium/src/main/kotlin/xtdb/debezium/CdcPayload.kt
+++ b/modules/debezium/src/main/kotlin/xtdb/debezium/CdcPayload.kt
@@ -1,0 +1,114 @@
+package xtdb.debezium
+
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonNull
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.booleanOrNull
+import kotlinx.serialization.json.doubleOrNull
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.longOrNull
+import xtdb.error.Incorrect
+import xtdb.indexer.OpenTx
+import xtdb.table.TableRef
+import xtdb.time.InstantUtil.asMicros
+import xtdb.util.asIid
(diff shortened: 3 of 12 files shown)


jarohen:
@Akeboshiwind starting fresh with a non-yolo impl :slightly_smiling_face:

PR #5459, opened 2026-04-14
feat(kafka): handle ProducerFencedException in single-writer mode

### LINKED ISSUE #5446: Ensure producer-fenced errors are handled correctly in single-writer mode
## Context

In single-writer mode, the leader opens an atomic Kafka producer with a stable `transactional.id` (e.g. `"${dbName}-leader"`) to write to the replica log.
Kafka's transactional semantics mean that if a new producer initialises with the same `transactional.id`, the old producer is **fenced** — any subsequent send from the old producer raises a `ProducerFencedException`.

This can happen in practice during:
- Rolling deployments where the new instance starts before the old one fully shuts down
- Leader failover / consumer group rebalance races
- Network partitions that cause Kafka to see a "zombie" producer

## Current state

- `KafkaCluster.kt` creates idempotent, transactional producers with `enable.idempotence=true` and `acks=all`
- `LogProcessor.kt` opens atomic producers with `"${dbState.name}-leader"` as the transactional ID
- `BlockUploader.kt` and `LeaderLogProcessor.kt` use `replicaProducer.withTx()` for atomic writes
- There is **no explicit handling** of `ProducerFencedException` or `OutOfOrderSequenceException`
- An unhandled fencing exception would likely propagate as an unrecoverable error, potentially crashing the node without clean shutdown

## Acceptance criteria

- [ ] Audit all paths where the atomic/transactional producer is used (`LeaderLogProcessor`, `BlockUploader`, `LogProcessor`)
- [ ] Decide on the correct recovery behaviour when a producer is fenced (e.g. graceful shutdown, step down to follower, retry with new epoch)
- [ ] Implement explicit catch-and-handle for `ProducerFencedException` (and related `InvalidProducerEpochException`, `OutOfOrderSequenceException`)
- [ ] Add test coverage for the fencing scenario
- [ ] Document the expected behaviour in the Allium spec if applicable

## Progress

**Status**: completed

- [x] Audit all paths where the atomic/transactional producer is used
- [x] Decide on correct recovery behaviour when a producer is fenced
- [x] Implement explicit catch-and-handle for ProducerFencedException and related exceptions
- [x] Add test coverage for the fencing scenario
- [x] Document expected behaviour in Allium spec if applicable

### Known Gap

- [ ] No test that `FencingAwareProcessor` converts fencing to `StepDownException` (InMemoryLog's subscription loop doesn't catch it; real Kafka test has timing issues)

mbutlerw:
## Summary

In single-writer mode, Kafka's transactional producer can be fenced when another node opens a producer with the same `transactional.id` during a consumer group rebalance, or when a transaction times out on the broker.
Previously this propagated unhandled and crashed the node.

A fenced leader now throws `StepDownException`, which the subscription loop catches — nulling out the processor and continuing to poll.
The in-progress rebalance fires `onPartitionsRevoked` on the next `poll()`, which handles the actual follower transition via the existing path.

Fencing during leader transition (`onPartitionsAssigned`) is handled separately — the exception fires inside `poll()` itself, so the subscription loop can't catch it.
The transition catch reverts to follower using the old follower's watermarks.

## Key decisions

**Step down, don't crash.**
The fenced producer is dead regardless, but the node can continue as a follower serving reads and be re-elected via the next rebalance.

**`StepDownException` as a signal, not self-managed step-down.**
Originally `FencingAwareProcessor` managed the transition itself (closing the leader, opening a follower, absorbing records).
Simplified to just throwing `StepDownException` — the subscription loop catches it, and the existing `onPartitionsRevoked` path handles the real transition.
This avoids duplicating step-down logic.

**No `enforceRebalance`.**
Fencing only occurs when a rebalance is already in progress (another node opened a producer after being assigned the partition).
The in-progress rebalance handles reassignment — forcing another is redundant.

**`LeaderLogProcessor` skips `watchers.notifyError` for fencing.**
Without this, clients waiting on transaction confirmations would be poisoned during a graceful recovery.

## Also fixes

`LeaderSystem.close()` now closes the `replicaProducer` — previously leaked on every `onPartitionsRevoked`.

## Test gap

No test that `FencingAwareProcessor` converts fencing to `StepDownException`.
`InMemoryLog`'s subscription loop doesn't catch `StepDownException` (only `KafkaCluster`'s does), and a real Kafka integration test has no clean observable signal for "the exception was caught" without timing hacks.
Covered by code review for now.

Resolves #5446

### DIFF
--- a/core/src/main/kotlin/xtdb/api/log/Log.kt
+++ b/core/src/main/kotlin/xtdb/api/log/Log.kt
@@ -139,6 +139,12 @@ interface Log<M> : AutoCloseable {
     interface AtomicProducer<M> : AutoCloseable {
         fun openTx(): Tx<M>
 
+        /**
+         * Returns true if the given exception (or any in its cause chain) indicates
+         * that this producer has been fenced by a newer producer with the same transactional.id.
+         */
+        fun isProducerFenced(e: Throwable): Boolean = false
+
         interface Tx<M> : AutoCloseable {
             fun appendMessage(message: M): CompletableDeferred<MessageMetadata>
             fun commit()
--- a/core/src/main/kotlin/xtdb/api/log/StepDownException.kt
+++ b/core/src/main/kotlin/xtdb/api/log/StepDownException.kt
@@ -1,0 +1,8 @@
+package xtdb.api.log
+
+/**
+ * Thrown by a [Log.RecordProcessor] to signal that the processor has been fenced
+ * and the subscription should continue polling (to receive rebalance callbacks)
+ * rather than terminating.
+ */
+class StepDownException(cause: Throwable) : Exception("Processor stepped down", cause)
--- a/core/src/main/kotlin/xtdb/database/Database.kt
+++ b/core/src/main/kotlin/xtdb/database/Database.kt
@@ -228,12 +228,14 @@ class Database(
                                 val demux = ExternalSource.Demux(leaderProc, extSource, 0, afterToken, txHandler)
                                 object : LogProcessor.LeaderSystem {
                                     override val proc get() = demux
-                                    override fun close() { demux.close(); extSource.close(); leaderProc.close() }
+                                    override fun isProducerFenced(e: Throwable) = replicaProducer.isProducerFenced(e)
+                                    override fun close() { demux.close(); extSource.close(); leaderProc.close(); replicaProducer.close() }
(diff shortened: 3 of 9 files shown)


mbutlerw:
See https://github.com/xtdb/xtdb/issues/5446#issuecomment-4252512676 for more details.

PR #5464, opened 2026-04-16
feat: Introducing `listAfter` to object store implementations, increase speed of liveness checks.

### LINKED ISSUE #5460: Liveness checks can time out when GC is disabled due to expensive listAllObjects on blocks
## Problem

The `/healthz/alive` liveness check computes block lag for each database via `->block-lag`, which calls `BufferPoolKt/getLatestAvailableBlockIndex`.

This property (`BufferPool.kt:101-108`) works by calling `listAllObjects("blocks")` and taking the last entry — listing **every** block object in storage just to find the latest one.

With GC turned off, the number of block files grows without bound. On remote storage (S3/Azure/GCS), this list operation has to paginate through all objects and gets progressively slower. The result is cached in a Caffeine cache with a 1-minute TTL, so after each expiry the full listing is repeated.

This can cause the liveness probe to exceed its timeout, leading k8s to restart the pod — which then has to do the same expensive listing again on startup, potentially entering a crash loop.

## Relevant code

- `core/src/main/kotlin/xtdb/storage/BufferPool.kt` lines 98-108 — `latestAvailableBlockIndex` and its 1-minute cache
- `core/src/main/clojure/xtdb/healthz.clj` line 29 — `->block-lag` calls `getLatestAvailableBlockIndex`

## Possible approaches

- Avoid the full listing: derive the latest available block index from the block catalog or another source that already tracks it, rather than listing all objects in storage.
- Use a narrowed listing (e.g. list with a starting-after marker based on a known recent block index) to avoid paginating from the beginning.
- Increase the cache TTL as a short-term mitigation (currently 1 minute).

danmason:
Resolves #5460 
Github actions: https://github.com/danmason/xtdb/actions/workflows/build.yml?query=branch%3Afix%2Ffaster-liveness-checks

## Background 
On XTDB deployments with a large amount of block files - we're seeing instances of liveness checks timing out after a number of seconds - primarily because the main mechanism for judging liveness (aside from checking for ingestion stopped) is to fetch the latest block index and compare against the current block index:
- We currently do this by doing a listAllObjects call on the 'blocks' directory and then filtering for the last value (as listAllObjects returns values sorted lexicographically).
- We _do_ currently cache this value for a minute - so not _all_ liveness checks will timeout, but it is still a potentially very expensive and slow operation to do every minute.

## Summary

This PR addresses this problem by drastically cutting down the number of blocks we need to list to fetch the latest block index:
- We introduce "listAfter" on the BufferPool/ObjectStore implementations:
  - This takes a given path and returns all objects from storage that are lexicographically AFTER the given path.
  - We also add tests against this within the object store tests, so we can test it against the various object store implementations.
- In order to implement the above against azure - we had to bump the current version of `azure-storage`.
- We then replace the current "latestAvailableBlockIndex" implementation within one that takes a current block index:
  - If this is null, we fall back to fetching ALL objects and returning the last value, as before.
  - Otherwise, we make use of listAfter() to fetch all blocks after the given block.
- Reduce the duration of the cache to ten seconds - keeps block date more up to date, though prevents the endpoint being spammed/making many synchronous calls to listObjects.
- In healthz, we then pass the current block index (from the BlockCatalog) into the latestAvailableBlockIndex method, so we only search for any blocks after the given current block index.


core/src/main/kotlin/xtdb/storage/BufferPool.kt
```diff
@@ -95,17 +96,16 @@ private fun Path.parseBlockIndex(): Long? =
         ?.groups?.get(1)
         ?.value?.fromLexHex
 
-private val latestAvailableBlockCache =
```
danmason, line 98: Opted to drop the caffeine cache here - this DOES potentially mean more listObjects calls, though I reason that these should be relatively cheap given the minimal amount of stuff being returned? (And make the check more responsive to block lag going above 5). Happy for counter arguments/adding this back in, however.

src/test/clojure/xtdb/gcp_test.clj
```diff
@@ -73,6 +73,16 @@
                                              :prefix (str "xtdb.gcp-test." prefix)}]}]
     :disk-cache {:path local-disk-cache}}))
 
+(t/deftest ^:google-cloud list-after-test
+  (let [os (object-store (random-uuid))]
```
mbutlerw: missing `with-open`

modules/aws/src/main/kotlin/xtdb/aws/S3.kt
```diff
@@ -218,6 +218,30 @@ class S3(
...
+    private fun listAfter0(listPrefix: Path, afterKey: Path) =
+        sequence {
+            var continuationToken: String? = null
+
+            while (true) {
+                val listResp = runBlocking {
+                    client.listObjectsV2 {
+                        it.bucket(bucket)
+                        it.prefix("$listPrefix/")
+                        it.startAfter(prefix.resolve(afterKey).normalize().toString())
```
mbutlerw, line 230: Could probably roll this and listAllObjects0 above into a single internal fn with startAfter conditionally applied.

core/src/main/kotlin/xtdb/storage/BufferPool.kt
mbutlerw: Given the simplicity the cache, I'd be inclined to keep it with a short lifespan (10s), not because of any $ or load concerns, but because in its current form any request to this endpoint kicks off an S3 call which is probably best not exposed/coupled.

mbutlerw (changes requested):

modules/aws/src/main/kotlin/xtdb/aws/S3.kt
danmason: This is perfect for a bit of cleanup - can run through that after this is on main.

PR #5469, opened 2026-04-17
fix: duplicate scan columns when requesting `_valid_time`/`_system_time`

egg-juxt:
Fix respects the prior order, by adding needed columns to the end (otherwise it could have been simpler code).

(I also included an essential re-indent of involved files which was breaking parinfer).


mbutlerw (approved):
LGTM :shipit: 

Side note, XTDB considers the columns in relations to be unordered semantically, therefore the simpler fix IMO would have been acceptable.

PR #5471, opened 2026-04-17
fix: skip attach/detach-db processing on read-only secondaries

Akeboshiwind:
## Summary

- Read-only secondaries watching another node's primary log NPE on `AttachDatabase`/`DetachDatabase` messages because `dbCatalog` is only present on the primary database's `SourceLogProcessor`
- Guard the catalog mutation behind `!readOnly` — read-only processors record the message as a committed no-op tx (keeping tx-ids aligned) without attempting the side-effect
- Added regression test: two nodes sharing a log, where the primary attaches a secondary database and the read-only follower processes the message without error

## Test plan

- [x] New test `secondary-node-handles-attach-db-in-shared-log` in `multi_db_test.clj` — verified it fails before fix (NPE) and passes after


Akeboshiwind:
Recreating with a feature branch instead of main

PR #5513, opened 2026-04-27
fix(indexer): block (not suspend) during parallel table-block uploads

mbutlerw:
`coroutineScope { launch(blockUploadDispatcher) }` *suspends* the calling coroutine while IO children run, then the IO threads dispatch the continuation back to the caller's dispatcher. In the sim tests, that dispatcher is the single-threaded `DeterministicDispatcher`, which is not thread-safe — the cross-thread callback can be lost, hanging the test.

`runBlocking { async(blockUploadDispatcher) }` achieves the same IO parallelism but *blocks* the calling thread instead of suspending.
The `DeterministicDispatcher`'s drain loop simply pauses until all uploads finish — no cross-thread dispatch needed.
This matches the existing pattern in `LiveTable.finishBlock`.

Every call site is a sequential record-processing pipeline where the next record can't start until `finishBlock` returns, so the thread has nothing else to do during the wait — no practical cost from blocking vs suspending.

### DIFF
--- a/core/src/main/kotlin/xtdb/indexer/BlockUploader.kt
+++ b/core/src/main/kotlin/xtdb/indexer/BlockUploader.kt
@@ -6,8 +6,9 @@ import java.nio.ByteBuffer
 import java.time.Instant
 import kotlinx.coroutines.Dispatchers.IO
 import kotlinx.coroutines.ExperimentalCoroutinesApi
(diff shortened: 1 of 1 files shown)


mbutlerw:
Closed in favour of https://github.com/xtdb/xtdb/pull/5515

PR #5519, opened 2026-04-28
ops: support arbitrary logger names via XTDB_LOGGER_LEVEL_*

egg-juxt:
Replaces XTDB_LOGGING_LEVEL_* (kept for backwards compatibility), which enforced the `xtdb` logger-prefix, so no third-party loggers allowed (Kafka, Netty, AWS SDK...)

The new XTDB_LOGGER_LEVEL_* is case-sensitive, allows for both Clojure namespaces with `-` and Kotlin packages with `_` separators.

Examples:
```
  XTDB_LOGGER_LEVEL_xtdb_garbage__collector: debug
  XTDB_LOGGER_LEVEL_software_amazon_awssdk: debug
```

### DIFF
--- a/main/src/main/clojure/xtdb/logging.clj
+++ b/main/src/main/clojure/xtdb/logging.clj
@@ -33,14 +33,27 @@
 (defmacro with-log-level [ns level & body]
   `(with-log-levels {~ns ~level} ~@body))
 
-(defn- env-var->log-ns [env-var]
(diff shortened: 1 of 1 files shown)


egg-juxt:
If this is introduced, we should probably just replace the existing option XTDB_LOGGING_LEVEL, instead of introducing a new one.

Breaking backwards compatibilty for such an option shouldn't be an issue, as it's going to be used for troubleshooting only.

PR #5526, opened 2026-04-29
feat(adbc): in-process AdbcStatement prepare/bind/execute (#5270)

### LINKED ISSUE #5270: ADBC: implement AdbcStatement.prepare() on XtdbConnection
## Context

`AdbcStatement.prepare()` currently throws `TODO(\"Not yet implemented\")` on the in-process ADBC path.
Every ADBC tutorial and client library expects this to work — it's how parameterized queries are used through ADBC.

The FlightSQL path has its own prepared statement machinery (`XtdbProducer.createPreparedStatement`, the `PreparedStatement` class, `stmts` map) which works, but it duplicates concerns that should live in XtdbConnection.

## Goal

Implement `prepare()` on `XtdbConnection`'s statement, and refactor FlightSQL's `XtdbProducer` to delegate to it.

This continues the pattern established in the metadata endpoints work: XtdbConnection is the canonical server-side API, FlightSQL delegates to it.

## Current state

- `XtdbConnection.createStatement().prepare()` → `TODO`
- `XtdbConnection.prepareSql(sql)` → works, returns `PreparedQuery` (used by FlightSQL)
- `XtdbProducer` has its own `PreparedStatement` class wrapping `PreparedQuery` + cursor + txHandle + params
- FlightSQL prepared statement flow: `createPreparedStatement` → `acceptPutPreparedStatementQuery` (bind params) → `getFlightInfoPreparedStatement` / `getStreamPreparedStatement` (execute)

## Approach

`prepare()` should set up the `PreparedQuery` on the statement, so that subsequent `executeQuery()`/`executeUpdate()` use it with bound parameters.

The FlightSQL `PreparedStatement` class and its lifecycle (`stmts` map, handles) are FlightSQL-protocol concerns (ticket management, gRPC streaming) — those stay in `XtdbProducer`. But the core prepare→bind→execute logic should move to XtdbConnection.

## Acceptance criteria

- `AdbcStatement.prepare()` works for both queries and DML
- Parameter binding via `AdbcStatement.bind()` works with prepared statements
- FlightSQL prepared statement path still works (delegate, don't break)
- Tests for in-process ADBC prepared statement flow

## Progress

**Status**: in-progress

- [ ] Implement `prepare()` on XtdbConnection's AdbcStatement (queries + DML)
- [ ] Wire parameter binding via `AdbcStatement.bind()` for prepared statements
- [ ] Move shared prepare→bind→execute logic out of XtdbProducer; FlightSQL delegates
- [ ] Tests for in-process ADBC prepared statement flow
- [ ] Verify FlightSQL prepared statement path still works

tggreene:
## Summary

Implements the `AdbcStatement` prepare→bind→execute lifecycle on the in-process ADBC path (`XtdbConnection`).
A JVM caller using `xtdb.connect()` can now drive the canonical ADBC tutorial shape — `setSqlQuery → prepare → bind → executeQuery|executeUpdate` — for both queries and DML.
Previously every step except `setSqlQuery` either threw `TODO("Not yet implemented")` or returned `NOT_IMPLEMENTED`.

This is the first half of #5270.
The second half — folding FlightSQL's `XtdbProducer.PreparedStatement` onto this same implementation so FSQL becomes a thin protocol shim — is a follow-up PR (#5554).

## Changes

Commits, smallest viable slice first:

1. `feat(adbc): implement AdbcStatement.prepare on XtdbConnection` — the no-param query case; smallest atomic step that closes the TODO and gives us a regression net.
2. `feat(adbc): support bound parameters` — `bind(VSR)` and `bind(RelationReader)`.
3. `feat(adbc): implement getParameterSchema and rename bound params to planner convention` — discoverability, plus aligns the in-process path with the FlightSQL `$N` external naming.
4. `feat(adbc): support bound parameters in DML` — `TxOp.SqlBytes` plumbing so parameterised `INSERT/UPDATE/DELETE` actually carries its args.
5. `test(adbc): use XT relations in in-process binds` — test rewrite using `xtdb.arrow.Relation` end-to-end, per review feedback ("convert to VSR as late as possible / from VSR as early as possible").
6. `refactor(adbc): use openSlice / openDirectSlice for bind and per-execution args` — addresses Matt's review: re-execution becomes zero-copy on the args path (`openSlice` is a ref-count bump when source and dest share an allocator root).
7. `fix(adbc): bind requires prepare for executeQuery, restore cross-root copy in bind` — addresses the silent-drop-of-bind footgun spotted in review (`bind` without `prepare` now throws); restores the row-by-row copy in `bind` because `openDirectSlice` is only valid same-root, and the ADBC contract doesn't let us dictate the caller's allocator.

## Implementation notes

**Args stored as an `xtdb.arrow.Relation` owned by `node.allocator`.**
`bind(VSR)` delegates to `bind(RelationReader)` (per James's review).
`bind(RelationReader)` row-copies the caller's reader into a fresh `Relation` allocated against `node.allocator`.
This is cross-root safe — Arrow's `transferOwnership` (the cheaper path) only works when the caller's allocator shares a root with ours, and the ADBC contract doesn't let us dictate that.
The row copy is the one-shot cost paid per `bind()`.

**Per-execute args path is zero-copy.**
`openQueryArgs` (called inside `executeQuery` / `executeUpdate`) does `args.openSlice(node.allocator)` — same-root by construction, so just bumps ref counts on the underlying buffers.
Re-executing the same prepared plan with the same bind doesn't copy data.

**`$N` parameter naming.** External clients see columns named `$0, $1, ...` in the parameter schema; internally the planner expects `?_0, ?_1, ...`.
`bind` renames by position.
This matches what `XtdbProducer` already does over the FlightSQL wire.
The 0-indexed `$N` is an XTDB choice, not Postgres-faithful — comment at the rename site flags this so a future reader doesn't try to "fix" it.

**Bind without prepare throws on `executeQuery`.**
`node.openSqlQuery` has no `args` parameter — silently falling through would drop the bind and return a result that ignored the caller's parameters.
We throw `IllegalStateException` upfront when args are bound but `prepared` is null, so the failure mode is loud and immediate rather than a wrong result.

**Sticky bind state.** Per the ADBC contract, parameters are bound until either a new `bind()` call is made or `close()` is called.
Two tests lock this in: re-execute with the same bind returns identical results, and `setSqlQuery` clears prior bind so stale args from an earlier SQL don't leak into a new prepare/execute.

**Ownership on the execute path.** `PreparedQuery.openQuery(args, opts)` takes ownership of `args` on success — the cursor wraps them in its closeables list (`xtdb/query.clj`'s `wrap-closeables`).
We only close `queryArgs` ourselves on the throw path.

## Usage

```kotlin
xtdb.connect().use { conn ->
    conn.createStatement().use { stmt ->
        stmt.setSqlQuery("INSERT INTO users (_id, name) VALUES (?, ?)")
        stmt.prepare()

        VectorSchemaRoot.create(stmt.parameterSchema, allocator).use { params ->
            // populate "$0" with id, "$1" with name
            stmt.bind(params)
            stmt.executeUpdate()
        }
    }
}
```

## Out of scope

- **FlightSQL refactor**: `XtdbProducer.PreparedStatement` still has its own parallel copy of prepare/bind/execute.
  Follow-up PR #5554 folds it onto `XtdbStatement` so the in-process and over-the-wire paths share one implementation.
- **Parameter type inference**: `getParameterSchema()` returns the `Null` placeholder type pre-bind; concrete inferred types would need `PreparedQuery.getColumnFields(paramFields)` integration.
- **`affectedRows`**: `executeUpdate` returns `-1`, not the actual row count.
- **Multi-row "execute many"**: bind state holds a single batch; ADBC supports binding N rows for N executions of the same plan, but XTDB's `openQuery` doesn't surface that today.

## Test plan

- New `xtdb.InProcessAdbcTest` (in-process surface, exercising `XtdbConnection.createStatement()` directly via `xtdb.connect()`):
  - `prepare and execute simple query`
  - `prepare with bound parameter`
  - `getParameterSchema reports columns in $N convention`
  - `prepared DML with bound parameters`
  - `bind state is sticky across executes` — re-execute returns identical results without re-binding
  - `setSqlQuery clears prior bind` — stale args from previous SQL don't leak
  - `executeQuery without prepare fails loud when parameters are bound` — locks in the no-silent-drop contract
- `xtdb.api.AdbcTest` (over-the-wire FlightSQL coverage) — all green.
- Full root suite (`./gradlew :test`) — green except `xtdb.read-only-node-test/read-only-node-updates-table-catalog-on-block-flush`, which reproduces on `main` and is unrelated to this branch.
- No reflection or boxed-math warnings.
- ktlint clean on changed files.

### DIFF
--- a/core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
+++ b/core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
@@ -28,6 +28,7 @@ import xtdb.arrow.unsupported
 import xtdb.database.DatabaseName
 import xtdb.table.TableRef
 import xtdb.query.PreparedQuery
+import xtdb.query.QueryOpts
 import xtdb.tx.TxOp
 import xtdb.tx.TxOpts
 import xtdb.util.useAll
@@ -44,16 +45,23 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
 
     override fun createStatement() = object : XtdbStatement {
         private var sql: String? = null
+        private var prepared: PreparedQuery? = null
 
         override fun setSqlQuery(sql: String) {
             this.sql = sql
+            this.prepared = null
+        }
+
+        override fun prepare() {
+            val sql = this.sql ?: error("SQL query not set")
+            prepared = node.prepareSql(sql, dbName)
         }
 
         override fun executeQuery(): QueryResult {
-            val sql = this.sql ?: error("SQL query not set")
-            val cursor = node.openSqlQuery(sql, dbName)
+            val cursor = prepared
+                ?.openQuery(null, QueryOpts())
+                ?: node.openSqlQuery(sql ?: error("SQL query not set"), dbName)
             val schema = Schema(cursor.resultTypes.map { (name, type) -> type.toField(name) })
-
             return QueryResult(-1, cursorToArrowReader(cursor, schema))
         }
 
@@ -62,9 +70,9 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
             return UpdateResult(-1)
         }
 
-        override fun prepare() = TODO("Not yet implemented")
-
-        override fun close() = Unit
+        override fun close() {
+            prepared = null
+        }
     }
 
     internal fun executeDml(op: TxOp) {
--- a/core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
+++ b/core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
@@ -14,13 +14,11 @@ import org.apache.arrow.vector.VarCharVector
 import org.apache.arrow.vector.VectorLoader
 import org.apache.arrow.vector.VectorSchemaRoot
 import org.apache.arrow.vector.VectorUnloader
-import org.apache.arrow.vector.ipc.ArrowStreamWriter
-import java.io.ByteArrayOutputStream
-import java.nio.channels.Channels
 import org.apache.arrow.vector.complex.DenseUnionVector
 import org.apache.arrow.vector.complex.ListVector
 import org.apache.arrow.vector.complex.writer.VarCharWriter
 import org.apache.arrow.vector.ipc.ArrowReader
+import org.apache.arrow.vector.ipc.ArrowStreamWriter
 import org.apache.arrow.vector.types.pojo.Schema
 import xtdb.ResultCursor
 import xtdb.api.Xtdb
@@ -31,12 +29,15 @@ import xtdb.arrow.VectorType.Companion.ofType
 import xtdb.arrow.unsupported
 import xtdb.arrow.withName
 import xtdb.database.DatabaseName
-import xtdb.table.TableRef
 import xtdb.query.PreparedQuery
 import xtdb.query.QueryOpts
+import xtdb.table.TableRef
 import xtdb.tx.TxOp
 import xtdb.tx.TxOpts
 import xtdb.util.useAll
+import java.io.ByteArrayInputStream
+import java.io.ByteArrayOutputStream
+import java.nio.channels.Channels
 
 class XtdbConnection(private val node: Node) : AdbcConnection {
 
@@ -51,16 +52,16 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
     override fun createStatement() = object : XtdbStatement {
         private var sql: String? = null
         private var prepared: PreparedQuery? = null
-        // Bound params kept in two forms: a reader view for `openQuery`, IPC bytes
-        // for `TxOp.SqlBytes`. We can't go via `Relation.load` (the obvious choice)
-        // because Arrow's `transferOwnership` fails across allocator roots, and
-        // callers can pass a VSR from an independent RootAllocator. Bytes don't care.
-        private var boundArgs: RelationReader? = null
+        // Arrow IPC bytes — allocator-agnostic, caller can close their VSR
+        // straight after bind. Cleared only by another bind() or close().
         private var argBytes: ByteArray? = null
 
         override fun setSqlQuery(sql: String) {
             this.sql = sql
+            // New SQL invalidates both the compiled plan and any prior bind —
+            // the parameter shape can have changed.
             this.prepared = null
+            this.argBytes = null
         }
 
         override fun prepare() {
@@ -77,24 +78,10 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
         }
 
         override fun bind(root: VectorSchemaRoot) {
-            boundArgs?.close()
+            argBytes = null
             // External `$N` → planner-internal `?_N`, by position.
-            // `$N` is 0-indexed by XTDB convention, not Postgres-faithful.
+            // `$N` is 0-indexed
             val renamedFields = root.schema.fields.mapIndexed { idx, f -> f.withName("?_$idx") }
-
-            // fromRoot doesn't transfer — the buffers stay in the source's allocator,
-            // openQuery reads them, and the cursor releases the wrappers on close.
-            val rel = Relation.fromRoot(node.allocator, root)
-            boundArgs = try {
-                RelationReader.from(
-                    rel.vectors.mapIndexed { idx, v -> v.withName("?_$idx") }.toList(),
-                    rel.rowCount
-                )
-            } catch (t: Throwable) {
-                rel.close()
-                throw t
-            }
-
             argBytes = ByteArrayOutputStream().use { baos ->
                 val renamedRoot = VectorSchemaRoot(renamedFields, root.fieldVectors, root.rowCount)
                 ArrowStreamWriter(renamedRoot, null, Channels.newChannel(baos)).use { writer ->
@@ -107,26 +94,30 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
         }
 
         override fun bind(rel: RelationReader) {
-            boundArgs?.close()
-            // XTDB-internal; assumes the source shares a root with node.allocator.
-            val owned = rel.openDirectSlice(node.allocator)
-            boundArgs = try {
-                RelationReader.from(
-                    owned.vectors.mapIndexed { idx, v -> v.withName("?_$idx") }.toList(),
-                    owned.rowCount
-                )
-            } catch (t: Throwable) {
-                owned.close()
-                throw t
-            }
-            // DML through this path isn't wired — executeUpdate will raise on the
-            // unbound `?` rather than silently dropping params.
             argBytes = null
+            rel.openDirectSlice(node.allocator).use { sliced ->
+                val renamedFields = sliced.vectors.mapIndexed { idx, v -> v.field.withName("?_$idx") }
+                Relation(node.allocator, renamedFields).use { renamed ->
+                    sliced.openArrowRecordBatch().use { rb -> renamed.load(rb) }
+                    argBytes = renamed.asArrowStream
+                }
+            }
         }
 
+        private fun deserialiseArgs(bytes: ByteArray): Relation =
+            Relation.StreamLoader(node.allocator, Channels.newChannel(ByteArrayInputStream(bytes))).use { loader ->
+                Relation(node.allocator, loader.schema).also { rel ->
+                    try {
+                        loader.loadNextPage(rel)
+                    } catch (t: Throwable) {
+                        rel.close()
+                        throw t
+                    }
+                }
+            }
+
         override fun executeQuery(): QueryResult {
-            // openQuery takes ownership on success; null our ref to avoid double-close.
-            val args = boundArgs.also { boundArgs = null }
+            val args = argBytes?.let(::deserialiseArgs)
             val cursor = try {
                 prepared?.openQuery(args, QueryOpts())
                     ?: node.openSqlQuery(sql ?: error("SQL query not set"), dbName)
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
```diff
@@ -44,27 +51,95 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
...
+
+        override fun getParameterSchema(): Schema {
+            val pq = prepared ?: error("call prepare() first")
+            // placeholder type until bind supplies real ones
+            return Schema(
+                (0 until pq.paramCount).map { idx -> "\$$idx" ofType VectorType.fromLegs() }
+            )
+        }
+
+        override fun bind(root: VectorSchemaRoot) {
```
jarohen, line 85: delegate this to bind(RelationReader) asap, save yourself writing an impl using Arrow Java

core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
```diff
@@ -44,27 +51,95 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
...
+                    writer.end()
+                }
+                baos.toByteArray()
+            }
+        }
+
+        override fun bind(rel: RelationReader) {
+            argBytes = null
+            rel.openDirectSlice(node.allocator).use { sliced ->
+                val renamedFields = sliced.vectors.mapIndexed { idx, v -> v.field.withName("?_$idx") }
```
jarohen: you shouldn't have to create fields of this then copy the data over - vectors have a `withName` on them too. should just™ be able to create a second relation out of those.

src/test/kotlin/xtdb/AdbcTest.kt
```diff
@@ -186,6 +193,204 @@ class AdbcTest {
...
+    @Test
+    fun `prepare with bound parameter (in-process)`() {
+        insertData("INSERT INTO foo RECORDS {_id: 1}, {_id: 2}, {_id: 3}")
+
+        xtdb.connect().use { ipConn ->
+            ipConn.createStatement().use { stmt ->
+                stmt.setSqlQuery("SELECT _id FROM foo WHERE _id = ?")
+                stmt.prepare()
+
+                val schema = Schema(listOf(
```
jarohen: use xtdb.arrow here, save yourself all of this manual creation - convert to vsr as late as possible/from vsr as early as possible. I might even have been tempted to write a couple of test helper extension functions so that it appears to us that FlightSql returns xtdb.arrow types.

src/test/kotlin/xtdb/AdbcTest.kt
```diff
@@ -186,6 +193,204 @@ class AdbcTest {
...
+            ipConn.createStatement().use { stmt ->
+                stmt.setSqlQuery("SELECT _id FROM foo WHERE _id = ? AND name = ?")
+                stmt.prepare()
+
+                val paramSchema = stmt.parameterSchema
+                assertEquals(2, paramSchema.fields.size, "expected 2 parameter slots")
+                assertEquals("\$0", paramSchema.fields[0].name)
+                assertEquals("\$1", paramSchema.fields[1].name)
+                // Types are deliberately the Null placeholder pre-bind.
+                assertEquals(ArrowType.Null.INSTANCE, paramSchema.fields[0].type)
```
jarohen: yeah, you def want to be using XT types by the time you get here.

jarohen (changes requested):
thanks @tggreene 🙏 

couple of notes in the impl itself. more generally, I'm not mad keen on the idea of storing bytes here - part of the point of in-process is that it stays as Arrow data throughout, so serializing to/from bytes kinda defeats the point somewhat 😅 

I appreciate you can't `transferOwnership`, which would have been ideal, but there's likely other methods that don't involve serializing. it might be that we're using the wrong allocator in the tests, for example.

re "AdbcTest" is mis-named: I agree - I'd usually fix this beforehand in a 'tidy-first' commit, so that the PR becomes easier to understand and you don't have to muddy the waters with qualifiers in your description 🙂

core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
tggreene: I realise you're probably quite busy but I've updated the PR to address your comments (hopefully).

I've opted for the incoming VSR as being owned by the caller and just copying rows into a new relation owned by the statement. I think that's reasonable?

Just going through and rebasing latter PRs too.

Thanks James.

tggreene: I've commented here as if to the root review, apologies if the ordering doesn't make sense.

core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
```diff
@@ -44,27 +47,94 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
...
     override fun createStatement() = object : XtdbStatement {
         private var sql: String? = null
+        private var prepared: PreparedQuery? = null
+        private var args: Relation? = null
+
+        private fun clearArgs() {
+            args.also { args = null }?.close()
+        }
+
+        private fun copyArgs(rel: RelationReader): Relation {
```
mbutlerw: I think this can be replaced with `rel.openDirectSlice`

EDIT: `openDirectSlice` not suitable here, requires shared root allocator.

core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
```diff
@@ -44,27 +47,94 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
...
+            return try {
+                rel.rowCopier(copied).copyRange(0, rel.rowCount)
+                copied
+            } catch (t: Throwable) {
+                copied.close()
+                throw t
+            }
+        }
+
+        private fun openArgs(): Relation? =
```
mbutlerw: Similarly here, I think `openSlice` should provide you with a copy of the relation.

However in this case the real win is that it will only inc the ref count, rather that copying the relation on every execution. Unlike for `bind` this is possible because we own the allocator.

mbutlerw (changes requested):

core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
tggreene: Claude posted here autonomously, since deleted as they didn't add anything other than creeping dread. I've overextended my gh permissions here, rethinking a lot 🤔

core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
```diff
@@ -44,27 +48,83 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
...
+            val queryArgs = openQueryArgs()
+            val cursor = try {
+                prepared?.openQuery(queryArgs, QueryOpts())
+                    ?: node.openSqlQuery(sql ?: error("SQL query not set"), dbName)
+            } catch (t: Throwable) {
+                queryArgs?.close()
+                throw t
+            }
+            val schema = Schema(cursor.resultTypes.map { (name, type) -> type.toField(name) })
             return QueryResult(-1, cursorToArrowReader(cursor, schema))
```
mbutlerw, line 113: `queryArgs` leaks on the success path — the `try/catch` only handles the throw case, but `openQuery` doesn't take ownership.
The planner reads from args during cursor evaluation, so they have to outlive the cursor and then need closing alongside it.
With `openSlice` semantics this is a refcount leak on the stored Relation's buffers — bounded, but visible.

Fix: pass `queryArgs` into `cursorToArrowReader` and close it in `closeReadSource` next to the cursor close.

Blocker.

core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
```diff
@@ -44,27 +48,83 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
...
+            // allocator — a ref-count bump for same-root allocators, a real copy when
+            // they differ. Caller can close their reader as soon as bind() returns.
+            args = rel.openDirectSlice(node.allocator)
+        }
 
+        override fun executeQuery(): QueryResult {
+            val queryArgs = openQueryArgs()
+            val cursor = try {
+                prepared?.openQuery(queryArgs, QueryOpts())
+                    ?: node.openSqlQuery(sql ?: error("SQL query not set"), dbName)
```
mbutlerw: If `bind()` has been called but `prepare()` hasn't, this elvis falls through to `openSqlQuery` with no args.
The bind is silently ignored *and* the slice opened by `openQueryArgs()` is leaked.

Two reasonable fixes:
- Enforce `prepared != null` here when args are present (error if not).
- Have `bind()` itself error when `prepared` is null, pushing the contract one step earlier.

Either way, worth a test in `InProcessAdbcTest` for the shape you pick.

Blocker.

core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
```diff
@@ -44,27 +48,83 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
...
             this.sql = sql
+            // New SQL invalidates both the compiled plan and any prior bind —
+            // the parameter shape can have changed.
+            this.prepared = null
+            clearArgs()
         }
 
-        override fun executeQuery(): QueryResult {
+        override fun prepare() {
             val sql = this.sql ?: error("SQL query not set")
```
mbutlerw: Project convention is `xtdb.error` for new code (see AGENTS.md → "Errors").
These are all caller-fault (this one and the others at lines 85, 108, 118), so `err.incorrect(...)`.

Not blocking, easy enough while in the file.

mbutlerw (commented):
A few more points:

- **executeQuery args leak** — see inline. Blocker.
- **Silent drop of args when bind precedes prepare** — see inline. Blocker.
- **`error(...)` → `xtdb.error`** — see inline. Not blocking, "while you're here".
- **PR description is stale** — the "Implementation notes → Bytes as the canonical bind form" section still describes the bytes commit that was reverted by `8335ea59`. Worth a quick rewrite before merge so the description matches the diff.

PR #5555, opened 2026-05-05
feat(azure): resolve BlobStorage credentials via !AzureBlob remote alias

danmason:
PR #5481 added the `remotes` pattern so node-local credentials never land on the source log; that PR covered Postgres but explicitly deferred storage-side credentials.
This closes the Azure side.

Both Azure credential paths — connectionString and storageAccountKey — can now be supplied via a node-local !AzureBlob remote, so the source log carries only an alias:

```yaml
remotes:
  az: !AzureBlob
    connectionString: !Env AZURE_BLOB_CONN_STR

storage: !Remote
  objectStore: !Azure
    storageAccount: my-account
    container: my-container
    remote: az
```

On the existing credential paths:
- `connectionString` has been removed from the factory config entirely
  - This path was always broken in practice (see bug below)
  - More importantly, allowing it inline meant credentials could be serialized into the source log
  - `connectionString` is now only supported via remote, ensuring it remains node-local
- `storageAccountKey` is now deprecated on the factory
  - Inline usage still works for backwards compatibility
  - Users are encouraged to migrate to remote to avoid credentials being serialized into logs

Setting both a remote alias and an inline storageAccountKey is rejected at openObjectStore time (xtdb.azure/conflicting-config); silent precedence would be a footgun.

This is partially breaking:
- ❌ Configs using inline `connectionString` must migrate to remotes
- ⚠️ Configs using inline `storageAccountKey` continue to work (deprecated)

## Additional fix
This PR also fixes a pre-existing latent bug in BlobStorage's credential setup:
- The builder always installed `DefaultAzureCredential` after any `connectionString(...)` call, silently overriding it
- The connectionString path therefore never actually worked — it was effectively a no-op shadowed by `DefaultAzureCredential`, which would then fail at request time with `CredentialUnavailableException` unless `storageAccountKey` happened to also be set (which would in turn override `DefaultAzureCredential`)

Given this, removing inline `connectionString` is both a correctness fix and a security improvement.

## Tests

- BlobStorageFactoryTest (default ./gradlew test)
  - Proto round-trip on both paths
  - Asserts the alias path strips credentials from the serialized proto
  - Verifies:
    - missing-remote
    - wrong-remote-type
    - conflicting-config
    - YAML decoding of !AzureBlob under remotes:
- BlobStorageTest (Azurite, @Tag("integration"))
  - Covers connectionString via remote (previously untested and effectively broken)
  - Alias-path round-trips for both credential types

Nightly Azure (`azure_test.clj`) wasn't extended — alias resolution is unit-tested, and once resolved the path goes through the same `BlobServiceClientBuilder` already exercised by the existing `^:azure` tests, so nothing new to verify against real Azure.

Out of scope: S3 / GCP migration to the same pattern — natural follow-ups now that the `Factory.openObjectStore` plumbing is in place.

### DIFF
--- a/core/src/main/kotlin/xtdb/api/storage/ObjectStore.kt
+++ b/core/src/main/kotlin/xtdb/api/storage/ObjectStore.kt
@@ -3,6 +3,8 @@ package xtdb.api.storage
 import kotlinx.serialization.modules.PolymorphicModuleBuilder
 import kotlinx.serialization.modules.SerializersModule
 import kotlinx.serialization.modules.polymorphic
+import xtdb.api.Remote
+import xtdb.api.RemoteAlias
 import java.nio.ByteBuffer
 import java.nio.channels.FileChannel
 import java.nio.file.Path
@@ -18,7 +20,7 @@ interface ObjectStore : AutoCloseable {
     }
 
     interface Factory {
-        fun openObjectStore(storageRoot: Path): ObjectStore
+        fun openObjectStore(storageRoot: Path, remotes: Map<RemoteAlias, Remote> = emptyMap()): ObjectStore
         val configProto: ProtoAny
 
         companion object {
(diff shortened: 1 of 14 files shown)


jarohen:
* `!AzureBlob` :gavel:
* if connString didn't work, then remove it
* in the protobuf, mark the original 'reserved' (with a comment as to what it was)

danmason:
- Opt to remove connection string since it was broken, keep the storageAccountKey but deprecate it.

Issue #5132, opened 2026-01-07
ADBC/FlightSQL support


jarohen:
## Overview

ADBC (Arrow Database Connectivity) and FlightSQL provide columnar-native access to XTDB, complementing the existing pgwire interface.
Two public API surfaces:

1. **ADBC in-process** — `XtdbConnection` + `AdbcDriverFactory` for JVM apps embedding XTDB. Arrow-native, no network overhead.
2. **FlightSQL over-the-wire** — `XtdbProducer` gRPC server for non-JVM clients (Python, Go, Rust, R, C++, etc).

### Why

- **Columnar efficiency**: pgwire serializes row-by-row; FlightSQL/ADBC stream Arrow record batches natively.
- **Language breadth**: ADBC drivers exist for Python, Go, Rust, R, Java, C++ — broadens XTDB's reach beyond JVM and psql-compatible tools.
- **Bulk ingestion**: standardized bulk-load path that preserves Arrow types end-to-end.

### What's Shipped (on `main`)

- **FlightSQL server** (`core/.../flight_sql/XtdbProducer.kt`) — queries, DML, prepared statements, transactions
- **ADBC connection** (`core/.../adbc/XtdbConnection.kt`) — queries, updates, bulk ingest (CREATE_APPEND only)
- **Driver factory** (`core/.../api/AdbcDriverFactory.kt`) — SPI-registered
- **Integrant wiring** (`core/.../flight_sql.clj`)
- **Tests**: Clojure FlightSQL tests + Kotlin ADBC tests

#### Completeness
- [x] ADBC: Implement `XtdbConnection.getInfo()` (currently TODO)
- [x] FlightSQL: Metadata endpoints (getCatalogs, getSchemas, getTables, getTableTypes) for tool discovery

#### Validation & Conformance

Use https://github.com/adbc-drivers/validation to verify and validate the completeness of the ADBC implementation.
The validation suite tests conformance against the ADBC specification and ensures compatibility with other ADBC consumers.

## Current Status

Baseline shipped on `main`. Triaging remaining work into independently committable sub-issues.

## Open Questions

- [x] Do we need FlightSQL metadata endpoints for tooling compatibility (DBeaver, DataGrip)?

jarohen:
### Session — 2026-02-20

**Status**: in-progress

- [x] Stage 3: #5082 DML error fix
- [x] Stage 4: Metadata endpoints
- [x] Code review fixes
- [x] Consumer readiness audit
[...]

Issue #5080, opened 2025-12-04
FlightSQL: Implement bulk ingestion (CommandStatementIngest)


jarohen:
## Summary

Implement Flight SQL bulk ingestion command (added in Arrow 17.0.0) to allow clients to efficiently insert data by streaming Arrow record batches directly.

## Method to implement

Override `acceptPutStatementIngest` in `XtdbProducer`:

```kotlin
override fun acceptPutStatementIngest(
    command: CommandStatementIngest,
    context: CallContext?,
    flightStream: FlightStream,
    ackStream: StreamListener<PutResult>
): Runnable
```

## CommandStatementIngest fields

- `tableDefinitionOptions` - create table / append behavior
- `table` - target table name
- `schema` - optional target schema
- `catalog` - optional target catalog/database
- `temporary` - create temp table?

## Implementation notes

- Client streams Arrow record batches
- Convert to XTDB insert operations (similar to how we handle prepared statement updates)
- Much more efficient than SQL INSERT statements for bulk loading
- Semantics follow ADBC bulk ingestion closely

## References

- [Flight SQL Format Specification](https://arrow.apache.org/docs/format/FlightSql.html)
- [CommandStatementIngest proposal](https://github.com/apache/arrow/issues/38255)
- Arrow 17.0.0 release notes

PR #5578, opened 2026-05-06
feat(flight-sql): wire bulk ingest via CommandStatementIngest (#5080)

tggreene:
## Summary

`cur.adbc_ingest(arrow_table, mode='create_append')` now lands a `pyarrow.Table` (or any Arrow-shaped data) in XTDB in a single round trip — the headline Arrow ergonomic that XTDB has been missing for non-JVM clients.

## Context

Documented in [`adbc-bugs.md` #3b](https://github.com/xtdb/driver-examples/blob/adbc-testing/adbc-bugs.md) on the driver-examples branch.
Without bulk ingest, callers had to shred their Arrow data into rows and round-trip through `executemany("INSERT … VALUES (?, …)")` — paying parameterised-INSERT overhead per row, losing some Arrow type fidelity, and giving up the "Arrow stream stays Arrow" pitch that's most of why someone reaches for ADBC over pgwire in the first place.

Iteration toward umbrella #5132; advances sub-issue #5080.

## Implementation

`XtdbProducer.acceptPutStatementBulkIngest` overrides the FSQL surface and delegates to the existing in-process `XtdbConnection.bulkIngest(targetTableName, BulkIngestMode.CREATE_APPEND)`.
Each Arrow batch from the `FlightStream` is bound to the statement and executed; the per-batch lifecycle keeps us off a "hold the whole stream in memory" path and matches how `XtdbStatement` already works for parameterised DML.

**Mode handling is intentionally narrow, and rejections are loud.** XTDB has implicit table creation, so the meaningful FSQL-mode distinctions collapse to:

- `ifExists`: `UNSPECIFIED` / `APPEND` accepted (XTDB upserts on `_id`). `FAIL` and `REPLACE` rejected with `INVALID_ARGUMENT` — would need an existence check / `ERASE` step we don't have today.
- `ifNotExist`: `UNSPECIFIED` / `CREATE` accepted. `FAIL` rejected — we can't honour "fail if not exist" because XTDB will auto-create on insert; silently accepting it would violate the FSQL contract.
- `cmd.catalog` per-call override: rejected. Connection-scoped catalog (headers / session) stays the only source of truth.
- `cmd.schema` per-call override: honoured (falls back to `"public"` when unset).

Rejections go through `CallStatus.INVALID_ARGUMENT.withDescription(...).toRuntimeException()` so the explanatory message reaches the client instead of being flattened to a generic gRPC INTERNAL.

## Test plan

- `AdbcTest.test FlightSQL bulk ingest via executeIngest` — round-trips a 2-row Arrow batch through `FlightSqlClient.executeIngest`, asserts the rows come back via `SELECT`.
- `AdbcTest.test FlightSQL bulk ingest within an FSQL transaction` — begins an FSQL transaction, ingests within it, commits, asserts visibility.
- `AdbcTest.test FlightSQL bulk ingest rejects per-call catalog override` — sets `cmd.catalog`, asserts `INVALID_ARGUMENT` with descriptive message.
- `AdbcTest.test FlightSQL bulk ingest rejects fail-if-not-exist` — sets `ifNotExist == FAIL`, asserts `INVALID_ARGUMENT`.
- 13/13 `xtdb.AdbcTest` tests green.
- End-to-end verified against a freshly-built integration node:

  ```python
  cur.adbc_ingest("people", arrow_table, mode="create_append")
  # → 5 rows of mixed types (string / int32 / double-with-null / date32 / bool)
  # → land in XTDB in ~12 ms; SELECT round-trips with types preserved.
  ```

No reflection or boxed-math warnings.

## Scope

- Read-only on table-existence policy.
  `replace` and `create-with-fail` reject upfront; deferring to a follow-up that wires existence/erase plumbing.
- The validation suite's bulk ingest tests stay skipped (`statement_bulk_ingest=False` in `xtdb.py`) because the suite's fixtures build tables without `_id` and XTDB rejects rows without it.
  This is an XTDB data-model concern rather than an ADBC plumbing concern; not in scope of this PR.

### DIFF
--- a/core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
+++ b/core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
@@ -172,31 +172,32 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
             override fun prepare(): Unit =
                 throw Incorrect("Bulk ingest does not support prepare", "xtdb.adbc/bulk-ingest-no-prepare")
 
-            private var rel: Relation? = null
+            private var docs: Relation? = null
 
             override fun close() {
-                rel.also { rel = null }?.close()
+                docs.also { docs = null }?.close()
             }
 
             override fun bind(root: VectorSchemaRoot) {
-                this.rel?.close()
-                rel = Relation.fromRoot(node.allocator, root)
+                this.docs?.close()
+                docs = Relation(node.allocator, root.schema).closeOnCatch { copy ->
+                    Relation.fromRoot(node.allocator, root).use { src ->
+                        src.rowCopier(copy).copyRange(0, src.rowCount)
+                    }
+                    copy
+                }
             }
 
             override fun bind(rel: RelationReader) {
-                this.rel = rel.openDirectSlice(node.allocator)
+                this.docs = rel.openDirectSlice(node.allocator)
             }
 
             override fun executeUpdate(): UpdateResult {
-                return (rel.also { this.rel = null } ?: return UpdateResult(0))
-                    .use { rel ->
-                        // TODO valid-from/valid-to for the batch
-                        node.executeTx(dbName, listOf(TxOp.PutDocs(schemaName, tableName, docs = rel)))
-
-                        close()
-
-                        UpdateResult(rel.rowCount.toLong())
-                    }
+                val docs = docs.also { this.docs = null } ?: return UpdateResult(0)
+                val rowCount = docs.rowCount.toLong()
+                // TODO valid-from/valid-to for the batch
+                executeDml(TxOp.PutDocs(schemaName, tableName, docs = docs))
+                return UpdateResult(rowCount)
             }
         }
     }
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
mbutlerw, line 162: I think the bulkIngest should mirror the SQL path wrt bind semantics.

core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
```diff
@@ -188,15 +188,10 @@ class XtdbConnection(private val node: Node) : AdbcConnection {
             }
 
             override fun executeUpdate(): UpdateResult {
-                return (rel.also { this.rel = null } ?: return UpdateResult(0))
-                    .use { rel ->
-                        // TODO valid-from/valid-to for the batch
```
mbutlerw, line 193: Did this TODO get resolved?

core/src/main/kotlin/xtdb/adbc/XtdbConnection.kt
mbutlerw: NIT: docs might be a more descriptive name here, mirroring `args` from SQL path.

mbutlerw (changes requested):
Overall LGTM, few fixes required. 

Might be nice to have a test which creates bind args on a standalone alloc (as a user would do) to prove our ownership model is correct.

PR #5605, opened 2026-05-11
feat(adbc): implement AdbcStatement.executeSchema (resolves #5604)

### LINKED ISSUE #5604: ADBC: implement AdbcStatement.executeSchema()
## Summary

`AdbcStatement.executeSchema()` (ADBC API rev 1.1.0) returns the result-set schema for a prepared query *without* executing it.
We currently inherit the default which throws `notImplemented`.

## Context

Part of the M1 ADBC closure umbrella (#5132).
The validation suite at https://github.com/adbc-drivers/validation has a `test_execute_schema_*` group which errors at setup against our server.
Apache ADBC's Python/Go drivers also rely on it for query-shape introspection (e.g. DuckDB-style integrations).

## How it's called over the wire

For the reader's mental model: the upstream `FlightSqlStatement.executeSchema()` has two paths —

- **Prepared**: calls `FlightSqlClient.PreparedStatement::getResultSetSchema`, which reads the `datasetSchema` bytes returned by `createPreparedStatement` at prepare time.
  Server-side this means populating `ActionCreatePreparedStatementResult.datasetSchema` (currently we only populate `parameterSchema`).
- **Ad-hoc**: calls `client.getExecuteSchema(query)` → server handler is `FlightSqlProducer.getSchemaStatement(CommandStatementQuery, ...)`.
  We don't implement this today.
- **Also affected**: `getSchemaPreparedStatement(CommandPreparedStatementQuery, ...)` — used by tools that issue `GetSchema` directly against a prepared-statement descriptor, currently inherits the default `UNIMPLEMENTED`.

## In scope

- **In-process**: `executeSchema()` override on `xtdb.adbc.XtdbConnection.XtdbStatement`, delegating to `PreparedQuery.getColumnFields` (the same path pgwire uses at `xtdb/pgwire.clj:996` for pre-execution column shape).
- **FSQL**: populate `datasetSchema` on `createPreparedStatement`, plus implement `getSchemaPreparedStatement` and `getSchemaStatement`.
- **Tests** in `xtdb.AdbcTest` covering in-process, the prepared FlightSQL ADBC path, and the ad-hoc `getExecuteSchema(query)` path.

## Out of scope

- `getSchemaSubstraitPlan` — separate code path, no current driver consumer.
- Behaviour when parameters affect the projected schema (e.g. `SELECT ? + 1` with unknown param type).
  We'll pass placeholder/null param fields by default — same compromise pgwire makes when oids are unknown.
  Document the trade-off in the PR; revisit only if a validation test exercises it.

Iterates toward umbrella #5132.

tggreene:
## Summary

Wires ADBC's `AdbcStatement.executeSchema()` end-to-end — in-process and over the FSQL wire — closing the validation suite's `test_execute_schema_*` group.

## Context

ADBC's `executeSchema()` (API rev 1.1.0) returns a query's result-set shape *without* executing it.
The default throws `notImplemented`, which the adbc-drivers/validation suite errors at setup on.

Iterates toward umbrella #5132.

## Stack

Built on top of #5526 (`tim/5270-adbc-prepare`).
This PR is the next layer up — depends on the prepared-statement plumbing landed there.
Conflicts with #5554 (`tim/5270-fsql-refactor`) if that merges first; rebase will be straightforward (the FSQL hooks added here move from the parallel `PreparedStatement` impl onto the post-refactor `XtdbStatement` delegation, but the in-process piece survives unchanged).

## The interesting decision: pre-bind result schemas

Result schemas can genuinely depend on parameter types — `SELECT ?` is the obvious example.
But `executeSchema()` runs before any bind, so we don't have the types.

XTDB's planner doesn't infer them either.
`PreparedQuery.getColumnFields(paramFields)` *takes* the param types as input rather than deriving them from query structure.
pgwire feeds it from PG oids when the client sends them, and falls back to "return every output column as the default type" when any oid is unknown (`xtdb/pgwire.clj:1008-1017` — `(some nil? param-types) → PG_DEFAULT × column-names`).

We go one step further: synthesise null-typed placeholder param fields and call `getColumnFields` with them, letting the planner type non-`?`-projected columns properly.

- Accurate for `SELECT cols FROM t WHERE _id = ?` (the common shape — projection doesn't depend on `?`'s type). More precise than pgwire's wholesale default-typing fallback for this shape.
- Placeholder-typed for queries that project the parameter directly (`SELECT ?`, `SELECT ? + 1`) — same answer pgwire would give.

A planner-side inference pass — walking the conformed plan to collect per-param constraints (`WHERE _id = ?` ⇒ param coerces to `_id`'s type; `LIMIT ?` ⇒ int; `?+1` ⇒ numeric) — would close the gap for both pgwire and ADBC.
Out of scope here — meaningful planner work, deserves its own issue.

## Changes

Four pieces, all small:

1. `XtdbConnection.XtdbStatement.executeSchema()` — calls `prepared.getColumnFields` with the placeholder param fields described above.
2. `XtdbProducer.createPreparedStatement` — also sets `datasetSchema` (in addition to `parameterSchema`) on the prepare response, for queries.
3. `XtdbProducer.getSchemaPreparedStatement` — for clients that issue `GetSchema` against a prepared-statement descriptor.
4. `XtdbProducer.getSchemaStatement` — ad-hoc path; opens a transient `PreparedQuery` to read the result shape.

The `datasetSchema` bit (2) is non-obvious but load-bearing:
the upstream FSQL ADBC driver's prepared `executeSchema()` reads `FlightSqlClient.PreparedStatement::getResultSetSchema`, which is just the bytes the server returned at prepare time.
Without populating `datasetSchema`, the driver hands back an empty schema and never reaches `getSchemaPreparedStatement`.
Both surfaces are needed: `datasetSchema` for the ADBC driver's prepared path, the two `getSchema*` hooks for tools that issue `GetSchema` directly.

DML prepares deliberately skip `datasetSchema` — the result is a row count, not a tabular shape.

## Scope

Out:

- Substrait variant (`getSchemaSubstraitPlan`) — no current driver consumer.
- Real param-type inference — see the planner-feature note above.

## Test plan

In-process (`xtdb.api.AdbcTest`):

- `executeSchema after prepare returns projected column names`.
- `executeSchema before prepare throws Incorrect` — pins the `xtdb.adbc/not-prepared` error path.
- `executeSchema on parameterised query exposes one column per projection` (`SELECT ? AS p`).

Over the FSQL wire (`xtdb.FlightSqlAdbcTest`):

- ADBC FSQL client prepared `executeSchema()` — exercises the `datasetSchema` round-trip.
- ADBC FSQL client ad-hoc `executeSchema()` (no `prepare()`) — exercises `getSchemaStatement`.
- Raw `FlightSqlClient.getExecuteSchema(query)` — direct ad-hoc wire path.
- `getExecuteSchema rejects DML` — pins the `CallStatus.INVALID_ARGUMENT` branch.

No reflection or boxed-math warnings.

## Known follow-up: #5608

Validation against [adbc-drivers/validation](https://github.com/adbc-drivers/validation) surfaced a deeper issue in our prepare path that's *not* fixed here:
`XtdbConnection.prepareSql` doesn't await the connection's own writes before opening the planning snapshot, so an `INSERT` followed by `executeSchema` on the same ADBC connection races the indexer.
Planner falls back to `_column_N` for tables it hasn't seen flushed.

Tracked in #5608 — fix is a separate plumbing change through `XtdbConnection.Node.prepareSql` (mirror pgwire's `await-token` pattern).

`executeSchema`'s wiring here is correct — once #5608 lands, the validation suite's `test_execute_schema_*` group flips from "error: `expected res != actual _column_1`" to green without any change to this PR's code.


core/src/main/kotlin/xtdb/flight_sql/XtdbProducer.kt
```diff
@@ -353,20 +353,46 @@ class XtdbProducer(private val node: Xtdb) : NoOpFlightSqlProducer(), AutoClosea
...
+            .setParameterSchema(
+                ByteString.copyFrom(xtdbStmt.parameterSchema.serializeAsMessageInterruptibly())
             )
-        )
 
+        if (!isDml(sql, dbName)) {
+            resultBuilder.setDatasetSchema(
+                ByteString.copyFrom(xtdbStmt.executeSchema().serializeAsMessageInterruptibly())
+            )
+        }
```
mbutlerw: Narrow resource leak: `stmts[psId] = …` happens at line 354, but the client only learns `psId` from `listener.onNext` below at line 368. If `executeSchema()` throws here — planner paths that survive `prepare()` but trip on `getColumnFields` — the entry stays in `stmts` indefinitely. Client never sees the handle, so it can't call `closePreparedStatement`.

Pre-existing pattern; this PR widens the window by adding a second throw point post-`stmts.put`. Either build the full result before inserting into `stmts`, or wrap this block in try/catch with `stmts.remove(psId)` + `xtdbStmt.close()` on throw.

src/test/kotlin/xtdb/FlightSqlAdbcTest.kt
```diff
@@ -178,6 +179,48 @@ class FlightSqlAdbcTest {
...
+    @Test
+    fun `test ADBC executeSchema on prepared query`() {
+        insertData("INSERT INTO foo RECORDS {_id: 1, n: 'one'}")
+
+        conn.createStatement().use { stmt ->
+            stmt.setSqlQuery("SELECT _id, n FROM foo")
+            stmt.prepare()
+            assertEquals(listOf("_id", "n"), stmt.executeSchema().fields.map { it.name })
+        }
+    }
```
mbutlerw, line 207: `getSchemaPreparedStatement` has no direct test. This test goes through the ADBC FSQL driver, which reads cached `datasetSchema` bytes off `PreparedStatement` and never round-trips to the server, so the new server-side override is unexercised.

Forcing the RPC is one line via `FlightSqlClient.PreparedStatement.fetchSchema()` (distinct from `getResultSetSchema()`, which is the cached path):

```kotlin
@Test
fun `test FlightSQL fetchSchema on prepared query`() {
    insertData("INSERT INTO foo RECORDS {_id: 1, n: 'one'}")
    fsqlClient.prepare("SELECT _id, n FROM foo", *emptyCallOpts).use { prepared ->
        assertEquals(listOf("_id", "n"), prepared.fetchSchema(*emptyCallOpts).schema.fields.map { it.name })
    }
}
```

Without this, `getSchemaPreparedStatement` could be returning anything and the suite wouldn't notice.

src/test/java/xtdb/api/AdbcTest.kt
```diff
@@ -225,4 +227,46 @@ class AdbcTest {
...
+            db.connect().use { conn ->
+                conn.createStatement().use { stmt ->
+                    stmt.setSqlQuery("SELECT ? AS p")
+                    stmt.prepare()
+                    stmt.executeSchema().fields.map { it.name } shouldBe listOf("p")
+                }
+            }
+        }
+    }
 }
```
mbutlerw: Missing trailing newline.

mbutlerw (commented):
Three small items from a closer read — none blocking, but the resource-leak window is worth tightening while the call site is open. Param-naming feedback is filed separately as #5651 — it's `getParameterSchema`'s surface, not `executeSchema`'s, so cleaner as its own change.

mbutlerw (changes requested):

Issue #5618, opened 2026-05-13
postgres-source: follower's snapshot rows missing despite latestCompletedTxId showing they were imported


Akeboshiwind:
## Summary

In a multi-node deployment, the leader's view of a postgres-source CDC database shows the full snapshot + streaming history, but a follower querying the same database sees only the streaming row.
Both nodes' `latestCompletedTxId` gauge agrees on the latest tx id, so as far as the indexer's bookkeeping is concerned the follower applied every ResolvedTx — yet the corresponding rows are not visible at query time.

## Context

- XTDB build: `d6f1943` (includes `e9c670392` and `2ac2569ba` — the recent split-msg-id fixes).
- Deployment: multi-node, Kafka log + Postgres logical-replication source via `!Postgres`.
- Affected database: a single attached CDC db (`cdc` in this report; the real name is environment-specific).
- 14 tables in the publication, ~422 rows total across the snapshot.
- No cdc block files exist in the object store (verified).

## Symptoms

On the leader:

- `SELECT * FROM xt.txs` on `cdc` returns all 12 txs (ids `0..11`).
- All snapshot tables return their snapshot rows, plus one table receives the streamed row.

On the follower:

- `SELECT * FROM xt.txs FOR ALL VALID_TIME` on `cdc` returns one row only — tx `11` (the streamed tx).
- The table that received the streaming insert returns only that one streaming row.
- All other snapshot tables return empty.

### Replica-log contents

The cdc replica topic contains 15 user-visible records, offsets 14–42:

| offset | type | timestamp (UTC) |
|---|---|---|
| 14 | NoOp | 14:48:38 |
| 16 | NoOp | 14:57:54 |
| 18 | NoOp | 15:01:19 |
| 20 | ResolvedTx (txId 0) | 15:01:20.172 |
| 22 | ResolvedTx (txId 1) | 15:01:20.229 |
| 24 | ResolvedTx (txId 2) | 15:01:20.250 |
| 26 | ResolvedTx (txId 3) | 15:01:20.283 |
| 28 | ResolvedTx (txId 4) | 15:01:20.320 |
| 30 | ResolvedTx (txId 5) | 15:01:20.341 |
| 32 | ResolvedTx (txId 6) | 15:01:20.381 |
| 34 | ResolvedTx (txId 7) | 15:01:20.397 |
| 36 | ResolvedTx (txId 8) | 15:01:20.426 |
| 38 | ResolvedTx (txId 9) | 15:01:20.470 |
[...]

PR #5619, opened 2026-05-14
test: Add test against snapshot process on multiple nodes running Kafka and S3.

danmason:
Trying to replicate #5618 via a more complicated end to end level test.

### DIFF
--- a/modules/kafka/src/test/kotlin/xtdb/api/log/KafkaClusterTest.kt
+++ b/modules/kafka/src/test/kotlin/xtdb/api/log/KafkaClusterTest.kt
@@ -9,7 +9,12 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.test.runTest
 import org.apache.kafka.clients.admin.AdminClient
 import org.apache.kafka.clients.admin.NewTopic
+import org.apache.kafka.clients.admin.RecordsToDelete
+import org.apache.kafka.clients.producer.KafkaProducer
+import org.apache.kafka.clients.producer.ProducerRecord
+import org.apache.kafka.common.TopicPartition
 import org.apache.kafka.common.errors.RecordTooLargeException
+import org.apache.kafka.common.serialization.ByteArraySerializer
 import org.junit.jupiter.api.*
 import org.junit.jupiter.api.Assertions.assertArrayEquals
 import org.junit.jupiter.api.Assertions.assertEquals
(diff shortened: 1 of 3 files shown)


danmason:
### Chalk — #5618 postgres-source follower-snapshot-skipped

- [x] Multi-node integration test harness
- [x] Reproduce #5618 in a regression test
- [x] Root cause identified

<details><summary>Test harness — baseline lifecycle</summary>

3 XTDB nodes against Kafka + MinIO + Postgres testcontainers. Leader attaches the postgres-source, all nodes converge on snapshot, late joiners catch up (both pre- and post-streaming-tx). Establishes the happy path each regression test deviates from. Passes locally. Per-test container lifecycle (`@BeforeEach`/`@AfterEach`) — see Kafka teardown hang note below.

</details>

<details><summary>Truncated replica topic prefix — reproduces #5618</summary>

Pre-seeds noops onto the replica topic, then advances the partition low-watermark past 0 via `AdminClient.deleteRecords` (same mechanism as retention expiry). Leader takes the snapshot; a fresh follower joins post-snapshot with a `-1` recovery token (no block on object store).

**Currently fails** — the follower shows `rows=0, txs=0` while the leader has the full snapshot. Will pass once the seek fix lands.

</details>

## Root cause

`KafkaCluster.tailAll` (and the parallel `SharedGroupConsumer.onPartitionAssigned` path) does:

```kotlin
c.seek(tp, afterMsgIdToOffset(epoch, afterMsgId) + 1)
```

When no block has been written to object store yet for this database, the recovery token is `-1` and the seek resolves to absolute offset `0`. On a replica topic where retention or `deleteRecords` has advanced the partition's log-start-offset past `0`, that seek is out-of-range.

The bug shape is stronger than "skip snapshot, catch the next streamed tx". The consumer dies entirely:

- consumer has `isolation.level=read_committed` and `auto.offset.reset=latest`
- the OOR fetch throws `OffsetOutOfRangeException` rather than silently auto-resetting (auto-reset interacts badly with the transactional-fetch state on an explicit out-of-range seek)
- `KafkaCluster.pollRecords` only catches `WakeupException` and `InterruptException` — `OffsetOutOfRangeException` propagates out of `tailAll`, kills the follower's replica-log subscription coroutine, and there's no respawn

Locally, the follower then stays at `rows=0, txs=0` even after subsequent streamed inserts — confirmed in the regression test. The qa observation of pod-1 "waking up" at the streamed tx ~8 minutes later was almost certainly a Kafka session-timeout / consumer-group rebalance restarting the polling coroutine fresh against the live position, not the original consumer recovering on its own.

## Proposed fix

Two complementary changes:

- **Primary**: in `tailAll` (and the matching `SharedGroupConsumer` path), clamp the seek offset to the partition's current `beginningOffsets` when the resolved offset is below it. "Start from the beginning" on a truncated log should mean current low-watermark, not absolute offset `0`. Avoids the OOR entirely.
- **Defence in depth**: catch `OffsetOutOfRangeException` in `pollRecords` and either re-seek to `beginningOffsets` and continue, or surface as a hard error — silently killing the polling coroutine is the worst outcome.

The truncated-prefix regression test will flip to passing once the primary fix lands.

## Side note — Kafka teardown hang

When a test assertion fails inside the `use { node -> ... }` block, the node's Kafka consumer doesn't release cleanly and the test process hangs for several minutes after the assertion already fired. Mitigated in this test suite by per-test containers, but the underlying behaviour — a partially-closed XTDB node leaves dangling Kafka consumer threads — would bite operators bouncing nodes too. Worth a separate look once the seek fix is in.

Refs #5618

danmason:
## Investigation note — close-during-rebalance deadlock (separate issue)

While running this branch's multi-node test (`postgres-source snapshot converges with truncated replica topic prefix`) end-to-end, the test hangs indefinitely after the expected #5618 assertion failure. The hang is **not** part of #5618 — it's a separate, pre-existing deadlock that #5618's late-follower scenario happens to trigger reliably because the assertion failure interleaves a `node.close()` with an in-flight Kafka rebalance.

### What's happening

Four-way wait cycle, fully closed:

```
Test worker — Database.close() → runBlocking { job.cancelAndJoin() }
  └─ waits for the source-log subscription job to complete
       └─ its finally block tries: sharedGroupConsumer.unregister(topic)
            └─ suspends on commandCh.send + consumer.wakeup()
                 └─ waits for SharedGroupConsumer pollingJob to reach its select
                      └─ pollingJob is stuck inside consumer.pollRecords()
                           └─ Kafka is mid-rebalance, calling onPartitionsAssigned
                                └─ runBlocking inside the rebalance listener
                                     └─ LogProcessor.onPartitionsAssigned →
                                        FollowerLogProcessor.awaitReplicaMsgId
                                          └─ suspends on replicaState.first { … }
                                               └─ would advance from tailAll.processRecords
                                                    └─ tailAll was just cancelled by job.cancelAndJoin
```

The leader-transition path in `LogProcessor.onPartitionsAssigned` awaits a `replicaState` StateFlow on the follower; the StateFlow is updated by the `FollowerSystem.tailAll` job on `Database.scope`; `Database.close()` cancels that scope, severing the only path that could complete the await. Meanwhile the `SharedGroupConsumer` poll thread is sitting in `consumer.poll()` running the rebalance listener synchronously, so it can never reach the `select` that would process `unregister` — so `cancelAndJoin` can never return.

The baseline lifecycle test doesn't hit this: it only closes nodes after every transition has settled, so there's no rebalance-handler-in-flight during close.

### Minimal reproducer (Kafka only, no Postgres / S3 / nodes)

Verified locally that this is reproducible at the `KafkaCluster` layer in isolation — the rebalance-handler-doesn't-return condition is sufficient on its own. The shape:

```kotlin
@Test
fun `cancelAndJoin deadlocks when rebalance handler does not return`() = runTest(timeout = 60.seconds) {
    val topic = "test-deadlock-${UUID.randomUUID()}"

    withClusterAndLogs(listOf(topic)) { _, logs ->
        val log = logs[0]

        val gotAssignment = CompletableDeferred<Unit>()
        val stuckListener = object : SubscriptionListener<SourceMessage> {
            override suspend fun onPartitionsAssigned(partitions: Collection<Int>): TailSpec<SourceMessage>? {
                gotAssignment.complete(Unit)
                // runBlocking inside the rebalance listener creates its own scope,
                // so cancelling the outer subscription job does NOT cancel this.
                awaitCancellation()
            }
            override suspend fun onPartitionsRevoked(partitions: Collection<Int>) {}
        }

        val job = launch { log.openGroupSubscription(stuckListener) }
        gotAssignment.await()

        // Mirrors Database.close()'s runBlocking { job.cancelAndJoin() }.
        // Should return promptly; under the deadlock, it never returns and
        // withTimeout fires.
        withTimeout(15.seconds) { job.cancelAndJoin() }
    }
}
```

In production, the rebalance handler suspends on `replicaState.first { … }` whose update path gets cancelled. In the repro it suspends on `awaitCancellation()`, which would only complete if the `runBlocking`'s own coroutine were cancelled (which nothing does). Both produce identical effects on the outer cycle: handler doesn't return → poll loop can't process `unregister` → subscription job can't terminate → `cancelAndJoin` deadlocks.

### Mitigations (not on this PR)

1. **Race `awaitReplicaMsgId` against a close signal.** Per-`LogProcessor` `CompletableDeferred<Unit>` that close completes early; the `first { }` becomes a `select`. Smallest diff.
2. **Don't block the rebalance handler.** Restructure so `awaitReplicaMsgId` runs between polls, not inside the listener. Right long-term fix; bigger change.
3. **Break the close-vs-rebalance ordering at the `Database` level.** `Database.close()` signals the LogProcessor to drop any in-flight leader transition before cancelling the scope.

(1) and (3) are equivalent in effect; (1) is closer to where the await is. (2) removes the fundamental fragility but is a deeper restructure.

Suggest filing this as a separate issue and progressing #5618 independently.

### Outcome (this PR)

Test-side workaround committed in `c288409e4` (`test(postgres-source): unblock #5618 multi-node test from close-during-rebalance hang`):
the multi-node regression test now closes the late follower explicitly, sleeps one `SETTLE_TIME` so node1's post-rebalance leader transition completes, and then rethrows any captured assertion.
The test now fails on the actual #5618 assertion rather than hanging.

The underlying close-during-rebalance deadlock is **not addressed on this PR** — it's a pre-existing indexer/Kafka issue and should be filed and tracked separately so #5618 can land independently.

Akeboshiwind:
### Chalk — log-layer regression tests for #5618

The seek-to-zero-on-truncated-log behaviour from #5618 lives in generic Kafka-log code — `KafkaCluster.tailAll` and `SharedGroupConsumer.onPartitionAssigned` both compute `seek(tp, afterMsgIdToOffset(epoch, afterMsgId) + 1)`, resolving to absolute offset 0 when `afterMsgId == -1`.
The postgres-source replica topic is where it surfaced operationally; the bug itself isn't postgres-source specific.

Two regression tests folded into `modules/kafka/src/test/kotlin/xtdb/api/log/KafkaClusterTest.kt`, alongside the happy-path `tailAll` and `openGroupSubscription` tests they're the regression siblings of:

- `tailAll picks up source messages appended past a truncated prefix` — source-log path.
- `openGroupSubscription picks up replica messages appended past a truncated prefix` — replica-log path via `SharedGroupConsumer`.

The harness is intentionally minimal — `KafkaCluster.LogFactory` + a raw `KafkaProducer` for seeding, `AdminClient.deleteRecords` for truncation, mockk/inline listener for capture.
No XTDB node, no MinIO, no postgres-source.
`deleteRecords` advances the partition's low-watermark via the same mechanism Kafka retention uses internally, so the precondition is reproduced faithfully without time-based fixtures.

Pre-fix both fail via `runTest(timeout = 30.seconds)` wall-clock — polling coroutine is dead on `OffsetOutOfRangeException`, wait loop never observes the appended records, runTest's outer timeout fires.
Confirmed locally — ~30s each, 69s total run time for the pair.
Post-fix (clamp the resolved seek up to `beginningOffsets`) both should pass in ~1s.

Useful as a placement check for the fix: a `tailAll`/`SharedGroupConsumer` change that doesn't flip these tests green means the fix isn't general enough.

Landed in `0e1e00d04`.

Refs #5618

jarohen:
Closing as superseded — #5618 is fixed and closed, and the parts of this branch that mattered landed by other routes.

- **The fix shipped in four commits** — 891858567 and 5e52b6098 (both `fix(kafka)!:`), plus e5f47c779 and c78ad7adb, which name #5618's Case 1 and Case 2 directly.

- **The log-layer regression tests landed separately**, folded into `modules/kafka/src/test/kotlin/xtdb/api/log/KafkaClusterTest.kt` alongside the happy-path `tailAll` and `openGroupSubscription` tests they guard. That is the right home for them: the seek-to-zero-on-truncated-log behaviour lives in generic Kafka-log code rather than anything postgres-source specific.

- **The close-during-rebalance deadlock traced in the investigation note above is captured as Instance 5 of #5711**, with the same cycle — the `runBlocking` bridge in `onPartitionsAssigned` driving `awaitReplicaMsgId` against a replica `StateFlow` that teardown has stopped advancing.

What remains unique to this branch is the three-node Kafka + MinIO + Postgres harness in `PostgresSourceMultiNodeTest.kt`, which has no equivalent on `main`. That is being let go rather than rebased: `main` has since inserted around 430 lines across this branch's insertion point in `KafkaClusterTest.kt`, so the conflict is a rewrite rather than a resolve. Noting it here so the harness is findable if we want that shape of test later.

PR #5641, opened 2026-05-25
feat(durable-streams): proto, atomic producer, integration test

timhogan:
## Summary

Fills in the deferred pieces from the initial `xtdb-durable-streams` module so it covers the same surface as Kafka for multi-node deployments and atomic submission.

- **`writeTo` / `fromProto`** — wired through a new `DurableStreamsLogConfig` proto (`cluster_alias`, `topic`, `epoch`, `replica_topic`, `tenant`).
  Follower nodes can reconstruct the log from the database catalog without recreating the underlying DS stream.

- **`openAtomicProducer`** — follows `InMemoryLog`'s buffered-tx model: `appendMessage` queues into the tx; `commit()` flushes synchronously with monotonic `Producer-Seq` headers so transport retries dedupe per DS §5.2.1; `abort()` drops the buffer without hitting the network.
  `Producer-Id` is suffixed with a UUID per instance to avoid stale-epoch fencing across restarts (DS treats `Producer-Epoch` monotonically per `Producer-Id`).

- **`DurableStreamsLogIntegrationTest`** — exercises the real WorkerDO via the scratchpad Node shim (`slice-1/worker/ds_node_shim.mjs`) as a subprocess.
  The shim re-uses the production DO implementation against in-memory state, so this exercises the same logic CF deploys minus the DO storage layer.
  Tests auto-skip via `assumeTrue` when `node` or the shim are not present, so this stays green on developer machines without the spike checkout.

## Test plan

- [x] `./gradlew :modules:xtdb-durable-streams:test` — 20 passed / 0 failed / 0 skipped (10 unit, 10 integration).
- [ ] `./gradlew :xtdb-core:test :xtdb-api:test` — adjacent module check (running).
- [ ] No reflection or boxed-math warnings.

### DIFF
--- a/modules/durable-streams/build.gradle.kts
+++ b/modules/durable-streams/build.gradle.kts
@@ -1,0 +1,37 @@
+plugins {
+    `java-library`
+    alias(libs.plugins.clojurephant)
+    `maven-publish`
+    signing
+    alias(libs.plugins.kotlin.jvm)
+    alias(libs.plugins.kotlin.serialization)
+    alias(libs.plugins.dokka)
+}
+
+publishing {
+    publications.create("maven", MavenPublication::class) {
+        pom {
+            name.set("XTDB Durable Streams")
+            description.set("XTDB Durable Streams log module")
+        }
+    }
+}
+
+java.toolchain.languageVersion.set(JavaLanguageVersion.of(21))
+
+dependencies {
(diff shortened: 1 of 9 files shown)


timhogan:
Closing — pushed in error to the wrong target.

Issue #5079, opened 2025-12-04
FlightSQL: Implement session management


jarohen:
## Summary

Implement Flight SQL session management extensions (added in Arrow 16.0.0) to allow clients to set session-level options.

## Context

Currently `XtdbProducer` improvises connection management in two opposite directions and neither matches the FlightSQL session model. `defaultConnections: ConcurrentHashMap<DatabaseName, XtdbConnection>` (`XtdbProducer.kt:146`) shares one `XtdbConnection` across every autocommit RPC for a database, regardless of client. `beginTransaction` (`XtdbProducer.kt:612`) calls `node.connect()` per transaction with no continuity across txs or with prior autocommit work. The autocommit path is under-isolated; the tx path is over-isolated.

This was previously workable because the shared default connection had no live mutable state — `autoCommit` stayed `true`, `pendingOps` stayed empty, `dbName` was set once. #5609 adds `awaitToken` as the first such field, surfacing the race: concurrent FlightSQL writers lose token-merge updates against each other, and `awaitAll`'s target gets dragged forward by other clients' writes so reads can hit `tx-timeout` even when their own writes are long indexed. #5648 moves the field onto an embedded `xtdb.Session` but doesn't change the sharing.

The FlightSQL spec defines the right primitive: "Servers may also combine other connection state opaquely with the session token... e.g. CloseSession would also invalidate any authentication context persisted via the session context." `awaitToken`, `autoCommit`, current schema, time zone, and any future session parameter are all examples of state that should live on a session, not a connection.

## Methods to implement

Override these methods in `XtdbProducer`:

- `setSessionOptions` - Set server session option(s)
- `getSessionOptions` - Get server session option(s)
- `closeSession` - Close/invalidate the session

## Session options to support

Check pgwire implementation for the session options we currently support there - we should aim for parity.

## Implementation notes

Arrow Flight Java ships `ServerSessionMiddleware` as the reference mechanism — it handles cookie parsing (`arrow_flight_session_id`), session-id generation, lookup, and `Set-Cookie` emission. It gives session identity at call time; XTDB-side state hangs off our own map keyed by `middleware.getSession().id`.

- Add `ServerSessionMiddleware.Factory` to the `FlightServer.Builder` setup.
- Replace `defaultConnections` with a per-session `xtdb.Session` map (post-#5648, `Xtdb$XtdbInternal.openSession(dbName)` is the constructor). pgwire already wires sessions this way in `cmd-startup`.
[...]

PR #5648, opened 2026-05-26
feat: extract per-connection await-token into shared Session layer

tggreene:
## Summary

Extracts per-connection await-token tracking from `XtdbConnection` and `pgwire` into a shared `xtdb.Session` primitive.
One implementation, two consumers.

Addresses @jarohen's review on #5609:
> if pgwire and ADBC both require the same behaviour (and a hypothetical third protocol might too) then it's worth instead moving this functionality down a layer.

## Stack

Stacked on #5609 (base shows 4 commits until that lands; the last two are this PR).
Merges cleanly once #5609 lands.

## Shape

`xtdb.Session` owns a connection-scoped `awaitToken: String?`, refreshes it after every `submitTx` / `executeTx`, and threads it through `openSqlQuery` / `prepareSql` / `openSnapshot`.

`Session.Node` is a pure executor SPI — `submitTx`, `executeTx`, `openSqlQuery`, `prepareSql`, `openSnapshot`.
The await-token bookkeeping lives on `Session` itself (`recordTx`), implemented natively in Kotlin over `Basis.mergeTxBasisTokens` / `encodeTxBasisToken` — no Clojure round-trip, and implementors of `Node` never see token mechanics.
`XtdbConnection.Node` extends `Session.Node`.

New `Xtdb.XtdbInternal.openSession(dbName)` factory; `impl.clj` grows a shared `->session-node` helper used by both `connect` and `openSession`, so the ADBC and pgwire bridges can't drift.

Two commits:

1. **`feat(adbc): extract per-connection await-token into shared Session layer`** — `Session` class + `SessionTest`, `XtdbConnection` delegates, `->session-node` helper + `connect` rewrite.
2. **`feat(pgwire): migrate per-connection await-token to shared Session layer`** — `openSession` factory, 13 `:await-token` references in pgwire migrated to a `:xt-session` in conn-state, `xtdb.basis` require dropped (merge math now lives behind `Session`).

## Review actions applied

Pre-empting the review patterns from #5605/#5609:

- **`mergeAwaitToken` removed from the `Session.Node` SPI** — it was a token-string helper masquerading as an executor method.
  Merge is now a `Session` state-transition (`recordTx(dbName, txId)`), so every `Node` implementor stays ignorant of token format.
- **Reify duplication collapsed** — `connect` and `openSession` shared a near-identical `Session.Node` reify; extracted `->session-node`.
- **`AutoCloseable` honoured at the ownership sites** — ADBC `XtdbConnection.close()` and pgwire `:msg-terminate` now close the session, rather than the type advertising a close contract nobody called.
- **`cmd-commit` local `tx-id` → `submitted-tx`** — it binds a `SubmittedTx`, not an id.

## Test plan

- `xtdb.SessionTest` (new): 5/5 — pins the read-your-writes token contract, monotonic accumulation, cross-db merge (the attach/detach primitive), and dbName retargeting.
- `xtdb.pgwire-test` + `xtdb.pgwire-protocol-test`: 192/192 green.
- `xtdb.api.AdbcTest` + `xtdb.FlightSqlAdbcTest`: 28/28 green.
- No reflection / boxed-math warnings.

## Scope

Out:

- `autoCommit` / `pendingOps` / `currentCatalog` state still lives in each consumer (`XtdbConnection`, pgwire's `:conn-state`).
  This PR scopes to await-token tracking only — that's what @jarohen's comment named, and it's the cleanest narrow win.
  Broader session-state extraction can follow.
- pgwire's `:session` config map (clock, parameters, characteristics) is unrelated and untouched — kept as `:xt-session` alongside to avoid conflating protocol session state with XTDB session state.

## Follow-ups

- Unblocks #5079 (FlightSQL session management): `openSession` is the factory it needs to replace `XtdbProducer.defaultConnections` with a per-FSQL-session `Session` map, fixing the await-token race on shared FSQL connections.

### DIFF
--- a/core/src/main/clojure/xtdb/node/impl.clj
+++ b/core/src/main/clojure/xtdb/node/impl.clj
@@ -119,24 +119,28 @@
        (submitTx [_ db-name ops opts] (.submitTx this-node db-name ops opts))
        (executeTx [_ db-name ops opts] (.executeTx this-node db-name ops opts))
 
-       (openSqlQuery [_ sql db-name]
-         (let [query-opts (-> {} (with-query-opts-defaults this-node db-name))]
+       (openSqlQuery [_ sql db-name await-token]
+         (let [query-opts (-> {:await-token await-token} (with-query-opts-defaults this-node db-name))]
            (-> (xtp/prepare-sql this-node sql query-opts)
                (.openQuery nil (QueryOpts. nil (:default-tz query-opts))))))
 
-       (prepareSql [_ sql db-name]
-         (let [query-opts (-> {} (with-query-opts-defaults this-node db-name))]
+       (prepareSql [_ sql db-name await-token]
+         (let [query-opts (-> {:await-token await-token} (with-query-opts-defaults this-node db-name))]
            (xtp/prepare-sql this-node sql query-opts)))
+
+       (mergeAwaitToken [_ existing-token db-name tx-id]
+         (basis/merge-tx-tokens existing-token (basis/->tx-basis-str {db-name [tx-id]})))
 
        (getColumnTypes [_ table snap]
          (when-let [^Database db (.databaseOrNull db-cat (.getDbName table))]
            (.getColumnTypes db table snap)))
 
-       (openSnapshot [_ db-name]
-         (-> (or (.databaseOrNull db-cat db-name)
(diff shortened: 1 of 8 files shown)


jarohen:
@tggreene let's review this on my return, I hadn't envisaged it would have this kind of impact on the public API 🙂

PR #5654, opened 2026-05-28
fix(kafka): isolate rebalance-listener failure to the failing topic

Akeboshiwind:
Follows on from #5650 (which made `SharedGroupConsumer` evict every subscription on poll-loop failure to unblock shutdown). That fix was right for "the whole poll loop died" but used the wrong granularity for the most common trigger: a single `SubscriptionListener.onPartitionsAssigned` failure (e.g. `LogProcessor`'s leader transition throwing) would propagate through Kafka's `ConsumerRebalanceListenerInvoker` as `KafkaException: User rebalance callback throws an error`, land in `pollingJob`'s outer catch, and tear down *every* subscriber sharing the consumer — taking down Kafka ingestion for every database on the node because one database's leader transition failed.

This PR moves listener failures from "tank everything" to "evict only this topic".

## Approach

The rebalance wrapper now runs the per-topic loop to completion (so healthy topics in the same rebalance still get set up — relevant on startup when several databases register before the first poll), collects failures into a `Map<String, Throwable>`, and throws a private `ListenerAssignmentFailures` carrier. The poll-loop's new `catch (KafkaException)` branch unwraps the carrier and evicts the named topics with their own per-topic causes via the existing `evictSubscriptions`. A `KafkaException` *without* the carrier (some other rebalance failure) is rethrown to preserve the existing `cleanupOnFailure` path.

## Decisions worth flagging

**Exception-as-carrier rather than a flag or a queued `Evict` command.**
The constraint forcing the choice is that `consumer.subscribe` (inside `applySubscriptions`) is unsafe from within Kafka's rebalance callback — Kafka's docs say so, and nesting `subscribe` inside an in-flight rebalance deadlocks the join state machine. *Something* has to communicate "evict these topics" out of the callback to a safe point. A flag-plus-pending-evictions list works but needs mutable state in both the callback and the poll loop, and is the shape an earlier draft had that we explicitly moved away from. A queued `Evict` command would leave a race window between the callback returning and the command being processed, during which `processor` would be null and records could NPE in the records-loop. The exception approach is synchronous — by the time we're back in the poll-loop catch, the rebalance has finished, no records have flowed, and we can call `evictSubscriptions` freely.

**Walking the cause chain to find the carrier.**

```kotlin
val failures = generateSequence<Throwable>(e) { it.cause }
    .filterIsInstance<ListenerAssignmentFailures>()
    .firstOrNull() ?: throw e
```

kotlinx coroutines' stacktrace-recovery copies exceptions as they cross suspend boundaries (`runInterruptible`, `Deferred.await`), with the original as the *copy's* cause — so our carrier ends up at depth two, not one. This is [documented kotlinx behaviour](https://github.com/Kotlin/kotlinx.coroutines/blob/master/docs/topics/debugging.md) and matches the pattern Spring Kafka's [`ErrorHandlingUtils.findRootCause`](https://docs.spring.io/spring-kafka/api/org/springframework/kafka/listener/ErrorHandlingUtils.html) uses for the same shape of problem. A single-level `e.cause` check landed amber in TDD before the chain walk turned it green.

## Tidy first

The first commit adds a `Map<String, Throwable>`-shaped overload to `evictSubscriptions` and makes the existing `Collection<String>, Throwable` form delegate to it via `topics.associateWith { cause }`. No behaviour change; lets the new caller hand its `Map` straight in instead of looping the existing form N times (N `applySubscriptions` round-trips per rebalance failure).

## Test plan

- [x] `./gradlew :modules:xtdb-kafka:integration-test --rerun-tasks` — all kafka integration tests pass on a fresh run.
- [x] Replaced `poll loop failure evicts all subscriptions with cause and unblocks shutdown` (premise no longer holds — listener throw doesn't tank loop) with `processRecords failure tears down poll loop and evicts all subscriptions`, an equivalent that exercises the same `cleanupOnFailure` path via a different trigger (a `RecordProcessor` that throws on first record). Catch-and-cleanup coverage preserved.
- [x] Added `listener failure on assignment evicts only that topic, healthy subscribers keep working` as the green test for the new behaviour — verifies the throwing subscriber surfaces its own failure (with the listener exception in its cause chain) while a sibling healthy subscriber on a different topic keeps processing fresh records.

## Out of scope

- Listener failures during `onPartitionsRevoked` (symmetric per-topic treatment would apply, but the revoked path doesn't tank the loop the same way today, and we don't have a failure mode in production driving it).
- Surfacing `CancellationException` stack provenance via `-Dkotlinx.coroutines.debug` — operational change, not code; same note as #5650.

---

Generated with Claude Code

### DIFF
--- a/modules/kafka/src/main/kotlin/xtdb/api/log/KafkaCluster.kt
+++ b/modules/kafka/src/main/kotlin/xtdb/api/log/KafkaCluster.kt
@@ -29,6 +29,7 @@ import org.apache.kafka.clients.consumer.OffsetAndMetadata
 import org.apache.kafka.clients.consumer.OffsetOutOfRangeException
 import org.apache.kafka.clients.producer.KafkaProducer
 import org.apache.kafka.clients.producer.ProducerRecord
+import org.apache.kafka.common.KafkaException
 import org.apache.kafka.common.TopicPartition
 import org.apache.kafka.common.errors.InterruptException
 import org.apache.kafka.common.errors.UnknownTopicOrPartitionException
@@ -152,6 +153,14 @@ class KafkaCluster(
         class Unregister(val topic: String, val cont: CancellableContinuation<Unit>) : GroupCommand
     }
 
+    private class ListenerAssignmentFailures(val byTopic: Map<String, Throwable>) :
+        RuntimeException(
+            byTopic.entries.joinToString(", ") { (topic, e) ->
+                "$topic (${e.javaClass.simpleName}: ${e.message})"
+            },
+            byTopic.values.firstOrNull(),
+        )
+
     @Suppress("UNCHECKED_CAST")
     private inner class SharedGroupConsumer : AutoCloseable {
         private val subscriptions = mutableMapOf<String, TopicSubscription<*>>()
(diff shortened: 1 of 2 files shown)


Akeboshiwind:
Closing in favour of https://github.com/xtdb/xtdb/commit/dc24fec16ac900e7154142cd666b7d7f44a640d0 (logging the specific partition)

PR #5662, opened 2026-06-01
refactor: scope-managed lifecycle for Database services

jarohen:
**Draft / WIP — an umbrella PR that will grow.**

XTDB's Database-owned components carry two different lifecycles tangled together. This PR moves them onto one uniform model:

- **State** (catalogs, the live index) — a synchronous `AutoCloseable` holding resources hydrated at startup. Unchanged.
- **Service** (the background processes) — torn down by cancelling the `CoroutineScope` that owns it, rather than via its own `close()` / `AutoCloseable`.

Two slices have landed so far:

1. **Garbage collectors.** `BlockGarbageCollector` and `TrieGarbageCollector` become a `class` + `runOn(scope)`, launched into and cancelled via `LeaderLogProcessor`'s `gcScope` — a `SupervisorJob` child so one GC's failure doesn't take the other down — with no self-owned `Job` and no `AutoCloseable`.
2. **Compactor.** `Compactor.ForDatabase`'s loop becomes a `runOn(scope)` Service owned by `Database`. Its `close()` shrinks to freeing the driver's child Arrow allocator — the one piece of genuine State it holds. Teardown order is load-bearing: the loop is cancelled-and-joined before the allocator closes, and `Database` registers that teardown with `safelyOpening` so an aborted `open` stops the loop too.

This is a **behaviour-preserving refactor**. Later slices extend the same pattern to the persister and `Database.open`/`close`.

The full reasoning for each slice — the State/Service distinction, the deterministic-dispatcher constraint that shaped the dispatcher wiring, the allocator teardown ordering, and the alternatives tried and rejected — is in the per-slice comments below.
## Progress

**Status**: in-progress

- [x] Slice 1 — garbage collectors (`Block`/`Trie`GC → `runOn`, owned by `gcScope`)
- [x] Slice 2 — compactor (`ForDatabase` loop → `runOn`, owned by `Database`)
- [x] Slice 3 — `LeaderLogProcessor` → scope-managed Service (`runOn`, owner-cancelled, pure uniform cancellation)
- [x] Slice 4 — follower/transition procs + `LogProcessor` wrapper: per-term `SupervisorJob`, handler-owned cancels, de-suspend the `close()` chain
- [ ] Follow-on — resignation-on-failure (unify leader subtree + resign hook + Kafka relinquish)
- [ ] Follow-on — make `Database.close()` suspend, removing the last `runBlocking` teardown bridge

### Open Questions

- [ ] `compactAll()` parks its waiter in a single shared `compactAllPromise` that isn't cancelled if the loop is torn down mid-flight, so an in-flight caller waits out its `compactAllSync` timeout instead of failing fast. Fixing it is a genuine behaviour change (the GC loop already cancels its waiters), so it's deferred to its own slice — see the compactor comment.

### DIFF
--- a/core/src/main/kotlin/xtdb/compactor/Compactor.kt
+++ b/core/src/main/kotlin/xtdb/compactor/Compactor.kt
@@ -26,7 +26,6 @@ import xtdb.trie.*
 import xtdb.util.*
 import java.nio.channels.ClosedByInterruptException
 import java.time.Duration
-import kotlin.time.Duration.Companion.seconds
 import kotlin.use
 
 typealias JobKey = Pair<TableRef, TrieKey>
@@ -52,6 +51,9 @@ interface Compactor : AutoCloseable {
     }
 
     interface ForDatabase : AutoCloseable {
+        /** Launches the compaction loop into [scope]; the caller owns its lifetime — cancelling [scope] stops it. */
+        fun runOn(scope: CoroutineScope)
+
         fun signalBlock()
         suspend fun compactAll()
 
@@ -167,9 +169,6 @@ interface Compactor : AutoCloseable {
 
         override fun openForDatabase(allocator: BufferAllocator, dbStorage: DatabaseStorage, dbState: DatabaseState, watchers: Watchers) = object : ForDatabase {
 
-            private val dbJob = Job()
-            private val scope = CoroutineScope(dbJob + dispatcher)
-
             private val trieCatalog = dbState.trieCatalog
             private val liveIndex = dbState.liveIndexOrNull
 
@@ -197,10 +196,12 @@ interface Compactor : AutoCloseable {
                 }
             }
 
-            init {
-                val jobsScope = CoroutineScope(SupervisorJob(dbJob) + jobsDispatcher)
+            // Pinned to the compactor's own [dispatcher] (the owner scope supplies only the lifetime),
+            // so the seeded `DeterministicDispatcher` injected in the simulation tests drives the loop.
+            override fun runOn(scope: CoroutineScope) {
+                val jobsScope = CoroutineScope(SupervisorJob(scope.coroutineContext.job) + jobsDispatcher)
 
-                scope.launch(CoroutineName("outer loop")) {
+                scope.launch(dispatcher + CoroutineName("outer loop")) {
                     while (true) {
                         availableJobs =
                             jobCalculator.availableJobs(trieCatalog)
@@ -273,15 +274,10 @@ interface Compactor : AutoCloseable {
                 liveIndex?.refreshSnap()
             }
 
+            // State teardown only: the loop is a Service, stopped by the owner cancelling its scope
+            // before this runs (so no buffer is live when the driver's child allocator closes).
             override fun close() {
-
-                runBlocking {
-                    withTimeoutOrNull(10.seconds) { dbJob.cancelAndJoin() }
-                        ?: LOGGER.warn("failed to close compactor cleanly in 10s")
-                }
-
                 driver.close()
-
                 LOGGER.debug("compactor closed")
             }
         }
@@ -293,6 +289,7 @@ interface Compactor : AutoCloseable {
         @JvmField
         val NOOP = object : Compactor {
             override fun openForDatabase(allocator: BufferAllocator, dbStorage: DatabaseStorage, dbState: DatabaseState, watchers: Watchers) = object : ForDatabase {
+                override fun runOn(scope: CoroutineScope) = Unit
                 override fun signalBlock() = Unit
                 override suspend fun compactAll() = Unit
                 override fun close() = Unit
--- a/core/src/main/kotlin/xtdb/database/Database.kt
+++ b/core/src/main/kotlin/xtdb/database/Database.kt
@@ -52,6 +52,7 @@ import io.micrometer.core.instrument.MeterRegistry
 import java.time.Duration
 import java.util.*
 import kotlin.concurrent.Volatile
+import kotlin.time.Duration.Companion.seconds
 
 private val LOG = Database::class.logger
 
@@ -62,6 +63,7 @@ class Database(
     val isIndexing: Boolean,
     val partitions: Map<Int, DatabasePartition>,
     private val meterRegistry: MeterRegistry?,
+    private val compactorScope: CoroutineScope? = null,
     private val job: Job? = null,
     private val processor: LogProcessor? = null,
     private val registeredGauges: List<Gauge> = emptyList(),
@@ -133,6 +135,11 @@ class Database(
     override fun close() {
         meterRegistry?.let { reg -> registeredGauges.forEach { reg.remove(it) } }
         runBlocking {
+            // Stop the compaction loop + its jobs before `compactorOrNull.close()` frees the
+            // driver's child allocator below — an outstanding compaction buffer would otherwise
+            // outlive its allocator. Bounded so a stuck job can't hang shutdown.
+            withTimeoutOrNull(10.seconds) { compactorScope?.coroutineContext?.job?.cancelAndJoin() }
+                ?: LOG.warn("compactor did not stop within 10s")
             job?.cancelAndJoin()
             processor?.close()
         }
@@ -267,6 +274,20 @@ class Database(
                 else compactor.openForDatabase(allocator, storage, state, watchers)
             }
 
+            // Independent root owned by this Database: cancelled (bounded) in `close()` before the
+            // compactor's driver allocator is freed. NOOP's `runOn` is a no-op (read-only nodes).
+            // Registered with `safelyOpening` so a failure later in `open` also stops the loop —
+            // it runs against the driver, which `safelyOpening` would otherwise free out from under it.
+            val compactorScope = CoroutineScope(Job())
+            open {
+                AutoCloseable {
+                    runBlocking {
+                        withTimeoutOrNull(10.seconds) { compactorScope.coroutineContext.job.cancelAndJoin() }
+                    }
+                }
+            }
+            compactorForDb.runOn(compactorScope)
+
             var processor: LogProcessor? = null
             val job = Job()
             val scope = CoroutineScope(job + CoroutineExceptionHandler { _, e ->
@@ -372,6 +393,7 @@ class Database(
                 isIndexing = indexerConfig.enabled,
                 partitions = mapOf(0 to partition),
                 meterRegistry = meterRegistry,
+                compactorScope = compactorScope,
                 job = job,
                 processor = processor,
                 registeredGauges = gauges,
--- a/core/src/main/kotlin/xtdb/database/Database.kt
+++ b/core/src/main/kotlin/xtdb/database/Database.kt
@@ -330,9 +330,23 @@ class Database(
                             flushTimeout = indexerConfig.flushDuration,
                         )
 
+                        // This Database owns the leader's process tree: a SupervisorJob child of
+                        // its scope, so a leader failure is isolated from the rest of the node.
+                        // Step-down cancels it (bounded), then frees the processor's resources —
+                        // cancellation leaves the watermarks consistent-or-behind, so the follower
+                        // the transition opens next replays any gap idempotently.
+                        val leaderScope =
+                            CoroutineScope(scope.coroutineContext + SupervisorJob(scope.coroutineContext.job))
+                        leaderProc.runOn(leaderScope)
+
                         return object : LogProcessor.LeaderSystem {
                             override val proc get() = leaderProc
-                            override suspend fun close() { leaderProc.close(); replicaProducer.close() }
+                            override suspend fun close() {
+                                withTimeoutOrNull(10.seconds) { leaderScope.coroutineContext.job.cancelAndJoin() }
+                                    ?: LOG.warn("[$dbName] leader did not stop within 10s")
+                                leaderProc.close()
+                                replicaProducer.close()
+                            }
                         }
                     }
 
--- a/core/src/main/kotlin/xtdb/database/Database.kt
+++ b/core/src/main/kotlin/xtdb/database/Database.kt
@@ -330,20 +330,14 @@ class Database(
                             flushTimeout = indexerConfig.flushDuration,
                         )
 
-                        // This Database owns the leader's process tree: a SupervisorJob child of
-                        // its scope, so a leader failure is isolated from the rest of the node.
-                        // Step-down cancels it (bounded), then frees the processor's resources —
-                        // cancellation leaves the watermarks consistent-or-behind, so the follower
-                        // the transition opens next replays any gap idempotently.
-                        val leaderScope =
-                            CoroutineScope(scope.coroutineContext + SupervisorJob(scope.coroutineContext.job))
-                        leaderProc.runOn(leaderScope)
-
+                        // The wrapper drives this system's lifecycle: `runOn(termScope)` starts the
+                        // leader's process tree, the wrapper cancels that scope on step-down / teardown,
+                        // and `close()` then frees the leader's resources (the loops have already stopped,
+                        // so the watermarks are final).
                         return object : LogProcessor.LeaderSystem {
                             override val proc get() = leaderProc
-                            override suspend fun close() {
-                                withTimeoutOrNull(10.seconds) { leaderScope.coroutineContext.job.cancelAndJoin() }
-                                    ?: LOG.warn("[$dbName] leader did not stop within 10s")
+                            override fun runOn(scope: CoroutineScope) = leaderProc.runOn(scope)
+                            override fun close() {
                                 leaderProc.close()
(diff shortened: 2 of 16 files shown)


jarohen:
### Chalk — scope-managed Database service lifecycle (context)

Background reasoning for this PR, since the description is intentionally thin. This is an umbrella for **one lifecycle pattern applied across the Database**; the first commit does the garbage collectors.

- [x] The State / Service distinction
- [x] Chosen shape — construct-inert + `runOn(scope)`, owner-cancelled
- [x] Dispatcher wiring & the `DeterministicDispatcher` constraint
- [x] Why the persister stays graceful (abort-safety vs memory-safety)
- [x] Alternatives considered and rejected
- [ ] Remaining slices

<details><summary>The State / Service distinction</summary>

The migration rests on splitting Database-owned components into two kinds, by a single test — **what does it own?**

- **State** (catalogs, the live index): a *container* of resources hydrated at startup. Teardown frees what it holds, **synchronously**. Stays `AutoCloseable`. (`LiveIndex.close()`'s 60s `snapRefCounter` wait is the extreme case — still a synchronous blocking wait, not a coroutine join, so it stays on the State side.)
- **Service** (the background processes): torn down by **cancelling the `CoroutineScope` that owns it**, never via its own `close()` / `AutoCloseable`.

The discriminator is the *shape of teardown*: free-held-resources-synchronously → State; cancel-a-running-process → Service. It's self-correcting — the moment a "State" component needs a *suspending* teardown, that's the tell it's actually running a process and belongs on the Service side.

</details>

<details><summary>Chosen shape — construct-inert + runOn(scope), owner-cancelled</summary>

A Service is constructed **inert** (channels only — nothing that needs closing) and started with `runOn(scope)`, which launches its loop into a `CoroutineScope` the owner provides and returns the `Job`. No self-created root `Job`, no `AutoCloseable`.

For the GCs, the owner is `LeaderLogProcessor`. It holds `gcScope` — a `SupervisorJob` child of its own scope — `runOn`s both GCs into it, and cancels it on `close()` (bounded at 5s). One cancellation point tears both GCs down, on node shutdown *and* on leader→follower step-down (both route through `close()`).

**Failure isolation is the owner's responsibility**, not the service's — hence the `SupervisorJob` at `gcScope`, so a failure in one GC doesn't cancel its sibling. A service only owns isolation for its *own* sub-jobs.

</details>

<details><summary>Dispatcher wiring & the DeterministicDispatcher constraint</summary>

The GCs' bounded-parallelism IO pools (`tableDispatcher` / `deleteDispatcher`) are **constructor parameters defaulting to `Dispatchers.IO.limitedParallelism(n)`**, not pinned internally.

This is deliberate and easy to get wrong. The simulation tests run on a single-threaded, **seeded `DeterministicDispatcher`** with a hard cross-thread `check`. They inject that dispatcher so the GCs' delete work stays on the simulation's thread. Pinning the work to real `Dispatchers.IO` lets those coroutines escape the seeded scheduler and trips the cross-thread check — a **hard crash**, not a subtle flake (and it would surface the instant the sim runs, not at review).

Keeping `limitedParallelism` *in the default value only* (so it's never applied to a test dispatcher, which would throw) is what lets one parameter serve both the production bound and the deterministic simulation.

</details>

<details><summary>Why the persister stays graceful (abort-safety vs memory-safety)</summary>

A later slice touches `LeaderLogProcessor`'s persister, so we validated whether it could be abruptly cancelled. The finding:

- **Durable consistency is abort-safe**: the commit point is the replica-log append, *before* the in-memory live-index mutation; an aborted tx is cleanly redone by replay. So an abrupt abort keeps the *database* consistent.
- **But the binding constraint is in-process memory safety**: the persister holds Arrow buffers tied to the database allocator, and teardown must drain-then-`allocator.close()`. Abrupt cancel mid-tx would free the allocator out from under live off-heap buffers.

So the persister keeps its graceful drain (close channels → join → close allocator) and is **not** abruptly cancelled. The State/Service line holds; the persister's service-ness is just bounded by a real memory-lifecycle constraint.

</details>

<details><summary>Alternatives considered and rejected</summary>

- **`suspend fun close()` + a `closeSuspend()` `AutoCloseable` bridge** — make services suspend-closeable while staying `AutoCloseable`. Rejected: it reintroduces a `runBlocking` bridge per service (the smell we're removing) and invents a non-idiomatic `closeSuspend`. Scope-cancellation removes the need for a close method at all.
- **Dissolving each GC into `interface` + `companion fun CoroutineScope.runBlockGc(...)`** (a closure, no inert object). Tried, and it compiled cleanly — the imported companion member-extension resolves fine. Reverted for **consistency**: the trie GC's `garbageCollectTries(asOf)` is driven *directly* by the simulation with an explicit cutoff and can't collapse into a fire-and-forget closure. One pattern that fits both GCs (`class` + `runOn`) beats a tidier one that fits only the block GC.

</details>

<details><summary>Remaining slices</summary>

**Remaining slices:**

- [x] **GCs** — `3a4627c8`. The block + trie GCs became pure Services (scoped, no `AutoCloseable`).
- [x] **Compactor + `Database` takes a scope** — `56a820472`. Shipped as a *hybrid*, which refined the model (below).
- [ ] **Persister + uniform cancellation** — still gated behind the buffer-escape audit; the persister keeps its graceful drain until that clears.

**Model refinement (Slice 2):** the State/Service binary became a three-way split. The compactor is neither — it runs a process *and* owns a resource (its `driver` + a child allocator). So rather than dropping `AutoCloseable` as the GCs did, `Compactor.ForDatabase` keeps it: `close()` does the synchronous resource-free (`driver.close()`), while the process is torn down by cancelling the scope it's launched into. The three kinds are now: pure State (catalogs — `AutoCloseable`, no process), pure Service (the GCs — scoped, no resources), and hybrid (the compactor — both).

`Database.close()` now cancels its scope first (stopping the compactor's process, bounded at 10s), then `processor.close()` for the persister's graceful drain, then the synchronous `closeAll` of the resources (compactor included, child allocators before the parent). An earlier cut freed the driver via the outer loop's `invokeOnCompletion` and dropped `AutoCloseable` from `ForDatabase` entirely — reverted, because keeping `AutoCloseable` lets `safelyOpening` roll the compactor back on a partial-open failure, leaving the open-failure path only needing to cancel-and-join the process.

Behaviour-preserving: 1900+ seeded simulation cases (CompactorSimulationTest, NodeSimulationTest, LogProcessorSimTest) green with no teardown deadlock; the full `./gradlew test` is running as the final gate.

</details>

jarohen:
### Chalk — compactor slice + ownership handover

Picking this PR up from slice 2 onward. This comment covers the compactor slice and a fix to the GC slice; the context comment above covers slice 1's design.

- [x] Compactor `ForDatabase` → scope-managed Service
- [x] Fix a trie-GC leak in the GC slice (`runOn(scope)` → `runOn(gcScope)`)
- [x] Register open-failure teardown with `safelyOpening`
- [ ] Full `./gradlew test` re-run (draft gate)

<details><summary>Compactor ForDatabase → scope-managed Service</summary>

The compactor is the trickiest Service in this PR because it's a *composite*. The `while(true)` outer loop and its per-job coroutines are a Service, but the `driver` it holds owns a child Arrow allocator and a `SegmentMerge` temp dir — genuine State that must be freed synchronously. So the State/Service split runs *through* `ForDatabase`: the loop moves to `runOn(scope)`, and `close()` survives but shrinks to just `driver.close()`. Unlike the GCs, we don't drop `close()` — there's a real resource on the other side of it.

The allocator is what makes teardown ordering load-bearing. `driver.close()` runs `al.close()`, which throws if any compaction buffer is still live, so the loop must be cancelled-and-joined before the driver closes. `Database` owns that ordering: it holds the `compactorScope` (an independent root, mirroring the old self-owned `dbJob`) and cancels it with a bounded 10s wait at the top of `close()`, before `closeAll()` reaches the driver. `runOn` pins the loop to the compactor's own dispatcher, so the owner's scope supplies only the lifetime, not the thread — which is what preserves the simulations' seeded single-threaded `DeterministicDispatcher`.

</details>

<details><summary>Open-failure teardown via safelyOpening</summary>

The loop starts inside `Database.open`, well before the constructor returns, so a failure later in `open` would leave it running against a driver that `safelyOpening` has already freed. So the `compactorScope` teardown is registered with `safelyOpening` too; since those run in reverse order of registration, an aborted open cancels-and-joins the loop before the driver closes — the same ordering guarantee as the happy-path `close()`.

</details>

<details><summary>Fix: trie-GC leak in the GC slice</summary>

Running the full suite surfaced 6 `ExternalSourceTest` failures — `runTest` blocks timing out at ~60s with `UncompletedCoroutinesError`. The root cause was in the GC slice, not the compactor: the block GC was launched on `gcScope` (which `LeaderLogProcessor.close()` cancels) but the trie GC on the bare `scope`, which `close()` never cancels. In production the parent scope cancels it on shutdown, so this was invisible; under `runTest`, which owns the scope and only calls `lp.close()`, the trie-GC loop survived and tripped the leak check.

The fix: the trie GC runs on `gcScope` too, matching the block GC and the GC commit's own stated design ("a SupervisorJob child so one GC's failure does not take the other down"). It also closes a latent failure-isolation gap — a trie-GC failure was cancelling the shared `scope`, taking the persister and ext-source pump down with it. Folded into the GC commit; `ExternalSourceTest` now passes in 2.6s.

</details>

<details><summary>Tests</summary>

Slice-2 + GC-fix verification is green: `xtdb.compactor.*`, `NodeSimulationTest`, `LeaderLogProcessorTest`, `FollowerLogProcessorTest`, `LogProcessorSimTest`, the Clojure `compactor_test` (real `Database.open`/`close`), and `ExternalSourceTest` — no allocator-closed errors, no reflection or boxed-math warnings. A full `./gradlew test` re-run after the trie-GC fix is the remaining draft gate.

</details>

<details><summary>Handover note for slice 3</summary>

Ownership of this PR changed hands mid-flight; slice 3 (persister + `Database.open`/`close`) remains. One dead end worth recording: an earlier attempt to make `Database` *take* a scope from `DatabaseCatalog` made `Database`'s own scope inherit the catalog's dispatcher — which, under `DatabaseCatalogTest`'s injected `StandardTestDispatcher`, livelocked the `runBlocking`-bridge `close()` (the virtual clock can't advance while the bridge blocks the only thread). `Database.close()` fundamentally needs its teardown coroutines on a real dispatcher. The compactor slice sidesteps this by giving `compactorScope` an independent root (defaulting to `Dispatchers.Default`) rather than inheriting one; slice 3 should pin `Database`'s scope to a real dispatcher for the same reason.

</details>

jarohen:
### Chalk — slice 3: LeaderLogProcessor as a scope-managed Service

Landed in `781ac3bba`. Full `./gradlew test` green (1504 tests, 0 failures) — no `UncompletedCoroutinesError`, no allocator-closed errors, leader↔follower transitions and multi-node paths clean.

- [x] LLP: drop self-owned `scope`/`ctx`, add `runOn(scope)`, shrink `close()` to the allocator
- [x] LogProcessor / `Database.openLeaderSystem`: own a per-term `leaderScope`, cancel-then-free on step-down
- [x] Migrate tests (ExternalSourceTest, LeaderLogProcessorTest, LogProcessorTest, LogProcessorSimTest) off `ctx`/`close()` onto `runOn`/cancel
- [x] Full `./gradlew test`

<details><summary>The finding that shaped it — the persister is interruptible by construction</summary>

LLP used to self-own `scope`/`gcScope` and tear down via a `suspend close()` that drained channels → joined the persister → freed its allocator, reached from `Database.close()` through a `runBlocking` bridge. We checked whether that graceful drain is load-bearing. It isn't.

The only suspension point in any persister task is the replica-log append itself. `importTx`, `deleteTries`, `nextBlock`, `notifyTx`, `isFull` are all synchronous, so a cancel can't land between a commit and its in-memory effect; they run as one synchronous continuation. A cancelled append leaves `latestReplicaMsgId` *behind* the replica log (the `.also` that advances it never runs), never ahead — and behind is always safe, because the follower the transition opens next resumes from that watermark and replays the gap idempotently. The cross-task invariant (no `finishBlock` snapshotting the catalog between a GC's `TriesDeleted` append and its catalog mutation) is held by serialization — the single `select` loop, one task at a time — which cancellation doesn't disturb. The one post-commit suspension, the block upload, is covered by `pendingBlock` + idempotent re-finish on transition. So cancellation is just a clean crash, and the leader is already crash-safe at every instruction.

**Load-bearing invariant:** this rests on the live-index being mutated by a *single* `importTx` per tx (it used to be mutated incrementally across a tx — that's what historically forced the drain). If incremental live-index mutation is ever reintroduced, the guarantee breaks and interruptibility must be re-derived.

</details>

<details><summary>Shape — owner-cancelled subtree</summary>

LLP is constructed inert; `runOn(scope)` launches the leader's process tree — both GCs (on a `SupervisorJob` child so one GC's failure doesn't take the other down), the ext-source pump, and the persister — into the supplied scope. `Database.openLeaderSystem` owns that scope (a `SupervisorJob` child of the database scope, so a leader failure is isolated from the rest of the node); step-down cancels it (bounded) and then frees the processor's resources. `close()` shrank to State teardown only — close the external source, close the child allocator — with no cancellation of its own.

</details>

<details><summary>Scope vs follow-ons</summary>

Behaviour-preserving: GCs keep sibling-isolation, the failure hook stays `notifyError`, and `Database.close()` keeps its bounded `runBlocking` this slice. Still out, and coupled:

- **Resignation-on-failure** — a fatal leadership-process failure should *resign* leadership (step down to follower) rather than `notifyError` the node; that's what would justify collapsing the GC isolation so any failure cancels the whole leader subtree. Needs a Kafka mechanism to voluntarily relinquish the partition.
- **Consolidate cancellation to one point** — the end-state is no `runBlocking { … cancelAndJoin() }` and no `cancel()` in any `close()`, save a single cancel in `Database` itself (plus the transition-scoped cancels that live in the rebalance handlers, not in `close()`). Slices 4 (the other processors + the LogProcessor wrapper) and 5 (consolidate up to Database) get there.

</details>

jarohen:
### Chalk — slice 4: follower/transition procs + the `LogProcessor` wrapper

- [x] Processors are State → model them as `AutoCloseable`; adopt `use {}` at the call sites
- [x] Wrapper owns a per-term `SupervisorJob`; steady-state cancels relocated into the rebalance handlers
- [x] SubSystems carry `runOn` (the compactor split); eager follower retained — no `Inert`, no `lateinit`
- [x] Don't treat a cancelled follower as a failure (no spurious `replicaState` / watchers fault)
- [ ] Open question — the state/services back-edge (see below); not yet resolved

<details><summary>Shape — two commits</summary>

- `tidy(indexer): make the log processors AutoCloseable` — `Processor` extends `AutoCloseable`; its `close()` is synchronous State teardown (the `suspend` was a vestige of the deleted leader drain). Call sites take the payoff: `FollowerLogProcessorTest` gains a `withProcessor { … }` helper that frees via `use {}`, and the leader→follower handoff uses `openTransition(…).use { transition -> … }`.
- `refactor(indexer): make log-processor SubSystems scope-managed Services` — the structural change below.

</details>

<details><summary>SubSystems as runOn Services; the wrapper owns the term job</summary>

The leader/follower SubSystems follow the compactor's split — constructed inert, started by the owner via `runOn(scope)`, torn down by the owner cancelling the scope. `close()` is resource-only.

The wrapper owns a per-term `termJob` (a `SupervisorJob` child of the indexer scope); each role runs under it. A transition `termJob.cancelAndJoin()`s in the rebalance handler — transition logic, not a `close()` — and starts the next role under a fresh one. Rooting on a `SupervisorJob` (not a bare `launch` job) is load-bearing: a term-process failure surfaces via `notifyError` instead of cancelling the source-log subscription. The cancel already ran on the Kafka poll thread (it was inside `oldSys.close()`), so this relocates it rather than adding one.

</details>

<details><summary>Cancellation isn't failure</summary>

A cancelled follower is now the routine teardown (every transition cancels the term). Its replica-tail launch and `processRecords` guard `CancellationException` ahead of their generic `catch`, so a cancel doesn't fault the follower's `replicaState` or the shared watchers. The leader's ext-source pump and persister already did this; the follower didn't.

</details>

<details><summary>Open question — the state/services back-edge</summary>

`LogProcessor` is a role state-machine, and "the currently running term" — the live SubSystem plus its job — is its state. So the wrapper holds a Service *as* State, and reads it (the leader gauge, `gcAll()`). Owner-cancels sharpened this: to cancel from the handler, the owner must hold the term's job.

Options on the table, not yet landed: model the running term as one explicit value (e.g. `Term(sys, job)`, swapped atomically) and/or have `runOn` return its `Job`. Parked pending a call on whether to name it (it genuinely is a state machine — embrace it) or read it as a sign the lifecycle sits in the wrong place.

</details>

<details><summary>Status / out of scope</summary>

Verified: core suite + 300 `LogProcessorSimTest` rebalance iterations green. Branch not yet pushed. `Database.close()` and its `runBlocking` are untouched — slice 5. Pre-push gate still to run: full `./gradlew test` + Kafka `integration-test`.

</details>

PR #5663, opened 2026-06-02
fix(postgres-source): respect _valid_from/_valid_to on CDC rows

### LINKED ISSUE #5657: postgres-source: explicit `_valid_from` on a CDC row is ignored
The Postgres external source supports an explicit `_valid_from` column to stamp a row version's valid-time (`PostgresSource.writeOp`), but the supplied value is silently dropped for standard PG timestamp columns — the version ends up at the CDC ingest/system time instead of the instant we asked for.

### Root cause

`writeOp` reads the column with an `as? Instant` cast:

```kotlin
val explicitValidFrom = (docMap.remove("_valid_from") as? Instant)?.asMicros
...
explicitValidFrom ?: openTx.systemFrom
```

But no PG temporal type decodes to `java.time.Instant` in the CDC path: `PgType` decodes `TIMESTAMP` → `LocalDateTime` and `TIMESTAMPTZ` → `ZonedDateTime`/`OffsetDateTime`. So the `as? Instant` cast yields `null` and the explicit valid-time is never applied — it falls back to `openTx.systemFrom`.

### Repro (failing integration test, real Postgres)

```kotlin
@Test
fun `explicit _valid_from on a CDC row sets the version valid-time`() = runTest(timeout = 60.seconds) {
    val pubName = "test_pub_${UUID.randomUUID().toString().replace("-", "_")}"
    val slotName = "test_slot_${UUID.randomUUID().toString().replace("-", "_")}"
    val sourceTopic = "test-topic-${UUID.randomUUID()}"

    // empty at attach time, so the row streams through writeOp (where _valid_from is handled)
    pgExecute(
        "CREATE TABLE IF NOT EXISTS pg_vf (_id INT PRIMARY KEY, name TEXT, _valid_from TIMESTAMPTZ)",
        "CREATE PUBLICATION $pubName FOR TABLE pg_vf",
    )

    openNode(sourceTopic).use { node ->
        attachPostgresSource(node, slotName = slotName, publicationName = pubName)

        pgExecute("INSERT INTO pg_vf (_id, name, _valid_from) VALUES (1, 'Alice', TIMESTAMPTZ '2020-01-01 00:00:00+00')")

        awaitCondition("row ingested", timeout = 30.seconds) {
            runCatching { xtQueryDb(node, "cdc", "SELECT _id FROM public.pg_vf WHERE _id = 1").isNotEmpty() }
                .getOrDefault(false)
        }

        val document = xtQueryDb(node, "cdc", "SELECT *, _valid_from FROM public.pg_vf WHERE _id = 1").single()
        assertEquals("Alice", document["name"])

        val validFrom = (document["_valid_from"] as ZonedDateTime).toInstant()
        assertEquals(java.time.Instant.parse("2020-01-01T00:00:00Z"), validFrom)
    }
}
```

**Expected:** the version's `_valid_from` is `2020-01-01T00:00:00Z` (the value we supplied).
**Actual:** it's the CDC ingest/system time.

### Fix sketch

Coerce the decoded temporal (`ZonedDateTime`/`OffsetDateTime`/`LocalDateTime`) to an `Instant` in `writeOp` rather than casting `as? Instant`. Decide timezone handling for `TIMESTAMP` (no zone). The `_valid_to` path (also `as? Instant`) likely needs the same treatment.

Akeboshiwind:
Resolves #5657.

## Summary

`PostgresSource.writeOp` was reading `_valid_from`/`_valid_to` with `as? Instant`, but the pgoutput text path decodes `TIMESTAMPTZ` to `ZonedDateTime` and `TIMESTAMP` to `LocalDateTime`.
The cast yielded `null` and the explicit valid-time was silently dropped, falling back to `openTx.systemFrom`.

This PR coerces a `ZonedDateTime` to `Instant` directly.
Other types are rejected with `xtdb.postgres/invalid-explicit-valid-time`, which surfaces via `Database.ingestionError` rather than going on to do the wrong thing.

## Approach — why strict TIMESTAMPTZ-only

The pgwire ingest path *is* more permissive — `xtdb.time/->instant` resolves `LocalDateTime`/`LocalDate`/zone-less strings against the session timezone.
But the CDC pipeline has no session-zone equivalent: pgoutput is an async replication feed without any client-session context, so a zone-less temporal would force us to invent a default zone the user didn't supply.
That's exactly the silent-misinterpretation shape of the original bug, just better disguised, so we refuse at the boundary.

Operators wanting explicit valid-time on a CDC source declare the column as `TIMESTAMPTZ`.

## Test plan

Three integration tests in `PostgresSourceIntegrationTest`:

- `explicit _valid_from on a CDC row sets the version valid-time` — the issue's repro, now green.
- `explicit _valid_from and _valid_to on a CDC row sets the version bounds` — `_valid_to` coverage (the same `as? Instant` bug applied to it too).
- `non-TIMESTAMPTZ _valid_from fails the source loudly` — asserts `ingestionError` surfaces with our error code and the primary db survives.

## Future work

If a real demand for zone-less temporal columns appears, the accepted set can widen by letting the operator declare a default timezone in the database / Postgres source config.
That would mirror how pgwire uses the session timezone today: `LocalDateTime`/`LocalDate` would resolve against the configured zone via `xtdb.time/->instant`, and a string of zone-less ISO timestamps would parse against the same default.
Deliberately not done here — strict TIMESTAMPTZ-only is the right starting point, since it keeps the input unambiguous and gives us a clean, opt-in widening path later.

### DIFF
--- a/modules/postgres-source/src/main/kotlin/xtdb/postgres/PostgresSource.kt
+++ b/modules/postgres-source/src/main/kotlin/xtdb/postgres/PostgresSource.kt
@@ -31,6 +31,7 @@ import xtdb.time.InstantUtil.asMicros
 import xtdb.util.*
 import java.nio.ByteBuffer
 import java.time.Instant
+import java.time.ZonedDateTime
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.atomic.AtomicLong
 import com.google.protobuf.Any as ProtoAny
@@ -318,6 +319,16 @@ class PostgresSource(
     }
 }
 
+private fun explicitValidTimeMicros(name: String, value: Any?): Long? = when (value) {
+    null -> null
+    is ZonedDateTime -> value.toInstant().asMicros
+    else -> throw Incorrect(
+        "'$name' must be a TIMESTAMPTZ column",
+        errorCode = "xtdb.postgres/invalid-explicit-valid-time",
+        data = mapOf("column" to name, "type" to (value::class.qualifiedName ?: value::class.simpleName)),
+    )
+}
+
 private fun writeOp(
     openTx: OpenTx,
     dbName: String,
@@ -332,8 +343,8 @@ private fun writeOp(
             val id = docMap["_id"]
                 ?: throw Incorrect("Missing '_id' in row from ${op.schema}.${op.table}")
 
-            val explicitValidFrom = (docMap.remove("_valid_from") as? Instant)?.asMicros
-            val explicitValidTo = (docMap.remove("_valid_to") as? Instant)?.asMicros
+            val explicitValidFrom = explicitValidTimeMicros("_valid_from", docMap.remove("_valid_from"))
+            val explicitValidTo = explicitValidTimeMicros("_valid_to", docMap.remove("_valid_to"))
 
             if (explicitValidTo != null && explicitValidFrom == null)
                 throw Incorrect("'_valid_to' requires '_valid_from'")
--- a/modules/postgres-source/src/test/kotlin/xtdb/postgres/PostgresSourceIntegrationTest.kt
+++ b/modules/postgres-source/src/test/kotlin/xtdb/postgres/PostgresSourceIntegrationTest.kt
@@ -808,6 +808,105 @@ class PostgresSourceIntegrationTest {
         }
     }
 
+    @Test
+    fun `explicit _valid_from on a CDC row sets the version valid-time`() = runTest(timeout = 60.seconds) {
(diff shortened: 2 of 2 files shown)


Akeboshiwind:
Landed on `main` as [`b6b810d`](https://github.com/xtdb/xtdb/commit/b6b810df20117a4dc99915ad36ec4b80de0f09ca) — the PR shows "closed" rather than "merged" because the fork push happened after the upstream push, so GitHub never saw a merge event tied to the original PR head SHA. The fix is in: the SHA at the tip of `main` is identical to this PR's head.

PR #5666, opened 2026-06-02
feat(flight-sql): implement session management

### LINKED ISSUE #5079: FlightSQL: Implement session management

tggreene:
Resolves #5079.

Implements FlightSQL session management via Arrow's `ServerSessionMiddleware`, and replaces the improvised connection handling in `XtdbProducer` with per-session connection ownership.

## What it does

- Wires `ServerSessionMiddleware` into the FlightSQL server — session identity via the `arrow_flight_session_id` cookie (the mechanism the Go-driver-based ADBC clients already speak).
- `setSessionOptions` / `getSessionOptions` / `closeSession`:
  - `catalog` is settable and validated against known databases (how the Go-driver clients select a db, since they can't send the `x-xtdb-database` header).
  - `schema` is reported as `public`; only a confirming `public` no-op is accepted (XTDB resolves schema by qualification — settable schema/timezone params are a follow-up, not pretend-supported).
  - `getSessionOptions` is a pure getter and mints nothing.
  - `closeSession` evicts the session's connections and invalidates the cookie.
- DB resolution: `x-xtdb-database` header → session `catalog` option → default `xtdb`.
- Connection ownership: one autocommit/query connection per `(session, db)` for session-aware clients, with a bounded anonymous per-db fallback for cookieless clients. Transactions own a dedicated connection rather than flipping `autoCommit` on a pooled one — preventing cross-client buffering, abandoned-tx pool poisoning, and double-begin aliasing.

## Scope notes

- **No public node-API change.** Only consumes existing `node.connect()` / `allocator` / `XtdbInternal.dbCatalog`; the only `xtdb.api` edit is 2 lines of internal middleware wiring in `FlightSql.open()`.
- **Landing order.** Deliberately built on plain `main`, *before* the await-token stack (#5609 / #5648), so session-scoped state has a real session boundary to attach to before it's introduced. The per-`(session, db)` ownership is the seam #5648 can hang the await-token on without cross-client token drag. Recommend landing this first.

## Known follow-ups (not blocking)

- Abandoned-tx / cookieless-session reclamation (no TTL / disconnect hook today — bounded leak, strict improvement over the prior shared-connection corruption).
- `resolveDb` header-vs-catalog policy — harmless on this substrate (stateless conns), but wants deciding before #5648 puts a token on the connection.
- Concurrent DML on a single tx handle — belongs with #5648's per-connection serialisation.

## Tests

`xtdb.FlightSqlAdbcTest` (36) and `xtdb.flight-sql-test` (10) green, including regression guards for tx/autocommit isolation, open-tx not poisoning autocommit, double-begin isolation, `getSessionOptions` not minting, and catalog session routing.

### DIFF
--- a/core/src/main/kotlin/xtdb/flight_sql/XtdbProducer.kt
+++ b/core/src/main/kotlin/xtdb/flight_sql/XtdbProducer.kt
@@ -43,6 +43,7 @@ import xtdb.util.requiringResolve
 import xtdb.util.serializeAsMessageInterruptibly
 import xtdb.util.warn
 import java.util.*
+import java.util.concurrent.Callable
 import java.util.concurrent.ConcurrentHashMap
 
 private fun resultTypesToSchema(types: SequencedMap<String, VectorType>) =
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/xtdb/flight_sql/XtdbProducer.kt
```diff
@@ -459,19 +512,113 @@ class XtdbProducer(private val node: Xtdb) : NoOpFlightSqlProducer(), AutoClosea
...
+                        .toRuntimeException()
+                )
+                return
+            }
+
+            val sessionId = mw.session.id
+            // remove before close so an in-flight call can't resolve a connection being closed.
+            txConns.entries.removeIf { (_, txConn) ->
+                if (txConn.sessionId == sessionId) {
+                    txConn.conn.close()
```
mbutlerw, line 607: Nit: if `conn.close()` throws here (rare — connection mid-failure, etc.) the exception propagates out of the `removeIf` lambda and halts iteration. Later `txConns` entries get stranded in the map, the `sessionConns` cleanup below never runs, and `mw.closeSession()` never fires — the session ends up half-closed. Same shape on L612 (`sessionConns.remove(key)?.close()`).

Suggest wrapping each close in `runCatching` so cleanup runs to completion even when one connection fails to close:

```kotlin
runCatching { txConn.conn.close() }
    .onFailure { LOGGER.warn(it, "failed closing tx conn during closeSession") }
true
```

The test file already uses this pattern (`runCatching { fsqlClient.rollback(...) }` in the test cleanup paths), so it stays consistent with how the rest of the new code handles best-effort cleanup.

mbutlerw (approved):

tggreene:
**Model: `session = connection → transaction` (what this PR ships vs. the direction)**

Capturing where this lands after the discussion, so the stack stays coherent.

**Model.** A FlightSQL session maps onto the JDBC/general-database notion of a connection: a `(client-cookie, db)` pair → one `XtdbConnection`, holding `autoCommit`/`pendingOps` (and, per #5648, the await-token). A transaction is that connection in `autoCommit=false` mode.

**This PR ships the cross-session guard.** A tx handle resolves only under its owning session (`txConnFor`); presented under any other session it's `NOT_FOUND`. This blocks cross-session tx reference and keeps per-session cleanup tractable.

**Sessionless (cookieless) callers still work here**, resolved by handle alone (cookieless == cookieless). Deliberate: transactions predate sessions in Flight SQL (txs 10.0, sessions 16.0) and the reference example server resolves a tx by handle with no session check, so a sessionless multi-statement tx (its own dedicated connection, isolated by handle unguessability) is coherent — not something to refuse.

**Direction (follow-ups):** make "sessionless" an *explicit* mode with defined defaults rather than today's implicit shared-`defaultConns` fallthrough. A sessionless caller should get sensible defaults (db `xtdb`, schema `public`, latest-snapshot reads, no cross-call read-your-writes) and be isolated from other anonymous callers instead of sharing one connection. Read-your-writes and persisted options are session features, unavailable sessionless by definition — an explicit tradeoff, not an error. Forcing a session for a given operation stays a contingency, only if that op proves incoherent sessionless. This is downstream of #5648 (await-token primitive only) and the subsequent `XtdbProducer` rewire onto `openSession` — not part of either. The await-token must live on the connection, never the client-cookie.

PR #5676, opened 2026-06-04
refactor: job-owned lifecycle for Database services

jarohen:
Rebuilds the lifecycle of the Database's background services onto one model — **the coroutine job tree is the ownership tree.** Everything the Database owns is one of two kinds:

**Service** — has a background loop (the garbage collectors, the compactor, the log processors):

- takes a `CoroutineScope` from its owner and launches its loop in `init`;
- frees only what it itself opened via `job.invokeOnCompletion` — *not* a body `finally`, which leaks if the job is cancelled before its body runs;
- is **not** `AutoCloseable` and has **no** `close()` — owners cancel the scope, never closing a service by hand.

**State** — a synchronous holder with no loop (catalogs, `LiveIndex`, `TransitionLogProcessor`): stays `AutoCloseable`, closed explicitly.

Teardown then falls out of the tree: one cancel cascades leaf-first, each node freeing what it owns as it completes. The single synchronous bridge (`runBlocking { … cancelAndJoin() }`) migrates up one level per phase until only `Database.close()` holds it — `job.cancelAndJoin()`, then `closeAll()` of the State.

Builds on #5679 (merged), which made `SimLog` pure state and moved the simulation tests onto a single scope-cancellation teardown.

## Changes

1. **`refactor(gc)`** — the block & trie collectors take their lifetime scope from the owner instead of self-creating a `Job`, and drop `AutoCloseable`/`close()`. The loop runs on the injected scope; the delete fan-out keeps a separate injectable dispatcher.
2. **`refactor(compactor)`** — `openForDatabase(scope, …)`; the driver's child allocator + `SegmentMerge` temp dir are freed via `dbJob.invokeOnCompletion`; `ForDatabase` is no longer `AutoCloseable`; `compactorScope` is a child of the database `job`.
3. **`refactor(indexer)`** — `LeaderLogProcessor`/`FollowerLogProcessor` own their lifetime (injected term scope, `invokeOnCompletion`, no `cancelAndJoin()`); the follower absorbs its own replica tail. `TransitionLogProcessor` becomes synchronous `AutoCloseable` State (used via `.use`). The `LogProcessor` wrapper models the running term as one value — `SubSystem` carries its own `SupervisorJob`, swapped atomically on each role transition; the factory builds bare processors and the wrapper assembles the term.
4. **`refactor(database)`** — `Database.close()` becomes a single `job.cancelAndJoin()` that reaps the compactor, the source-log subscription and the live term leaf-first, then `closeAll()` of State. No Service retains a teardown method of its own.

Behaviour-preserving throughout: indexing, leader↔follower transitions, block finishing and GC behave exactly as before; only the lifecycle mechanism moves.

## Implementation notes

- **Unbounded teardown** — no `withTimeoutOrNull`/defensive bounds; structured concurrency does the ordering the hand-written teardown used to spell out.
- **Two dispatchers per IO service** — the loop rides the injected scope (in sims, the seeded deterministic dispatcher, so a cancel joins it on that same thread); the IO/delete fan-out takes a separate injectable dispatcher. Both default at the service boundary — the only caller that varies them (the simulations) injects there directly.
- **`invokeOnCompletion`, not `finally`** — a completion handler fires even if the job is cancelled before its body runs. Child allocators are freed `this@…`-qualified, since a bare `allocator` in an `init` lambda binds the constructor parameter (the parent), not the child field.

## Test plan

`:xtdb-core` unit suite + `ExternalSourceTest` + `DatabaseCatalogTest`; `LogProcessorSimTest` (≥300 rebalance iterations), `NodeSimulationTest`, `CompactorSimulationTest` — all green, no `UncompletedCoroutinesError`, allocator leaks, or dispatcher livelock. Whole-project `./gradlew test` + Kafka `integration-test` run on CI.

The migrating-entry-point model, the binding lessons, and a post-completion adversarial audit are in the comments below.

## Progress

**Status:** ready for review (base #5679 merged; full suite + Kafka integration on CI)

- [x] The pattern — job-tree-is-ownership-tree + migrating entry point
- [x] Phase 1 — garbage collectors
- [x] Phase 2 — compactor
- [x] Phase 3 — leader/follower processors + wrapper (the running term is `SubSystem`)
- [x] Phase 4 — `Database` single teardown cancel
- [x] Adversarial audit — aims confirmed; removed the orphaned `LogProcessor.cancelAndJoin()` (no Service retains a teardown method)
- [ ] Follow-up (separate, not blocking) — `NodeSimulationTest` over-specification + `CompactorMockDriver` fixture soundness

### DIFF
--- a/core/src/main/kotlin/xtdb/compactor/Compactor.kt
+++ b/core/src/main/kotlin/xtdb/compactor/Compactor.kt
@@ -26,7 +26,6 @@ import xtdb.trie.*
 import xtdb.util.*
 import java.nio.channels.ClosedByInterruptException
 import java.time.Duration
-import kotlin.time.Duration.Companion.seconds
 import kotlin.use
 
 typealias JobKey = Pair<TableRef, TrieKey>
@@ -51,7 +50,7 @@ interface Compactor : AutoCloseable {
         fun create(meterRegistry: MeterRegistry?, threads: Int): Compactor
     }
 
-    interface ForDatabase : AutoCloseable {
+    interface ForDatabase {
         fun signalBlock()
         suspend fun compactAll()
 
@@ -63,7 +62,7 @@ interface Compactor : AutoCloseable {
         }
     }
 
-    fun openForDatabase(allocator: BufferAllocator, dbStorage: DatabaseStorage, dbState: DatabaseState, watchers: Watchers): ForDatabase
+    fun openForDatabase(scope: CoroutineScope, allocator: BufferAllocator, dbStorage: DatabaseStorage, dbState: DatabaseState, watchers: Watchers): ForDatabase
 
     interface Driver : AutoCloseable {
         suspend fun executeJob(job: Job): TriesAdded
@@ -165,10 +164,14 @@ interface Compactor : AutoCloseable {
         private val jobsDispatcher = dispatcher.limitedParallelism(threadCount, "compactor")
         private val jobsSemaphore = Semaphore(threadCount)
 
-        override fun openForDatabase(allocator: BufferAllocator, dbStorage: DatabaseStorage, dbState: DatabaseState, watchers: Watchers) = object : ForDatabase {
+        override fun openForDatabase(scope: CoroutineScope, allocator: BufferAllocator, dbStorage: DatabaseStorage, dbState: DatabaseState, watchers: Watchers) = object : ForDatabase {
 
-            private val dbJob = Job()
-            private val scope = CoroutineScope(dbJob + dispatcher)
+            // Child of the owner scope: cancelling that scope stops the loop and its per-job
+            // coroutines, and the driver's resources (child allocator + SegmentMerge temp dir) are
+            // freed when this job completes — leaf-first, after every in-flight compaction buffer's
+            // `.use` has released, so no buffer ever outlives its allocator.
+            private val dbJob = Job(scope.coroutineContext.job)
+            private val loopScope = CoroutineScope(dbJob + dispatcher)
 
             private val trieCatalog = dbState.trieCatalog
             private val liveIndex = dbState.liveIndexOrNull
@@ -198,9 +201,16 @@ interface Compactor : AutoCloseable {
             }
 
             init {
+                // Free the driver's resources when the compactor job completes — fires even if the
+                // job is cancelled before the loop is ever scheduled, so the driver never leaks.
+                dbJob.invokeOnCompletion {
+                    driver.close()
+                    LOGGER.debug("compactor closed")
+                }
+
                 val jobsScope = CoroutineScope(SupervisorJob(dbJob) + jobsDispatcher)
 
-                scope.launch(CoroutineName("outer loop")) {
+                loopScope.launch(CoroutineName("outer loop")) {
                     while (true) {
                         availableJobs =
                             jobCalculator.availableJobs(trieCatalog)
@@ -272,18 +282,6 @@ interface Compactor : AutoCloseable {
                 // catalog, so refresh the snap here rather than at every test/dev call-site.
                 liveIndex?.refreshSnap()
             }
-
-            override fun close() {
-
-                runBlocking {
-                    withTimeoutOrNull(10.seconds) { dbJob.cancelAndJoin() }
-                        ?: LOGGER.warn("failed to close compactor cleanly in 10s")
-                }
-
-                driver.close()
-
-                LOGGER.debug("compactor closed")
-            }
         }
     }
 
@@ -292,10 +290,9 @@ interface Compactor : AutoCloseable {
     companion object {
         @JvmField
         val NOOP = object : Compactor {
-            override fun openForDatabase(allocator: BufferAllocator, dbStorage: DatabaseStorage, dbState: DatabaseState, watchers: Watchers) = object : ForDatabase {
+            override fun openForDatabase(scope: CoroutineScope, allocator: BufferAllocator, dbStorage: DatabaseStorage, dbState: DatabaseState, watchers: Watchers) = object : ForDatabase {
                 override fun signalBlock() = Unit
                 override suspend fun compactAll() = Unit
-                override fun close() = Unit
             }
 
             override fun close() = Unit
--- a/core/src/main/kotlin/xtdb/database/Database.kt
+++ b/core/src/main/kotlin/xtdb/database/Database.kt
@@ -133,11 +133,11 @@ class Database(
     override fun close() {
         meterRegistry?.let { reg -> registeredGauges.forEach { reg.remove(it) } }
         runBlocking {
-            // One cancel for the whole job tree. The compactor runs under `job`, so this also stops
-            // its loop and frees the driver's child allocator (via the compactor job's
-            // invokeOnCompletion) before `closeAll` frees the database allocator below.
+            // One cancel for the whole job tree: the compactor, the source-log subscription and the
+            // live leader/follower term all run under `job`, so cancelling and joining it stops them
+            // and runs their `invokeOnCompletion` frees (driver, allocator, replica producer, ext
+            // source) before `closeAll` frees the database allocator below.
             job?.cancelAndJoin()
-            processor?.cancelAndJoin()
         }
         (partitions.values + listOf(storage, allocator)).closeAll()
     }
--- a/core/src/main/kotlin/xtdb/database/Database.kt
+++ b/core/src/main/kotlin/xtdb/database/Database.kt
@@ -293,38 +293,34 @@ class Database(
 
                 val procFactory = object : LogProcessor.ProcessorFactory {
                     override fun openFollower(
+                        termScope: CoroutineScope,
                         pendingBlock: PendingBlock?,
                         afterReplicaMsgId: MessageId,
                     ) = FollowerLogProcessor(
-                        allocator, storage.bufferPool, state, compactorForDb,
+                        termScope, allocator, storage.replicaLog, storage.bufferPool, state, compactorForDb,
                         watchers, dbCatalog, pendingBlock, afterReplicaMsgId,
                         hasExternalSource = hasExternalSource,
                         meterRegistry = base.meterRegistry,
                     )
 
-                    override fun openLeaderSystem(
+                    // The leader term owns its replica producer and ext source; both are freed by
+                    // LeaderLogProcessor's `invokeOnCompletion` when `termScope` is cancelled.
+                    override fun openLeader(
+                        termScope: CoroutineScope,
                         replicaProducer: Log.AtomicProducer<ReplicaMessage>,
                         afterReplicaMsgId: MessageId,
-                    ): LogProcessor.LeaderSystem {
-                        val extSource = dbConfig.externalSource?.open(dbName, base.remotes, base.meterRegistry)
-
-                        val leaderProc = LeaderLogProcessor(
-                            allocator, base, storage, crashLogger,
-                            state, blockUploader, watchers,
-                            extSource = extSource,
-                            replicaProducer = replicaProducer,
-                            skipTxs = indexerConfig.skipTxs.toSet(),
-                            dbCatalog = dbCatalog,
-                            partition = 0,
-                            afterReplicaMsgId = afterReplicaMsgId,
-                            flushTimeout = indexerConfig.flushDuration,
-                        )
-
-                        return object : LogProcessor.LeaderSystem {
-                            override val proc get() = leaderProc
-                            override suspend fun cancelAndJoin() { leaderProc.cancelAndJoin(); replicaProducer.close() }
-                        }
-                    }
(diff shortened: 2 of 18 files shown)


jarohen:
### Chalk — job-owned lifecycle for Database services

- [x] The pattern — job-tree-is-ownership-tree + migrating entry point
- [x] Phase 1 — garbage collectors (`675571629`)
- [x] Phase 2 — compactor (`30e025709`)
- [ ] Phases 3–5 — processors → `LogProcessor` wrapper → `Database`
- [ ] `NodeSimulationTest` soundness — over-specification vs a real fixture defect

<details><summary>The pattern — job tree is the ownership tree; the entry point migrates up</summary>

A service takes a `CoroutineScope`, launches its loop in `init`, and frees whatever it owns or opened via `job.invokeOnCompletion` — not a body `finally` (that leaks when a job is cancelled before its body ever runs). It is not `AutoCloseable` and has no `close()`; an owner never calls `child.close()`. Children are structurally nested, so one cancel cascades leaf-first and each node frees its own resources as the tree unwinds. The only `AutoCloseable` survivors are genuine State with no job (catalogs, `LiveIndex`, `TransitionLogProcessor`).

There is always exactly one synchronous entry point — the `runBlocking { … cancelAndJoin() }` bridge into the job tree — at the top of the migrated subtree. Each phase sinks one more service into the tree and lifts that bridge one level higher; an explicit cancel at the current top is not a leak, it is the entry point at that level, and the next phase deletes it by nesting. End state: a single bridge in `Database.close()`.

</details>

<details><summary>Phase 1 — garbage collectors (675571629)</summary>

The block and trie GCs take their lifetime scope from the owner instead of self-creating a `Job`, and drop `AutoCloseable`/`close()`. They own no resources, so they're the simplest case — pure processes. `LeaderLogProcessor` builds a `gcScope` (a `SupervisorJob` child of its own scope, preserving sibling-failure isolation), runs both GCs on it, and does the bounded `cancelAndJoin` that each collector's `close()` used to do. The loop stays pinned to the collector's own dispatcher (`scope.launch(dispatcher)`), so the owner scope supplies only lifetime — that's what preserves the seeded `DeterministicDispatcher` in the sims. Behaviour-preserving; green incl. 100-iteration `NodeSimulationTest`.

</details>

<details><summary>Phase 2 — compactor (30e025709)</summary>

`Compactor.openForDatabase` now takes the owner's `CoroutineScope`; the compactor's `dbJob` is a child of it, and `dbJob.invokeOnCompletion { driver.close() }` frees the driver's child allocator + `SegmentMerge` temp dir when the job completes. `ForDatabase` is no longer `AutoCloseable`. `Database` owns an independent-root `compactorScope` (bounded 10s `cancelAndJoin` in `close()`, also registered with `safelyOpening` for the partial-open failure path); `DatabasePartition` no longer closes the compactor.

The allocator-throws-if-buffers-live ordering that used to be a hand-sequenced "cancel-join, then close" is now structural: `jobsScope` is a `SupervisorJob(dbJob)` child and the outer loop runs under `dbJob`, so `invokeOnCompletion` fires leaf-first, after every per-job `.use` has released its buffers. The synchronous bridge sits at `Database.close()` for this slice; Phase 5 folds it into the single root cancel.

Behaviour-preserving. Green in isolation: 813 `CompactorSimulationTest` + 800 `NodeSimulationTest` iterations, plus the Clojure `compactor_test` (real `Database.open`/`close`) — no allocator-closed errors, no coroutine leaks, no reflection/boxed-math warnings.

</details>

<details><summary>NodeSimulationTest — over-specification vs a real fixture defect</summary>

Putting the suite under more rigour (running the whole process tree under the seeded `DeterministicDispatcher`) surfaced two distinct categories, both TEST/fixture problems — not product bugs, and not caused by the lifecycle change:

1. **Over-specified tests.** `l1 compaction followed by garbage collection` and `staggered system startup` prescribe an ordering the autonomous compactor doesn't honour — the loop compacts any available job on every poll, including its first, so `signalBlock`/`compactAll` only nudge a parked loop. A real simulation test should assert interleaving-invariants, not a prescribed ordering; these encode a scheduling assumption that was only ever true by luck.

2. **A genuine fixture defect in `CompactorMockDriver`** (used by the multi-system `l0 to l3 with concurrent compaction + gc`). `publishTries` broadcasts each `TriesAdded` to the other systems asynchronously via a `MutableSharedFlow` consumed by a detached collector, then applies its own catalog mutation. Two faults: (a) `compactAll()`'s idle detection (`availableJobs.isEmpty() && queuedJobs.isEmpty()`) is blind to in-flight `sharedFlow` messages, so it can declare idle mid-broadcast — the cascade stalls at L2 and `l3Count == 0`; (b) `Instant.now()` real wall-clock inside a deterministic sim (~90/100 reproduction).

The rigour restructure that exposed these was rolled back; the suite on this branch is the unmodified `main` shape plus the minimal per-phase scope wiring. Reframing the over-specified tests as invariant assertions and fixing the `CompactorMockDriver` gaps is deferred (Phase-5 follow-up) so it doesn't block the lifecycle work.

</details>

<details><summary>Test-load finding — tight compact-all! timeouts flake under concurrent sims</summary>

Running the two 100-iteration property sims and the Clojure `compactor_test` in one loaded sequence, `test-l2+-compaction` and `test-l2-compaction-badly-distributed` timed out on their tight (1–2s) `compact-all!` wall-clock deadlines. In isolation, all 18 Clojure compactor tests pass deterministically (confirmed across two runs). This is CPU starvation against a real-time deadline, not a lifecycle regression — the change only gives the compactor's `dbJob` a parent, which can't slow `compactAll`'s promise completion. Recording it as pre-existing timeout fragility; not fixing it in this PR.

</details>

jarohen:
### Chalk — Phase 3 landed; rebased onto `sim-log-lifecycle`; at the Phase 4 gate

This supersedes the approach in the earlier comment. The work was rebased onto the `sim-log-lifecycle` branch (#5679, which lands first), and the pattern was refined: **unbounded** structured-concurrency cancels throughout (no `withTimeoutOrNull` bounds), and the compactor's scope is now a **child of the database `job`** rather than an independent root. The Phase 1/2 hashes in the earlier comment (`675571629`, `30e025709`) are obsolete — the re-landed, amended commits are `a8c24e212` (GC) and `dbfece864` (compactor).

- [x] Rebase onto `sim-log-lifecycle` — GC + compactor re-landed, amended to the refined pattern
- [x] Phase 3 — leader/follower processors own their lifetime (`3f092b61a`)
- [x] Test smells — leaf-processor and collector teardown folded onto the test scope
- [ ] Phase 4 — `LogProcessor` wrapper: model the running term as one value
- [ ] Phase 5 — `Database` single teardown cancel
- [ ] Deferred — `NodeSimulationTest` / `CompactorMockDriver` soundness (carried from the earlier comment)

<details><summary>Rebase onto sim-log-lifecycle + the refined pattern</summary>

The earlier "from `main`" lineage is abandoned. This branch now sits on `sim-log-lifecycle`, which already did the groundwork the old plan tried to fold in: `SimLog` is pure state (caller-owned scope, no `runBlocking` in a leaf `close`), and the sim tests tear down via a single scope cancellation. Two refinements to the pattern came out of that:

- **Unbounded teardown everywhere.** No `withTimeoutOrNull`, no defensive warns — trust structured concurrency. A `runBlocking { join }` inside a leaf `close()` deadlocks the single thread the deterministic dispatcher needs to drive a *suspending* `cancelAndJoin`, which is exactly why SimLog had to become pure state first.
- **Compactor nested under the database `job`** (not an independent root), so the single `job.cancelAndJoin()` in `Database.close()` reaps it.

The latest rebase also brought a base behaviour change that is **not** this PR's: `LeaderLogProcessor.handleResolvedTx` now stamps the leader's source-log watermark onto an ext-source `ResolvedTx.srcMsgId` (`-1` when there are no source records) instead of `null`, so a follower's `latestSourceMsgId` doesn't lag between block boundaries.

</details>

<details><summary>Phase 3 — leader/follower processors (3f092b61a)</summary>

`LeaderLogProcessor` and `FollowerLogProcessor` take their term scope from the `LogProcessor` wrapper, launch their loops in `init`, and free only what they opened via `invokeOnCompletion` — leaf-first, once their coroutines join. `cancelAndJoin()` is gone from both, and the explicit `gcScope` cancel introduced with the GC slice folds into the term cancel.

The follower now owns its replica tail (`replicaLog.tailAll`), absorbed from the wrapper. A cancelled tail must not be mistaken for a failure: `CancellationException` is rethrown rather than routed through `notifyError`, both in the tail launch and in `processRecords`, so a shutdown doesn't poison the shared `Watchers`.

`TransitionLogProcessor` goes the other way — no loop, driven synchronously through the transition, so it's genuine State: a plain `AutoCloseable` used via `.use`, with `cancelAndJoin()` dropped from the shared `Processor` interface and `TransitionProcessor` gaining `AutoCloseable`.

The wrapper holds a per-term `SupervisorJob`; a transition cancels the running term job and builds the next term on a fresh one. `sys` (role) and `termJob` are deliberately two fields for now — the state/services back-edge Phase 4 folds into a single `Term` value.

Verified: 18 unit tests + `LogProcessorSimTest` (≥300 rebalance iterations) + `NodeSimulationTest` + `CompactorSimulationTest`, all green — no `UncompletedCoroutinesError`, no allocator leaks, no `DeterministicDispatcher` cross-thread crashes, no reflection/boxed-math warnings.

</details>

<details><summary>Binding lessons (incl. a worktree gotcha)</summary>

1. **Unbounded `cancelAndJoin`** — no timeouts, no defensive warns.
2. **Two dispatchers for IO services** — the loop runs on the injected scope's dispatcher (`scope.launch { … }`), the IO/delete fan-out on a separate injectable `dispatcher`; the owner forwards an injectable dispatcher (GCs got `gcDispatcher`) so sims keep work on the seeded `DeterministicDispatcher`. Pinning the *loop* to a real pool caused a cross-thread teardown hang.
3. **No `runBlocking` bridge in a leaf `close`** — it deadlocks the deterministic dispatcher.
4. **Allocator init-context shadowing** — inside a ctor `init` lambda, bare `allocator` binds the ctor parameter (the parent), not the child field; free `this@…​.allocator` explicitly.
5. **Compactor shape** — `compactorScope` is a child of the database `job`.
6. **Worktree gotcha (cost real time)** — the main checkout is on a different branch; the work lives in the `.claude/worktrees/josl` worktree. All file edits MUST use the worktree path; a bare main-repo path silently reads the wrong branch's copy.

</details>

<details><summary>Test-smell status — what folded, what's still blocked</summary>

The leaf-processor unit tests (`Follower`/`Leader`/`Transition`), `ExternalSourceTest`, and the sims now fold into the sim-log teardown shape: an owned `backgroundScope` (or the sim's single `launch(dispatcher).invokeOnCompletion { cancel() }`) is the teardown — no explicit processor `cancelAndJoin`. The ad-hoc replica-log collectors the assertions read from were also moved onto `backgroundScope`, so the one scope cancel reaps them too. `FollowerLogProcessor`'s unit tests park a mock `replicaLog` (`tailAll` answers `awaitCancellation()`) so the self-launched tail idles while the test drives `processRecords`.

The one remaining smell is `LogProcessorTest` — it hand-rolls `CoroutineScope(SupervisorJob())` + `logProc.cancelAndJoin()` + `job.cancelAndJoin()`. That's load-bearing-blocked: the wrapper still *has* `cancelAndJoin()` (collapses in Phase 5) and the test owns plain `InMemoryLog`s closed explicitly rather than scope-managed like the sim's `SimLog`. It folds out as Phases 4–5 collapse the wrapper teardown.

</details>

<details><summary>Where this is — the Phase 4 gate</summary>

At the re-eval gate before Phase 4. Next: collapse the `sys` + `termJob` back-edge into a single `Term(sys, job)` value, swapped atomically — the leader gauge and `gcAll()` read the term's role, the rebalance handlers cancel `term.job`. Watch the `DatabaseCatalogTest` `StandardTestDispatcher` path: teardown coroutines must not inherit a virtual-clock dispatcher (the wrapper scope stays on `Dispatchers.Default`). Then Phase 5 nests the wrapper under `job` so a single `job.cancelAndJoin()` in `Database.close()` reaps everything.

</details>
---

**Update — migration code-complete.**
The wrapper amendment landed, folded into the Phase 3 commit: the running term *is* `SubSystem` now — it carries its own `SupervisorJob`, the factory returns bare processors (`openLeader`/`openFollower`), and the wrapper assembles the term. No `termJob` field, no `Term` layer.
The final slice also landed: `Database.close()` is a single `job.cancelAndJoin()` — the compactor, the source-log subscription and the live term all nest under the database `job`, so one cancel reaps the tree leaf-first and each service's `invokeOnCompletion` frees before the join returns.
Verified green: full `:xtdb-core:test` (278) + LogProcessorSim/Node/Compactor sims; `DatabaseCatalogTest` confirms no `StandardTestDispatcher` teardown livelock.
Pending before merge: whole-project `./gradlew test` + Kafka `integration-test`, then flip out of draft once #5679 lands.

PR #5679, opened 2026-06-05
tidy(indexer): hand SimLog's lifecycle to the caller's scope

jarohen:
## Summary

Make `SimLog` pure state and hand its lifecycle — and the simulation tests' — to the caller's coroutine scope, so teardown is structured-concurrency cancellation rather than a self-managed join.

## Motivation

`SimLog` is the in-memory `Log` the indexer simulation tests run against. It used to own its own coroutine machinery: it took a `CoroutineContext` in its constructor, built its own `Job` and `CoroutineScope`, started its group/plain-consumer delivery loops in an `init` block, and tore them down in `close()` with a `runBlocking { job.join() }`. That conflated two lifetimes behind one object — `SimLog` the `Log` (an `AutoCloseable` resource) and `SimLog` the *process* (its delivery loops) — and forced the synchronous `close()` to bridge into the coroutine world with `runBlocking`, which is the wrong thing to do on a `runTest` dispatcher.

## Approach

`SimLog` is now pure state. Construction takes just a name and the RNG; the delivery loops are launched explicitly onto a caller-owned scope via a `launchSimLog` extension, and `close()` becomes a no-op — there's nothing to release beyond the process, and the process belongs to the scope that launched it. Teardown is structured-concurrency cancellation of that scope rather than a self-managed join. This is the test-side mirror of the lifetime model the real `LogProcessor` uses (its `cancelAndJoin` rename landed separately and is already on main). `SimLogTest` is adapted to the new two-arg constructor and launches the loops on the test scope so a failing `processRecords` still propagates out through it.

With `SimLog` and `LogProcessor` both taking their scope from the caller, the simulation tests no longer track jobs by hand. Each test launches its `SimLog` loops, its group subscriptions (each opening its processor on its own coroutine), and its driver under a single `launch(dispatcher)` scope; when the driver quiesces, `invokeOnCompletion { cancel() }` cancels that scope and structured concurrency tears the whole tree down and joins it. The earlier per-node `CoroutineScope`s, per-subscription job cancellation, explicit `cancelAndJoin()` calls in `finally`, and the `withLaunchedSimLogs` helper are all gone. All three tests — `single node`, `stable leader`, `multi-node` — share this one shape now; the only difference between them is the scenario: node count, and rebalances vs a steady flush.

## Usage — writing a deterministic test against `SimLog`

Determinism rests on three things, and a new test stays reproducible only as long as it respects all three:

- **One seeded RNG.** `SimulationTestBase` hands you `rand`, and `@RepeatableSimulationTest` replays the test across seeds. Every nondeterministic choice must flow from `rand` — the action sequence (`buildActions(rand, …)`), and inside `SimLog` the delivery batch sizes (`rand.nextInt(1, lag + 1)`) and leader election (`groupConsumers.random(rand)`). Same seed in, same run out. Pull randomness from anywhere else — the wall clock, an unseeded `Random`, a real dispatcher — and replay breaks.
- **One test dispatcher.** Everything runs on `dispatcher` under `runTest`: the `SimLog` loops, every `LogProcessor`, and the driver. Interleaving is virtual-time and single-threaded, and `yield()` drives cooperative scheduling instead of real delays.
- **Quiesce, then cancel.** The driver does its work, waits for the system to settle (`simExtSource.awaitQuiescence()`, `replicaLog.awaitAllDelivered()`), and only then triggers teardown — so the assertions run against a stable final state.

The shape:

1. Create the logs as pure state (the `@BeforeEach` already does this): `srcLog = SimLog("src", rand)`.
2. Open storage and nodes: `MemoryStorage(...).use { bp → SimNode(...).use { node → … } }`.
3. Under a single `launch(dispatcher) { … }`:
   - `launchSimLog(srcLog)` / `launchSimLog(replicaLog)` to start the delivery loops on this scope;
   - one `launch { srcLog.openGroupSubscription(node.openLogProcessor(this)) }` per node;
   - a driver `launch { … }` that appends messages / triggers rebalances (with `yield()` between steps), then awaits quiescence — ending with `.invokeOnCompletion { cancel() }` so that finishing the driver cancels the whole scope.
4. `.join()` the outer launch, then assert against `replicaLog.topic`, the block catalogs, and the live-index snapshot.

```kotlin
@RepeatableSimulationTest
fun `my scenario`() = runTest(timeout = 5.seconds) {
    val actions = buildActions(rand, rand.nextInt(50, 100))
    val simExtSource = SimExtSource(actions)

    MemoryStorage(allocator, epoch = 0).use { bp →
        SimNode("test-db", bp, IndexerConfig(rowsPerBlock = rand.nextLong(15, 25)), simExtSource).use { node →
            launch(dispatcher) {
                launchSimLog(srcLog)
                launchSimLog(replicaLog)

                launch { srcLog.openGroupSubscription(node.openLogProcessor(this)) }

                launch {
                    repeat(rand.nextInt(20, 40)) {
                        yield()
                        if (rand.nextInt(100) < 50) srcLog.rebalanceTrigger.send(Unit)
                        else srcLog.appendMessage(SourceMessage.FlushBlock(null))
                    }
                    simExtSource.awaitQuiescence()
                    replicaLog.awaitAllDelivered()
                }.invokeOnCompletion { cancel() }
            }.join()

            assertReplicaTxInvariants()
            // … scenario-specific assertions …
        }
    }
}
```

## Follower-teardown fix

Building these deterministic tests surfaced a latent ordering bug in the leader/follower transition. `FollowerSystem.cancelAndJoin()` was tearing down the outgoing replica subscription with a fire-and-forget `cancel()`, so a trailing follower delivery could call `notifyMsg` after the node's new role had already advanced the shared `Watchers` — violating the source-msgId monotonicity the rebalance path assumes (`srcMsgId 17 < latestSourceMsgId 19`). It now `cancelAndJoin()`s the subscription, draining and unregistering the old follower before the transition proceeds — the "no concurrent processing during rebalance" invariant the code already documents. Latent since `cd8962983`; the teardown rework's scheduling shift is what made it reproduce (~50% of the time under a pinned seed, 10/10 with the fix).

## Source-watermark fix (found here, tracked in #5680)

The same simulation testing surfaced a second, deeper bug: in ext-source mode the source-log watermark can regress on a follower across rapid leadership churn. Both fixes — stamping ext-source `ResolvedTx`, and a `NoOp` carrying the source watermark on block-less flushes — are written up in full (mechanism, trace, rationale) in #5680.

## Behaviour

The `SimLog` and simulation-test changes are a pure refactor — same scenarios, same assertions. The production behaviour changes are the two fixes above (the follower-teardown `cancelAndJoin`, and the #5680 source-watermark stamping), each kept as its own commit.

### DIFF
--- a/core/src/test/kotlin/xtdb/indexer/LogProcessorSimTest.kt
+++ b/core/src/test/kotlin/xtdb/indexer/LogProcessorSimTest.kt
@@ -36,6 +36,7 @@ import xtdb.database.DatabaseStorage
 import xtdb.database.ExternalSource
 import xtdb.database.ExternalSourceToken
 import xtdb.error.Incorrect
+import xtdb.indexer.SimLog.Companion.launchSimLog
 import xtdb.indexer.TxIndexer.TxResult
 import xtdb.storage.MemoryStorage
 import xtdb.table.TableRef
@@ -62,12 +63,25 @@ class LogProcessorSimTest : SimulationTestBase() {
     fun setUp() {
         nodeBase = openBase(openMeterRegistry = false)
         allocator = nodeBase.allocator.newChildAllocator("test", 0, Long.MAX_VALUE)
+        srcLog = SimLog("src", rand)
+        replicaLog = SimLog("replica", rand)
     }
 
     @AfterEach
     fun tearDown() {
+        replicaLog.close()
+        srcLog.close()
         allocator.close()
         nodeBase.close()
+    }
+
+    private inline fun <R> CoroutineScope.withLaunchedSimLogs(block: () -> R): R {
+        val job = launch(dispatcher) {
+            launchSimLog(srcLog)
+            launchSimLog(replicaLog)
+        }
+
+        return block().also { job.cancel() }
     }
 
     private val docsTable = TableRef("test-db", "public", "docs")
@@ -174,7 +188,8 @@ class LogProcessorSimTest : SimulationTestBase() {
         val watchers = Watchers(latestTxId = -1, latestSourceMsgId = -1)
             .also { simExtSource.watch(it) }
         val dbStorage = DatabaseStorage(srcLog, replicaLog, bp, null)
-        val blockUploader = BlockUploader(dbStorage, dbState, mockk(relaxed = true), null, null, uploadDispatcher = dispatcher)
+        val blockUploader =
+            BlockUploader(dbStorage, dbState, mockk(relaxed = true), null, null, uploadDispatcher = dispatcher)
         val crashLogger = CrashLogger(allocator, bp, "sim-node")
 
         override fun openLeaderSystem(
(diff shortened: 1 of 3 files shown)


jarohen:
### Chalk — SimLog lifecycle + deterministic sim tests

- [x] Hand SimLog's lifecycle to the caller's coroutine scope
- [x] Tear the sim tests down via a single scope cancellation
- [x] Bugs surfaced by the deterministic tests

<details><summary>Hand SimLog's lifecycle to the caller's coroutine scope</summary>

SimLog is a `Log` subclass, so it expects to be `close()`d — but it owns no resources of its own; what it really owns is a *process*. Closing it recursively from outside that process split the lifecycle across two owners: the caller held the object, the coroutine held the work.

The caller's coroutine scope now owns the whole tree. SimLog launches its process into that scope, and a single cancellation tears down SimLog along with everything launched under it — no separate `close()` to forget, no ordering question between "stop the process" and "close the log".

</details>

<details><summary>Tear the sim tests down via a single scope cancellation</summary>

The pattern landed first on `single node processes txs and flush-blocks with rebalances`, then propagated across the file: every sim test now stands its node tree up under one scope and tears it down with one `cancel`, rather than each test closing its own pieces in its own order.

That's what makes the "no concurrent processing during rebalance" invariant checkable — teardown is a single deterministic event, not a race between several close paths.

</details>

<details><summary>Bugs surfaced by the deterministic tests</summary>

Making the sim deterministic (seeded `rand` + single-threaded dispatcher) turned three intermittent races into reproducible failures, each fixed as its own cherry-pickable commit on `main`:

- A follower-teardown race — `job.cancel()` left the follower's work unjoined; `cancelAndJoin()` closes the window.
- Two halves of an ext-source source-watermark regression across rapid leadership churn (`srcMsgId < latestSourceMsgId`) — stamping ext-source `ResolvedTx` with the leader's watermark, then propagating it on block-less flushes via `NoOp(srcMsgId)`. Both written up in #5680.

The refactor was the goal; the determinism it enabled is what made the underlying races legible.

</details>

PR #5694, opened 2026-06-08
feat: extract per-connection await-token into a node-level Connection

tggreene:
## Summary

Extracts per-connection await-token tracking out of `XtdbConnection` and `pgwire` into a shared `xtdb.Connection` primitive that lives at the node layer.
One implementation, two consumers, read-your-writes guaranteed by the `Connection` rather than re-derived by each frontend.

This is the **connection-level alternative to #5648** (the `Session` extraction), addressing @jarohen's review on #5609:
> if pgwire and ADBC both require the same behaviour (and a hypothetical third protocol might too) then it's worth instead moving this functionality down a layer.

#5648 is left open for side-by-side comparison.

## Shape

`xtdb.Connection` is a connection-scoped handle onto the node.
It owns an await-token, advances it on every write (`submitTx`/`executeTx`/`recordTx`), and threads it into every read (`prepareSql`/`openSqlQuery`/`openSnapshot`).
Callers issue token-free reads and writes and get read-your-writes for free; the token surfaces only via `awaitToken` for the pgwire `SET`/`SHOW AWAIT_TOKEN` SQL feature.

`Connection.Node` is a pure executor SPI implemented by a Clojure reify in `node/impl.clj` (where `db-cat` and the basis token math already live).
The factory is `xtp/open-connection` on the **internal** `xtdb.protocols/PNode` — alongside `attach-db`/`detach-db` — so it is reached the same way pgwire already reaches the node.

## How this differs from #5648

- **No public API growth.** Zero changes to `xtdb/api/Xtdb.kt`. #5648 added `openSession` to the `Xtdb.XtdbInternal` interface (the diff @jarohen reacted to); here the factory is on the internal `xtp/PNode` protocol, so the documented `Xtdb` surface is untouched.
- **No empty `AutoCloseable`.** `Connection` holds no resources, so it doesn't advertise a close contract nobody needs (a smell #5648 reintroduced).
- **pgwire `await_token` wire-surface is behaviour-preserving.** Every conn-state `:await-token` reader was rebound to source from the `Connection`; `SHOW AWAIT_TOKEN`, `SHOW latest_submitted_tx`, the per-transaction `WITH (AWAIT_TOKEN …)` override, and `SET AWAIT_TOKEN` are unchanged. The existing `test-show-await-token` / `in-tx-await-token` / `await-token+snapshot-token` tests are the regression guard and stay green.

## Commits

1. `feat(adbc): extract per-connection await-token into a node-level Connection` — `Connection` + `ConnectionTest`, `xtp/open-connection`, ADBC adopts it (`XtdbConnection.Node` extends `Connection.Node`, bespoke `awaitToken`/`runTx`/`mergeAwaitToken` deleted).
2. `feat(pgwire): migrate per-connection await-token to the shared Connection` — pgwire stores the `Connection` in `:conn-state`, routes writes through it, drops the inline `merge-tx-tokens` and the `xtdb.basis` require.

## Test plan

All green, no reflection / boxed-math warnings:

- `xtdb.ConnectionTest` (new): 6/6 — read-your-writes, monotonic accumulation, cross-db merge, dbName retargeting, direct token replacement.
- `xtdb.api.AdbcTest`: 11/11.
- `xtdb.FlightSqlAdbcTest`: 45/45.
- `xtdb.pgwire-test`: 144/144 (includes the `await_token` wire-surface tests).
- `xtdb.pgwire-protocol-test`: 11/11.

## Stack

Built on top of #5609 (`tim/5608-prepare-await-token`); base shows #5609's commits until it lands, then this rebases cleanly.
Iterates toward umbrella #5132.

## Scope

Out:

- Broader connection-state extraction (`autoCommit`, `pendingOps`, `currentCatalog`, pgwire session clock/params) — this PR scopes to await-token only, matching @jarohen's ask.
- Unifying with the node's `java.sql.Connection` — a larger piece, deliberately deferred.

### DIFF
--- a/core/src/main/clojure/xtdb/node/impl.clj
+++ b/core/src/main/clojure/xtdb/node/impl.clj
@@ -128,9 +128,6 @@
          (let [query-opts (-> {:await-token await-token} (with-query-opts-defaults this-node db-name))]
            (xtp/prepare-sql this-node sql query-opts)))
 
-       (mergeAwaitToken [_ existing-token db-name tx-id]
-         (basis/merge-tx-tokens existing-token (basis/->tx-basis-str {db-name [tx-id]})))
-
        (getColumnTypes [_ table snap]
          (when-let [^Database db (.databaseOrNull db-cat (.getDbName table))]
            (.getColumnTypes db table snap)))
--- a/core/src/main/clojure/xtdb/node/impl.clj
+++ b/core/src/main/clojure/xtdb/node/impl.clj
@@ -119,24 +119,28 @@
        (submitTx [_ db-name ops opts] (.submitTx this-node db-name ops opts))
        (executeTx [_ db-name ops opts] (.executeTx this-node db-name ops opts))
 
-       (openSqlQuery [_ sql db-name]
-         (let [query-opts (-> {} (with-query-opts-defaults this-node db-name))]
+       (openSqlQuery [_ sql db-name await-token]
+         (let [query-opts (-> {:await-token await-token} (with-query-opts-defaults this-node db-name))]
            (-> (xtp/prepare-sql this-node sql query-opts)
                (.openQuery nil (QueryOpts. nil (:default-tz query-opts))))))
 
-       (prepareSql [_ sql db-name]
-         (let [query-opts (-> {} (with-query-opts-defaults this-node db-name))]
+       (prepareSql [_ sql db-name await-token]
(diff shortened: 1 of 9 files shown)


tggreene:
Superseded by #5695 — reworked from the `Connection` extraction into a leaner `AwaitToken` accumulator (closed automatically when this branch was renamed).

PR #5696, opened 2026-06-08
feat(kafka-connect-source): external source via Kafka Connect

### LINKED ISSUE #5562: Kafka Connect sink connector
## Summary

Run XTDB as a target for Kafka Connect: KC owns partition assignment for the source topic(s), and XTDB hosts the connector in a stripped-down node that does nothing else.

Spec required.
Sub-issue of #5330.
Sibling of #5561 (`!KafkaTopic` external source — same problem from XTDB's side, but with XTDB owning partition assignment).

## Motivation

Kafka Connect is the standard way users wire Kafka into other systems.
A sink connector for XTDB lets operators land KC-managed topics into a database without running a separate consumer process — they register a connector against an existing KC cluster.

`!KafkaTopic` covers the case where XTDB owns the consumer; this card covers the case where KC does.
Both should share as much machinery as possible — `RecordIndexer`, failure-mode plumbing, the `!Docs` reference indexer.

## What changes vs `!KafkaTopic`

The big inversion: **partition assignment is decided by Kafka Connect**, not by XTDB's source-log leadership.

- `!KafkaTopic` — the framework calls `onPartitionAssigned(partition, ...)` when XTDB's source-log leader picks up that database's partition.
- KC sink — KC's worker rebalance calls `SinkTask.open(partitions)` / `SinkTask.close(partitions)`.
  XT's components have to take that as instruction, not produce it.

The leader/follower model that today decides who consumes the external source doesn't apply directly here.
The indexer entry point has to be driven by KC callbacks.

## Direction (rough sketch — to confirm in spec)

- Users start a stripped-down XTDB node configured for one database — no pgwire, no query, just enough of the indexer + source-log writer to host the connector.
- That node embeds a KC worker (or registers itself as one) and hosts the sink connector for the named database.
- KC's `SinkTask.open(partitions)` maps onto the same `onPartitionAssigned` hook #5561 introduces; `SinkTask.close(partitions)` to revocation.
- `SinkTask.put(records)` calls into a `RecordIndexer` (same SPI as #5561) wrapped to satisfy KC's flush/commit semantics.
- Offsets are KC's responsibility; XTDB indexes the records and signals durable-flush back to KC.

## Open questions

- **Deployment shape.** Embedded KC worker (full worker runtime in-process) or a thinner host that runs one connector + one task without the worker rebalance machinery? What does "stripped-down" exclude concretely?
- **Connector config surface.** KC connectors are configured via the KC REST API as a flat string-map. How do we project our `!Docs`-style indexer config onto that flat namespace?
- **Failover interaction.** If KC moves a task to a different worker, that worker is a different XTDB node. How does the new node resume indexing given KC owns the offset?
- **Leader/follower vs KC tasks.** Does the connector run on every XTDB node for the database, or only on a designated leader? Does KC's task count need to match the XTDB node count?
- **Offset durability boundary.** KC commits offsets after `flush()` returns. What does `flush()` wait for — source-log append, indexed-into-LiveIndex, block boundary?
- **External-source-token persistence.** `!KafkaTopic` stores a per-partition resume token in the block file. Does the KC sink also persist one (belt-and-braces against KC offset-store loss), or is KC's offset store the sole source of truth?
- **Schema / payload model.** KC has its own Connect-Schema type system. Does the indexer see Connect `Struct`s or plain JVM values? Where do SMT-style transforms apply — KC layer (standard SMTs) or XT layer?
- **`RecordIndexer` reuse.** Can we share the SPI with `!KafkaTopic` straight, or does KC's per-task-per-partition lifecycle want a different shape? Same question for `!Docs`.
- **Reverse direction.** XTDB-as-source connector (XT change events → Kafka) is a separate feature; out of this card.

## Scope

In, once spec is settled:

- Stripped-down XTDB node mode for connector hosting.
- KC sink connector + task that wraps a `RecordIndexer`.
- Reuse of #5561's `RecordIndexer` SPI and `!Docs` reference indexer where the lifecycle allows.
- Operator docs: installation, configuration, running an XTDB node in connector mode.

Out:

- KC source connector (XTDB → Kafka).
- Cross-DB connectors — one connector targets one database.

## Progress

**Status**: not-started

(Filled in once spec lands.)

Akeboshiwind:
Resolves #5562.

## Summary

Adds `!KafkaConnect` as a second external-source flavour for XT databases fed by a Kafka topic. The framework owns subscription, partition assignment, polling and resume-token lifecycle; bytes flow through a Kafka Connect `Converter` + `Transformation` chain to produce `SinkRecord`s, which a pluggable `RecordIndexer` writes to XT. The reference `!Docs` indexer (one record = one tx, value-as-doc, key-as-`_id` fallback) walks the resulting `Struct` against its `Schema` and emits Connect's logical types as their natural XT counterparts: `Decimal → BigDecimal`, `Timestamp → Instant`, `Date → LocalDate`, `Time → LocalTime`.

## Context / Motivation

`!Postgres` is the only external source today. Kafka topics are the next-most-common shape operators want to consume from, and a Connect-shaped pipeline gives us standardised handling across formats (Avro, JSON, JSON Schema, plus anything else with a `Converter`), free SMT chain support that mirrors how operators already shape data in their KC pipelines, and a SPI shape (`SinkRecord`) that lines up with what KC sink connectors consume — the same shape full KC compatibility would want anyway.

The alternative shape — `KafkaConsumer<K, V>` + raw `Deserializer`s — would put every indexer on the hook for format-specific JVM types (`Map` for JSON, `GenericRecord` for Avro, `JsonNode` for JSON Schema) and lose schema information at the indexer boundary. Connect already solved that problem; we re-use the solution.

## Usage

Node config — register the Kafka cluster as a remote in the YAML:

```yaml
remotes:
  my-kafka: !Kafka
    bootstrapServers: kafka.prod.internal:9092
```

Then attach a database against it at runtime, via SQL on a connected client:

```sql
ATTACH DATABASE orders WITH $$
  log: !Kafka
    cluster: my-kafka
    topic: orders-replica
  externalSource: !KafkaConnect
    remote: my-kafka
    topic: orders
    connectConfig:
      key.converter: org.apache.kafka.connect.storage.StringConverter
      value.converter: io.confluent.connect.avro.AvroConverter
      value.converter.schema.registry.url: http://schema-registry:8081

      # SMT chain (optional, same shape as KC):
      transforms: unwrap,maskEmail
      transforms.unwrap.type: org.apache.kafka.connect.transforms.ExtractField$Value
      transforms.unwrap.field: payload

      # Predicates gate transforms — only mask when the PII header is present:
      transforms.maskEmail.type: org.apache.kafka.connect.transforms.MaskField$Value
      transforms.maskEmail.fields: email
      transforms.maskEmail.predicate: hasPii
      predicates: hasPii
      predicates.hasPii.type: org.apache.kafka.connect.transforms.predicates.HasHeaderKey
      predicates.hasPii.name: PII
    indexer: !Docs
      table: orders
$$;
```

`connectConfig` accepts the same keys you'd pass to a KC sink connector: `key.converter`, `value.converter`, `value.converter.<prop>`, `transforms`, `transforms.<alias>.<prop>`, plus the predicate triplet (`predicates`, `predicates.<alias>.type`, `predicates.<alias>.<prop>`) and the `transforms.<alias>.predicate` / `transforms.<alias>.negate` keys that gate a transform on a predicate.

Three Confluent-supported formats are bundled into the published `aws`, `azure`, and `google-cloud` Docker images:

- `org.apache.kafka.connect.json.JsonConverter` — schemaless JSON, no SR.
- `io.confluent.connect.avro.AvroConverter` — Avro + Confluent SR.
- `io.confluent.connect.json.JsonSchemaConverter` — JSON Schema + Confluent SR.

The `standalone` image is *not* wired up — see the `build(docker)` commit body for why.

## Implementation notes

**Connect data model end-to-end.** `KafkaConnectSource` configures the consumer with `ByteArrayDeserializer` for both key and value, runs each record's bytes through the configured `Converter`s to produce a `SchemaAndValue`, and assembles a `SinkRecord`. The SMT chain (built from `transforms.*` config) operates on the `SinkRecord` exactly as it does inside a KC sink task. The indexer SPI receives a `List<SinkRecord>` and decides commit cadence.

**Schema-first walk in `DocsIndexer.convertValue`.** Matches the order Connect's own `JsonConverter.convertToJson` dispatches: null → `schema.name()` for logical-type promotion → `schema.type()` → schemaless fallback dispatching on JVM class. The schema is the primary source of truth because most Connect values can't self-describe — a `java.util.Date` could be a `Date`, `Time`, or `Timestamp` logical type, only `schema.name()` disambiguates.

**UTC for Date and Time.** Connect encodes `Date` as `Date(epochDay * 86_400_000)` and `Time` as `Date(millis_of_day)`, both anchored at the UTC epoch. Decoding requires re-interpretation in UTC — anywhere else and the calendar reading shifts by the JVM's local offset (a node running in Pacific time would read `2024-06-08` Date values as `2024-06-07`). `DocsIndexerTest` forces the JVM default zone to America/Los_Angeles for the assertion to prove this isn't accidentally satisfied by a UTC CI runner.

**ProtoAny for the indexer factory.** `RecordIndexer.Factory` mirrors `ExternalSource.Factory` — config flows through `google.protobuf.Any` to support cross-node ATTACH DATABASE. Considered a tag-plus-bytes shape parallel to 3763caa42's `ExternalSourceToken = ByteArray` refactor, but tokens are opaque per-source-instance (no polymorphism) while indexer configs are polymorphic across registered factories (need a discriminator). ProtoAny's `typeUrl` is doing real work. Third-party indexer authors take the protobuf dep — same trade-off `PostgresSource` accepted internally.

**Halt-on-decode-failure.** Matches KC's default (`errors.tolerance=none`): any converter / SMT / indexer exception propagates up, the source coroutine dies, the leader fails over, the consumer re-reads from the last good token. KC's `errors.tolerance=all` (DLQ) mode requires offset-ordered commits — see Dead ends.

## Dead ends

**DLQ-via-aborted-tx.** Earlier iterations carried a DLQ path: a failed decode wrote an aborted XT tx with the bad record's topic/partition/offset coords and advanced the resume token, then continued. Looked clean but had a subtle race — the aborted-tx commit advanced the token *before* the indexer's per-record commits, so on a mid-batch indexer crash the good records earlier in the batch were lost on restart (their token had been overwritten). KC handles this in `errors.tolerance=all` mode by reordering commits so good records always commit before any abort that supersedes them; doing that properly here is out of scope. Halt-on-failure for now, DLQ as a follow-up.

**Bespoke `MapJsonDeserializer`.** An earlier iteration shipped a custom test deserializer to bridge raw Kafka deserializers' output into `!Docs`'s expected `Map`. Survey of the ecosystem (Kafka Connect's `Converter` + `SchemaAndValue`, Pinot's `StreamMessageDecoder`, Druid's `inputFormat`, Materialize / RisingWave's `FORMAT ENCODE`, ClickHouse's `kafka_format`) showed all of them converge on per-format decoders producing a uniform schema-aware shape. Switched the source to KC's `Converter` directly; the bespoke deserializer went away.

## Out of scope, deferred

- Multi-partition consumption (Stage 3 constraint from #5330 — single-partition only).
- Multi-topic / topic patterns — one source = one topic.
- The `errors.tolerance=all` DLQ flow with offset-ordered commits.
- Additional reference indexers beyond `!Docs` (e.g. CDC-shaped record mapping for Debezium envelopes).
- Pluggable `_id` source.
- `standalone` Docker image wiring — would transitively pull `xtdb-kafka` log support into the playground image; separate decision.

## Changes

Two commits, intended reading order:

1. `feat(kafka-connect-source): external source via Kafka Connect data model` — the module itself: SPI, source, indexer, factory + integration tests.
2. `build(docker): bundle xtdb-kafka-connect-source in the cloud images` — wire the module into the `aws`, `azure`, `google-cloud` published images so operators can use `!KafkaConnect` without rebuilding.

## Test plan

7 unit tests (`./gradlew :modules:xtdb-kafka-connect-source:test`):

- `KafkaConnectSourceFactoryTest` (4) — YAML + proto roundtrip for `!KafkaConnect` + `!Docs`.
- `DocsIndexerTest` (3) — UTC convention for Connect `Date` / `Time` / `Timestamp` under a non-UTC default JVM zone.

13 integration tests (`./gradlew :modules:xtdb-kafka-connect-source:integration-test`, requires Docker for Kafka + Schema Registry containers):

- `JsonConverter` happy path, tombstone delete, all JSON types.
- `JsonSchemaConverter` with Schema Registry: tombstone delete, all JSON Schema types including the `title`-keyed path for typed temporals.
- `AvroConverter` with Schema Registry: tombstone delete, all Avro types, recursive logical-type promotion through nested compound shapes.
- Resume from token after node restart.
- Halt-on-decode-failure (consumer dies; good records before the bad one stay committed, source doesn't recover until restart).

All 20 tests pass on the branch.

### DIFF
--- a/modules/kafka-connect-source/src/main/kotlin/xtdb/kafkaconnectsource/KafkaConnectSource.kt
+++ b/modules/kafka-connect-source/src/main/kotlin/xtdb/kafkaconnectsource/KafkaConnectSource.kt
@@ -1,0 +1,380 @@
+package xtdb.kafkaconnectsource
+
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.currentCoroutineContext
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.runInterruptible
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.modules.PolymorphicModuleBuilder
+import kotlinx.serialization.modules.SerializersModule
+import kotlinx.serialization.modules.subclass
+import org.apache.kafka.clients.consumer.ConsumerRecord
+import org.apache.kafka.clients.consumer.KafkaConsumer
+import org.apache.kafka.common.TopicPartition
+import org.apache.kafka.common.errors.InterruptException
+import org.apache.kafka.common.errors.WakeupException
+import org.apache.kafka.common.header.Headers
+import org.apache.kafka.common.serialization.ByteArrayDeserializer
+import org.apache.kafka.connect.data.SchemaAndValue
+import org.apache.kafka.connect.header.ConnectHeaders
+import org.apache.kafka.connect.sink.SinkRecord
+import org.apache.kafka.connect.storage.Converter
+import org.apache.kafka.connect.storage.HeaderConverter
+import org.apache.kafka.connect.storage.SimpleHeaderConverter
+import org.apache.kafka.connect.transforms.Transformation
+import org.apache.kafka.connect.transforms.predicates.Predicate
+import io.micrometer.core.instrument.MeterRegistry
+import xtdb.api.Remote
+import xtdb.api.RemoteAlias
+import xtdb.api.log.KafkaCluster
+import xtdb.database.ExternalSource
+import xtdb.database.ExternalSourceToken
+import xtdb.database.proto.DatabaseConfig
+import xtdb.error.Incorrect
+import xtdb.indexer.TxIndexer
+import xtdb.kafkaconnectsource.proto.KafkaConnectSourceConfig
+import xtdb.kafkaconnectsource.proto.KafkaConnectSourceToken
+import xtdb.kafkaconnectsource.proto.kafkaConnectSourceConfig
+import xtdb.util.error
+import xtdb.util.info
+import xtdb.util.logger
+import xtdb.util.warn
+import java.time.Duration
+import com.google.protobuf.Any as ProtoAny
+
+private val LOG = KafkaConnectSource::class.logger
+
+private const val PROTO_TAG_PREFIX = "proto.xtdb.com"
+private val POLL_DURATION: Duration = Duration.ofSeconds(1)
+
+// Resume token is our offset; Kafka must not commit its own.
+private val HARDCODED_CONSUMER_CONFIG: Map<String, String> = mapOf(
+    "enable.auto.commit" to "false",
+    "key.deserializer" to ByteArrayDeserializer::class.java.name,
+    "value.deserializer" to ByteArrayDeserializer::class.java.name,
+)
+
+// We always seek explicitly (seekToBeginning when no token; seek(offset+1) otherwise).
+// `auto.offset.reset=none` makes it a hard failure if the explicit seek ever fails to happen,
+// rather than silently sliding to earliest/latest.
+private val DEFAULT_CONSUMER_CONFIG: Map<String, String> = mapOf(
+    "auto.offset.reset" to "none",
+)
+
+class KafkaConnectSource(
+    private val dbName: String,
+    private val cluster: KafkaCluster,
+    private val topic: String,
+    private val connectConfig: Map<String, String>,
+    private val indexer: RecordIndexer,
+) : ExternalSource {
+
+    @Serializable
+    @SerialName("!KafkaConnect")
+    data class Factory(
+        val remote: RemoteAlias,
+        val topic: String,
+        val connectConfig: Map<String, String> = emptyMap(),
+        val indexer: RecordIndexer.Factory,
+    ) : ExternalSource.Factory {
+
+        override fun open(
+            dbName: String,
+            remotes: Map<RemoteAlias, Remote>,
+            meterRegistry: MeterRegistry?,
+        ): ExternalSource {
+            val raw = remotes[remote]
+                ?: throw Incorrect(
+                    "no remote configured with alias '$remote' — add a '!Kafka' entry under 'remotes:' in node config",
+                    errorCode = "xtdb.kafka-connect-source/missing-remote",
+                    data = mapOf("alias" to remote),
+                )
+
+            val actualType = raw::class.simpleName ?: raw::class.qualifiedName ?: "unknown"
+
+            val cluster = raw as? KafkaCluster
+                ?: throw Incorrect(
+                    "remote '$remote' is a $actualType, expected a !Kafka remote",
+                    errorCode = "xtdb.kafka-connect-source/wrong-remote-type",
+                    data = mapOf("alias" to remote, "actualType" to actualType),
+                )
+
+            return KafkaConnectSource(dbName, cluster, topic, connectConfig, indexer.open(dbName))
+        }
+
+        override fun writeTo(dbConfig: DatabaseConfig.Builder) {
+            dbConfig.externalSource = ProtoAny.pack(kafkaConnectSourceConfig {
+                remote = this@Factory.remote
+                topic = this@Factory.topic
+                connectConfig.putAll(this@Factory.connectConfig)
+                indexer = this@Factory.indexer.toProto()
+            }, PROTO_TAG_PREFIX)
+        }
+
+        class Registration : ExternalSource.Registration {
+            override val protoTag: String
+                get() = "$PROTO_TAG_PREFIX/xtdb.kafkaconnectsource.proto.KafkaConnectSourceConfig"
+
+            override fun fromProto(msg: ProtoAny): ExternalSource.Factory {
+                val config = msg.unpack(KafkaConnectSourceConfig::class.java)
+                return Factory(
+                    remote = config.remote,
+                    topic = config.topic,
+                    connectConfig = config.connectConfigMap,
+                    indexer = RecordIndexer.Factory.fromProto(config.indexer),
+                )
+            }
+
+            override fun registerSerde(builder: PolymorphicModuleBuilder<ExternalSource.Factory>) {
+                builder.subclass(Factory::class)
+            }
+
+            override val serializersModule: SerializersModule = RecordIndexer.Factory.serializersModule
+        }
+    }
+
+    override suspend fun onPartitionAssigned(
+        partition: Int,
+        afterToken: ExternalSourceToken?,
+        txIndexer: TxIndexer,
+    ) {
+        LOG.info("[$dbName] Partition $partition assigned (topic=$topic)")
+
+        val keyConverter = openConverter(connectConfig, isKey = true)
+        val valueConverter = openConverter(connectConfig, isKey = false)
+        val headerConverter = SimpleHeaderConverter().also { it.configure(emptyMap<String, Any>()) }
+        val transformChain = openTransforms(connectConfig)
+
+        val mergedConsumerConfig: Map<String, Any> =
+            DEFAULT_CONSUMER_CONFIG + cluster.kafkaConfigMap + filteredConsumerConfig(connectConfig) + HARDCODED_CONSUMER_CONFIG
+
+        val consumer = KafkaConsumer<ByteArray?, ByteArray?>(mergedConsumerConfig)
+
+        try {
+            val tp = TopicPartition(topic, partition)
+            consumer.assign(listOf(tp))
+
+            if (afterToken != null) {
+                val offset = KafkaConnectSourceToken.parseFrom(afterToken).offset + 1
+                LOG.info("[$dbName] Resuming from offset $offset on $topic-$partition")
+                consumer.seek(tp, offset)
+            } else {
+                LOG.info("[$dbName] No token — seeking to beginning of $topic-$partition")
+                consumer.seekToBeginning(listOf(tp))
+            }
+
+            while (currentCoroutineContext().isActive) {
+                val records = try {
+                    runInterruptible(Dispatchers.IO) { consumer.poll(POLL_DURATION) }
+                } catch (_: WakeupException) {
+                    break
+                } catch (_: InterruptException) {
+                    break
+                }
+
+                if (records.isEmpty) continue
+
+                // Halt-on-failure: converter / transform exceptions propagate up. Mirrors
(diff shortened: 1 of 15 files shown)


modules/kafka-connect-source/src/main/kotlin/xtdb/kafkaconnectsource/KafkaConnectSource.kt
```diff
@@ -0,0 +1,380 @@
...
+                    indexer.indexRecords(sinkRecords, txIndexer)
+                }
+            }
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            LOG.error(e, "[$dbName] External source failed")
+            throw e
+        } finally {
+            runCatching { transformChain.close() }
```
jarohen, line 198: nit: `openFoo().use { foo -> ...}`. does mean nesting but saves this explicit `finally`/`runCatching` pattern.

modules/kafka-connect-source/src/main/kotlin/xtdb/kafkaconnectsource/KafkaConnectSource.kt
```diff
@@ -0,0 +1,380 @@
...
+    override fun close() {
+        LOG.info("[$dbName] Closing external source")
+        runCatching { indexer.close() }
+            .onFailure { LOG.error(it, "[$dbName] Indexer close failed") }
+    }
+}
+
+// -- Converter / Transform loaders -----------------------------------------------------------------------
+
+private fun openConverter(connectConfig: Map<String, String>, isKey: Boolean): Converter {
```
jarohen, line 251: don't use booleans here - no-one ever remembers which one's true and which one's false. prefer a small enum class instead

modules/kafka-connect-source/src/main/kotlin/xtdb/kafkaconnectsource/KafkaConnectSource.kt
```diff
@@ -0,0 +1,380 @@
...
+        .mapKeys { (k, _) -> k.removePrefix("predicates.$alias.") }
+
+    val base = (Class.forName(type) as Class<Predicate<SinkRecord>>)
+        .getDeclaredConstructor().newInstance()
+        .apply { configure(predicateConfig) }
+
+    return if (negate) NegatedPredicate(base) else base
+}
+
+private class NegatedPredicate<R : org.apache.kafka.connect.connector.ConnectRecord<R>>(
```
jarohen, line 373: imports, avoid fully qualified class names as they just add noise

modules/kafka-connect-source/src/test/kotlin/xtdb/kafkaconnectsource/DocsIndexerTest.kt
```diff
@@ -0,0 +1,64 @@
...
+        TimeZone.setDefault(TimeZone.getTimeZone("America/Los_Angeles"))
+        try {
+            return block()
+        } finally {
+            TimeZone.setDefault(original)
+        }
+    }
+
+    @Test
+    fun `Date logical type does not shift under non-UTC JVM default zone`() = withNonUtcDefaultZone {
```
jarohen, line 31: overly verbose, Claude needs reining in on these kinds of things.
you also don't _need_ to faff about with the tz because our CI tests set a random one at startup.

I'd probably go for: 
- helper fn to create the schema
- one test:

```kotlin
// set tz
try {
  assertEquals(
    LocalDate.of(2024, 6, 8),
    convertValue(Date(...), i32Schema(NAME)),
    "encodes date as LocalDate"
  )

  assertEquals(
    // same for time
    "connect encodes time as Date(millis_of_day)"
  )
} finally {
  // reset tz
}

File: modules/kafka-connect-source/src/test/kotlin/xtdb/kafkaconnectsource/KafkaConnectSourceIntegrationTest.kt [hunk 5]
```diff
@@ -0,0 +1,1080 @@
...
+
+            // Without the predicate, MaskField would crash on the tombstone's null value
+            // and we'd see an aborted tx for that record.
+            val aborted = xtQueryDb(node, "events", "SELECT committed FROM xt.txs WHERE committed = false")
+            assertTrue(aborted.isEmpty(), "no aborted txs — tombstone should bypass MaskField via the predicate, got: $aborted")
+        }
+    }
+
+    // Helpers that paper over JDBC's "could be any of these" behaviour for date / time / timestamp.
+    private fun toInstant(v: Any?): java.time.Instant = when (v) {
```
jarohen, line 449: XT's driver should be kicking out better types than this, worth checking what's going on here. or use the node query API rather than JDBC, you're in process

modules/kafka-connect-source/src/test/kotlin/xtdb/kafkaconnectsource/KafkaConnectSourceIntegrationTest.kt
```diff
@@ -0,0 +1,1080 @@
...
+        // `*_str` fields: documents the format-metadata-only behaviour (arrives as strings).
+        val timestampStr = "2024-06-08T12:34:56.789Z"
+        val dateStr = "2024-06-08"
+        // JSON Schema's `time` format requires a zone designator (RFC 3339 full-time).
+        val timeStr = "12:34:56Z"
+
+        // `*_val` fields: typed temporals via `title` + integer. Producer emits Connect's wire
+        // encoding for each logical type: Timestamp → millis since epoch, Date → days since epoch,
+        // Time → millis since midnight.
+        val expectedTs = java.time.Instant.parse("2024-06-08T12:34:56.789Z")
```
jarohen, line 592: no fully qualified class names pls

jarohen (commented):
thanks @Akeboshiwind 🙏 review comments below.

For user-facing functionality, updating/adding documentation is also part of the definition-of-done unless it's part of a wider block-of-work that all gets documented at the end - otherwise it just never gets done 😅 

As mentioned, better raised as an 'ask' given it's public API - we don't have the same kind of flexibility in making changes once this goes out, so it's always worth a pause for an explicit review.

Otherwise, great functionality to add, thank you - and looking forward to seeing it in action ☺️

Issue #4467, opened 2025-05-21
Table-not-found/column-not-found → error


jarohen:
Currently these throw warnings, largely because in 'normal' SQL you can create a table and then query the empty table immediately - in XT, without a CREATE statement, you can't.

That said, it's obviously a complete PITA to only get warnings for this when it's most likely a mistake.

IIRC, it was the Auctionmark benchmark test that unearthed this one.

related: #3411## Progress

**Status**: in-progress

- [x] empty `CREATE TABLE foo` — table existence, persists and replica-propagates (commit 307af0b26)
- [x] table-not-found → error; Auctionmark passes on empty `CREATE TABLE`s (commit 7012181d2)
- [x] memory leak in the `mixed-ops-across-boundaries` property test — root-caused to a pre-existing latent leak in `SourceLogTxIndexer.forEachArgRow` (a per-row arg relation orphaned when a tx-op query errors); fixed on main (commit eb908c438)
- [x] column declarations — `CREATE TABLE foo (col, …)` as `Nothing`, plus the `Nothing` round-trip wiring (commit 519b33937; PR #5698)
- [x] column-not-found → error on a user table; internal-schema columns tolerated (commit ea013b034; PR #5698)
- [x] docs (additive half) — new `reference/main/sql/schema.md` documenting `CREATE TABLE`; schemaless reframed as opt-in (commit 7dcb93df3)
- [ ] docs (schema-on-read half) — the unknown-table/column errors in their own right, the internal-schema tolerance, the schema-on-write vs schema-on-read framing, and the changelog blocks on the affected pages
- [ ] per-column types — `CREATE TABLE foo (a INT)` + write-type enforcement (deferred to the path-based-schema epic #3411)

jarohen:
### Chalk — `CREATE [OR ALTER] TABLE` + table/column-not-found errors

Re-scoped: ship **table**-not-found → error now (empty `CREATE TABLE` already unblocks it), and defer **column**-not-found → error plus the column-type machinery to the path-based-schema epic (#3411). Branch renamed `missing-tables-as-errors-4467`.

- [x] empty `CREATE TABLE foo` (commit 307af0b26)
- [ ] table-not-found → error; Auctionmark on empty `CREATE TABLE`s
- [ ] column declarations + `Nothing` round-trip wiring — deferred to #3411
- [ ] column-not-found → error — deferred

<details><summary>Why this shape — declare tables into existence, then error on the unknown</summary>

[...]

PR #5697, opened 2026-06-08
feat(sql)!: empty CREATE TABLE; reads of unknown tables error (#4467)

jarohen:
Part of #4467.

## Summary

Two steps toward making an unknown-table reference a hard error rather than a silent empty result. First, an empty `CREATE TABLE foo` so a table can exist in the catalog ahead of any data. Then the flip itself: reads and `UPDATE`/`DELETE`/`ERASE` against a never-created table now abort planning with `Table not found`.

## Context

Querying a table that was never created returned an empty result with a DEBUG warning, because XTDB had no `CREATE` step — you couldn't declare a table before inserting into it, so the only sane fallback for an unresolved reference was "empty". In practice that masks typos: `SELECT * FROM uesrs` silently yields nothing. The Auctionmark benchmark originally surfaced this, and is the acceptance here.

The empty-`CREATE TABLE` increment removes the reason the warning existed — you can now declare a table up front — which unblocks the error.

## Changes

1. **CREATE [OR ALTER] TABLE registers an empty table** (307af0b26) — an empty `CREATE TABLE foo` as a transaction op (not a pgwire control-plane message), so it's atomic within a transaction and mixable with DML. Both `CREATE TABLE` and `CREATE OR ALTER TABLE` spellings are accepted, and creating an existing table is a no-op. The table is a schema-bearing 0-row relation that persists through a block flush and replica-propagates to followers, so the dataless-table path had to be cleared at the live index, the replica-log serialisation, and the 0-row page's temporal metadata.

2. **reads of unknown tables error instead of warning** (2f3dafbdc) — the warn→error flip for reads + `UPDATE`/`DELETE`/`ERASE`, plus the Auctionmark setup and test migration below.

## Behaviour change (the `!`)

A query against a table that doesn't exist used to return an empty result; it now errors. To adopt: declare the table with `CREATE TABLE`, or fix the typo. `INSERT` is unaffected — it still auto-creates, matching "the write-type is set on first insertion" — so the error is for the read side only.

Two classes of unresolved reference stay warnings, because empty is the right answer:

- data-backed `xt` system tables like `xt.txs`, absent from the catalog until the first transaction lands — `SELECT COUNT(*) FROM xt.txs` on a fresh node must return 0;
- the optional `pg_catalog` / `information_schema` tables that client tools (Metabase, psql) probe whether or not XTDB implements them.

`system-table-chain?` draws the line: a reference into an internal schema (`xt`, `pg_catalog`, `information_schema`), or a `pg_*` table name in any chain position, falls through to an empty scan; a user-schema table errors.

## Usage

```sql
CREATE TABLE users;                  -- declare an empty table
SELECT * FROM users;                 -- [] (no warning)
SELECT * FROM uesrs;                 -- ERROR: Table not found: uesrs
INSERT INTO posts (_id) VALUES (1);  -- still auto-creates `posts`
```

## Out of scope — deferred to #3411

Column declarations (`CREATE TABLE foo (col, …)`) and column-not-found → error. Erroring on an unknown column needs a way to declare columns first, plus the `Nothing` bottom-type round-trip wiring those columns require — that work belongs to the path-based-schema epic.

## Test plan

Full suite green (507 tests). Auctionmark passes — a `create-tables!` step declares its schema before the load phase, since its workers read tables (`item_max_bid`, `item_bid`, …) before the load writes them. ~26 existing tests that relied on warn-and-continue were migrated: declare-first where the missing table was incidental, expect-error where the missing-table behaviour is the subject. A psql regression (`xt.txs` empty on a fresh node erroring through to a stderr-only response) is what pulled `xt` into the internal-schema exception.


jarohen:
The property-test CI failure here (`mixed-ops-across-boundaries`, an 8-byte allocator leak) was root-caused to a **pre-existing latent leak** in `SourceLogTxIndexer.forEachArgRow` — a per-row argument relation orphaned when a tx-op query errors — not to anything on this branch. The table-not-found error flip just gave it a new way to trigger.

Fixed separately on `main` with `closeOnCatch` (commit eb908c438). Rebasing this branch onto updated main picks up the fix and the property-test job goes green.

jarohen:
User-facing docs for this change (the `CREATE TABLE` statement and the unknown-table error) aren't in this PR — they land in the subsequent PR, #5698, where they're written alongside the column work so the full schema story (table existence, columns, and the read-side errors) is documented together. The docs punch list is tracked on #4467.

Issue #5711, opened 2026-06-11
Teardown hangs: cancellation-immune waits surfaced by job-owned lifecycle (CI timeouts since 2026-06-05)


mbutlerw:
## Summary

Since 2026-06-05, CI test jobs intermittently hang until their timeout. Investigation — culminating in a live coroutine dump from a hung run — shows this is not one bug but one **bug class**: teardown paths waiting, now honestly and unboundedly, on *waits that cancellation cannot reach*. The job-owned-lifecycle work (#5676, then #5703 and the compactor follow-up) progressively removed the layers of forgiveness — bounded joins, detached cleanup, abandoned coroutines, process kill — that had been silently absorbing these waits for a long time. Each layer removed surfaces another set as permanent hangs. The refactors created none of the waits; they audit them.

## The model — a two-stage unveiling

Kotlin cancellation only lands at suspension points; a thread parked in blocking code, or a wait on a counterparty outside the cancelled tree, is unreachable.

**Stage 0 (pre-#5676):** teardown was hand-sequenced explicit closes — some with bounded joins (e.g. the GC collectors' 10s `cancelAndJoin`), some never awaited — and at the outer layer the test-worker JVM exits and Gradle reaps orphan processes. Cancellation-immune waits ended in a timeout warning, a leaked thread, or `kill` — invisibly.

**Stage 1 (#5676, Jun 5):** `Database.close()` becomes a single unbounded `cancelAndJoin` over every coroutine **body** in the tree. Any body that can't take cancellation now hangs close forever. `d751ac91f` (#5680), the same morning, made every block-less `flush-block!` drive the persister through a replica append — so teardown now routinely arrives while bodies are mid-work. Hangs begin appearing on main that day.

**Stage 2 (#5703 + the compactor port):** moving resource cleanup from `invokeOnCompletion` into the coroutine body's `finally` — the correct fix for the handler/join race — also pulls **cleanup code** inside the join. Cleanup that blocks forever, previously a detached silent leak, becomes a hang. This stage's hangs appear only on branches carrying the fences.

[...]

PR #5712, opened 2026-06-11
fix(indexer): single persister task channel, not selectUnbiased

mbutlerw:
Part of #5711 — the issue carries the full deadlock analysis and the CI-timeline evidence; this PR lands the first (and for now only) remediation.

The persister received tasks via `selectUnbiased` over three per-source rendezvous channels, with `onUndeliveredElement` as the loss-protection for cancellation races. That protection is only reliable for plain send/receive — a task handed off in a select/cancellation race is owned by nobody: the persister's catch-chain never sees it and `onUndeliveredElement` never fires, so the task's `onComplete` is never resolved and its submitter awaits forever. When the submitter is the shared log-consumer loop (outside the database job tree, so unreachable by `Database.close`'s cancel), that await stops the loop polling and deadlocks close until the CI action timeout.

The fix collapses the three channels into one `Channel<PersisterTask>` received with plain `receive()`, making the lost-handoff state unrepresentable. A single FIFO queue preserves the cross-source fairness the unbiased select was approximating — no task source can starve another. All failure-path behaviour is otherwise unchanged: a closed/cancelled channel lands in the same catch/finally the select did.

Deliberately one fix at a time: this is the only candidate that explains all three observed hang shapes (including the in-memory-log hangs, where the other suspect — uninterruptible blocking Kafka producer calls — cannot apply). #5711 stays open to watch CI for recurrence; the producer-side `runInterruptible` wrap is queued there if needed.

Test plan: `xtdb.indexer.*` + `xtdb.database.*` unit tests, `LogProcessorSimTest`/`NodeSimulationTest` property runs, and the Clojure `xtdb.indexer-test`/`xtdb.api-test` suites — all green. The real verification is CI integration runs over the coming days, per the issue.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/src/main/kotlin/xtdb/indexer/LeaderLogProcessor.kt
+++ b/core/src/main/kotlin/xtdb/indexer/LeaderLogProcessor.kt
@@ -4,7 +4,6 @@ import io.micrometer.core.instrument.Counter
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.channels.Channel.Factory.RENDEZVOUS
-import kotlinx.coroutines.selects.selectUnbiased
 import org.apache.arrow.memory.BufferAllocator
 import xtdb.Metrics.withSpan
 import xtdb.NodeBase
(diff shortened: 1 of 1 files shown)


jarohen:
@mbutlerw presumably this'd be considered a bug in kotlinx.coroutines, given you don't seem to be able to access the element?

at least, I wonder if anyone else has documented having run into it?

mbutlerw:
I think Claude was overeager here, closing for now.

PR #5713, opened 2026-06-11
test: hang watchdog — dump coroutines/threads when a test runs long

mbutlerw:
Part of #5711. The close-path hangs die silently: the test stops logging, the CI action times out 15 minutes later, and the log names neither the test nor the stuck coroutine. Six occurrences have each cost an investigation, and the in-memory shape remains unattributed — so the diagnostic lands ahead of further speculative fixes.

A JUnit Platform `TestExecutionListener` (new `modules/test-watchdog`, ServiceLoader-registered, on every `Test` task's runtime classpath via the root's shared test config) tracks in-flight tests. When one passes the threshold — default 6 minutes: above any legitimate single test, below the action timeout — it prints kotlinx `DebugProbes.dumpCoroutines()` plus a full thread dump to stderr, which Gradle captures into the test-results XML. It reports at the threshold and again at 2×, so movement between the dumps distinguishes stuck from slow, then goes quiet.

Design notes: dump-only — it never fails or interrupts a test, so a slow-but-progressing test costs log noise, not a failure. Coroutine creation stack traces are disabled (a throwable per launch is far too hot for the sims). `jdk.attach` joins the jlink module list — DebugProbes installs by dynamic self-attach, which the custom test JRE otherwise lacks; if install fails anyway, the watchdog degrades to thread dumps rather than failing the run. The module is deliberately not `java-library`, since the root's shared test-config block assumes clojurephant.

Test plan: unit tests for the threshold/reporting logic (fake clock); end-to-end verification with a deliberately hanging scratch test (since deleted) — reports fired at threshold and 2×, the sleeping coroutine appeared in the dump as `BlockingCoroutine{Active}`; `:xtdb-api:test` green on the rebuilt custom JRE with probes installing cleanly and zero watchdog noise on a healthy run.

### DIFF
--- a/build-logic/jlink/build.gradle.kts
+++ b/build-logic/jlink/build.gradle.kts
@@ -26,6 +26,8 @@ val jlinkModules = listOf(
     // clojureRepl
     "jdk.compiler",
     "jdk.javadoc",
+    // DebugProbes dynamic self-attach (test hang watchdog, #5711)
+    "jdk.attach",
 )
 
 val customJreDir = layout.buildDirectory.dir("custom-jre")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -256,6 +256,10 @@ allprojects {
             testRuntimeOnly(libs.junit.jupiter.engine)
             testRuntimeOnly(libs.junit.platform.launcher)
 
+            // dumps coroutines + threads when a test exceeds the hang threshold — see #5711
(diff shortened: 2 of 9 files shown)


mbutlerw:
Superseded by #5721 — same branch, rewritten to the final single-commit design (quiet-period trigger, companion-object state, `-javaagent` probes). Closing-state was unreopenable, hence a fresh PR.

PR #5717, opened 2026-06-15
feat(datasets): GLEIF LEI dataset + Postgres-source federation

jarohen:
## Summary

A securities-reference dataset for XTDB built on the public GLEIF LEI register, and the Postgres-source federation that populates XTDB from it. GLEIF lands in a plain Postgres securities-master (current state, no history); a custom Postgres-source indexer (`GleifPgIndexer`) follows it into XTDB over CDC, deriving a bitemporal history the source-of-record never holds.

The end-to-end run-through — bring up the supporting cast, `ATTACH`, load, query — lives with the demo harness under `modules/datasets/gleif`: a README, plus a `(comment …)` block at the bottom of `xtdb.datasets.gleif`. The supporting fixes the work surfaced — `:remotes` Clojure config, two pgwire host bugs, a sub-minute-offset timestamp parser fix — landed separately on `main`.

## Why GLEIF

It's redistribution-clean (CC0) and genuinely bitemporal — its `LegalEntityEvents` carry both an *effective* and a *recorded* date, so the bitemporal story is real data, not synthesised. The licensed vendor feeds (S&P, Compustat, Capital IQ) were ruled out: they can't be redistributed.

## What a GLEIF publish actually is

This shaped every modelling decision, and it's the least obvious thing about the feed. A GLEIF "delta" is **not a diff**. Each touched entity is emitted as a full snapshot — current state plus the entity's complete event history for all time — re-sent verbatim on every publish that touches it (verified: one entity identical across 20 consecutive daily deltas). The anchor full-file is the same shape.

GLEIF also splits one logical change across three places: the *new value* is the top-level snapshot field (`legal_name`); the *when* is a dated marker event (`change-legal-name` carries only effective/recorded dates, **never the value**); the *old values* sit under `other-names`. A single snapshot can't reconstruct which former value held in which interval. Forward replay can, because an event only *appears* in the feed at the publish where it was recorded — verified: an address change effective 2025-05-28, recorded 2025-06-17, first appearing in the 2025-06-18 publish. So **system-time = publish date** (when XTDB learned it), **valid-time = effective date** (when it became true), and the divergence between them is the demonstration.

## Modelling — state vs event

The deepest decision. GLEIF files everything under one `LegalEntityEvents` list, but the events are two different kinds of fact:

- **State transitions** (the five `change-*` types). A `change-legal-name` isn't an event to log — it's a *transition* of the entity's name. The new value lives on the snapshot; the event only dates it. In XTDB — whose effective primary key is `_id` + `_valid_from` — these key by `lei`, and successive changes become **valid-time versions of one document**: the bitemporal name/address history. There is no separate "name-change log"; the valid-time axis *is* the change history.
- **Occurrences** (the other fourteen: dissolution, merger, liquidation, …). Things that happened at a point, that don't supersede a prior value, and an entity can have several. These key by `lei__effective_date`, one immutable document each.

The test: *does this supersede a previous value (→ state) or did it just happen (→ event)?* Most "event" feeds are a mix, and the source's own taxonomy won't tell you which is which.

## The two sinks diverge

The stores want opposite things, so each projects the records its own way:

**Postgres** (relational, current-state SOR): a `lei` table (UPSERT by LEI, nested bits as `jsonb`), plus a table per `LegalEntityEventType`. The five `lei_*_changes` tables carry the then-current attribute value, baked in from the snapshot at load — a record only ever shows the one change that produced its current value, so the pairing is 1:1. The fourteen occurrence tables (`lei_dissolutions`, …) are date-only. Enums are native PG enum types declared from the XSD lists. `(lei, effective_date)` PK throughout.

**XTDB** (schemaless, bitemporal): populated *only* via `GleifPgIndexer` over CDC — there is no native XTDB loader; the federation is the population story. It derives `_id` and valid-time per table family (entity → `lei` valid from system-time; state changes → `lei` valid from effective date; occurrences → `lei__effective_date`), re-keys PG enum text to keyword values, and passes `jsonb` through (string-keyed maps collapse to the same XT struct fields). The indexer carries a `!Gleif` registration so its factory travels in `ATTACH DATABASE` config — letting a full, queryable node run it.

## Notable findings & decisions

- **Produce/consume split.** `xtdb.datasets.gleif.mirror` is a REPL-only produce ritual: it fetches the GLEIF API, decodes the LEI-CDF 3.1 JSON, curates to GB-jurisdiction entities (~228k), and writes idiomatic-record transit to S3. The load path reads plain transit with the stock serde handlers and never touches the mirror or knows GLEIF JSON exists. The anchor full-file is ~1GB; the mirror streams it with Jackson tokens (one record's subtree at a time) — slurp-then-filter OOM'd.
- **Windowed event inserts.** Because GLEIF re-sends an entity's whole event history on every publish, a naive loader re-UPSERTs the same events ~20× for a frequently-touched entity. Each event is instead inserted only in the publish whose `(prev, this]` recorded-date window first carries it.
- **Identity must be byte-stable across the federation boundary.** Event `_id`s use the effective date's canonical `Instant` form, not the `timestamptz` column's own `toString` — a `ZonedDateTime` renders the same instant differently, which would split one event into two identities.
- **pgjdbc can't bind a `PGobject` to a custom enum** ("Unknown type", even schema-qualified) → bind enum values as text with a SQL cast (`?::entity_category`); Postgres coerces server-side. `jsonb` is fine via `PGobject` (registered natively).
- **in-progress events have a null effective-date** (recorded, not yet legally in force) → default `effective_date` to `recorded_date` so the `(lei, effective_date)` PK holds.
- **Schema ownership.** `modules/datasets/gleif/init/postgres/01-gleif.sql` owns the schema (tables, enums, `FOR ALL TABLES` publication); the loader only UPSERTs.
- **Dataset location.** The curated transit downloads to `src/test/resources/data/gleif` alongside the other datasets, loaded by filesystem path (not `io/resource`). The demo harness dir shares the `gleif` name with the S3 prefix, so `download-dataset.sh` disables shell globbing — otherwise the unquoted `gleif/*` include would expand against the local harness dir before `aws` saw it.

## Usage (for an XTDB developer)

**You do not run the mirror.** `xtdb.datasets.gleif.mirror` is a one-time produce job that's already been run — it fetched the GLEIF API and wrote the curated transit to `s3://xtdb-datasets/gleif/`. You only consume what's there. (The mirror exists so the dataset can be regenerated or re-curated later; for using the dataset, ignore it.)

**1. Get the dataset from S3** (curated GB slice, ~228k entities, 2025-06-15 anchor + a 12-month delta tail — lands under `src/test/resources/data/gleif`, gitignored, alongside the other datasets):

```sh
modules/datasets/scripts/download-dataset.sh --gleif
```

**2. Bring up the supporting cast** — Kafka, Postgres, Metabase, Prometheus, Grafana (XTDB itself runs from a REPL, since `GleifPgIndexer` has to be on the node's classpath):

```sh
cd modules/datasets/gleif && docker compose up -d
```

The compose network is pinned to `172.30.0.0/16`; the host node binds that network's gateway (`172.30.0.1`) — reachable from the containers but not the LAN — and every published port binds `127.0.0.1`. A container reaches the node *via the host*, so a `ufw`-style host firewall (default `INPUT DROP`) needs a rule allowing that subnet to the node's ports (`8081` metrics, `5432` pgwire) — see the README's Troubleshooting. (Postgres has no data volume, so `down` resets it; the step-3 load repopulates it.)

Grafana (`http://localhost:3000`, `admin`/`admin`) comes up pre-provisioned with the repo's canonical XTDB metrics dashboards (over Prometheus, scraping the node's `/metrics`) and an XTDB pgwire datasource for querying the GLEIF data directly.

**3. Run the demo** from the `(comment …)` at the bottom of `xtdb.datasets.gleif` (prose walkthrough in the harness README) — `./gradlew :clojureRepl`, then work through it:

1. Start a node — pgwire + `healthz` on the pinned gateway (`172.30.0.1`), Kafka log, the `pg` Postgres remote.
2. `ATTACH DATABASE gleif … indexer: !Gleif` while Postgres is still empty; CDC begins.
3. `submit-gleif-pg!` loads Postgres; watch `SELECT count(*) FROM gleif.lei` on the node climb to ~228k as it streams in over CDC.
4. The bitemporal queries — name history as valid-time, the system-time-as-of divergence — against the live XTDB node.

The node + `ATTACH` config in that comment is reconstructed from the tests (the original driver namespace wasn't committed), so sanity-check it before leaning on it.

## Verification

The Postgres sink is verified end to end against the demo Postgres: **227,922 `lei` rows** load, events fan out across the per-type tables, enums/jsonb/timestamptz land as the right types, the null-effective-date fallback holds, re-runs are idempotent. The **full federation path runs against a live node**: `ATTACH`, CDC snapshot, all 227,922 rows mirrored into XTDB, and the bitemporal queries return — including a real GB entity (`984500Z4F0C6A9BE0594`) that renamed itself twice, where XTDB reconstructs the valid-time history and the system-time-as-of divergence that Postgres structurally cannot express. The observability surfaces run live too — Prometheus scrapes the node, Grafana's provisioned XTDB dashboards populate, and the pgwire datasource queries the GLEIF tables (once the host firewall allows the subnet).

## Scope / known limitations

- **Relationships** (the `rr` file: parent hierarchy, periods, qualifiers/quantifiers) are deferred — the temporal hierarchy deserves its own modelling pass; the mirror can fetch the files when we get there.
- **No automated test** — this is demo material; the manual end-to-end run is the test.

### DIFF
--- a/modules/datasets/build.gradle.kts
+++ b/modules/datasets/build.gradle.kts
@@ -4,6 +4,10 @@ plugins {
 
     `maven-publish`
     signing
+    alias(libs.plugins.kotlin.jvm)
+    // GleifPgIndexer.Factory is @Serializable so it can travel as persisted
+    // secondary-database config (its Registration round-trips via protobuf Struct).
+    alias(libs.plugins.kotlin.serialization)
 }
 
 java.toolchain.languageVersion.set(JavaLanguageVersion.of(21))
@@ -14,10 +18,19 @@ dependencies {
     api(project(":xtdb-core"))
 
     api(libs.tpch)
-    api(libs.clojure.data.csv)
-
-    api(libs.aws.s3)
+    api(libs.jsonista)
     api(libs.next.jdbc)
+
+    // the GLEIF mirror fetches the API (hato) and writes idiomatic records as transit
+    api(libs.hato)
+    api(libs.transit.clj)
+
+    // the GLEIF PgIndexer (Kotlin) writes Postgres-source rows into XT; brings
+    // PgIndexer / RowOp / the Postgres ExternalSource.
+    implementation(project(":modules:xtdb-postgres-source"))
+    implementation(kotlin("stdlib"))
+    // the Postgres sink references PGobject directly (enum/jsonb binding).
+    implementation(libs.pgjdbc)
 }
 
 publishing {
--- a/modules/datasets/gleif/README.md
+++ b/modules/datasets/gleif/README.md
@@ -1,0 +1,104 @@
+# GLEIF federation demo
+
+This demo loads the public GLEIF LEI register (legal-entity reference data) into plain Postgres — a current-state system-of-record — then has XTDB *follow Postgres over CDC*, deriving a bitemporal history the source never held.
+GLEIF is the vehicle because it's redistribution-clean (CC0) and genuinely bitemporal: its events carry both an effective and a recorded date, so the temporal story is real data, not synthesised.
+
+The federation *is* the XTDB-population story — there's no native XTDB loader.
+XTDB attaches Postgres as an external source, and a custom `GleifPgIndexer` shapes each CDC row into a bitemporal document.
+
+## What runs where
+
+XTDB itself runs from a **REPL on the host**, not as a container — it's the only way to put `GleifPgIndexer` on the node's classpath.
+`docker-compose.yml` brings up only the supporting cast:
+
+| Service | Host port | Role |
+| --- | --- | --- |
+| Kafka | `127.0.0.1:9092` | XTDB log |
+| Postgres | `127.0.0.1:5432` | GLEIF sec-master (db `gleif`, schema from `init/postgres/01-gleif.sql`) |
+| Metabase | `127.0.0.1:3001` | queries XTDB pgwire at `172.30.0.1:5432` |
+| Prometheus | `127.0.0.1:9090` | scrapes XTDB `/metrics` at `172.30.0.1:8081` |
+| Grafana | `127.0.0.1:3000` | dashboards (`admin`/`admin`) |
+
+Every published port binds `127.0.0.1`, so nothing is reachable from an untrusted LAN — only from the host.
+The host XTDB node needs the same care: bind it to `172.30.0.1` — the gateway of this stack's pinned network (`docker-compose.yml`), i.e. the host as the containers see it, which they can reach but the LAN can't — **not** `*`.
+
+## Running it
+
+The REPL snippets live in a `(comment …)` block at the bottom of the relevant namespace.
+This README is the map; those comments are the copy-paste.
+
+1. **Get the dataset** — it's not checked in.
+   Sync it from S3 (needs AWS creds) with `modules/datasets/scripts/download-dataset.sh --gleif` — it lands under `src/test/resources/data/gleif/`.
+   Or produce it locally with `xtdb.datasets.gleif.mirror/mirror!`, which fetches the GLEIF API.
+   The anchor is ~1GB, so producing it is slow.
+
+2. **Bring up the supporting cast.**
+   ```bash
+   cd modules/datasets/gleif && docker-compose up -d
+   ```
+   Postgres creates the GLEIF tables, enum types and CDC publication on first boot from `init/postgres/`.
+
+3. **Start XTDB and federate**, then **load GLEIF into Postgres and watch it stream in**, then **query the bitemporal history** — all from the root REPL (`./gradlew :clojureRepl`).
+   The end-to-end run-through is the `(comment …)` at the bottom of `xtdb.datasets.gleif` (`src/main/clojure/xtdb/datasets/gleif.clj`).
+   It starts a node with the `pg` remote and `:kafka` log-cluster, `ATTACH`es `gleif` while Postgres is still empty, then loads Postgres — which streams into XTDB over CDC — and runs the demo queries.
+   The Postgres-load step has its own rich comment, with `psql` verification queries and how to watch per-batch progress, at the bottom of `xtdb.datasets.gleif.pg`.
+
+After the load, `SELECT count(*) FROM lei;` should return **227,922** — that alone confirms the dataset and loader.
+Once the `gleif` database is attached, the `q-entity-as-of-system-time` / `q-entity-as-of-valid-time` / `q-legal-name-history` queries (defined in `xtdb.datasets.gleif`) show truth-as-of-then versus corrected-truth-now — the point the federation makes.
+
+## Exploring the data
+
+Metabase and Grafana both query the live XTDB node over pgwire — the same endpoint the run-through binds to the compose network gateway — so the node has to be up with `gleif` attached and loaded (the steps above).
+They connect with the same details:
+
+- Host `172.30.0.1` (this stack's network gateway — the host node, reachable from the containers), port `5432`.
+- Database `gleif`, user `xtdb`, no password, SSL disabled.
+
+### Metabase — http://localhost:3001
+
+1. Open Metabase and complete the first-run admin account setup.
+2. Add a database of type **PostgreSQL** with the connection details above.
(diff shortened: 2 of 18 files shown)


jarohen:
## Progress

**2026-06-16 (later) — federation runs live end-to-end; state/event model; demo extracted**

The whole "XT follows Postgres" path now works against a live node, and running it surfaced the modelling and several core bugs.

- **State vs event split.** GLEIF's `LegalEntityEvents` are two kinds of fact. The five `change-*` types are *state transitions* — the value lives on the snapshot, the event only dates it — so in XT (PK = `_id` + `_valid_from`) they key by `lei` and successive changes are valid-time versions of one doc (the bitemporal name/address history). The fourteen occurrences key by `lei__effective_date`. Postgres mirrors this: `lei_*_changes` carry the snapshot value baked in at load (1:1 — a record only shows the change that produced its current value); occurrences stay date-only.
- **GleifPgIndexer is the sole XT-population path** — the native XTDB loader was removed; the CDC stream *is* the federation story. Realigned to the `lei` / `lei_*_changes` / occurrence table set, `_id`+valid-time per family, jsonb passed through. Made `@Serializable` with a `!Gleif` registration so it travels in `ATTACH DATABASE` config — a full queryable node runs it (an `IngestNode` could too, but has no pgwire, so couldn't serve the demo queries).
- **Verified live.** `ATTACH DATABASE … indexer: !Gleif`, CDC snapshot, **227,922 rows** mirrored into XT. The shiny query: GB entity `984500Z4F0C6A9BE0594` renamed twice (TRADEWAVE CAPITAL PLC → TRADE HEDGE PLC → TRADE HEDGE LIMITED); XT reconstructs the valid-time history *and* the system-time-as-of divergence (last rename effective 2025-08-06, recorded ~2025-09-10), which the Postgres SOR structurally can't express.
- **Bugs the run surfaced — fixed + landed on `main` separately, not in this PR:** `:remotes` Clojure node config (you couldn't name a Postgres source remote from Clojure); pgwire `:host` rejecting a host string; pgwire's own `DataSource` hardcoding `localhost` instead of the bound host (which silently routed the node's queries to the docker Postgres → a misleading SCRAM error); and the SQL-timestamp parser rejecting sub-minute zone offsets (GLEIF's pre-1847 dates render through Europe/London LMT as `-00:01:15`, straight from tzdata).
- **Demo extracted.** The REPL run-through moved to the `xt26` dev namespace on its own branch; the standalone README was folded into it. This PR is now just the dataset + indexer.

Next: relationships (the `rr` file — parent hierarchy, periods) remain deferred to their own modelling pass.


**2026-06-16 — mirror refactor + Postgres model landed (verified end-to-end)**

Reworked the data layer into a produce/consume split and reshaped both sinks around what a GLEIF publish actually is.

- **Produce/consume split.** `xtdb.datasets.gleif.mirror` now owns all GLEIF-API fetch + JSON decode as a REPL-only produce ritual writing curated transit to S3; the load path reads plain transit with stock serde handlers. Replaces the old shared `parse.clj` + on-load-path JSON decode. Streams the ~1GB anchor with Jackson tokens (the slurp-then-filter cut OOM'd).
- **Understood the feed.** A delta is a *full snapshot + complete event history re-sent verbatim*, not a diff (verified: one entity identical across 20 daily deltas). The change *value* is the snapshot field, the *when* is a dated marker event, the *old values* are under `other-names`. system-time = publish date, valid-from = effective date; the divergence is real (address change effective 2025-05-28, recorded 2025-06-17, published 2025-06-18).
- **Postgres model verified end-to-end.** `lei` table (renamed from `lei_record`) + 19 event tables (`lei_*_changes` + pluralised occurrences), `(lei, effective_date)` PK. **227,922 `lei` rows** load, events fan out correctly, re-runs idempotent. Event inserts are *windowed* by recorded-date so we don't re-UPSERT whole histories ~20×. Schema owned by `init/postgres/01-gleif.sql` (dropped `create-tables!`).
- **Two PG traps fixed.** pgjdbc can't bind a `PGobject` to a custom enum → bind text + `?::enum` cast. in-progress events have null effective-date → default to recorded-date for the PK.
- **Demo dir** renamed `demo/` → `xt26-demo/`.

Next pass: split events into their own XT tables (XT sink is snapshot-only for now), realign `GleifPgIndexer` to the new table names, run the XT sink + indexer CDC path end-to-end.

PR #5720, opened 2026-06-15
feat(datasets): SEC EDGAR fundamentals — mirror → transit → Postgres → XT

jarohen:
## Summary

Another federation dataset for XTDB, a sibling of GLEIF: load a curated slice of SEC EDGAR company fundamentals into plain Postgres as a current-state system-of-record, then have XTDB *follow Postgres over CDC*, deriving the bitemporal history the source never held — point-in-time fundamentals and restatements visible on the system-time axis. No tracking issue; this description carries the full context.

## Why EDGAR

EDGAR earns its place because its bitemporality is *real*, not synthesised: companies restate — re-filing corrected figures for a period they already reported — so the system-time axis carries a genuine "what we believed, and when". Energy Focus, for instance, first reported FY2024 net income as +$1.6M (10-Q, May 2025), restated it to −$1.6M that August, then refined it to −$1.582M in the March 2026 10-K. Instant balances that a later as-of supersedes give the valid-time axis a real timeline. And it's public-domain SEC data, so it ships without redistribution constraints.

## The model (the load-bearing design)

Facts project into three wide tables split by temporality, because a flow and a balance are different kinds of fact:

- `issuer` — static reference, one row per filer.
- `income_statement` — duration flows; a figure is fixed for the window it reports, so it keys on `(cik, period_start, period_end)`, and a restatement re-files the same period on the **system-time** axis. A `period_end` can carry both a quarterly and a year-to-date duration — distinct statements — which is why `period_start` is part of the key.
- `balance_sheet` — instant balances; an as-of value a later as-of supersedes, keyed `(cik, period_end)` — a genuine **valid-time** timeline.

Valid-time models *when the fact is valid/known*, not the period it describes: `income_statement` valid-from = `filed` (when reported), `balance_sheet` valid-from = `period_end` (as-of). `period_start`/`period_end` stay as plain domain columns.

Two ways in, one model:

- **Native XT loader** (`submit-edgar!`) replays filings oldest-first, one transaction per `filed` date stamped at that date's system-time, reconstructing the calendar-accurate vintage history a node would have built consuming each filing as it landed.
- **Postgres sink** (`edgar.pg`) UPSERTs current state (latest-`filed`-wins); XTDB then follows it over CDC via `EdgarPgIndexer`, which derives `_id` and valid-from per row to match the native loader exactly — so a bulk load and a CDC stream produce identical identities.

`EdgarPgIndexer` carries an `!Edgar` `Registration` so it travels in `ATTACH DATABASE` config onto a full, queryable node (the demo queries EDGAR over pgwire, which a bare IngestNode wouldn't serve). Its empty proto carrier is `google.protobuf.Empty`, not GLEIF's `google.protobuf.Struct`: the proto type-URL is the sole discriminator when a persisted secondary is read back, and both indexers register in one `META-INF/services`, so a shared carrier would let one database deserialise as the other.

## Usage

The demo is a self-contained sibling of GLEIF's under `modules/datasets/edgar/` — README as the map, the copy-paste REPL run-through in the `(comment …)` at the bottom of `xtdb.datasets.edgar`. It shares GLEIF's host ports, so run one demo at a time.

1. Get the dataset (not checked in): `scripts/download-dataset.sh --edgar` (S3, needs AWS creds), or produce it locally via `xtdb.datasets.edgar.mirror/mirror!`.
2. `cd modules/datasets/edgar && docker compose up -d` — Postgres creates the tables + `edgar_pub` on first boot from `init/postgres/`.
3. From the root REPL: load Postgres, start a node bound to the pinned network gateway `172.30.0.1`, `ATTACH DATABASE edgar` with the `!Postgres` external source + `!Edgar` indexer, then run the `q-*` demo queries.

Two operational notes are baked into the config: a filing coalesces into one transaction that can exceed Kafka's 1 MB default, so the demo raises the broker + client record caps to 16 MB; and the containers reach the host node at `172.30.0.1` (a pinned compose subnet), which a `ufw` host may need explicit firewall rules for (README Troubleshooting).

## Verified

- 10 unit tests green (`edgar_test`, `edgar_tsv_test`, `edgar_snapshot_test`, `edgar_restatement_test`): parse/pivot, a transit round-trip, and the restatement-on-system-time centrepiece end-to-end through mirror → transit → load. No reflection or boxed-math warnings.
- Live end-to-end: a full node attached `edgar` and tailed Postgres CDC, replicating **6,609 issuers / 73k current income rows / 92.6k income vintages** across system-time. Restatements are queryable — Energy Focus's FY2024 sign-flip trail via `FOR ALL SYSTEM_TIME` — and CDC-derived `_id`s match the native loader.

## Scope / not covered

- Over CDC, system-time is *ingestion* time (when XT sees each Postgres change), so `FOR SYSTEM_TIME AS OF <historical date>` returns nothing — the restatement story shows via `FOR ALL SYSTEM_TIME` ordered by `filed`. Only the native loader back-dates system-time to `filed`. This is inherent to federation, not a gap to close.
- Entity relationships (issuer hierarchies) are deferred — the temporal hierarchy deserves its own modelling pass.
- Edge case: two distinct accessions reporting the same `(cik, period_start, period_end)` on the same `filed` date collide on `_id` within one transaction (the second wins). Rare in practice; acceptable for the demo.

## Test plan

From the worktree:

```
./gradlew :modules:xtdb-datasets:compileKotlin
./gradlew :test --tests 'xtdb.datasets.edgar_test' --tests 'xtdb.datasets.edgar_tsv_test' --tests 'xtdb.datasets.edgar_snapshot_test' --tests 'xtdb.datasets.edgar_restatement_test'
```

All 10 green.


jarohen:
### Checkpoint — pipeline aligned to GLEIF, runs end-to-end

This session brought the EDGAR pipeline onto the same rails as GLEIF and proved the Postgres half against a live database.

**`mirror!` now fetches from SEC itself.** Previously a bash script (`download-edgar-tsv.sh`) pulled the raw quarterly ZIPs and `mirror!` read them off disk. That split diverged from GLEIF, whose `mirror!` fetches its API directly — download and transform in one Clojure pass. EDGAR's `mirror!` now does the same: pulls each quarter's ZIP from `sec.gov` (hato), reads sub.txt + num.txt out of it (via `ZipFile`, for the random access the num→sub join needs), filters to the registered consolidated observations, writes `<quarter>.transit.json.gz`. The bash downloader is gone. Tests can't drive `mirror!` any more (it hits the network), so the local-file work moved to a shared seam — `write-quarter-transit!` (sub-reader + num-reader → transit) — which the fetch path and the tests both use.

**The xt26-demo compose stack is in**, copied from GLEIF and pointed at EDGAR: Kafka, Postgres (`wal_level=logical`, the three tables + `edgar_pub` created by `init/postgres/01-edgar.sql` on first boot), Metabase, Prometheus, Grafana — same 127.0.0.1 port-binding hygiene.

**Running the PG sink for real surfaced and fixed a bug.** `upsert-sql` built column names straight from the doc keys, but the fixed keys are kebab (`:period-start`) while only the line items are snake — so the generated SQL carried `period-start` and Postgres rejected the hyphen (`syntax error at or near "-"`). A `col-name` helper now munges any key to its snake column name for the SQL, while the param lookup still uses the doc's own key.

With that, the full pipeline runs end-to-end — **mirror (SEC → transit) → S3 → download → `submit-edgar-pg!`** — loading **6,609 issuers / 73,362 income-statement / 57,073 balance-sheet rows** across the four quarters (2025q2–2026q1). Latest-filed-wins collapses re-sends to one row per `(cik, period)`; spot-checked Apple's annual + quarterly income statements and balance sheets directly.

One thing the spot-check made concrete, worth recording: in Postgres, Apple's pre-2025-06 balance-sheet rows show only `stockholders_equity`, with `assets`/`liabilities` null. That's not a loader bug — it's latest-filed-wins keeping the most recent filing to mention each `period_end`, which for older balances is a *comparative* in a later 10-Q that only carried equity. The fuller earlier vintage (the original 10-K's complete balance sheet) is exactly what XTDB retains under `FOR ALL SYSTEM_TIME` and Postgres drops — the federation demo's point in miniature.

Still open: the live CDC run through an `IngestNode` + `EdgarPgIndexer` (wiring in `CDC_WIRING.md`; Postgres is now standing by with the publication and `wal_level=logical`); the actual S3 upload of the mirrored transit; pointing Metabase/Grafana at the demo queries.

10 tests green, no reflection/boxed-math warnings. Branch `jarohen:edgar-dataset` pushed.

PR #5724, opened 2026-06-17
fix(kafka): cancel in-flight delivery on subscription teardown (#5711)

mbutlerw:
The shared Kafka poll loop drives each subscription's records by calling `processRecords` synchronously and joining it; for a leader that call sends into the term's persister over a rendezvous channel. When a database closes, its source-log subscription is cancelled and unregisters under `NonCancellable` — which needs the poll loop to process the command. If the loop is parked on that rendezvous send (the persister stopped draining mid-teardown), it never returns to its `select`, the unregister is never processed, and close hangs forever. The loop lives on the cluster scope, a different job tree, so the closing database's own cancellation can't reach it. Dump-confirmed as the Kafka instance of #5711.

Give each subscription its own delivery job — a child of the poll loop — and run `processRecords` under it. Teardown cancels that job before the NonCancellable unregister, interrupting the parked send so the loop's `join()` returns and the loop is free to process the unregister. The rescue reaches the wedged loop sideways rather than through the command channel, so it works precisely when the loop can't help itself.

Parenting to the loop preserves the existing semantics: cluster shutdown reaps the job, and a genuine delivery error still propagates up and tanks the loop. Cancellation is sticky, so a delivery that races teardown is born cancelled — a no-op — which is why no separate torn-down flag is needed.

This deliberately does not depend on the persister's cleanup closing the channel to free the send: the earlier attempt down that path (`leader-term-channel-cleanup`) left the CI hang identical, because the channel stayed open for reasons still not pinned. Cancelling the delivery sidesteps it.

The test reproduces the deadlock deterministically at the SharedGroupConsumer boundary — a processor that blocks in `processRecords`, then a cancelled subscription — using `runBlocking` so the timeout is real.

Refs #5711

### DIFF
--- a/modules/kafka/src/main/kotlin/xtdb/api/log/KafkaCluster.kt
+++ b/modules/kafka/src/main/kotlin/xtdb/api/log/KafkaCluster.kt
@@ -164,6 +164,9 @@ class KafkaCluster(
             val epoch: Int,
             val listener: Log.SubscriptionListener<M>,
             var processor: Log.RecordProcessor<M>?,
+            // Delivery runs under this job — a child of the poll loop: cluster shutdown reaps it, a
+            // delivery error still tanks the loop, and teardown cancels just this subscription. #5711
+            val deliveryJob: Job,
             val completion: CompletableDeferred<Unit> = CompletableDeferred(),
         ) {
             // expecting a load of change here for multi-part.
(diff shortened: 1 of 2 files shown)


mbutlerw:
Closing for now, think this might be too narrow a fix.

Issue #5703, opened 2026-06-09
LeaderLogProcessor.invokeOnCompletion handler violates kotlinx contract (slow, blocking), allowing parent allocator to close mid-handler


mbutlerw:
`LeaderLogProcessor`'s `init` block registers an `invokeOnCompletion` handler on the term's `SupervisorJob` that performs slow synchronous I/O — `extSource?.close()` (Postgres / Kafka external-source teardown), `replicaProducer.close()` (Kafka transactional producer flush+close), then `allocator.close()` on the leader's child allocator. The kotlinx contract for `CompletionHandler` requires handlers to be **"fast, non-blocking, and thread-safe"** and explicitly permits them to **"be invoked concurrently with the surrounding code."** LP's handler violates that contract.

The symptom is `IllegalStateException: Attempting operation on allocator when allocator is closed` thrown from inside the handler at `LeaderLogProcessor.kt:357`, wrapped as `CompletionHandlerException` and reported via the database scope's `CoroutineExceptionHandler` as `"ingestion stopping"`.

The exception is intermittent in its consequences. Most fires are absorbed by the supervisor's exception handler and just spam the log. Occasionally one wedges the test thread enough that the Kafka consumer-coordinator loses its heartbeat lease and the 15-minute `Integration Test` CI cap kills the job with no visible progress.

## Symptoms

Recent integration-test runs intermittently fail with this pattern:

- Active test progress, leader/follower transitions logging cleanly.
- A multi-minute silence with no test output.
- A `[kafka-coordinator-heartbeat-thread] WARN ConsumerCoordinator: consumer poll timeout has expired` line.
- The 15-minute `Integration Test` GitHub Action cap kills the job.

Underneath, the test process is generating exceptions like:

```
Caused by: kotlinx.coroutines.CompletionHandlerException: Exception in completion handler
  InvokeOnCompletion@... for SupervisorJobImpl{Cancelled}@...
Caused by: java.lang.IllegalStateException: Attempting operation on allocator when allocator is closed.
  Allocator(database/xtdb) ...
  at org.apache.arrow.memory.BaseAllocator.assertOpen
  at org.apache.arrow.memory.BaseAllocator.childClosed
  at org.apache.arrow.memory.BaseAllocator.close
  at xtdb.indexer.LeaderLogProcessor._init_$lambda$1(LeaderLogProcessor.kt:357)
```

[...]

PR #5728, opened 2026-06-24
refactor: two-phase teardown, retiring launchWithCleanup

jarohen:
## Summary

Replaces coroutine-completion-driven resource cleanup (the `launchWithCleanup` helper) with an explicit two-phase teardown across the database, indexer and compactor lifecycle: cancel-and-join the coroutines, then free resources synchronously. Behaviour-preserving on the success path — the motivating symptom is `Database.close()` hanging.

## Context

There's no separate tracking issue, so the context lives here. `Database.close()` could hang. Since the #5662 lifecycle refactor, each database runs under one job tree with a single `job.cancelAndJoin()` in `close()`, and resource cleanup was welded onto coroutine completion via `xtdb.util.launchWithCleanup` — the cleanup ran in the coroutine's `finally` under `NonCancellable`, so the parent's join had to wait on it. That coupling is the fragile part: it depends on `CoroutineStart.ATOMIC` to survive a cancel-before-first-dispatch and on `invokeOnCompletion` ordering subtleties (#5703) to fence the parent, and it makes shutdown hard to reason about.

## The model

Teardown is now two explicit phases:

1. **Stop** — cancel-and-join the job tree. One global cancel at the database level (compactor, source-log subscription and the live term all run under one `job`); a scoped per-term cancel on Kafka rebalance.
2. **Free** — a synchronous `AutoCloseable.close()` on each unit, in allocator-parentage order (children before parents, which Arrow requires).

Because phase 1 has joined every coroutine, phase 2 runs with nothing live touching those resources — the cancel-before-dispatch leak, the `invokeOnCompletion` race, and close-order races all disappear, and `launchWithCleanup` goes with them.

The shape is symmetric with construction: resources a unit owns from instantiation (created there, or handed to it) are freed at its `close()`; resources a coroutine creates *during* the job stay in that job's own `finally`/`use`. The unit's lifetime owns the standing resources, so we stop making the job's end the unit's end. `Database.close()` already had this split at the database level (`job.cancelAndJoin()` then `closeAll`); this pushes it down to the term (`LogProcessor.SubSystem`, modelled as `(job, closeable)` so a job-less or unfreeable term is unrepresentable) and the compactor (`Compactor.ForDatabase`).

## Open-failure unwind

`safelyOpening` unwinds last-registered-first. A single global `job.cancelAndJoin()` closeable registered last would leave a window: a failure *between* resources could free a child allocator while its loop is still live. So each coroutine-owning resource instead registers its own closeable that cancel-joins its scope and then frees it.

## Behaviour change

One deliberate change: a unit that crashes internally with no following rebalance now holds its resources until the next teardown (rebalance or database close) rather than freeing them on the spot. A self-crashed term has already stopped ingestion via `notifyError`, and wiring crash→close would reinstate exactly the completion-driven path this PR removes.

## Out of scope

- Root-causing the specific `Database.close()` hang — this PR is the architectural change that stops cleanup riding on coroutine completion; the hang is the motivation.
- The pre-existing, narrow allocator leak if a processor constructor throws mid-init.

## Test plan

- Full `./gradlew test`: 1554 tests green, no reflection/boxed-math warnings.
- Property/simulation suites (LogProcessor / Node / Compactor): 1,913 iterations green, specifically watched for Arrow allocator leaks and shutdown hangs — none.
- A code-review pass over the diff caught one real bug in the first cut — the open-failure ordering described above — which is fixed in this PR.

`dev/doc/coroutines.adoc` is rewritten to document the two-phase model, keeping the #5703 analysis as the rationale for moving off completion-driven cleanup.


jarohen:
Some context from the design discussion behind this PR, for reviewers.

The starting point was a `Database.close()` hang plus a higher-level question: is it more idiomatic to give each component its own structured-concurrency scope, or to run the whole system under one? The conclusion was that it's a bit of a false choice — scopes nest, so we already have per-component scopes under a per-database root. The friction wasn't the tree shape; it was that a single cancel-cascade gives you no ordering, and no "free this *after* its own join" hook.

`launchWithCleanup` is what bridges that gap today, and it does it correctly: it ties a resource's free to its coroutine's completion so a parent's `cancelAndJoin` waits for it, and it handles the two genuinely tricky cases there — the cancel-before-first-dispatch window (via `ATOMIC` start) and the fact that `invokeOnCompletion` doesn't fence a parent's join (#5703). None of that is wrong; the question this PR pokes at is whether, given a slightly different teardown shape, we still need that machinery at all.

Two-phase teardown is the answer it lands on: once Phase 1 has cancel-joined the whole tree, Phase 2 frees synchronously with nothing live — so the cases `launchWithCleanup` carefully guards against can't arise, and the helper falls away rather than being replaced. The framing that made it click (jarohen's): it's symmetric with construction. Resources a unit owns from instantiation are freed at its `close()`; only resources a coroutine creates *during* the job belong in that job's `finally`/`use`. We'd drifted into making the job's end the unit's end; this puts the standing resources back on the unit's lifetime.

A few things worth a careful look in review:

- The open-failure path was the subtle bit. The first cut used a single global `job.cancelAndJoin()` closeable registered last; a code-review pass caught that a failure *between* resources could then have `safelyOpening` free a child allocator while its loop was still live. It's now per-resource cancel-then-close, which closes that window.
- The one deliberate behaviour change — a self-crashed unit frees lazily at the next teardown rather than immediately — is called out in the description. Re-adding eager free would mean re-introducing the completion-driven path we're removing, so it felt like the right trade, but it's worth a second opinion.
- Verification leaned on the property/sim suites (1,913 iterations) because they're what actually exercise rebalance-driven term transitions and shutdown, where an allocator leak or a hang would show up.

Happy to talk through any of it.

PR #5729, opened 2026-06-24
Docs: config updates

Akeboshiwind:
Makes a bunch of updates to the [configuration](https://docs.xtdb.com/ops/config.html) page, plus related fixes to the Kafka and Azure pages.

**Config page**

- Fix broken `!Env`
- Trim and correct Log & Storage
- Change structure to have sections related to config at the top and other concepts at the bottom
- Moved some config under a new Log & Storage heading
- Split "Other Configuration"
- Added config for: FlightSQL, Node ID, Default TZ, Indexer, Garbage Collector
- Added a "Caching" section (`memoryCache` & `diskCache`)
- Added a "Remotes" section — documents `remotes`, deprecates `logClusters` (+ changelog entry)
- Added more specific links to other pages
- Added a "Troubleshooting Configuration" section with: readOnlyDatabases & skipDbs

**Related pages**

- Kafka: migrate examples from `logClusters` to `remotes` (+ changelog entry; fix missing `cluster:` in the SASL example)
- Azure: document storage-account-key / connection-string auth via an `!AzureBlob` remote

A (GitHub rendered) preview can be seen [here](https://github.com/ake-forks/xtdb/blob/docs-config-updates/docs/src/content/docs/ops/config.md)

### DIFF
--- a/docs/src/content/docs/ops/config.md
+++ b/docs/src/content/docs/ops/config.md
@@ -3,68 +3,141 @@ title: Configuration
 ---
 
 <details>
-<summary>Changelog (last updated v2.1)</summary>
+<summary>Changelog (last updated v2.2)</summary>
+
+v2.2: `remotes` replaces `logClusters`
+
+: Named connections to external systems are now configured under [`remotes`](#remotes).
+
+  Previously these lived under `logClusters`, which was scoped to transaction-log clusters. `remotes` generalises it to any external connection — Kafka clusters, cloud identities, Postgres databases, etc.
+
+  `logClusters` is deprecated but still honoured: entries under both names are merged, so existing config keeps working — rename to `remotes` when convenient.
+
+  The pre-existing `!Kafka` can simply be moved under `remotes` with no other changes.
 
 v2.1: multi-database support
 
-: The log and storage configurations were changed as part of 2.1's multi-db support. 
+: The log and storage configurations were changed as part of 2.1's multi-db support.
 
   For more details on those changes, see the [Transaction Logs](config/log) and [Object Storage](config/storage) documentation.
-    
+
 </details>
 
 XTDB nodes are configured using YAML files.
 
-The two main pluggable components of XTDB are transaction logs and object storage - these can be configured in the `databases` section of the configuration file.
+All config options have default values, it is therefore valid to not specify a config file or not specify any part of the top-level config.
+
+## Log & Storage
+
+The two main pluggable components of XTDB are [transaction logs](config/log) and [object storage](config/storage).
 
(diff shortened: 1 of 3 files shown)


docs/src/content/docs/ops/config.md
```diff
@@ -3,68 +3,142 @@ title: Configuration
...
+
+# ...then referenced by alias, e.g. by a Kafka transaction log:
+log: !Kafka
+  cluster: my-kafka
+  topic: xtdb_topic
+```
+
+For the fields each remote type takes, and how a given component references its alias, see that component's documentation — e.g. [transaction log](config/log), [object storage](config/storage).
+
+`logClusters` is the former name for this option and is deprecated — use `remotes`.
```
jarohen: good point, but shift it to the changelog - keep the main body the current state

docs/src/content/docs/ops/config.md
```diff
@@ -89,16 +163,128 @@ server:
   # Default is -1, to not start a read-only server.
   # Set to 0 to have the server choose an available port.
   readOnlyPort: -1
+```
 
+### Flight SQL Server
```
jarohen, line 167: debatably should live with the ADBC configuration. fine here for now unless you strongly agree 🙂

docs/src/content/docs/ops/config.md
```diff
@@ -89,16 +163,128 @@ server:
...
+### Garbage Collector
+
+Reclaims object-store space by deleting files left behind once compaction has superseded them.
+
+Disabled by default.
+
+```yaml
+garbageCollector:
+  # Set to true to enable garbage collection
+  enabled: false # default
```
jarohen, line 250: default true now, IIRC? double check

jarohen (approved):
thanks @Akeboshiwind 🙏

docs/src/content/docs/ops/config.md
Akeboshiwind: `logClusters` is still valid. Do we want deprecated options out of the text?

docs/src/content/docs/ops/config.md
Akeboshiwind: Apparently not: https://github.com/xtdb/xtdb/blob/6e6c002f39695f6fb988e09dfa12600f96bd476a/core/src/main/kotlin/xtdb/api/GarbageCollectorConfig.kt#L14

docs/src/content/docs/ops/config.md
Akeboshiwind: Chatted out of band. Answer: yes

PR #5736, opened 2026-06-25
refactor(adbc): consolidate the connection transaction onto a coalescing DML buffer

jarohen:
## Summary

The unified `Xtdb.Connection` is becoming the SQL front door for every frontend (pgwire, ADBC, Flight SQL): session state, prepare, and query execution already route through it. This moves the connection's *transaction* there too, starting with ADBC, and reshapes how buffered DML is held so that coalescing — previously a pgwire-only trick — becomes a property of the connection that all frontends share.

This is the first of two increments: it wires ADBC onto the new model; a follow-up ports pgwire onto the same buffer (see *Out of scope*).

## Context

Per-frontend transaction machinery is the largest remaining duplication between pgwire and ADBC. pgwire keeps tx-state in a Clojure map and coalesces consecutive same-SQL DML into one multi-row op; ADBC buffered manual-mode DML as a flat list of eagerly-opened `TxOp`s — one per `executeUpdate`. So a prepared `INSERT` run in a loop produced N separate ops on the ADBC path, each holding its own Arrow buffers and planned independently at commit, where pgwire would have folded them into one.

## Approach

The connection now owns the DML buffer. It keeps its ops in a list and coalesces in place: a DML op whose SQL matches the last buffered op appends its argument rows into that op's own relation, so a prepared `INSERT` run in a loop becomes one multi-row `TxOp.Sql` rather than N ops. Other ops, and SQL with no arguments, stay standalone and ordered. The op carries a growing relation from its first row, so there is no separate seal step or deferred/eager conversion — at commit the list is already the submittable op list.

`maybePromote` on the relation writer absorbs arguments whose types differ across executes, so the buffer needs no up-front schema — it widens as rows arrive, the same way pgwire's commit-time `sql->static-ops` merges types across rows.

The transaction is modelled as a sealed `AccessMode`: a read-only tx carries the read basis, a read-write tx carries the buffer and its commit metadata. In XTDB a "read-write" transaction is effectively write-only — queries are rejected inside a DML tx and DML inside a read-only one — so a read basis on a write tx is now unrepresentable rather than merely unused.

## Behaviour

This is behaviour-preserving for results: a coalesced multi-row op applies each row exactly as N single-row ops would, which is the semantics pgwire has always relied on. The one deliberate change is JDBC correctness — `commit`/`rollback` now key off the sticky autocommit flag and throw when autocommit is on, rather than silently flushing.

Arrow lifecycle is single-owner throughout: each incoming argument batch is appended into the op's relation and the incoming reader closed; `drain` hands the ops to the committer and empties the buffer so a later close cannot double-free. A throw mid-append (e.g. OOM opening a relation) closes the arguments and leaves the buffer consistent.

## Out of scope

pgwire keeps its own buffer for now; porting it onto this connection-owned one — feeding rows a row at a time — is the next increment, and will populate the `systemTime`/`userMetadata`/`async` and `failed` fields the model carries for that migration. ADBC/Flight SQL begin-time snapshot isolation, and gating session-state setters mid-transaction, are separate later increments.

## Test plan

`InProcessAdbcTest` gains real-node cases: coalescing (a prepared insert looped five times lands every row), seal-on-SQL-change (a run, an interleaved statement, then a fresh run all commit in order), and rollback discarding a buffered run without leaking. The full ADBC / Connection / Flight SQL suites and the pgwire suite stay green — pgwire is unaffected, as it submits through `executeTx`/`submitTx`, not `executeDml` — with no reflection or boxed-math warnings.

### DIFF
--- a/core/src/main/kotlin/xtdb/api/Xtdb.kt
+++ b/core/src/main/kotlin/xtdb/api/Xtdb.kt
@@ -36,6 +36,7 @@ import xtdb.api.module.XtdbModule
 import xtdb.api.storage.Storage
 import xtdb.arrow.Relation
 import xtdb.arrow.RelationReader
+import xtdb.arrow.RelationWriter
 import xtdb.arrow.VectorType
 import xtdb.arrow.VectorType.Companion.ofType
 import xtdb.arrow.unsupported
@@ -138,7 +139,7 @@ interface Xtdb : DataSource, AdbcDatabase, AutoCloseable {
         private val awaitTimeout: Duration = Duration.ofMinutes(1)
 
         private var autoCommit = true
-        private val pendingOps = mutableListOf<TxOp>()
+        private var tx: Transaction? = null
 
         private fun db(dbName: DatabaseName): Database =
             dbCat.databaseOrNull(dbName)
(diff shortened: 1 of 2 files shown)


jarohen:
### Chalk — model correction: removed AccessMode dead speculation

<details><summary>Why the amend</summary>

Building #5737 (pgwire onto the connection) on top of this surfaced a shape problem in the transaction model here, and tightening this PR was cleaner than working around it downstream.

As shipped, `AccessMode.ReadOnly` carried a `Basis` (the pinned read snapshot), and the commit message claimed this made "a read basis on a write tx unrepresentable". But in this PR's world — ADBC only — `ReadOnly` is never constructed: `executeDml` only ever resolves the mode to `ReadWrite`, and the `openQuery` basis-merge that would *read* a pinned basis doesn't land until #5737. So the `Basis` class and `ReadOnly`'s field were speculation for a caller this PR doesn't ship.

Removed both — `ReadOnly` is now a bare `data object`, `Basis` is gone. The sealed read-only/read-write model and the "can't write in a read-only tx" guard stay; those are the domain model, not speculation. All 14 InProcessAdbcTest cases pass — it's a pure deletion. Commit body updated to match (amended, force-pushed).

</details>

<details><summary>The model we settled on (for #5737)</summary>

When the snapshot/commit state does arrive in #5737, placement follows *when each fact is known*:

- the read snapshot (current-time, snapshot-token, snapshot-time) is pinned at BEGIN, before the mode is decided, so it lives on `Transaction`; a read-write tx ignores it.
- the commit-opts (system-time, user-metadata, async) are only set once write intent is declared, so they stay on `ReadWrite` — exactly where this PR already had them.
- the mode is late-bound: Undecided → ReadOnly (a query ran) or ReadWrite (a DML ran), resolved and locked on the first statement; once read-only, a later DML is rejected rather than promoted.

</details>

Issue #4501, opened 2025-05-29
Add ability to COPY Arrow data directly into XT


jarohen:
see #3856, #4500 

After this card, we should be able to move Arrow data all the way from client, through pgwire, directly onto the tx-log without any serde - i.e. just transferring bytes.

This will likely require refactoring to the xtdb.log namespace, to allow `:put-docs` to contain Arrow (or maybe a new operation would be worthwhile)

Issue #4415, opened 2025-05-07
Too easy to get keys wrong for INSERT...RECORDS (Clojure + pg2 + transit)


egg-juxt:
When doing `INSERT INTO my_table RECORDS $1` with a transit parameter, it works only when keys are all strings, and id must be "_id". In any other case, the INSERT doesn't work correctly. In some cases, the insertion dangerously succeeds, leaving the table not queryable:

```
(defn insert-and-query [doc]
  (with-open [db (pg/connect {:host "localhost"
                              :port 5432
                              :user "xtdb"
                              :database "xtdb"})]
    (pg/execute db
      "INSERT INTO my_table RECORDS $1"
      {:params [(-> doc (serde/write-transit :json) (String.))]
       :oids [(int 16384)]})

    (pg/execute db "SELECT * FROM my_table")))

(comment
  (insert-and-query {"_id" 1, "value" 1}) ; => [{:xt/id 1, :value 1}]
  (insert-and-query {"_id" 1, :value 1})  ; => Server error response: {severity=ERROR, code=08P01, message=RECORDS ? not supported outside of DML, verbosity=ERROR}, sql: INSERT INTO my_table RECORDS $1
  (insert-and-query {:xt/id 1, "value" 1}) ; => Server error response: {severity=ERROR, code=08P01, message=RECORDS ? not supported outside of DML, verbosity=ERROR}, sql: INSERT INTO my_table RECORDS $1
  (insert-and-query {"xt/id" 1, "value" 1}) ; (Insert succeeds!!!) => Assert failed: Field my_table.1/xt/id not found in relation, available (my_table.1/_id my_table.1/value my_table.1/id)
  (insert-and-query {:xt/id 1, :value 1}) ; (Insert succeeds!!!) => Assert failed: Field my_table.1/xt/id not found in relation, available (my_table.1/value my_table.1/id)
  ,)
```

jarohen:
@egg-juxt could you re-check this one in light of the #4403 fix?

egg-juxt:
The 2 last cases still fail, same error.

tggreene:
Related to https://github.com/xtdb/xtdb/issues/4469

tggreene:
I attempted to apply a change that allows the keyword :xt/id in transit encoded insert operations (in particular put docs) but found that it introduced stricter validation around valid to/from data types (requiring all values to be UTC) so I backed it out with the reasoning that the overall change still improves things.

[...]

PR #5737, opened 2026-06-26
refactor: move the connection transaction off pgwire onto Xtdb.Connection

jarohen:
Part of the SHOW-into-engine arc. Builds on #5736 — review that one first. Until it merges this PR's diff shows both changes; it collapses to just the pgwire migration once #5736 lands on `main`.

**Draft / work in progress.** The connection-side API has landed (commit 1); the pgwire migration that consumes it (commits 2–3) is in progress. Raised early for visibility into the shape.

## Summary

#5736 gave `Xtdb.Connection` a coalescing DML buffer and wired ADBC onto it. This retires pgwire's *own* transaction machinery so the connection becomes the single home for transaction state, leaving `pgwire.clj` with pg-protocol concerns only.

## Context

pgwire keeps everything about a transaction in its `:transaction` conn-state map: the DML buffer (`:dml-buf` plus a coalescing swap-fn), the read basis (`:current-time`/`:snapshot-token`/`:snapshot-time`, pinned at `BEGIN` and consumed when binding a query), the commit options (`:system-time`/`:user-metadata`/`:async?`), and protocol flags (`:implicit?`, `:failed`, `:access-mode`). ADBC, since #5736, keeps the same notion on the connection. That's two competing homes for one concept, and the duplicated commit path is where they drift — pgwire assembles `TxOpts` and chooses sync-vs-async submission itself, reimplementing what the connection now does.

So this moves the whole notion at once rather than half-migrating into a state with two tx homes.

## Approach

The connection owns tx *content* — existence, access-mode, the pinned read basis, the DML buffer, commit metadata, and the failed flag. pgwire owns protocol *lifecycle* only — the parse/bind/execute/sync loop, implicit-transaction (auto-begin) tracking, auto-commit at `Sync`, `ReadyForQuery` status, the pg-specific permissibility and aborted-tx errors, command tags, and COPY framing. pgwire *drives* the connection's transaction and *queries* it for status to shape its protocol responses; the connection never learns what an implicit transaction is.

The read basis is the semantic crux. The connection's `openQuery` becomes transaction-aware: when a read-only tx is active it supplies that tx's pinned basis, so pgwire stops threading basis through its query options. The merge rule is "active read-only tx's basis wins, else the caller's per-statement options, else the connection default". This is behaviour-preserving: pgwire's per-statement basis overrides are vestigial (always nil today), so "tx wins" matches current behaviour, and the clock-split is respected — pgwire still supplies current-time from its session clock for tx-less reads and at `BEGIN`. ADBC opens no read-only tx, so `openQuery` sees no active tx and behaves exactly as before.

## Out of scope

A genuine `SETTING SNAPSHOT_TOKEN` per-statement override channel — noted, not built; the merge rule leaves room for it. Begin-time snapshot isolation for ADBC/Flight SQL, and gating session-state setters mid-transaction (the original motivation for all this), are separate later increments.

## Test plan

The full pgwire suite (`xtdb.pgwire-test`, `xtdb.pgwire.*`) plus `xtdb.sql-test` are the behaviour-preservation oracle, exercising both protocols, PREPARE/EXECUTE, COPY, BEGIN/COMMIT/ROLLBACK, implicit transactions, aborted-tx rejection, and snapshot/basis reads. The ADBC/Connection/Flight SQL suites must stay green since the shared `openQuery`/buffer changed. No reflection or boxed-math warnings.

### DIFF
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -324,11 +324,11 @@
   If the status is omitted, the status is determined from whether a transaction is currently open."
   ([conn]
    ;; it would be good to look at ready status being automatically driven by pure conn state, this is a bit messy.
-   (if-some [transaction (:transaction @(:conn-state conn))]
-     (if (:failed transaction)
-       (cmd-send-ready conn :failed-transaction)
-       (cmd-send-ready conn :transaction))
-     (cmd-send-ready conn :idle)))
+   (let [cs @(:conn-state conn)]
+     (cond
+       (.getTxFailed ^Xtdb$Connection (:node-conn cs)) (cmd-send-ready conn :failed-transaction)
+       (:transaction cs) (cmd-send-ready conn :transaction)
+       :else (cmd-send-ready conn :idle))))
   ([conn status]
    (pgio/cmd-write-msg conn pgio/msg-ready {:status status})))
 
@@ -599,8 +599,8 @@
 
 (defn cmd-commit [{:keys [conn-state tx-error-counter tx-latency-timer tx-submit-timer tx-execute-timer] :as conn}]
   (let [{:keys [transaction ^Xtdb$Connection node-conn]} @conn-state
-        {:keys [failed async?]} transaction]
-    (if failed
+        {:keys [async?]} transaction]
+    (if (.getTxFailed node-conn)
       (throw (pgio/err-protocol-violation "transaction failed"))
 
       (try
@@ -714,10 +714,10 @@
 (defmethod handle-msg* :msg-sync [{:keys [conn-state] :as conn} _]
   ;; Sync commands are sent by the client to commit transactions
   ;; and to clear the error state of a :extended mode series of commands (e.g the parse/bind/execute dance)
-  (let [{:keys [implicit? failed]} (:transaction @conn-state)]
+  (let [{:keys [implicit?]} (:transaction @conn-state)]
     (try
       (when implicit?
-        (if failed
+        (if (.getTxFailed ^Xtdb$Connection (:node-conn @conn-state))
           (cmd-rollback conn)
           (cmd-commit conn)))
       (catch Throwable t
@@ -1151,7 +1151,7 @@
   (pgio/cmd-write-msg conn pgio/msg-command-complete {:command "SET SESSION CHARACTERISTICS"}))
 
 (defn- cmd-exec-dml [{:keys [node conn-state tx-error-counter ^BufferAllocator allocator] :as conn} {:keys [^ParsedStatement$Dml parsed args param-oids]}]
-  (when (get-in @conn-state [:transaction :failed])
+  (when (.getTxFailed ^Xtdb$Connection (:node-conn @conn-state))
     (throw (pgio/err-protocol-violation "current transaction is aborted, commands ignored until ROLLBACK is received")))
 
   (let [query (.getOriginalSql parsed)]
@@ -1227,7 +1227,7 @@
                        :as _portal}]
   ;; Create an implicit transaction if one hasn't already been started
   (let [transaction (get-in @conn-state [:transaction])]
-    (when (:failed transaction)
+    (when (.getTxFailed ^Xtdb$Connection (:node-conn @conn-state))
       (throw (pgio/err-protocol-violation "current transaction is aborted, commands ignored until ROLLBACK is received")))
 
     (when-not (or transaction (instance? ParsedStatement$ShowVariable parsed))
@@ -1490,13 +1490,13 @@
           (when-let [{:keys [implicit?]} (:transaction @conn-state)]
             (if implicit?
               (cmd-rollback conn)
-              (swap! conn-state util/maybe-update :transaction assoc :failed true, :err e)))
+              (.failTx ^Xtdb$Connection (:node-conn @conn-state) e)))
 
           (send-ex conn e))))
 
-    (let [{:keys [implicit? failed]} (:transaction @conn-state)]
+    (let [{:keys [implicit?]} (:transaction @conn-state)]
       (when implicit?
-        (if failed
+        (if (.getTxFailed ^Xtdb$Connection (:node-conn @conn-state))
           (cmd-rollback conn)
           (cmd-commit conn))))
 
@@ -1644,14 +1644,14 @@
 
     (catch Throwable e
       (log/debug e "error processing message: " (ex-message e))
+      ;; error seen while in :extended mode, start skipping messages until sync received
       (swap! conn-state
-             (fn [{:keys [transaction protocol] :as cs}]
+             (fn [{:keys [protocol] :as cs}]
                (cond-> cs
-                 ;; error seen while in :extended mode, start skipping messages until sync received
-                 (= :extended protocol) (assoc :skip-until-sync? true)
+                 (= :extended protocol) (assoc :skip-until-sync? true))))
 
-                 ;; mark a transaction (if open as failed), for now we will consider all errors to do this
-                 transaction (update :transaction assoc :failed true, :err e))))
+      ;; mark any open tx as failed (no-op if none) — all errors abort the tx for now
+      (.failTx ^Xtdb$Connection (:node-conn @conn-state) e)
 
       (send-ex conn e))))
 
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -578,14 +578,15 @@
                                            (->> (into {} (filter (comp some? val))))))
                                  (assoc :await-token await-token)))))))))
 
-  ;; mirror the write-side opts onto the connection's tx, so executeDml buffers and commit submits with them.
-  ;; the read basis (current-time/snapshot/await-token) still flows via the :transaction map + bind-stmt — that's C′.
+  ;; mirror the tx onto the connection: it pins the read basis (a read-only tx's queries read it via openQuery) and
+  ;; the commit metadata (a read-write tx's submit uses it).
   (let [{:keys [^Xtdb$Connection node-conn session]
-         {:keys [access-mode default-tz system-time user-metadata async?]} :transaction} @conn-state]
+         {:keys [access-mode current-time snapshot-token snapshot-time await-token
+                 default-tz system-time user-metadata async?]} :transaction} @conn-state]
     ;; pgwire's swap above already decided whether a (re-)begin is allowed (it throws on a double explicit BEGIN).
     ;; an explicit BEGIN is preceded by an empty implicit auto-begin (msg-simple-query) — discard it and re-begin.
     (.rollback node-conn)
-    (.beginTx node-conn nil nil nil nil
+    (.beginTx node-conn current-time snapshot-token (some-> snapshot-time time/->instant) await-token
               default-tz system-time (get-in session [:parameters "user"]) user-metadata
               (boolean async?)
               (case access-mode :read-only true :read-write false nil))))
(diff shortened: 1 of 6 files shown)


jarohen:
### Chalk — consolidate the connection transaction off pgwire

Tracking the tx-state-consolidation arc. This PR is the pgwire half; #5736 is the ADBC half it builds on.

- [x] PR1 (#5736) — connection-owned coalescing DML buffer, ADBC wired on
- [x] PR2-A — connection transaction API (this branch, commit 1)
- [ ] PR2-B — pgwire writes onto `executeDml`, drop `:dml-buf`
- [ ] PR2-C — pgwire lifecycle + reads onto the connection
- [ ] PR2-D — verify (full pgwire+sql+ADBC), code-review, curate commits, un-draft

<details><summary>The arc and its end state</summary>

`Xtdb.Connection` is becoming the single SQL front door for every frontend — pgwire, ADBC, Flight SQL. Prepare, query-open, tx-submit, and session-state get/set already route through it. The largest remaining duplication is the *transaction*: pgwire keeps tx-state in a Clojure `:transaction` conn-state map and coalesces consecutive same-SQL DML into one multi-row op; ADBC kept its own flat list of eagerly-opened ops.

End state: `pgwire.clj` carries pg-protocol concerns only (the parse/bind/execute/sync loop, implicit-tx tracking, auto-commit at `Sync`, `ReadyForQuery`, command tags, COPY framing); the connection owns all tx *content* (existence, access-mode, pinned read basis, the DML buffer, commit metadata, failed-flag). pgwire drives the connection's tx and queries it for status.

This is a sub-arc of the SHOW-into-engine work — the same "move it off the frontend onto the connection" pattern.

</details>

<details><summary>PR1 (#5736) — what it established</summary>

The connection gained a coalescing DML buffer and ADBC was wired onto it. A DML op whose SQL matches the last buffered op appends its argument rows into that op's own growing relation, so a prepared INSERT looped N times becomes one multi-row `TxOp.Sql` rather than N independently-planned ops. `maybePromote` on the relation writer absorbs cross-execute type divergence, so the buffer needs no up-front schema — it widens as rows arrive, mirroring commit-time `sql->static-ops` type-merging.

The buffer holds eager ops and coalesces in place via `lastOrNull` (no separate seal step — the op carries its relation from the first row, so at commit the list is already submittable). Arrow lifecycle is single-owner: each incoming batch is appended and its reader closed; `drain` empties the buffer so a later close can't double-free.

`commit`/`rollback` key off the sticky autocommit flag (pgjdbc semantics) and throw when autocommit is on.

</details>

<details><summary>PR2-A — connection transaction API (commit 1 on this branch)</summary>

Added the API pgwire will consume: `beginTx`, a basis-applying `openQuery`, `commitTx` that builds `TxOpts` and chooses async-submit vs sync-execute, and status accessors (`txOpen`/`txFailed`/`txReadOnly`/`failTx`).

The model reshaped from PR1. pgwire pins the read basis at `BEGIN`, *before* the first statement reveals read-vs-write, so the basis can't live on the access-mode subtype the way PR1 had it. Basis + commit-metadata now pin onto a begin-time `Transaction`; `AccessMode` reduces to a `ReadOnly` marker plus `ReadWrite(buffer)`. The first statement resolves an undecided tx — a query makes it read-only, DML read-write. (A "read-write" tx is write-only in XTDB: queries are rejected inside a DML tx, so no basis is ever read on the write path.)

ADBC reads now route through `openQuery`, but it opens no read-only tx, so the basis merge is a no-op for it — behaviour unchanged. Verified: 66 ADBC/Connection/Flight SQL tests green, no reflection or boxed-math warnings.

</details>

<details><summary>Read basis — the semantic crux (decided: connection applies it)</summary>

`openQuery` is transaction-aware: an active read-only tx supplies its pinned basis, so pgwire stops threading basis through its query options. Merge rule: active read-only tx's basis wins, else the caller's per-statement options, else the connection default.

This is behaviour-preserving. pgwire's per-statement basis overrides are vestigial (always nil today), so "tx wins" matches current behaviour. The clock-split is respected — pgwire still supplies current-time from its session clock for tx-less reads and at `BEGIN`; the connection only carries the timezone, not the now-source.

Chosen over the alternative (pgwire keeps assembling basis and passes it explicitly). The connection-applies approach is the purer, larger, more semantic surface — the right end-state for the front-door model, at the cost of a wider diff that wants close review.

</details>

<details><summary>What's left — PR2-B/C/D</summary>

B: `cmd-exec-dml` and the COPY paths build `TxOp.Sql(sql, args-rel)` (row args materialised via `vw/open-args`) and call `executeDml`; delete `:dml-buf` and its coalescing swap-fn — the connection coalesces now.

C: `cmd-begin`/`cmd-commit`/`cmd-rollback` drive the connection's lifecycle (pgwire passes current-time from its session clock, snapshot-token via `xtp/snapshot-token`, plus user/await-token/tz/commit-opts); `bind-stmt` drops basis assembly and lets the connection apply it; `verify-permissibility` reads `txReadOnly`; `cmd-send-ready`/`msg-sync` read `txOpen`/`txFailed`; the error handler calls `failTx`. `implicit?` and auto-commit-at-`Sync` stay pgwire-side — the connection never learns what an implicit tx is. The `:transaction` map collapses to roughly just `implicit?`.

D: full pgwire + sql + ADBC suites as the behaviour oracle, no warnings; code-review (basis merge + lifecycle boundary are the focus); curate/squash the wip commits; then un-draft.

Deferred past this arc: a genuine `SETTING SNAPSHOT_TOKEN` override channel, ADBC/Flight SQL begin-time snapshot isolation, and mid-tx setter gating (the original motivation) — each its own later increment.

</details>

jarohen:
### Chalk — PR2-B decision: drop `sql->static-ops`, unify on the relation buffer

Investigating PR2-B (route pgwire's writes through the connection's `executeDml`) surfaced a decision worth recording.

pgwire's commit currently runs `sql->static-ops` (`sql.clj:3567`), which statically expands **INSERT/PATCH only** into `PutDocs`/`PatchDocs` at submit time; everything else already falls through to a raw `TxOp.Sql` planned at apply time. ADBC has never used this path. It's the optimization's only caller, added for #4501 (`e2fb8cf5c`).

**Decision: drop it from pgwire** and submit raw `TxOp.Sql` for all DML, like ADBC. This is behaviour-preserving for results, and error-timing is unchanged (`sql->static-ops` returns nil on any error and the caller already falls back to the raw op, so it never surfaces errors earlier). The only effect is that INSERT/PATCH planning shifts from the submit thread to the indexer's apply thread.

It can't simply move onto the connection's buffer as-is: `sql->static-ops` is row-major (`->param-row` per arg-row, type-merge across rows), while the connection's coalescing buffer is a column-major `Relation`. Feeding a coalesced relation back into the row-based expander would mean a relation→rows round-trip that defeats the buffer.

**Deferred as a standalone follow-up**: re-home the optimization *inside* the connection's commit (a relation-aware static-ops variant), so both frontends benefit — ADBC would gain it for the first time. Validate the perf delta first; it may prove negligible and the unified raw-Sql path stands. Brief written up for that session.

jarohen:
### Chalk — PR2-B implemented (writes + lifecycle + static-ops), 5 lifecycle-edge tests remain

Combined PR2-B and PR2-D into one change (they couldn't separate — see below), implemented on `inline-adbc-pgwire`. **Uncommitted** (5 tests still red).

**Discovery that forced the merge:** dropping `sql->static-ops` from pgwire is NOT behaviour-preserving. The PutDocs static-expansion path emits user-facing error messages tests rely on ("missing '_id'", "Cannot put documents with columns: [...]") and correctly coerces UUID/varbinary literal IDs — the raw-SQL apply-time path differs ("INSERT does not contain mandatory _id column"; "Invalid ID type: [B"). So the connection's commit must apply static-ops in the *same* commit pgwire stops calling it.

**What landed:**

- pgwire tx lifecycle → connection: `cmd-begin`→`beginTx`, `cmd-exec-dml`→`executeDml(TxOp.Sql)`, `cmd-commit`→`commitTx`, `cmd-rollback`/`close-transaction`→`rollback`, COPY→`executeDml`. Connection opened in manual mode. `:dml-buf` and pgwire's `sql->static-ops` call deleted.
- `Xtdb.Connection`: `executeDml`/`commitTx` made public; `commitTx` returns the SubmittedTx/ExecutedTx for the frontend to report.
- Static-ops re-homed on the connection via a new `IQuerySource.sqlToStaticOps` (row-major bridge over the existing expander; column-major is a noted follow-up optimisation). `commitTx` expands each buffered `TxOp.Sql`, with disjoint close-sets for originals vs opened expansions.

**Two fixes found via the test oracle (32→19→16→5):**

- Explicit `BEGIN` is preceded by an empty implicit auto-begin in `msg-simple-query`, so `beginTx` threw "already started" — discard the empty implicit tx before re-beginning (mirrors the original's silent upgrade).
- `toTuples` must use `SNAKE_CASE_STRING` keys, not the default kebab keyword — record/map args need string `"_id"` keys for the PutDocs/PatchDocs writers. This took sql-test to fully green.

**State:** sql-test 100% green; pgwire 143/148; ADBC/Connection/Flight SQL green; no reflection/boxed-math warnings.

**5 remaining (all pgwire-test, NOT static-ops — transaction lifecycle edges):**

- `test-show-latest-submitted-tx` — await-token value/encoding differs (connection's await-token, advanced internally by `executeTx`'s `recordTx`, vs the test's expected `basis/->tx-basis-str`).
- `better-assert-error-code-and-message-3878`, `runtime-error-commit-test`, `test-assert-3445` — commit-time error surfacing through the new `commitTx`→`ExecutedTx.getError` path.
- `test-combines-dml` — DML coalescing expectation.

Next session picks these up (hypotheses + the full recipe are in the working note). Not committing until green.

jarohen:
### Chalk — pgwire transaction onto Xtdb.Connection

- [x] Migrate lifecycle + DML + reads/basis onto the connection
- [x] Move `failed` (aborted-tx state) onto the connection
- [x] BEGIN auto-begin simplification (uncommitted on branch)
- [ ] Correct the AccessMode model on rebase (commit-opts back to `ReadWrite`)
- [ ] Fix the pg2 transit-id regression
- [ ] access-mode migration (now unblocked)
- [ ] Verify (full pgwire + sql + ADBC), code-review, un-draft

<details><summary>Scope, pared back to the migration</summary>

The PR had sprawled, and the cause was two threads riding along that aren't the migration itself: a "move as much session state onto the connection as possible" sweep (`failed` → access-mode → `tz` → `implicit?`), and re-homing the static-ops expansion onto the connection.

Re-scoped to just *pgwire's transaction onto the connection* — lifecycle (begin/commit/rollback), writes (`executeDml`), reads (the pinned basis via `openQuery`), and `failed`. `tz` spins out to its own task: it's net-new transactional behaviour, not a move (a rollback must discard a mid-tx `SET TIME ZONE`, and a read-write tx that has already buffered ops must reject `SET`, because only one tz is submitted to the log). `implicit?` stays pgwire-side — it's protocol bookkeeping (auto-begin / auto-commit-at-Sync) the connection needn't know about.

</details>

<details><summary>AccessMode model — resolved, and this PR must partly revert onto PR1 (#5736)</summary>

Working through where each piece of transaction state lives settled the model — and showed PR1 had it half-right and this PR over-corrected.

The mode is the *late-bound* thing. A fresh BEGIN is Undecided; the first statement resolves and locks it — a query makes it read-only, a DML makes it read-write — and once it's read-only a later DML is rejected, not promoted (running a query commits you to reads). So three states: Undecided / ReadOnly / ReadWrite.

Placement follows *when each fact is known*:

- the read snapshot (current-time, snapshot-token, snapshot-time) is pinned at BEGIN, before the mode is decided, so it lives on `Transaction`; a read-write tx ignores it, which matches XTDB's write-only-tx semantics.
- the commit-opts (system-time, user-metadata, async) are only ever set once write intent is declared — which *is* the decision — so they belong on `ReadWrite`.

PR1 had the commit-opts on `ReadWrite` (right) but hung the snapshot off `ReadOnly` (wrong — it's pinned before the mode). This PR moved the snapshot onto `Transaction` (right) but dragged the commit-opts along with it (wrong — that makes "a read-only tx carrying a system-time" representable). On rebase onto the corrected PR1, move system-time / user-metadata / async back onto `ReadWrite`.

</details>

<details><summary>BEGIN auto-begin simplification</summary>

pgwire's `msg-simple-query` auto-begins an implicit tx for *every* statement before it knows what the statement is — so an explicit `BEGIN` was always preceded by a throwaway implicit tx. That forced two pieces of machinery: a `(.rollback node-conn)` in `cmd-begin` to discard the implicit connection-tx before re-beginning, and a double-begin guard that keyed off access-mode *presence* (the implicit tx had none, so it was upgradeable; a tx that had run a statement had one, so a second BEGIN threw).

Excluding `ParsedStatement$Begin` from the blanket auto-begin collapses both: no implicit tx in front of an explicit BEGIN, so the rollback is gone, and the guard simplifies to "a tx already exists → throw". One behaviour change to bless: a bare `BEGIN; BEGIN` now throws "transaction already started" where it previously silently re-began — closer to Postgres, which warns rather than silently re-pinning the basis.

This is the enabler for the access-mode migration: the guard no longer reads access-mode, so access-mode is free to move onto the connection. Currently uncommitted on the branch; the pgwire + sql suites are green on it (the one failure is the unrelated pg2 regression below).

</details>

<details><summary>pg2 transit-id regression (blocker)</summary>

`xtdb.pgwire.pg2_test/test-transit-id-formats` fails: a transit `INSERT INTO t RECORDS $1` of `{:xt/id 1, …}` should be rejected ("missing-id" — in SQL only `_id` is the id, `:xt/id` is XT1-legacy, tracked for #4415), but cases C and E stopped throwing.

Root cause is the static-ops bridge introduced here. On main, pgwire fed row-major arg-rows straight to `sql->static-ops`, and the resulting doc's id key was not keyword `:xt/id`, so `log.clj`'s PutDocs validator rejected it. PR2-B changed pgwire to feed writes through an Arrow `RelationReader`; `sqlToStaticOps` then *reconstructs* row-major via `getObject(…, SNAKE_CASE_STRING)`, and that round-trip leaves the doc with a keyword `:xt/id` key, which the validator's legacy clause (`(or (:xt/id doc) (get doc "_id"))`) accepts. `xtify-args` (the decode) is identical on both branches, so the Arrow round-trip is the whole difference.

Decision pending: a narrow fix (make the bridge reject `:xt/id` like the raw path) versus the proper column-major expander that operates on the `RelationReader` directly — no row-major reconstruction, matching the raw path by construction. The column-major version was the original intent of this work; leaning towards it, since the regression is evidence the row-major bridge is the wrong shape.

</details>

<details><summary>State-migration status</summary>

`failed` moved onto the connection cleanly — one accessor (`getTxFailed`), one mutator (`failTx`), eight touchpoints once `msg-sync`'s destructure was found. access-mode is materially more entangled (a session-default-vs-tx dimension via `SET SESSION CHARACTERISTICS`, the double-begin guard, the await-skip, a metrics counter) but the BEGIN simplification removes its hardest coupling, so it's now tractable. `tz` and `implicit?` are out of scope as above.

</details>

PR #5750, opened 2026-06-29
feat(adbc): transaction & session control on the connection (ADBC parity)

jarohen:
Part of #5132.

## Summary

Brings `Xtdb.Connection` — the connection handle shared by pgwire and ADBC/Flight SQL — to parity on the SQL transaction and session model, exercised through ADBC. Transaction control, access modes, commit options, read-only snapshot isolation, the `SET`/`SHOW` session surface (including `SHOW LATEST_SUBMITTED_TX` and `BEGIN … WITH (TIMEZONE)`), query tracing, and eager expansion of DML to static ops now all live on the connection rather than in pgwire. Reads open through a single gated `openQuery`, so a frontend that drives its own cursor no longer reconstructs the basis/tz/tracer nor re-implements the access-mode gate.

## Context / motivation

The transaction and session logic lived in pgwire (Clojure), so ADBC and Flight SQL clients couldn't open transactions, pin a read snapshot, or issue `SET`/`SHOW` — they could only auto-commit single statements, and even those skipped pgwire's commit-time DML expansion (so a literal UUID `_id` failed where pgwire accepted it).

The broader goal (#5132) is to consolidate *all* transaction state onto the connection, so pgwire becomes pg-protocol-only and both frontends share one implementation. This PR does the ADBC-parity-first half: build the connection's model out to parity with ADBC driving it — which forces it complete and correct — so the eventual pgwire migration is a thin delegation rather than a translation layer. It also unblocks transaction support for ADBC/Flight SQL in its own right.

## Usage

ADBC / Flight SQL clients can now drive the full transaction and session surface as SQL (via `executeUpdate` / `executeQuery`), alongside the programmatic `setAutoCommit` / `commit` / `rollback`:

\`\`\`sql
BEGIN READ WRITE WITH (SYSTEM_TIME = TIMESTAMP '2024-01-01Z', ASYNC = true);
INSERT INTO foo RECORDS {_id: 1};
COMMIT;

-- cross-statement snapshot isolation
BEGIN READ ONLY WITH (SNAPSHOT_TOKEN = '…');   -- or SNAPSHOT_TIME / CLOCK_TIME; a bare BEGIN READ ONLY auto-pins
SELECT ...;   -- every read in the transaction sees one consistent snapshot
COMMIT;

BEGIN READ WRITE WITH (TIMEZONE = 'America/New_York');   -- tz for this transaction only
SET TIME ZONE 'America/New_York';   -- session tz; transactional (reverted on ROLLBACK)
SET SESSION CHARACTERISTICS AS TRANSACTION READ ONLY;
SHOW AWAIT_TOKEN;
SHOW LATEST_SUBMITTED_TX;   -- this connection's own last write
\`\`\`

## Changes (reading order)

1. \`docs(tx)\` — the connection transaction-model spec (\`tx.allium\`).
2. \`feat(adbc)\` — drive \`BEGIN\` / \`COMMIT\` / \`ROLLBACK\` (with access mode and commit options) as SQL on the connection.
3. \`feat(adbc)\` — reject queries in a write transaction; resolve a bare \`BEGIN\` to read-only on its first query.
4. \`refactor(database)\` — a Kotlin time-basis token and \`Catalog.snapshotToken()\` (groundwork for begin-time snapshots).
5. \`feat(adbc)\` — pin a read basis for \`READ ONLY\` transactions (snapshot isolation).
6. \`feat(adbc)\` — answer the session \`SET\` / \`SHOW\` surface on the connection.
7. \`feat(adbc)\` — eager-expand connection DML to core tx-ops (\`INSERT\`/\`PATCH RECORDS\` → \`PutDocs\`/\`PatchDocs\`) through a \`SqlPlanner.toStaticOps\` adapter over the (now \`RelationReader\`-based) \`sql->static-ops\`, matching pgwire's commit-time expansion.
8. \`feat(adbc)\` — answer \`SHOW LATEST_SUBMITTED_TX\` from the connection's own last-write record (tx-id / system-time / committed / error / await-token).
9. \`feat(adbc)\` — set the transaction time zone via \`BEGIN … WITH (TIMEZONE = …)\`; tx-scoped, so the session tz is restored after the transaction, while a mid-transaction \`SET TIME ZONE\` persists on \`COMMIT\`.
10. \`feat(adbc)\` — expose the connection's default access mode (frontend introspection).
11. \`feat(adbc)\` — the connection owns query tracing: the node resolves the (config-gated) query-tracer once and hands it to every connection, so tracing follows the node's config uniformly across frontends. Previously only pgwire threaded it, so ADBC/Flight never traced.
12. \`refactor(adbc)\` — fold the read access-mode gate into \`openQuery\`; \`resolveForQuery\` and \`queryOpts\` go private, reached only through it.
13. \`feat(adbc)\` — expose \`openUncheckedQuery\`, the one sanctioned ungated read (a frontend's driver-compatibility probe, e.g. pgjdbc's type-metadata query issued mid-write-tx).

## Implementation notes

Transaction control is expressed as SQL and classified by the shared \`SqlParser\` / \`ParsedStatement\` (the same path pgwire will delegate through). A \`ReadWrite\` transaction is write-only (queries rejected) and a \`ReadOnly\` one is read-only (DML rejected), matching XTDB's existing pgwire semantics; a bare \`BEGIN\` resolves on its first statement.

A read-only transaction pins a begin-time basis — a snapshot token plus the current-time — so every read in it observes one consistent snapshot (hence the Kotlin time-basis token and a \`Clock\` on the connection); \`READ ONLY WITH (...)\` overrides the auto-pin.

**Session surface.** \`SHOW AWAIT_TOKEN\` / \`SHOW LATEST_SUBMITTED_TX\` / \`SHOW <param>\` are answered from connection state via a small in-memory result cursor, ahead of the access-mode gate (SHOW is valid in any transaction). \`SET TIME ZONE\` is transactional — reverted on \`ROLLBACK\`, kept on \`COMMIT\`, rejected once a write transaction has buffered ops; \`BEGIN … WITH (TIMEZONE)\` overrides the zone for that transaction only. The model is captured in \`tx.allium\`.

**DML expansion.** DML is eager-expanded to core tx-ops before submit — the way pgwire expands at commit — rather than submitted as a raw \`Sql\` op for index-time expansion: \`INSERT\` / \`PATCH RECORDS\` become \`PutDocs\` / \`PatchDocs\`, which coerces literal ids (a UUID \`_id\` round-trips, where the raw path failed at index time as \`Invalid ID type: [B\`) and surfaces planner-style errors; a non-expandable statement (\`DELETE\` / \`UPDATE\`) keeps its raw \`Sql\` op. \`expandStaticOps\` applies \`SqlPlanner.toStaticOps\` at both submit sites — the commit-buffer drain and the autocommit single op — closing each buffered op exactly once.

**Query open + tracing.** Reads open through \`openQuery(pq, args)\`: it resolves the tx access mode and basis, then opens at the connection's basis, tz and tracer. Every read routes through it (\`executeQuery\`, \`getObjects\`), so \`resolveForQuery\` and \`queryOpts\` are private — a caller reconstructs neither the \`QueryOpts\` nor the gate. \`openUncheckedQuery\` is the lone bypass, for driver-compat probes that must run in any transaction. Query tracing moved off pgwire: the node resolves the gated tracer and the connection threads it into every read's \`QueryOpts\`, so ADBC/Flight now trace when query-tracing is enabled. This is the read surface the pgwire flip will drive.

## Out of scope (follow-ups)

- **Connection \`user\` / authn** — its own increment. ADBC's parameter-less \`connect()\` can't supply a user, so pgwire will populate it at the handshake (threaded into \`TxOpts.user\` and authz).
- **pgwire wholesale flip** — route pgwire's statement dispatch and reads/writes through the connection (\`openQuery\` / \`openUncheckedQuery\` / \`executeDml\` / \`begin*\` / \`commitTx\`), reducing it to a protocol shim. pgwire's \`verify-permissibility\` keeps only the read-only-*server* rule (a pgwire deployment property, not connection state) and the pgjdbc carve-out (routed through \`openUncheckedQuery\`). The node's Clojure \`snapshot-token\` de-dupes onto \`Catalog.snapshotToken()\` then.

## Test plan

\`InProcessAdbcTest\` (transaction control, access-mode, snapshot isolation, session surface, static-ops expansion, tz), \`FlightSqlAdbcTest\`, \`ConnectionTest\`, \`AdbcTest\`; the \`pgwire\` and \`sql\` suites as a no-regression oracle (the \`pg2_test/test-transit-id-formats\` id-format gate included); full \`./gradlew test\`, no reflection/boxed-math warnings. \`allium check\` clean.

### DIFF
--- a/api/src/main/kotlin/xtdb/database/Basis.kt
+++ b/api/src/main/kotlin/xtdb/database/Basis.kt
@@ -2,14 +2,20 @@
 
 package xtdb.database
 
+import com.google.protobuf.Timestamp
 import xtdb.api.log.MessageId
 import xtdb.error.Incorrect
 import xtdb.proto.basis.*
+import java.time.Instant
 import kotlin.io.encoding.Base64
 import kotlin.math.max
 
 typealias DatabaseName = String
 typealias TxBasisToken = Map<DatabaseName, List<MessageId>>
+
+// a read snapshot's upper bound: per database, the latest-completed system-time of each partition
+// (null where a partition has completed no transactions).
+typealias TimeBasisToken = Map<DatabaseName, List<Instant?>>
 
 fun TxBasisToken.encodeTxBasisToken(): String =
     txBasis {
@@ -32,6 +38,39 @@ fun String.decodeTxBasisToken(): TxBasisToken =
         .let { basis ->
             basis.databasesMap.entries
                 .associate { it.key to it.value.messageIdsList }
+        }
+
+fun TimeBasisToken.encodeTimeBasisToken(): String =
+    timeBasis {
+        entries.forEach { (dbName, sysTimes) ->
+            databases.put(dbName, systemTimes {
+                systemTimes.addAll(sysTimes.map {
+                    it?.let { Timestamp.newBuilder().setSeconds(it.epochSecond).setNanos(it.nano).build() }
+                        ?: Timestamp.getDefaultInstance()
+                })
+            })
+        }
+    }
+        .toByteArray()
+        .let { Base64.encode(it) }
+
+fun String.decodeTimeBasisToken(): TimeBasisToken =
+    runCatching { TimeBasis.parseFrom(Base64.decode(this)) }
+        .getOrElse {
+            throw Incorrect(
+                message = "Invalid basis: ${it.message}",
+                errorCode = "xtdb.basis/invalid-basis",
+                cause = it
+            )
+        }
+        .let { basis ->
+            basis.databasesMap.entries
+                .associate { (dbName, sysTimes) ->
+                    dbName to sysTimes.systemTimesList.map {
+                        if (it == Timestamp.getDefaultInstance()) null
+                        else Instant.ofEpochSecond(it.seconds, it.nanos.toLong())
+                    }
+                }
         }
 
 fun mergeTxBasisTokens(l: String?, r: String): String {
--- a/core/src/main/clojure/xtdb/node/impl.clj
+++ b/core/src/main/clojure/xtdb/node/impl.clj
@@ -19,6 +19,7 @@
   (:import io.micrometer.core.instrument.composite.CompositeMeterRegistry
            io.micrometer.core.instrument.Counter
            (java.io Closeable Writer)
+           (java.time Clock)
            (java.util HashMap)
            [java.util.concurrent.atomic AtomicReference]
            (org.apache.arrow.memory BufferAllocator)
@@ -85,7 +86,7 @@
 (defn- ->node-connection ^Xtdb$Connection [^Database$Catalog db-cat ^BufferAllocator allocator
                                            ^IQuerySource q-src ^SqlPlanner sql-planner default-tz ^Counter tx-error-counter
                                            tx-await-timer tx-submit-timer tx-execute-timer db-name]
-  (Xtdb$Connection. allocator db-cat q-src sql-planner default-tz tx-error-counter tx-await-timer tx-submit-timer tx-execute-timer db-name))
+  (Xtdb$Connection. allocator db-cat q-src sql-planner (Clock/systemDefaultZone) default-tz tx-error-counter tx-await-timer tx-submit-timer tx-execute-timer db-name))
 
 (defrecord Node [^BufferAllocator allocator, ^Database$Catalog db-cat
                  ^IQuerySource q-src ^SqlPlanner sql-planner
--- a/core/src/main/clojure/xtdb/node/impl.clj
+++ b/core/src/main/clojure/xtdb/node/impl.clj
@@ -11,6 +11,7 @@
             [xtdb.protocols :as xtp]
             [xtdb.query :as q]
             [xtdb.serde :as serde]
+            [xtdb.sql :as sql]
             [xtdb.time :as time]
             [xtdb.tracer]
             [xtdb.util :as util]
@@ -27,7 +28,7 @@
            xtdb.api.module.XtdbModule$Factory
            (xtdb.database Database Database$Catalog DatabasePartition)
            xtdb.error.Anomaly
-           (xtdb.query IQuerySource PreparedQuery QueryOpts)))
+           (xtdb.query IQuerySource PreparedQuery QueryOpts SqlPlanner)))
 
 (set! *unchecked-math* :warn-on-boxed)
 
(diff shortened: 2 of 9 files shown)


jarohen:
### Chalk — Eager-expand connection DML to static ops

Part of the ADBC-parity arc (Refs #5132). Lands `f67f63526`.

- [x] Column-major `rel->static-ops` + `SqlPlanner.toStaticOps` SPI
- [x] `expandStaticOps` wired into `commitTx` and the autocommit branch of `executeDml`
- [x] Tests + parity/spec verification

<details><summary>Why — the connection skipped a load-bearing expansion</summary>

The connection DML path (ADBC and Flight SQL drive it directly) submitted a raw `TxOp.Sql` and left expansion to the indexer, where pgwire instead expands each buffered op through `sql->static-ops` at commit. That expansion isn't a perf tweak — it coerces literal ids (a UUID `_id` that otherwise fails at index time as `Invalid ID type: [B`) and surfaces planner-style errors rather than indexer-style ones. So the connection's writes diverged from pgwire's on exactly the cases that matter to a user. This brings it to parity, which ADBC and Flight SQL get today, and gives the eventual pgwire flip one already-correct path to join.

</details>

<details><summary>Mechanism — a column-major front-door reusing the shared visitor</summary>

The connection holds args column-major (a `RelationReader`, vectors named `?_0..?_n`), where `sql->static-ops` takes row-major arg-rows. `rel->static-ops` (xtdb.sql) is the column-major counterpart: it reads args straight off the reader — the vector fields are the arg-fields, so it drops the row-major path's type-merge loop — and reuses the same `SqlToStaticOpsVisitor`, keeping the expansion logic in one place.

Args are read with `KEBAB_CASE_KEYWORD`: `SNAKE_CASE_STRING` would id-normalize `:xt/id`→`_id` and wrongly accept docs the row-major path rejects (parity gate: `pg2_test/test-transit-id-formats`).

`SqlPlanner.toStaticOps` exposes it to the Kotlin connection, composing `rel->static-ops` with `open-tx-op` (via `safe-mapv`, which closes partial results if a later op throws) into core `TxOp`s — or nil when the statement isn't statically expandable, leaving `DELETE`/`UPDATE` to expand at index time as before.

The row-major `sql->static-ops` stays for pgwire — deliberate, temporary duplication that deletes when pgwire flips onto this path.

</details>

<details><summary>Resource ownership in `expandStaticOps` — the sharp edge</summary>

The buffered ops own Arrow `RelationReader`s, so `expandStaticOps` has to be exact. It runs on a drained, throwaway list and its result is `useAll`-d after submit, so every op is closed once: an expanded op's original `Sql` op is closed in place (it never reaches the output), a non-expandable op passes through to be closed by the caller, and if `toStaticOps` throws mid-list (a deliberate throw, e.g. a missing `_id`) the unconsumed tail is closed there while the accumulated output is closed by `closeAllOnCatch`. Wired at both submit sites — `commitTx` (drained buffer) and the autocommit branch of `executeDml` (the single op) — so manual-tx and autocommit DML take the same path.

</details>

<details><summary>Verification</summary>

284 tests green — `InProcessAdbcTest` (incl. 4 new: UUID-`_id` round-trip via INSERT and PATCH RECORDS, the static missing-`_id` error, the non-expandable DELETE fallback), `ConnectionTest`, `FlightSqlAdbcTest`, plus `sql_test`/`pg2_test`/`patch_test` confirming the row-major path is untouched (the `pg2_test/test-transit-id-formats` parity gate stays green). No reflection or boxed-math warnings. `allium:weed` over `tx.allium` confirmed no spec change — `TxOp` is opaque at the spec boundary, so where/when expansion happens is below it.

One caveat: this was an optimistic commit to get CI moving, so the local code-review pass was skipped — CI is the first gate. The spot worth a human eye is the resource-ownership in `expandStaticOps`.

</details>

PR #5752, opened 2026-06-29
refactor(postgres-source)!: move database from the !Postgres remote onto the source

Akeboshiwind:
## Summary

Moves the `database` field off the `!Postgres` remote and onto the `!Postgres` external source, so a single remote can back several sources.

## Motivation

`database` lived on the `!Postgres` remote, which pinned a remote to one upstream database. Two Postgres sources streaming different databases off the same server couldn't share a remote — you had to declare a near-duplicate remote per database, identical but for the `database` line. Hostname, port and credentials are genuinely reusable across sources; the database to replicate is the one part that varies per source, so it moves to where that variation lives.

## Config change

The connection details stay on the remote; the database moves to each source:

```yaml
remotes:
  pg: !Postgres
    hostname: pg.prod.internal
    username: xtdb
    password: secret

# database config
externalSource: !Postgres
  remote: pg
  database: app          # ← was on the remote
  slotName: my_slot
  publicationName: my_pub
```

To adopt: drop `database:` from each `!Postgres` remote under `remotes:`, and add it to each `externalSource: !Postgres` block.

## Breaking change

The `!Postgres` YAML shape is user-facing, hence the `!`.

On the wire, `PostgresSourceConfig` gains `database` as field 5 (1–4 untouched). A config persisted *before* this change has no field 5, so it decodes with `database = ""` and fails at connect time (`jdbc:postgresql://host:port/`). Unlike `indexer` — which back-fills to `!DirectMirror` when absent — `database` has no sensible default, so a loud failure is preferable to silently connecting to the wrong place.

## The counter-argument (why this is an ask)

There's a real case *against* this move: `database` is part of forming the JDBC connection — host, port, credentials **and** database together are what open a session, and without it we can't connect to Postgres at all. By that reading it belongs alongside the other connection parameters on the remote, and the source should only name *what* to stream (slot, publication), not *where*.

The view this PR takes is that the database is the unit a source replicates, not a property of the endpoint: the same server, reachable by one set of credentials, can host many databases, and we want one remote to serve all of them. The connection is still fully specified — it's just assembled from the remote (the reusable endpoint) plus the source (the per-source database) at `open` time, rather than being baked whole into the remote.

That's the judgement I'd like a second opinion on before merging.

## Test plan

- `:modules:xtdb-postgres-source:test` — passes (proto + YAML round-trips assert the field lands on the source).
- `:modules:xtdb-postgres-source:integration-test` — the replication path passes, confirming the database value reaches the real Postgres connection from its new home.
- One reroute test (`PostgresSourceIndexerTest`) is failing/hanging on a known `main` bug unrelated to this change, being fixed separately.

### DIFF
--- a/modules/postgres-source/src/main/clojure/xtdb/postgres.clj
+++ b/modules/postgres-source/src/main/clojure/xtdb/postgres.clj
@@ -6,5 +6,5 @@
   (:import [xtdb.postgres PostgresRemote$Factory]))
 
 (defmethod remote/->remote-factory ::remote
-  [_ {:keys [host port database username password] :or {port 5432}}]
-  (PostgresRemote$Factory. host port database username password))
+  [_ {:keys [host port username password] :or {port 5432}}]
+  (PostgresRemote$Factory. host port username password))
--- a/modules/postgres-source/src/main/kotlin/xtdb/postgres/PostgresRemote.kt
+++ b/modules/postgres-source/src/main/kotlin/xtdb/postgres/PostgresRemote.kt
@@ -9,7 +9,6 @@ import xtdb.api.Remote
 class PostgresRemote(
     val hostname: String,
     val port: Int,
-    val database: String,
     val username: String,
     val password: String,
 ) : Remote {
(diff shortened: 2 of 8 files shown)


Akeboshiwind:
Decision: Keep on the remote because we need the database to create the connection pool

Issue #5714, opened 2026-06-12
Compaction job fails with 'cannot promote DUV leg' across schema-evolution epochs; table becomes permanently unreadable ('vectorFor unsupported on NullVector')


JohanCodinha:
## Summary

A compaction job crashed with `xtdb.error.Unsupported: cannot promote DUV leg` while merging blocks that span a schema-evolution boundary (a string column present in one data generation, absent in the next). After the failed job, **every read of the table fails permanently** with `vectorFor unsupported on NullVector` — including `SELECT COUNT(*)` and `SELECT _id WHERE _id = ...`. Recovery required dropping the node's data volume.

This looks like the production-path manifestation of the DUV-promotion hole already tracked in #5570 / #5058 — filing separately because the consequence here is a wedged table rather than a failing property test, which raises the severity considerably.

## Environment

- XTDB `2.1.0` (`ghcr.io/xtdb/xtdb:2.1.0`), single standalone node, local volume, all access over pgwire (pgjdbc 42.7.7)
- Data written via `COPY ... FROM STDIN WITH (FORMAT 'transit-json')` in 100k-doc batches, plus small parameterized INSERTs

## Timeline that produced it

The `records` table went through several full data generations in one day (load-test churn):

1. `COPY` 1M docs — column set `{_id, dataset_id, name, sensitivity}` (`sensitivity` utf8, absent on ~5% of docs)
2. `ERASE FROM records WHERE true`; small INSERT generation (~10 rows, same columns)
3. several more ERASE + 1M-`COPY` generations
4. final 1M-`COPY` generation adds a new utf8 column `org_id` on every doc
5. `ERASE FROM records WHERE true`; small INSERT generation **without** `org_id` (and with one row carrying an explicit `NULL` `sensitivity` param)
6. within seconds, the compactor logs the error below; from then on all reads of `records` fail

```
13:24:29 | ERROR xtdb.compactor.Compactor | error running compaction job: public/records/l01-rc-b118, files in job: '["l01-rc-b117" "l00-rc-b118"]'
xtdb.error.Unsupported: cannot promote DUV leg
	at xtdb.arrow.DenseUnionVector.legVectorFor(DenseUnionVector.kt:270)
	at xtdb.arrow.DenseUnionVector.rowCopier0$xtdb_core(DenseUnionVector.kt:325)
	at xtdb.arrow.Vector.rowCopier(Vector.kt:121)
[...]

PR #5757, opened 2026-06-30
fix(arrow): reconcile null DUV legs instead of failing the merge (#5714)

mbutlerw:
Compaction merges that combined a put-less trie with one that still had documents threw `cannot promote DUV leg`, and the table stopped compacting. The `op` dense-union keys its legs by operation, so the `put` leg's type is data-dependent: `Struct` where a segment has documents, but `Null` once a put-less segment (deletes/erases only) is compacted — `dataRelSchema(mergeFields(...))` mints a present-but-`Null` `put` leg when no input carries a put schema. Merging that into a `Struct`-put trie reached `DenseUnionVector.legVectorFor`, which rejected any same-named leg whose arrow type differed.

`Null` is the bottom of the type lattice, though — always reconcilable, never a real clash. An incoming `Null` leg carries no typed values, so it absorbs into the existing leg; an existing `Null` leg promotes to the incoming type. Only two genuinely-distinct non-null types stay unpromotable — a `toLeg` naming collision, handled elsewhere — and that now throws with the leg name and both types so it's diagnosable.

This is the compaction-side of #5714. The read-side wedge (`vectorFor unsupported on NullVector`, the permanently-unreadable table) reproduces on 2.1.0 but not on main — already fixed between the release and now — so this covers what was still live.

### DIFF
--- a/core/src/main/kotlin/xtdb/arrow/DenseUnionVector.kt
+++ b/core/src/main/kotlin/xtdb/arrow/DenseUnionVector.kt
@@ -261,7 +261,18 @@ class DenseUnionVector private constructor(
         for (i in legVectors.indices) {
             val leg = legVectors[i]
             if (leg.name == name) {
-                if (leg.arrowType != arrowType) throw Unsupported("cannot promote DUV leg")
+                // a DUV leg keyed by op rather than type (e.g. `op`'s `put`) can be Struct
+                // in one segment and Null in a put-less one. null is the lattice bottom, so
+                // reconcile it instead of throwing; only two distinct non-null types are
+                // genuinely unpromotable. #5714
+                if (leg.arrowType != arrowType) {
+                    when {
+                        arrowType == NULL_TYPE -> leg.nullable = true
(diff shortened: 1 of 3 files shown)


mbutlerw:
Closing in favour of #5762.

Same fix — DenseUnionVector.legVectorFor reconciles the op.put leg (Struct in a doc-bearing segment, Null in a put-less one) instead of throwing cannot promote DUV leg. The difference is how the Null side is handled.

This branch eagerly set leg.nullable = true. That's redundant: writeNull already auto-widens a leg to nullable the moment an actual null is written — so a genuine null still nullables the leg either way. All = true adds is nullability when there's no null to justify it, and for op.put there never is (a put row always carries a document, so a Null put leg is always zero-row). #5762 takes the no-op instead: leave the existing typed leg alone and let writeNull widen it if and when a real null lands. So put stays a non-nullable Mono(STRUCT) — the shape logRelTypes assumes and where #4467's Nothing is headed — with no loss of safety.

Superseded; nothing unique to carry over.

PR #5759, opened 2026-06-30
docs: external source pages (Postgres + Kafka Connect)

Akeboshiwind:
Documents XTDB's external sources — secondary databases that mirror an upstream feed rather than accepting client writes — which had shipped without user-facing docs.

## Summary

Adds documentation for both external sources, organised under `ops/external-sources/` with one directory per source and wired into the ops sidebar:

- `external-sources/overview` — what an external source is (a read-only mirror of an upstream) and how you attach one.
- `external-sources/postgres/` — the Postgres CDC source: a setup guide and a reference.
- `external-sources/kafka-connect/` — the Kafka Connect source: a setup guide and a reference.

## Shape

The pages follow the docs' Diataxis split: the overview is explanation (no task), each reference is the exhaustive lookup (config surface, indexers, type mappings), and each setup guide walks one concrete path end to end — e.g. topic `orders` into a `kc_orders` database, masking the `email` field via a transform gated by a tombstone predicate.

The per-source directory layout mirrors the nav tree, following the `ops/backup-and-restore/` precedent (a feature directory with an overview plus its sub-pages) rather than splitting how-tos into `ops/guides/`.

## Conventions

Config is shown as inline-commented YAML — the field docs live in comments inside a copy-pasteable block, matching the other config pages (`ops/config/log/kafka` and friends) rather than a separate definition list or table.

The bundled transforms and predicates defer to the versioned Kafka Connect docs (`kafka.apache.org/41/…`), which document each class with its config; the version catalog carries a note to update the page on a Kafka bump. Converters stay listed with links, since Kafka has no per-converter config page whereas Confluent's guide documents each.

## Accuracy

The pages track the config as it currently stands, including the `!Docs` indexer's single dotted `table` (`schema.table`) value after that field settled from an earlier separate schema/table shape. The Kafka Connect pages were reviewed against the source code twice, so the `connectConfig` key space, the `_id` derivation rules, and the type mappings all match the implementation.

### DIFF
--- a/docs/src/content/docs/ops/external-sources/kafka-connect/setup.md
+++ b/docs/src/content/docs/ops/external-sources/kafka-connect/setup.md
@@ -1,0 +1,82 @@
+---
+title: Setting up a Kafka Connect external source
+---
+
+In this guide we will set up a [Kafka Connect external source](/ops/external-sources/kafka-connect/reference) from the Kafka topic `orders` into a database in XTDB called `kc_orders`.
+Along the way we will apply a transform to mask the `email` field, gated by a predicate so it skips tombstones.
+
+To do so we will:
+
+1. Configure the Kafka cluster credentials on the XTDB node and redeploy
+1. Run `ATTACH DATABASE` in XTDB
+
+## Prerequisites
+
+As with other [external sources](/ops/external-sources/overview) you will need:
+
+- A [transaction log](/ops/config/log)
+- An [object store](/ops/config/storage)
+
+:::caution
+Ensure that you have a transaction log and object store that do not conflict with other databases.
+:::
+
+Additionally you will need a single-partition Kafka topic carrying the records to sync.
+
+## Deploy the Kafka cluster credentials
+
+Configured under the [`remotes`](/ops/config#remotes) section of the node config like so:
+
+```yaml
+remotes:
+  kafka_remote: !Kafka
+    bootstrapServers: "localhost:9092"
+```
+
+For nodes to pick up this config change a rolling re-deploy is required.
+
+:::note
+If your [transaction log](/ops/config/log/kafka) is already a `!Kafka` remote, you can reuse those credentials for the source rather than declaring a new remote — and since the config is unchanged, no rolling re-deploy is needed.
+:::
+
+## Run `ATTACH DATABASE` in XTDB
+
+Finally to attach the secondary database with the external source:
+
+```sql
+ATTACH DATABASE kc_orders WITH $$
+# Set up in the prerequisites
+log: !Local
+  path: 'kc-orders/log'
+storage: !Local
+  path: 'kc-orders/storage'
+
+externalSource: !KafkaConnect
+  remote: kafka_remote
+  topic: orders
+  connectConfig:
+    key.converter: org.apache.kafka.connect.storage.StringConverter
+    value.converter: org.apache.kafka.connect.json.JsonConverter
+    value.converter.schemas.enable: "false"
+    transforms: mask
+    transforms.mask.type: org.apache.kafka.connect.transforms.MaskField$Value
+    transforms.mask.fields: email
+    transforms.mask.replacement: REDACTED
+    transforms.mask.predicate: notTombstone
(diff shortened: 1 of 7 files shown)


docs/src/content/docs/ops/kafka-connect-external-source.md
```diff
@@ -0,0 +1,203 @@
...
+
+:::caution
+Records on partitions other than partition 0 are not consumed, so the topic should have a single partition.
+:::
+
+## Configuration
+
+Configured by [attaching a secondary database](/about/dbs-in-xtdb#attachingdetaching-secondary-databases-v21) with the additional `externalSource` options:
+
+remote
```
jarohen: let's have this consistent with the other YAML reference - i.e. YAML comments within the example. happy to discuss migrating it to a different format if you think there's value in it, but for this PR let's go for consistency

docs/src/content/docs/ops/kafka-connect-external-source.md
```diff
@@ -0,0 +1,203 @@
...
+  my_kafka_remote: !Kafka
+    bootstrapServers: "localhost:9092"
+```
+
+### Connect config
+
+`connectConfig` is a map of [Kafka Connect](https://kafka.apache.org/documentation/#connect) options applied to each record before it is indexed.
+XTDB accepts the keys below; any other key is rejected.
+
+| Key | Description |
```
jarohen: same here, inline in the YAML so that people can copy-paste it as default config, they can leave the comments in as their own documentation

docs/src/content/docs/ops/kafka-connect-external-source.md
```diff
@@ -0,0 +1,203 @@
...
+
+Converters, transforms, and predicates are loaded from the node's classpath by their fully-qualified class name.
+The classes listed below are bundled with XTDB.
+
+To use any other Kafka Connect class, for example [`io.confluent.connect.json.JsonSchemaConverter`](https://docs.confluent.io/platform/current/connect/userguide.html#json-schema-and-protobuf), add its jar to the node's classpath.
+How you add a jar to the classpath depends on how you deploy XTDB, for example as a dependency in your build or bundled into your XTDB image.
+
+A custom class must implement the relevant Kafka Connect interface:
+- Converters: `org.apache.kafka.connect.storage.Converter`
+- Transforms: `org.apache.kafka.connect.transforms.Transformation`, parameterised over `org.apache.kafka.connect.sink.SinkRecord`
```
jarohen: Transformation&lt;SinkRecord&gt; rather than wordy

docs/src/content/docs/ops/kafka-connect-external-source.md
```diff
@@ -0,0 +1,203 @@
...
+Converters deserialize record keys and values.
+The following are included:
+- [`org.apache.kafka.connect.storage.StringConverter`](https://docs.confluent.io/platform/current/connect/userguide.html#string-format-and-raw-bytes)
+- [`org.apache.kafka.connect.json.JsonConverter`](https://docs.confluent.io/platform/current/connect/userguide.html#json-without-sr)
+- [`io.confluent.connect.avro.AvroConverter`](https://docs.confluent.io/platform/current/connect/userguide.html#avro)
+
+### Transforms
+
+[Single Message Transforms](https://kafka.apache.org/43/kafka-connect/user-guide/#transformations) modify each record before it is indexed.
+The following are included:
```
jarohen: defer to the Kafka docs for this list, assuming you have a permalink to the version we're using. also worth adding a note to the Gradle versions file that this file also needs updating when we bump KC

docs/src/content/docs/ops/external-sources/kafka-connect/setup.md
```diff
@@ -0,0 +1,82 @@
+---
```
jarohen, line 1: nit: I'd go for file path docs/ops/external-src/kafka/setup.md or similar given that's the hierarchy in the nav tree (similar for other files)

jarohen (approved):
thanks mate - nits below, fine to merge without further review 🙏

PR #5761, opened 2026-07-01
fix(database): bound DatabaseCatalog teardown so a stalled cancel-join fails loud, not hung (#5711)

mbutlerw:
Since "lift the teardown scope tree to DatabaseCatalog", node shutdown cancel-joins the whole catalog job tree in one step — `DatabaseCatalog.close`'s `dbJob.cancelAndJoin()`. A wait in some database's tree that cancellation can't reach (the #5711 class) hangs that join forever, and the JVM only dies at the CI action timeout hours later.

Bound the cancel-join with a generous 60s timeout. On a stall we skip Phase 2 and raise a `Fault`: freeing a database's Arrow allocator while a live, wedged coroutine still holds its buffers would be a use-after-free (or an allocator "leaked memory" throw that masks the real timeout), so we abandon the trees — their state leaks, unavoidable while a coroutine is stuck — rather than free under them. A stall becomes a fast, loud, diagnosable failure. The healthy path is unchanged: Phase 1 joins, then Phase 2 `closeAll` frees state as before.

The escape would otherwise cost the diagnostic: a bounded teardown makes the test *finish*, so the watchdog's no-activity scanner never trips and the coroutine dump naming the stuck wait is lost. So the timeout fires the dump itself, through a `TeardownStallProbe` SPI — a no-op in production, implemented by the test-watchdog (reusing its `DebugProbes` dump). The interface lives in `:xtdb-core` alongside the other SPIs (`ExternalSource`, `ObjectStore`); the watchdog depends on `:xtdb-core` to implement it.

Scope: this bounds and surfaces the stall; it doesn't remove the immune waits — those are the per-instance fixes tracked on the issue. `IngestNode.close` carries the same unbounded `rootJob.cancelAndJoin()` and is left as-is (test-only in-repo; the same bound applies if it surfaces). With no watchdog in production a stall throws without a coroutine dump.

Refs #5711

### DIFF
--- a/core/src/main/kotlin/xtdb/database/DatabaseCatalog.kt
+++ b/core/src/main/kotlin/xtdb/database/DatabaseCatalog.kt
@@ -5,14 +5,18 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.TimeoutCancellationException
 import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
+import kotlinx.coroutines.time.withTimeout
 import xtdb.NodeBase
 import xtdb.compactor.Compactor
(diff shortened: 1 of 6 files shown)


mbutlerw:
Closed in favour of https://github.com/xtdb/xtdb/pull/5763

PR #5762, opened 2026-07-01
fix(arrow): reconcile null DUV legs instead of failing the merge (#5714)

mbutlerw:
Compaction merges that combined a put-less trie with one that still had documents threw `cannot promote DUV leg`, and the table stopped compacting. The `op` dense-union keys its legs by operation, so the `put` leg's type is data-dependent: `Struct` where a segment has documents, but `Null` once a put-less segment (deletes/erases only) is compacted — `dataRelSchema(mergeFields(...))` mints a present-but-`Null` `put` leg when no input carries a put schema. Merging that into a `Struct`-put trie reached `DenseUnionVector.legVectorFor`, which rejected any same-named leg whose arrow type differed.

`Null` is the bottom of the type lattice, though — always reconcilable, never a real clash. An existing `Null` leg promotes to the incoming type; an incoming `Null` leg yields to the existing one *without* widening it to nullable. Nullability tracks the nulls actually written (`writeNull` auto-nullables), not a leg's declared type — and a `put` row is never null, so `op.put`'s `Null` leg is always empty. The reconciled `put` thus stays a non-nullable `Mono(STRUCT)`, the shape `logRelTypes` requires of it and the direction the #4467 `Nothing` work is taking, rather than a nullable struct no value justifies — one that would contradict that invariant and, once written, stick across re-merges via `mergeFields`. Two genuinely-distinct non-null types stay unpromotable — a `toLeg` naming collision handled elsewhere — and that throws with the leg name and both types, so it's diagnosable.

The same null-handling gap has a second call site: the scan reconciling a `Null`-put trie against a `Struct`-put one, which threw `vectorFor unsupported on NullVector` — the permanently-unreadable table in #5714. That read path was hardened separately between 2.1.0 and main; this commit fixes the compaction call site, and the e2e test guards both — a read across the un-merged tries, and the L2 merge that folds them together.

Refs #5714

### DIFF
--- a/core/src/main/kotlin/xtdb/arrow/DenseUnionVector.kt
+++ b/core/src/main/kotlin/xtdb/arrow/DenseUnionVector.kt
@@ -261,7 +261,21 @@ class DenseUnionVector private constructor(
         for (i in legVectors.indices) {
             val leg = legVectors[i]
             if (leg.name == name) {
-                if (leg.arrowType != arrowType) throw Unsupported("cannot promote DUV leg")
+                // `op` keys legs by operation, so `put` is a Struct where a segment has documents and
+                // Null in a put-less one. Null is the lattice bottom — reconcile rather than throw. We
+                // deliberately don't assert nullability: an incoming Null yields to the existing type,
+                // and real nulls (if any) fold in via writeNull, which auto-nullables. So op.put's
+                // zero-row Null leg leaves `put` a non-nullable Mono(STRUCT) — the shape logRelTypes
+                // asserts and the #4467 Nothing work relies on — rather than spuriously widening it.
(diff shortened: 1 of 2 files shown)


mbutlerw:
Holding rather than merging.

This is the compaction-side shim for legacy data: legVectorFor reconciles a Null op.put leg instead of throwing cannot promote DUV leg. #5766 stops new put-less compactions creating a Null put leg at all, so only tries written before that fix can still trip the merge.

Needed solely to unwedge a deployment that already holds legacy Null-put tries on disk — fresh data never hits it. Two mutually-exclusive ways to clear those: merge this shim (ordinary compaction then reconciles them in place), or recompact-from-L0 (discards and rebuilds them, no shim needed). Consider merging if a deployment needs unwedging before it can recompact.

PR #5769, opened 2026-07-02
fix(indexer): abort in-flight leader promotion when a database shuts down

jarohen:
Related to #5711 — the bounded-teardown safety net whose trigger this removes.

## Summary

A node could hang on close when shutdown raced a follower→leader promotion. `DatabaseCatalog.close`'s cancel-join never completed, timed out, and faulted with "database catalog did not shut down in time". This removes the underlying deadlock.

## Context

Surfaced while investigating hangs in `PostgresSourceIntegrationTest`, which hangs the whole class indefinitely. It's amplified by the tests sharing one Kafka consumer group (the default `xtdb`): a single wedged teardown leaves a zombie node that starves every subsequent node in the group, so one stuck test cascades into a whole-suite hang. The deadlock itself is per-node and independent of the shared group.

## Root cause

Shutdown races a follower→leader promotion:

```
[poll]  Kafka fires onPartitionsAssigned in a detached runBlocking
        (a root coroutine — not in any database's cancel-join tree)
          └─ LogProcessor promotion → awaitReplicaMsgId   ← SUSPENDED, awaiting replica catch-up

[close] node.close() → dbJob.cancelAndJoin()
          ├─ cancels the db's own coroutines: follower tail, source-log subscription, term
          └─ does NOT reach the promotion — it's on the detached [poll] coroutine   ← divergence

[poll]  still wedged in awaitReplicaMsgId → never returns to its select loop
          └─ never drains the queued Unregister command

[close] source-log openGroupSubscription's finally → NonCancellable unregister
          └─ awaits the poll loop that will never run it → cancelAndJoin blocks forever
```

Kafka drives its rebalance callbacks from a `runBlocking` on the shared consumer's poll thread — a root coroutine with no parent, so the promotion (its `awaitReplicaMsgId`, and the preceding replica-log `withTx { NoOp }.await()`) isn't in any database's cancel-join tree. `dbJob.cancelAndJoin()` walks the database's own coroutines but never reaches that await. The poll thread stays wedged, so it never drains the queued group-unregister, whose `NonCancellable` wait is the child `cancelAndJoin` is actually stuck on.

## The fix

Re-parent the promotion into the database's job at the transport→database boundary, so ordinary structured cancellation reaches it:

1. The `Database` subscription listener runs `onPartitionsAssigned` under `withContext(scope.coroutineContext.job)`. A shutdown mid-promotion now cancels its awaits. Revocation is deliberately left unwrapped — it also runs as part of teardown (the unregister invokes it), when that job is already cancelled, so re-parenting it would throw before the unregister completes and wedge teardown instead.

2. `LogProcessor.onPartitionsAssigned` rethrows the `CancellationException` rather than routing it through `watchers.notifyError` — that would wrap it in an `IngestionStoppedException`, log "ingestion stopping" at ERROR and permanently fail the watcher, relabelling a clean shutdown as an ingestion failure and hiding the type from awaiters. Matches the existing leader/follower convention.

3. The Kafka rebalance handler contains the `CancellationException` per-subscription so it doesn't reach `SharedGroupConsumer.pollingJob`, whose `cleanupOnFailure` would evict every database on the shared consumer. Freeing the poll thread lets the queued unregister drain; whole-node shutdown stops the poll loop through its own channel-close path regardless.

Also folds in a small tidy: `processCommand` is now `suspend` and calls `onPartitionsRevoked` directly, dropping the now-unused `onPartitionsRevokedSync` `runBlocking` bridge.

## Implementation notes

- **Why re-parent at the `Database` listener, not deeper.** It's the transport→database boundary and the first frame with the db `scope`; it's the single chokepoint for every log type and normalises the invocation context (in-memory/local logs already run under `scope`, so it's a no-op there; only Kafka's detached poll-thread `runBlocking` needs it). Burying it in `awaitReplicaMsgId` would leave the preceding `withTx().await()` exposed to the same hang.
- **The assign/revoke asymmetry is deliberate.** The revoke path *is* teardown — tying it to the job teardown cancels is self-defeating. An earlier attempt that wrapped it wedged the unregister (confirmed via coroutine dump: `openGroupSubscription` stuck in a `NonCancellable` unregister that `onPartitionsRevoked` never let complete).
- **Cancellation isn't laundered through `notifyError`.** `notifyError` wraps into `IngestionStoppedException`, hiding the `CancellationException` type from downstream `catch (CancellationException)` sites — so a clean shutdown would masquerade as an ingestion failure. Handling it separately is the existing house convention.

## Test plan

- Core transition suite (`xtdb.indexer.*`, `ExternalSourceTest`): 35/35 pass — no regression on the normal leader↔follower path.
- `PostgresSourceIntegrationTest`: 20/20 pass, class completes in ~2m14s (previously hung forever), with the shared consumer group still in place — confirming the fix removes the cascade at its root.
- No reflection / boxed-math warnings.

## Out of scope

- The tests' shared Kafka consumer group is a separate test-hygiene issue (a unique `groupId` per node would isolate them); unnecessary once the deadlock is fixed, so left as-is.
- `onPartitionsAssignedSync` / `onPartitionsLostSync` are pre-existing dead code (no callers); left for a separate tidy.

### DIFF
--- a/core/src/main/kotlin/xtdb/api/log/Log.kt
+++ b/core/src/main/kotlin/xtdb/api/log/Log.kt
@@ -160,7 +160,6 @@ interface Log<M> : AutoCloseable {
         suspend fun onPartitionsAssigned(partitions: Collection<Int>): TailSpec<M>?
         fun onPartitionsAssignedSync(partitions: Collection<Int>): TailSpec<M>? = runBlocking { onPartitionsAssigned(partitions) }
         suspend fun onPartitionsRevoked(partitions: Collection<Int>)
-        fun onPartitionsRevokedSync(partitions: Collection<Int>) = runBlocking { onPartitionsRevoked(partitions) }
         suspend fun onPartitionsLost(partitions: Collection<Int>) = onPartitionsRevoked(partitions)
         fun onPartitionsLostSync(partitions: Collection<Int>) = runBlocking { onPartitionsLost(partitions) }
     }
--- a/core/src/main/kotlin/xtdb/database/Database.kt
+++ b/core/src/main/kotlin/xtdb/database/Database.kt
@@ -397,10 +397,20 @@ class Database(
 
             if (indexerConfig.enabled && !readOnly) {
                 val listener = object : Log.SubscriptionListener<SourceMessage> {
+                    // Re-parent the promotion into this database's job. The Kafka shared consumer drives
+                    // rebalance callbacks from a detached `runBlocking` on its poll thread, so without this
+                    // the leader-transition await isn't in the database's cancel-join tree and a shutdown
+                    // that lands mid-promotion hangs. (In-memory/local logs already run us under `scope`,
+                    // so this is a no-op for them.)
                     override suspend fun onPartitionsAssigned(partitions: Collection<Int>) =
-                        partitions.singleOrNull()
-                            ?.let { db.partitions[it]?.logProcessor?.onPartitionsAssigned(listOf(it)) }
+                        withContext(scope.coroutineContext.job) {
+                            partitions.singleOrNull()
+                                ?.let { db.partitions[it]?.logProcessor?.onPartitionsAssigned(listOf(it)) }
+                        }
 
+                    // Deliberately NOT re-parented: revocation also runs as part of teardown (the unregister
+                    // path invokes it), by which point this job is already cancelled — re-parenting here
+                    // would throw before the unregister completes and wedge teardown.
                     override suspend fun onPartitionsRevoked(partitions: Collection<Int>) {
                         for (idx in partitions) db.partitions[idx]?.logProcessor?.onPartitionsRevoked(listOf(idx))
                     }
--- a/core/src/main/kotlin/xtdb/indexer/LogProcessor.kt
+++ b/core/src/main/kotlin/xtdb/indexer/LogProcessor.kt
@@ -2,6 +2,7 @@ package xtdb.indexer
 
 import io.micrometer.core.instrument.Gauge
 import io.micrometer.core.instrument.MeterRegistry
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
@@ -176,6 +177,14 @@ class LogProcessor(
                 } catch (e: InterruptedException) {
                     throw e
                 } catch (e: Interrupted) {
+                    throw e
+                } catch (e: CancellationException) {
+                    // Shutdown cancelled this database's job mid-promotion — the transition runs under it
+                    // (re-parented at the Database subscription listener), so an await here was cancelled.
+                    // Rethrow rather than fall through to the notifyError catch below: a clean shutdown
+                    // must not be surfaced as an ingestion failure. The transport contains this
+                    // per-subscription so it doesn't evict a whole shared consumer — see the Kafka
+                    // SharedGroupConsumer's rebalance handler.
                     throw e
                 } catch (e: Throwable) {
                     LOG.error(e, "[$dbName] transition: failed to transition to leader")
--- a/modules/kafka/src/main/kotlin/xtdb/api/log/KafkaCluster.kt
+++ b/modules/kafka/src/main/kotlin/xtdb/api/log/KafkaCluster.kt
@@ -47,6 +47,7 @@ import xtdb.util.MsgIdUtil.afterMsgIdToOffset
 import xtdb.util.MsgIdUtil.msgIdToEpoch
 import xtdb.util.MsgIdUtil.msgIdToOffset
 import xtdb.util.close
+import xtdb.util.debug
 import xtdb.util.error
 import xtdb.util.info
 import xtdb.util.logger
@@ -212,6 +213,14 @@ class KafkaCluster(
                                 val sub = subscriptions[topic] as? TopicSubscription<Any?> ?: continue
                                 try {
                                     sub.onPartitionAssigned(tps.single())
+                                } catch (e: CancellationException) {
+                                    // A topic's log processor aborted a leader transition because its
+                                    // database is shutting down. Swallow it here rather than let it reach
+                                    // the poll loop: this runBlocking is on the poll thread shared by every
+                                    // database on the cluster, and an exception escaping into `pollingJob`
+                                    // runs cleanupOnFailure and evicts them all. Returning frees the thread
+                                    // so this topic's queued unregister can drain and its teardown finish.
+                                    LOG.debug(e) { "onPartitionsAssigned: '$topic' assignment aborted — subscription shutting down" }
                                 } catch (e: Throwable) {
                                     LOG.error(e) { "onPartitionsAssigned($partitions): failed handling assignment for topic '$topic' ($tps)" }
                                     throw e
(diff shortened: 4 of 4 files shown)


jarohen:
## Approaches tried (and why they didn't land)

This went through a few iterations before the version above — recording the dead ends so they're not re-attempted.

**Misdiagnosis: snapshot socket timeout.** The first hypothesis was the Postgres snapshot read hanging with no socket timeout when the upstream dies. Wrong turn — the thread dump showed the hang is during `ATTACH DATABASE`, *before* Postgres is killed, in Kafka teardown. The failing test's name ("source fails when postgres dies during snapshot") is misleading about where it actually wedges.

**Poison `replicaState` on follower-tail cancellation.** The first *working* fix: the follower's cancellation path set `replicaState = Failed(CancellationException)` so an in-flight `awaitReplicaMsgId` would throw. Green (20/20), but rejected — it fires on *every* follower-tail cancellation, including a benign follower→leader promotion (`oldSys.cancelAndJoin()`, after the await has already returned), poisoning a to-be-discarded follower's state. Harmless in practice (`latestReplicaMsgId` tolerates `Failed`, and the follower is discarded), but fragile, and it contradicts the existing "cancellation is not a processing failure — don't poison" convention two methods up in `FollowerLogProcessor`.

**Re-parent both `onPartitionsAssigned` and `onPartitionsRevoked`.** Wrapping *both* in `withContext(scope.coroutineContext.job)` re-broke teardown, on a *different* test. Revocation also runs as part of the unregister (`processCommand` → `onPartitionsRevoked`), by which point the db job is already cancelled → the `withContext` throws before `cmd.cont.resume(Unit)` runs → the `NonCancellable` unregister continuation is orphaned and teardown hangs (and the escaping `CancellationException` propagates into `pollingJob`, whose `cleanupOnFailure` would evict *every* database on the shared consumer). Confirmed via coroutine dump: `openGroupSubscription` stuck in "Cancelling". Lesson: the revoke path *is* teardown, so it must not be tied to the job teardown cancels — hence the deliberate assign/revoke asymmetry.

**`onPartitionsRevokedSync → onPartitionsRevoked` (suspend `processCommand`) alone.** Considered whether just dropping that `runBlocking` bridge fixes it. No — the wedge is in the *assigned* path (`awaitReplicaMsgId` blocking the poll thread); `onPartitionsRevokedSync` is never even reached in the hang, because the poll thread never returns to the select loop to process commands. A good cleanup (folded in), not the fix.

**Tie the await to the follower's scope inside `awaitReplicaMsgId`.** Considered wrapping just the await in `withContext(followerJob)`. Too deep — it rescues that one await but leaves the preceding replica-log `withTx { NoOp }.await()` exposed to the same detached-`runBlocking` hang. The right boundary is the transport→database handoff (the `Database` listener), where the whole promotion is re-parented at once.

**Landed:** re-parent `onPartitionsAssigned` only, rethrow the cancellation in `LogProcessor` (don't launder it through `notifyError`), and contain it per-subscription in the Kafka rebalance handler.

jarohen:
## Closing to start fresh — full context for the next attempt

Closing this PR: the approach it settled on (re-parenting the promotion into the database job) turned out to be **structurally hazardous for teardown** — three separate teardown regressions, the last one flaky. The bug and the reasoning are sound; the *mechanism* was wrong. Recording everything here so the next attempt starts from knowledge, not scratch.

### The bug (unchanged — still worth fixing)

Node close can hang: `DatabaseCatalog.close` cancel-joins the database's job tree, but the join never completes → `Fault: database catalog did not shut down in time`. Shared-consumer-group test suites cascade one stall into a whole-suite hang.

```
[poll]  Kafka fires onPartitionsAssigned in a detached runBlocking
        (a root coroutine — not in any database's cancel-join tree)
          └─ LogProcessor promotion → awaitReplicaMsgId   ← SUSPENDED, awaiting replica catch-up
[close] dbJob.cancelAndJoin()
          ├─ cancels the db's own coroutines (follower tail, source-log subscription, term)
          └─ does NOT reach the promotion — it's on the detached [poll] coroutine   ← divergence
[poll]  still wedged → never drains the queued Unregister command
[close] source-log openGroupSubscription's NonCancellable unregister → awaits the poll loop forever
```

The core difficulty: the promotion runs on Kafka's shared poll thread, in a `runBlocking` with no parent, so it's outside every database's cancellation tree.

### Approaches tried, and why each failed

1. **Poison `replicaState`** (FollowerLogProcessor's tail-cancel sets `Failed(CancellationException)`; `onPartitionsAssigned` returns null on cancel). **Worked — postgres 20/20.** Moved off it for aesthetics: it poisons a follower's state on *every* tail-cancel, including a benign follower→leader promotion. We established that's harmless (it fires *after* the await, on a to-be-discarded follower; `latestReplicaMsgId` tolerates `Failed`). **Moving off it was the wrong trade — see below.** Never tested against connect-source (that version was based on `origin/main`, which doesn't have it).

2. **Re-parent `onPartitionsAssigned` into `dbJob`** via `withContext(scope.coroutineContext.job)` at the `Database` subscription listener. Looked cleaner (honest structured cancellation). It caused **three** teardown regressions:
   - Wrapping `onPartitionsRevoked` too → the unregister invokes revoke *during teardown* when `dbJob` is already cancelled, so `withContext` threw before `cont.resume`, stranding the `NonCancellable` unregister. (Fixed by not wrapping revoke.)
   - The `processCommand` suspend / `RevokedSync → Revoked` cleanup → an in-flight `Unregister` continuation could be stranded when the poll loop was cancelled mid-`processCommand`. Deterministically failed **4 connect-source tests**. (Fixed by reverting to `onPartitionsRevokedSync`.)
   - Even with both reverted, the re-parent *itself* → a **flaky** teardown hang across connect-source tests (different test each run; always `openGroupSubscription` stuck "Cancelling" at the `NonCancellable` unregister, `DatabaseCatalog.close` > 60s).

   Root flaw: `withContext(dbJob)` makes the promotion a **child of `dbJob` that runs on the shared poll thread**. So `cancelAndJoin()` gains a poll-thread-bound child, while the unregister (another `dbJob` child) *also* needs that poll thread — a teardown race. Benign for postgres; connect-source's timing exposes it.

### Learnings for the fresh start

- **Don't re-parent poll-thread (KafkaCluster) work into a database's job.** It couples `cancelAndJoin` to the shared poll loop and races the unregister. The fix must connect the promotion to the db lifecycle *without* making it a `cancelAndJoin` child on the poll thread — which is exactly what the poison does (observe shared state; no join dependency).
- **Verify against BOTH `PostgresSourceIntegrationTest` and `KafkaConnectSourceIntegrationTest`.** They share the `LogProcessor`/`Database`/`KafkaCluster` machinery; connect-source has a flaky teardown race that postgres doesn't surface. Postgres-only green is not enough.
- **CI-log red herrings.** The `KafkaClusterTest` "listener boom" stacks are the test's own instrumentation (it passes). The `Incorrect: record has no usable key` errors are another test's expected path. The real failures were teardown timeouts (`43 tests completed, 4 failed`; grep `> .* FAILED`).
- **The teardown hang is flaky under load** (many integration tests in one JVM + shared Kafka). A single green run isn't proof; the watchdog signal is `DatabaseCatalog.close exceeded 60s`.

### Recommended next approach

Go back to the **poison** (poison `replicaState` on follower-tail cancel; `onPartitionsAssigned` returns null on `CancellationException`), and verify it against **postgres + connect-source + KafkaCluster** before landing. No poll-thread `dbJob` child → none of the teardown races above.

PR #5774, opened 2026-07-06
refactor(pgwire): answer SHOW from the connection

jarohen:
Behaviour-preserving for pgwire; a fix for ADBC.

## Summary

`SHOW <var>` is now answered entirely by the connection. pgwire treats it like any other query — prepare, then execute the cursor — instead of reimplementing it.

## Context

pgwire reimplemented every state-dependent `SHOW` as a synthetic relational-algebra query: a per-variable RA template, a per-variable column-type table, and a bind-time step that injected the values as query args. The connection meanwhile carried a second, half-built SHOW implementation for its ADBC path — it covered the connection-state variables but not the node-status ones (`latest_completed_txs`, `latest_submitted_msg_ids`, `latest_processed_msg_ids`), which fell through to a session-parameter lookup and answered wrong over ADBC.

The parser already sorts SHOWs: the ones the engine can answer from the query basis (`TIMEZONE`, `SNAPSHOT_TOKEN`, `CLOCK_TIME`) are classified `Query` and were never pgwire's to special-case; the state-dependent ones are `ParsedStatement.ShowVariable`. Those turn out to be answerable entirely from the connection — the node-status readers only need `Database.Catalog`, which the connection already holds, not a node handle.

## What changes

`Connection.prepare(parsed)` is the single entry pgwire uses for queries, `EXECUTE` and `SHOW` alike. A query-like statement plans through the engine; a `ShowVariable` builds a zero-param `ShowPreparedQuery` whose column types are fixed per variable and whose rows snapshot connection/catalog state at open time. pgwire hands over the `ParsedStatement` it already parsed and drives the resulting cursor exactly as it drives a query — so `prep-stmt` and `bind-stmt` collapse the query / execute / show branches into one, and the synthetic-RA path, the per-variable type table and the bind-time injection are deleted.

As a side effect, the node-status SHOWs now answer correctly over ADBC/Flight SQL, not just pgwire.

## Gate exemption

`SHOW` is introspection, not a basis read, so it must stay valid in any transaction and bypass the access-mode gate. Rather than have the gate sniff a concrete prepared-query type or carry a bespoke flag, a `PreparedQuery` now remembers the `ParsedStatement` it was prepared from, and `openQuery` gates on that kind: a `ShowVariable` opens unchecked, everything else resolves as before. The distinction lives with the classification that already draws it.

## What stays pgwire's

Statement dispatch (which wire sequence to run), the wire framing (RowDescription, DataRow, command tags, ParameterStatus), and the SET / tx-control acknowledgements — none of it SHOW-specific.

## Test plan

Full unit suite green (1,599 tests, 0 failures, no reflection / boxed-math warnings). The pgwire, playground (multi-db `SHOW LATEST_COMPLETED_TXS`), types, pg2, `ConnectionTest` and `InProcessAdbcTest` suites exercise the touched paths.

### DIFF
--- a/core/src/main/clojure/xtdb/pgwire.clj
+++ b/core/src/main/clojure/xtdb/pgwire.clj
@@ -5,7 +5,6 @@
             [cognitect.anomalies :as-alias anom]
             [cognitect.transit :as transit]
             [integrant.core :as ig]
-            [xtdb.api :as xt]
             [xtdb.authn :as authn]
             [xtdb.db-catalog :as db]
             [xtdb.error :as err]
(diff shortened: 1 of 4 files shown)


jarohen:
Rebased onto `main` now that #5776 (the typed prepare boundary) has landed — force-pushed. The rework is leaner than the original commit; three things the new base gives us for free or lets us drop:

- **Reuses `Connection.prepareSql(parsed)`.** #5776 already types it, so `prepare(parsed)`'s query/EXECUTE arm is just that — no re-plumbing.
- **Drops `PrepareOpts.parsed`.** The `ParsedStatement` now flows into `prepareQuery`/`prepareRa` directly, so the reify reads it from its `stmt` local to populate `PreparedQuery.parsed` — no smuggling through the opts.
- **Removes the interim `Connection.prepareRa`** that #5776 had added for pgwire's SHOW route. SHOW now prepares through `prepare(parsed)` → `ShowPreparedQuery`, and `Connection.prepareRa` had no other caller. (`IQuerySource.prepareRa` stays — `prepareTxSql`/`preparePatchDocsQuery` use it.)

SHOW behaviour is unchanged — verified against the pre-rebase commit (`898848909`): same per-variable column types, same rows, same `awaitAll`-before-read for the node-status variables. Full `./gradlew test` green (1599 tests, no reflection/boxed-math warnings); code-review pass clean.

Issue #5507, opened 2026-04-24
Batch and pipeline the leader's replica-log writes


jarohen:
## Context

In `LeaderLogProcessor`, every resolved transaction is forwarded to the replica log inside its own Kafka transaction:

```kotlin
private suspend fun appendToReplica(message: ReplicaMessage): Log.MessageMetadata =
    replicaProducer.withTx { tx -> tx.appendMessage(message) }.await()
```

`withTx` opens a Kafka transactional context, writes a single message, commits, and the leader awaits the ack before moving on.
Every path in `processRecords` does this — `Tx`, `LegacyTx`, `AttachDatabase`, `DetachDatabase`, `TriesAdded`, and `finishBlock`'s `BlockBoundary`.

Now that single-writer discipline has landed (#5212), the leader's replica-log hop is the dominant per-tx overhead on the write path.

## Problem

Two costs, both paid per XT transaction:

1. **Commit tax.**
   A Kafka transaction carries fixed overhead — add-partitions-to-txn, end-txn coordinator round-trip, two control records — that amortises away when many messages share a transaction.
   We currently send one.
   At high tx rates, most of the coordinator traffic is waste.

2. **Head-of-line round-trip.**
   The leader awaits each replica commit before processing the next source record.
   While that round-trip is in flight the indexer sits idle; it could already be indexing the next tx.

## Proposed approach

Two knobs, related but independent:

**Batching.**
Accumulate several resolved messages into one `AtomicProducer.Tx` and commit once.
`Tx.appendMessage` already returns a `CompletableDeferred<MessageMetadata>` that resolves on commit, so correlating a per-tx ack with a batched commit is already expressible in the API.

**Pipelining.**
Let `processRecords` keep indexing into the next batch while a previous batch's commit is in flight.
`notifyTx` for a given tx still has to wait for its `CompletableDeferred` (a client observing tx results must not see a tx committed before its replica-log entry lands), but the indexer itself doesn't.

They compose — batching reduces the number of round-trips, pipelining hides the ones that remain — but either alone is worth having.

## Invariants to preserve

- **Block boundary atomicity.**
  A batch MUST NOT carry a `BlockBoundary` partway through — the follower relies on the boundary committing together with the block's contents.
  Simplest: a boundary closes the current batch.
- **Watcher notification ordering.**
  `notifyTx` runs only after its replica-log entry has committed.
  A client that writes-then-reads on a downstream reader must not see the reader lagging past an acked write.
[...]

PR #5783, opened 2026-07-09
feat(indexer): accumulate external-source transactions (#5507)

jarohen:
Part of #5507.

## Summary

The leader's external-source ingest (Postgres mirror, Kafka connect sink) committed one replica-log transaction per CDC tx. This lets ext-source txs accumulate into a shared staging batch and commit together when the resolver goes idle — the same treatment the source-log poll batch already gets (#5782).

## Context

#5507 batches and pipelines the leader's replica-log writes; #5782 landed the source-log side (accumulate a poll batch, one fenced commit). This is the external-source follow-up: the benchmark-agnostic foundation both CDC drivers need before any driver-specific tuning. A colleague is benchmarking FHIR ingest through multiple sources offline, and this is what lets the ext-source path benefit from batching at all.

Postgres-driver pipelining (decoupling the WAL slot-ack from `nextTransaction`) and cross-batch overlap (a committer coroutine resolving ahead of an in-flight commit) are explicitly out of scope. This keeps `drainStaging` synchronous, so the resolver never runs ahead of an in-flight commit — it's unaffected by the demote-off-thread question.

## Changes

1. `refactor(indexer)` — add a dormant idle-drain arm to the resolver's select loop. Behaviour-preserving: the arm never fires while the per-tx drain still stands, so it lands as a pure no-op.

2. `feat(indexer)` — remove the per-tx drain so ext-source txs accumulate; `submitTx` returns a durability handle; `executeTx` = submit + await; the GC drain-before-append barrier.

## Implementation notes

**Drain when idle.** A zero-delay `onTimeout` arm on the resolver's `selectUnbiased`, registered only while staging is non-empty (no busy-spin). A synchronously-ready `onReceive` is selected before the zero timeout dispatches, so it fires only when all channels are momentarily empty. There's no cap or timer — a new commit can't start before the previous one finishes (single fenced producer), and the synchronous drain already waits for that, so the batching window is free.

**Durability handle.** `submitTx` returns as soon as the tx is queued — ordering is guaranteed by the hand-off FIFO and the single resolver, not by staging — and hands back a `Deferred<Unit>` that completes on durability. `executeTx` is exactly `submitTx` plus awaiting it. A fire-and-forget source (the Kafka connect sink) drops the handle; the Postgres mirror awaits it. The handle is signalled after the *whole* drain, including any block cut: signalling mid-loop let a caller race ahead of an in-flight `finishBlock`, which a term teardown could then cancel between the `BlockBoundary` and its `BlockUploaded`, leaving an orphan boundary on the replica log.

The handle must be completed on *every* terminal path a staged tx can take, or an awaiting `executeTx` hangs. The resolver swallows faults and is a supervisor child, so its death doesn't cancel the caller — so the channel's `onUndeliveredElement` and the resolver's teardown (`StagingIndex.failPending`) close the two gaps the old `onComplete`-await path covered but the narrower `durable` handle didn't.

**GC ordering.** A GC `TriesDeleted` appends to the replica log directly, so it now drains first — like the existing control-message barrier — because appending ahead of earlier-staged txs would invert the log order once txs accumulate.

## Out of scope / follow-ups

On a mid-batch drain fault the whole batch's durability handles are failed, including txs already promoted. This is harmless while `executeTx` callers are serial (batch-of-one, as Postgres is today); per-tx granularity is a follow-up for pipelined callers.

Coalescing is emergent — how much a burst batches depends on how much accumulates before the resolver idles, which the capacity-1 hand-off channel bounds. That capacity is the throughput knob the FHIR benchmark will size; it's deliberately left at 1 here.

## Test plan

`ExternalSourceTest` (real `LeaderLogProcessor` + in-memory replica log) covers a lone `executeTx` becoming durable off the idle drain, a burst of `submitTx` accumulating and all committing in send order (awaiting the real durability handles, no blind delay), and `executeTx`/`submitTx` handles throwing on a drain fault rather than hanging. `LogProcessorSimTest` (300 property iterations) and `LeaderLogProcessorTest` guard block-boundary/upload consistency and source-log behaviour. Full `./gradlew test` is green, with no new reflection/boxed-math warnings.

Generated with Claude Code

### DIFF
--- a/core/src/main/kotlin/xtdb/indexer/LeaderLogProcessor.kt
+++ b/core/src/main/kotlin/xtdb/indexer/LeaderLogProcessor.kt
@@ -112,12 +112,22 @@ class LeaderLogProcessor(
     }
 
     private sealed interface ExtSourceTask : PersisterTask {
+        // Completed when the tx is promoted (durable) in drainStaging. submitTx returns this handle;
+        // executeTx awaits it. It must be completed on *every* terminal path a task can take — drained,
+        // faulted, thrown pre-staging, left un-drained at term exit, or dropped undelivered by the channel
+        // — or an awaiting executeTx hangs. onComplete no longer implies durability for ext-source txs.
+        val durable: CompletableDeferred<Unit>
+
         class IndexTx(
             val externalSourceToken: ExternalSourceToken?,
             val systemTime: Instant?,
             val writer: suspend (OpenTx) -> TxResult,
         ) : ExtSourceTask {
+            // Completed at staging, once the writer has run and the outcome is known; executeTx reads it
+            // after `durable`. Distinct from durability — the tx is resolved but not yet on the replica log.
             val result = CompletableDeferred<TxResult>()
+
+            override val durable = CompletableDeferred<Unit>()
 
             override val onComplete = CompletableDeferred<Unit>()
         }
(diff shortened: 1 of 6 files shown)


jarohen:
Superseded by the double-buffer landing (`f8ad8b6ab`, #5507), which took the committer-coroutine offload approach for the source-log path and deliberately kept ext-source synchronous — deferring ext-source coalescing to its own increment on the new `InFlightBatch`/settle machinery. This PR's ext-source rework (accumulate + drain-when-idle + durability handles) was built on the pre-offload model and would be a full rewrite on that machinery, not a rebase. What we learned is captured on #5507. Not merging.

Issue #5781, opened 2026-07-09
perf: reduce per-transaction overhead for small-batch ingestion


jarohen:
Small-batch ingestion carries a lot of per-transaction overhead that doesn't get amortized. The `ingest-tx-overhead` benchmark ingests the same total row count at different batch sizes; the smaller the batch, the worse per-row cost dominates:

| Stage | Duration |
| --- | --- |
| Batch 1000 | 9.79s |
| Batch 100 | 19.01s |
| Batch 10 | 1.7m |
| Batch 1 | 15.4m |

Profiling one-row transactions shows significant time spent constructing `RowCopier`s to append vectors — work that's set up per-transaction rather than reused.

A few avenues to chase, roughly independent:

- [ ] **Ingest pipelining (#5507)** — already in flight; moves per-tx work off the critical path by running it in parallel.
- [ ] **Faster append** — `RowCopier` construction (`core/src/main/kotlin/xtdb/arrow/RowCopier.kt`, appended via `LiveTable`/`LiveIndex`) shows up hot in the profiler; cache or otherwise avoid rebuilding copiers per transaction.
- [ ] **Avoid re-serialising for `importData`** — `LiveIndex.importData`/`LiveTable.importData` take a serialised data structure, but the deserialised format appears to already be available at the call site; pass that through instead of round-tripping.
- [ ] **On-demand snapshots** — `LiveIndex.refreshSnap0` rebuilds the shared `Snapshot` on *every* transaction (through `importTx`, on the leader, follower and transition paths) and on every `nextBlock` — slicing each live table's relation and rebuilding the column-type maps — even when nothing is reading it, which is the common case under bulk ingest or high write throughput. Build the snapshot lazily in `openSnapshot()` under the existing read lock (as the `openTx` path at `LiveIndex.kt:151` already does) instead of eagerly on the write path; writers then take their write lock only to mutate, and the compactor's out-of-band `refreshSnap()` drops out too.

## Progress

**Status**: in-progress

- [ ] Ingest pipelining (#5507)
- [x] ~~Faster append (RowCopier reuse)~~ — explored, no perf gain (copy work relocates, isn't removed); closed as PR #5784. ADBC `bind` tidy (`8c5e1e452`) salvageable.
- [ ] Avoid re-serialising for `importData`

jarohen:
### Chalk — Faster append: direct per-vector append, shrink RowCopier to copyRow

[...]

PR #5784, opened 2026-07-09
refactor(arrow): direct per-vector append, bypassing RowCopier (#5781)

jarohen:
Part of #5781 (the "faster append" avenue).

**Outcome: closing this unmerged — it didn't move the benchmark.** Opening it anyway to record the attempt and why the premise was wrong.

## What it does

Appending an Arrow vector routes through `RowCopier`: `VectorWriter.appendRange` built a fresh copier tree per call (`r.rowCopier(this).copyRange(...)`), used it to copy a contiguous slice, then threw it away. The hypothesis in #5781 was that this per-call *construction* — recursing the vector tree, allocating a closure per node, ~9 objects for a one-row `{_id}` put — was the per-transaction overhead dominating small-batch ingestion.

This adds a direct, copier-free path. `Vector.appendRange0` does the buffer-level copy per type (memcpy / `writeBits`, offset rebasing for the var-width and list types, a whole-vector fast path for dense unions), and `VectorReader.appendRangeTo` is the source-dispatched entry that mirrors `rowCopier` exactly — a `DenseUnionVector` dest absorbs a leg-typed source, a `LegVector` dest routes to `appendRangeFrom`, otherwise a shallow `arrowType` mismatch throws `InvalidCopySourceException` for the caller to promote and retry. `VectorWriter.appendRange` is then repointed onto it. `RowCopier` stays for the genuinely row-at-a-time callers (merge, sort, hash-distinct); `appendRows` (scatter) was left for a follow-up.

## Why it's closing

The `ingest-tx-overhead` benchmark showed no measurable change at any batch size.

The premise was wrong: the expensive part isn't *constructing* the copier, it's the copy work itself — walking the vector tree and moving the bytes. `appendRange0` does that same walk and the same byte-moves; it just does them from a method instead of from a closure the copier built. The work `RowCopier` was doing is relocated into `appendRange`, not removed, so the construction overhead we blamed turns out to be noise next to the copy itself — and, for one-row transactions, next to whatever else dominates the per-tx path outside the append.

No reason to carry a second copy path parallel to `RowCopier` for no gain.

## Salvageable independently

The first commit — `tidy: use RelationWriter.append for ADBC bind copies` — is a standalone equivalence cleanup unrelated to the perf question: it replaces a hand-rolled `rowCopier(...).copyRange(0, n)` with `append` at two `bind` sites. Worth keeping on its own, and can be cherry-picked to `main` separately.

## Test plan

Behaviour-preserving throughout. The direct path is verified equivalent to the copier by a new `AppendRangeTest` (each vector shape copied both ways and compared) and by the full unit suite (1599 tests green). The benchmark run was `ingest-tx-overhead` via `do-ingest-without-pgwire`.

### DIFF
--- a/core/src/main/kotlin/xtdb/api/Xtdb.kt
+++ b/core/src/main/kotlin/xtdb/api/Xtdb.kt
@@ -359,7 +359,7 @@ interface Xtdb : DataSource, AdbcDatabase, AutoCloseable {
                 clearArgs()
                 args = Relation(allocator, root.schema).closeOnCatch { copy ->
                     Relation.fromRoot(allocator, root).use { src ->
-                        src.rowCopier(copy).copyRange(0, src.rowCount)
+                        copy.append(src)
                     }
                     copy
                 }
@@ -918,7 +918,7 @@ interface Xtdb : DataSource, AdbcDatabase, AutoCloseable {
                     this.docs?.close()
                     docs = Relation(allocator, root.schema).closeOnCatch { copy ->
                         Relation.fromRoot(allocator, root).use { src ->
-                            src.rowCopier(copy).copyRange(0, src.rowCount)
+                            copy.append(src)
                         }
                         copy
                     }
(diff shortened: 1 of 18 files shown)


jarohen:
Closing as documented above — the direct append path is behaviour-equivalent but the copy work simply relocates from `RowCopier` into `appendRange`, so `ingest-tx-overhead` is unchanged. The ADBC `bind` tidy (first commit) is worth keeping independently.

PR #5805, opened 2026-07-17
Model the pg_catalog surface, resolve internal-schema columns strictly

### LINKED ISSUE #5804: SQL: bare column across a join of two internal-schema tables reported as ambiguous
## Summary

A bare (unqualified) column reference across a join of two internal-schema tables (`pg_catalog`, `information_schema`, `xt`) is rejected as ambiguous, even when the column exists on only one of the joined tables.

## Symptoms

Metabase's enum-type discovery probe fails to plan against XTDB:

```sql
SELECT nspname, typname
FROM pg_type t JOIN pg_namespace n ON n.oid = t.typnamespace
WHERE t.oid IN (SELECT DISTINCT enumtypid FROM pg_enum e)
```

```
xtdb.api.error.Incorrect: Errors planning SQL statement:
  - Ambiguous column reference: nspname
  - Ambiguous column reference: typname
```

`nspname` exists only on `pg_namespace`; `typname` only on `pg_type` — so neither reference is genuinely ambiguous. Qualifying them (`n.nspname`, `t.typname`) plans without error, so tools that emit unqualified names trip over this while others don't. The `pg_enum` subquery is incidental — the bare join on its own reproduces it.

Origin: `metabase/src/metabase/driver/postgres.clj`, the enum-type discovery query run during schema sync.

## Root cause

The column-level tolerance introduced in #4467 keys on the table's schema, not on whether the column exists: a reference against any `xt` / `pg_catalog` / `information_schema` table resolves (to null) rather than erroring, so tools can probe optional catalog columns XTDB doesn't model. Because that tolerance accepts *any* name, every internal-schema table effectively matches every column. In a join of two such tables, a bare column that is real on one side is therefore also matched — spuriously — on the other, and the two matches read as ambiguous.

## Scope

Any join of two internal-schema tables that selects a bare column defined on only one side. User tables are unaffected — an undeclared column there remains a hard error (#4467).

## Progress

**Status**: in-progress — Part 1 (seed) committed & ready to ship; Part 2 (strict internal-schema resolution) verified, CI green, PR #5808 open for review. Merge order: land Part 1, rebase Part 2, merge #5808.

- [x] Reproduce + root-cause
- [x] Model pg_catalog / information_schema surface + empty probe tables (landed on main)
- [x] Migrate pg_user.username → usename (landed on main)
- [x] Part 1: seed xt.txs / xt.role_membership — validated, reviewed, drift-guard test added (ready to ship)
- [x] Part 2: strict internal-schema resolution — column + table tolerance removed; warning tests repointed; postgres-source poll tolerance (#5733); Metabase/Grafana smoke tests. CI green (PR #5808).
- [ ] Land Part 1, rebase Part 2, merge PR #5808

### Open Questions

- [ ] Follow-up issue for deferred index-introspection (array slicing, `_pg_expandarray`, `pg_has_role`, 2-arg `current_setting`)

jarohen:
Resolves #5804.

## Summary

Bare (unqualified) column references across a join of two internal-schema tables were falsely reported ambiguous — Metabase's enum probe `SELECT nspname, typname FROM pg_type t JOIN pg_namespace n …` failed to plan. This models the full Postgres catalog surface those tools probe, then removes the tolerance that caused the false positive, so internal-schema columns resolve exactly like user-table columns.

## Context

The #4467 column tolerance let `BaseTable/-find-cols` resolve *any* column name on an `xt` / `pg_catalog` / `information_schema` table, so tools could probe catalog columns we hadn't modelled. That made every internal-schema table match every name — so in a join, a bare column real on one side was also matched on the other, and the two matches read as ambiguous. Qualifying the columns worked, so tools emitting unqualified names hit this while others didn't.

We surveyed the Metabase and Grafana Postgres drivers for the catalog columns and tables they actually probe. Both select unmodelled columns on tables we *do* model (`information_schema.columns.is_generated`, `pg_user.usename`), so narrowing the tolerance to unknown-tables-only would break them — the surface has to be modelled.

## Approach

Four commits, tidy-first:

1. `fix(serde):` — `->type` handles `:?`-wrapped composite specs (`[:? :list :i32]`), needed for nullable array columns like `pg_constraint.conkey`. Previously a nullable composite fell through to `->arrow-type`, which can't carry a list's element type.
2. `feat(sql)!:` — completes `pg_type` / `pg_class` / `pg_attribute` / `information_schema.columns` to their real-Postgres column sets, and adds the empty tables tools probe (`pg_enum`, `pg_constraint`, `pg_extension`, `information_schema.{table_constraints, key_column_usage, constraint_column_usage}`). Unpopulated columns declared `[:? …]` read as null via `Relation.endRows` null-padding, so the generators are unchanged. Renames `pg_user.username` → `usename`.
3. `fix(sql):` — deletes the column tolerance. Internal-schema columns now resolve strictly; an undeclared one is a hard error. The table-level unknown-table warn stays, so a typo'd table warns while a typo'd column errors.
4. `test(sql):` — Metabase/Grafana introspection smoke tests.

## Decision rationale

Model-and-delete, over ranking real matches above tolerated ones: the ranking approach is a safety layer over a tolerance that shouldn't fire on modelled tables at all. Since drivers probe unmodelled columns on modelled tables, the surface must be modelled regardless — and once it is, the tolerance is dead weight, so deleting it is one mechanism instead of two. Declared column types were verified against a real Postgres 16 catalog.

## Breaking change

`pg_user.username` → `usename`. Postgres names that column `usename`; the old name meant a driver's `WHERE usename = ?` probe silently matched nothing. Anyone querying `pg_user.username` must switch to `usename`.

## Out of scope

Index-introspection queries that need array slicing (`indkey[0:n]`), `information_schema._pg_expandarray`, `pg_has_role`, or 2-arg `current_setting` — these cluster in the drivers' index sync and need ANTLR grammar work. Deferred to a follow-up. Grafana's xorm backing-store queries target Grafana's own database, not XTDB, so they're excluded too.

## Test plan

Green across `information_schema`, `sql`, `pgwire`, `pgwire.metabase`, `pgwire.grafana`, `metrics`, `node`, `pgwire.pg2`, `pgwire.dbeaver`, `pgwire.playground`, `create_table`, and `types` (~464 tests). Full suite runs in CI.

Deliberate behaviour change to confirm in review: an undeclared column on an internal-schema table now errors instead of returning a silent null — the accepted whack-a-mole tradeoff for retiring the tolerance.

### DIFF
--- a/api/src/main/clojure/xtdb/serde/types.clj
+++ b/api/src/main/clojure/xtdb/serde/types.clj
@@ -1,5 +1,6 @@
 (ns xtdb.serde.types
-  (:require [clojure.pprint :as pp])
+  (:require [clojure.pprint :as pp]
+            [xtdb.error :as err])
   (:import (java.io Writer)
            (org.apache.arrow.vector.types DateUnit FloatingPointPrecision IntervalUnit TimeUnit Types$MinorType UnionMode)
            (org.apache.arrow.vector.types.pojo ArrowType ArrowType$Binary ArrowType$Bool ArrowType$Date ArrowType$Decimal ArrowType$Duration ArrowType$FixedSizeBinary ArrowType$FixedSizeList ArrowType$FloatingPoint ArrowType$Int ArrowType$Interval ArrowType$List ArrowType$Map ArrowType$Null ArrowType$Struct ArrowType$Time ArrowType$Time ArrowType$Timestamp ArrowType$Union ArrowType$Utf8 Field FieldType Schema)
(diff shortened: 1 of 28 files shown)


jarohen:
Superseded — this branch's work has been split into smaller, independently-reviewable pieces:

- the `pg_catalog` surface modelling and the `pg_user.username` → `usename` rename (with the serde `->type` fix) landed on `main` directly;
- seeding the data-backed system tables (`xt.txs`, `xt.role_membership`) so they resolve on a fresh node ships as its own change;
- the strict internal-schema resolution that actually fixes #5804 is now #5808.

Closing in favour of #5808.

Issue #5838, opened 2026-07-31
export-snapshot! is silently partition-0-only for a multi-partition database


mbutlerw:
## Summary

`export-snapshot!` assembles the minimal durable file set for a database and copies it into a fresh single-partition store. It reads the buffer pool, block catalog and trie catalog off partition 0, so for a multi-partition database it would emit a snapshot of partition 0 alone — incomplete rather than corrupt, since it never sees the other partitions, and silent either way.

This has to be settled before the multi-partition gate lifts, because the failure mode is a successful-looking export.

## Current state

The export opens a deliberately inert node — indexer disabled, no compactor threads, no server, no healthz — so the durable file set can't move while it reads. It asks the block catalog for the latest completed block index, then assembles the block file, each table's block-metadata file, and the meta and data files of every trie the trie catalog reports as live or nascent. Each file is copied through the buffer pool's copy operation, which resolves both source and destination against the pool root, so the snapshot lands inside the store it came from, arranged the way a fresh node expects to find it.

## Invariants

**The output root is deliberately unmarked, and must stay that way.** An exported snapshot is a fresh single-partition store, so it builds its root without a partition marker — that is correct, not an oversight. A mechanical sweep converting root construction to the partition-aware form would silently re-key every export under a partition subtree and break restore. The reason needs to sit at the call site, not only in this issue.

## Open questions

- [ ] Which shape does a multi-partition export take? Three candidates, none obviously right:
  - **Per-partition subtrees**, restored into a node with the same partition count. Preserves the copy-only property, but the snapshot is then only usable at a matching partition count.
  - **Merge every partition into one single-partition snapshot.** The most useful output, but it means renumbering block indices and trie keys — a genuine data operation rather than a copy, with a different risk profile from everything else this tool does.
  - **Refuse for partitioned databases**, and offer per-partition export explicitly. Cheapest and loudest; pushes the composition problem to the operator.

Related: #4725 (extend the export to extract only subparts of a database) and #4857 (export a table's snapshot to a Debezium-format topic). If the third candidate wins, #4725 is the natural implementation home for per-partition selection.

Ship/Show/Ask: **Ask** for the decision, then Show.

## Progress

**Status**: not-started

- [ ] Decide the multi-partition export shape
- [ ] Record at the call site why the export's storage root is unmarked
- [ ] Implement the chosen shape, or reject partitioned databases with a clear error

Issue #5837, opened 2026-07-31
Multi-DB stage 4 unit 9: enable multi-partition


mbutlerw:
## Summary

The switch. `Database.open` currently throws `xtdb/multi-partition-not-yet-enabled` for any database configured with more than one partition; this unit lifts that and constructs N partitions for real.

## Current state

The gate was planted in unit 5 so the type system could widen ahead of the wiring. `Database.open` hardcodes a single partition at index 0, and `PartitionStorage` carries a note that single-partition is the only reachable shape until the gate lifts.

Most of the plumbing is already in place. The leader processor reads its partition from its storage handle, and the follower and consumer paths tail and append through a partition-bound log view, so the index flows through automatically once the construction loop lands. There are no leader or follower internals to touch here.

## Future state

- **Drop the gate** and construct N partitions in a loop, each with its own storage handle carrying its partition index.
- **Reject more than one partition for XT-format databases.** The primary database and XT-format secondaries stay single-partition through stage 4: there's no upstream topic to validate against, and the per-partition writer and replica-log machinery doesn't apply.
- **Enforce the source's declared partitioning capability.** Without it the gate lifts with a single-partition-only source silently accepted at four partitions, failing later and obscurely. Tracked as its own sub-issue.
- **Strip the defaulted partition argument from the log interface.** Unit 5 kept `partition = 0` defaults on the offset, message and append operations so existing single-partition callers didn't have to change. Once multi-partition is live, defaulting to partition 0 is the hack-value pattern the project's guidance warns against: a caller writing the one-arg form might mean "I intended partition 0" or might have forgotten to think about partitions, and the compiler can't tell. Call sites to bump span the Clojure interop, `Database`, the database catalog, tests, and the follower and leader processors.
- **Remove the internal gate-bypass opener** added for end-to-end testing ahead of the flip. The existing end-to-end tests re-target the public API.

## Invariants

**A single-partition database must upgrade with no operator intervention.** That is the requirement — not byte-identity of every persisted artefact. Unit 5 already emits a new proto field unconditionally for every attach, shifting the golden block file, and that's harmless because the reader coerces the absent field back to a single partition. Code-side defaulting is a legitimate tool; what must not happen is an upgrade that needs a manual migration step.

The one place where identity really is load-bearing is the Kafka transactional id, because fencing is keyed on the id string — covered in the Kafka partition-awareness sub-issue.

[...]

Issue #5833, opened 2026-07-30
tidy(storage): cache keys omit the storage root, and the namespacing is duplicated across both BufferPools


mbutlerw:
The node-shared memory and disk caches namespace their entries by `dbName/<partition>`, which omits the storage root that every object key actually lives under. The derivation is also duplicated verbatim across the two `BufferPool` implementations, and only one of the two copies is test-covered.

## Context

The caches became node-scoped as part of multi-database support — `memoryCache` and `diskCache` were lifted out of the per-database local/remote storage config, so a single cache instance now holds entries for every database on the node. `4c36a34c9` added the `dbName/<partition>` namespacing to keep them from colliding, writing it into both pool implementations in one commit. #5557 unit 6 (#5815) then changed the placeholder `0` to a real partition index — again in both copies. Every change to this concept so far has had to touch both sites.

## Current state

The derivation is byte-identical in the two pools:

```kotlin
// core/src/main/kotlin/xtdb/storage/LocalStorage.kt:51
// core/src/main/kotlin/xtdb/storage/RemoteBufferPool.kt:60
private val cacheRootPath = dbName.asPath.resolve(partition.toString())
```

Object keys, by contrast, are resolved under a storage root: `v06` today, `v06_e02` after an epoch bump, `parts/1/v06_e02` at N>1. None of that reaches the cache key.

Because the cache hands the full namespaced key to its loader, each pool then has to undo its own namespacing to get back to an object key — `cacheRootPath.relativize(...)` at `LocalStorage.kt:69` and `RemoteBufferPool.kt:120`, `:129` and `:148`.

## Why it matters

**Duplication that drifts.** Two copies of one invariant, edited in lockstep twice, and only the remote one is pinned by a test. `RemoteStorageTest.partitions get their own entries in the node-shared caches` covers it there; the local copy can't be covered the same way, because `MemoryCache` releases an entry once the last reference drops, so a sibling partition's read is a fresh miss that reloads from its own root. Change one copy and forget the other and remote fails loudly while local regresses in silence.

**Possible staleness across a storage root change.** `DiskCache` is durable across restarts and its keys don't move when the storage root does, so a read of a given pool-relative key could in principle be served an entry written under a previous version or epoch. This needs confirming rather than assuming — see Open questions. `MemoryCache` is immune by construction: it dies with the process, so it can only ever hold entries from the current root. Any exposure here is disk-cache-only.

## Future state

[...]

Issue #5557, opened 2026-05-05
Multi-DB Stage 4: external source logs (multi-partition)


jarohen:
## Overview

Stage 3 (#5330) put XTDB on top of a single-partition external source log.
Stage 4 lifts the single-partition restriction: one database can ingest from a multi-partition Kafka topic, with partitions processed concurrently by separate leaders.
The single-leader-per-database architecture from stage 2 (#5188) generalises naturally to one-leader-per-partition.

## Motivation

External source logs (Debezium, Kafka Connect, …) are typically multi-partition in the Kafka world — that's how operators scale ingestion throughput.
As long as XT can only attach to a single-partition topic, users either repartition upstream (often impractical, sometimes impossible) or accept the throughput ceiling of a single XT leader.
Both are friction we can remove.

## Configuration

A new top-level `partitions` field on `ATTACH DATABASE` declares how many partitions a database has.

- Default: `1` — existing behaviour, no migration for current users.
- Set at attach time, immutable thereafter.
- Database-level, not external-log-level — sits alongside the existing log/source config rather than nested under it.
- Validated at attach against the source's actual partition count; a mismatch fails the attach rather than being silently accepted. The count has to live somewhere durable regardless — the basis-token wire format errors on a size mismatch, so the partition count is a stable property of the database.

If the upstream topic is later altered to change its partition count, the database stops working until it is re-attached with the new count. Repartitioning is out of scope (see Scope), so that is the intended behaviour for now.

## Architecture

What was called `Database` is now `DatabasePartition`. A `Database` aggregate wraps a set of partitions and is what the rest of the system (pgwire, info schema, query engine, healthz) addresses.

Each `DatabasePartition` owns:

- Its own leader election.
  Kafka's consumer group with the cooperative-sticky assignor already distributes partitions across nodes, so no new election mechanism is needed.
- Its own `LogProcessor`, follower processors, and replica-log subscription.
- Its own block / compaction / GC cadence — partitions are entirely independent for these purposes.
- Its own `Watchers` instance — one per `(database, partition)`, not per database.
- Its own subtree of the object store.

The remaining single-partition validator is `KafkaCluster`'s topic-partition-count check, which becomes a mismatch check against the configured count — see #5841.

## Consistency

Within a partition, XT's existing serialisability guarantees apply unchanged.

Between partitions there is no defined order.
Events on different partitions are independent, and any guarantee that would require cross-partition coordination is not offered.
[...]

PR #5848, opened 2026-08-04
feat(query): partition-aware read path (#5835)

### LINKED ISSUE #5835: Multi-DB stage 4 unit 7: partition-aware read path
## Summary

Every partition-scoped read still resolves through partition 0. This unit makes the read path walk partitions: a database-level `TableCatalog` view that merges across them, a UNION-at-scan emit so each branch reads its own partition's snapshot, per-branch temporal bounds, and an emit-cache key that accounts for every partition's block index.

Single-partition behaviour is preserved exactly — at one partition the wrapper is a passthrough and the scan emits one cursor, as today.

## Current state

`Database` exposes a `partition0` delegate and routes eleven getters through it — `storage`, `queryState`, `watchers`, `compactorOrNull`, `blockCatalog`, `tableCatalog`, `trieCatalog`, `liveIndex`, `bufferPool`, `metadataManager`, `compactor`. All of them are genuinely per-partition values, so each caller has to learn *which* partition it means.

On the Clojure side, `scan.clj`'s `scan-vec-types` and `->cursor` carry `TODO (#5557 unit 7)` markers where they take the only partition's snapshot. The temporal-bounds setup in the same file indexes the basis at slot 0 with no marker at all, so at more than one partition every branch would apply partition 0's bound — silently wrong rows rather than an error. `Database.getColumnTypes` has the same single-partition read for ADBC and Flight SQL schema introspection.

## Future state

Two narrow aggregation points, established by walking the call sites:

- **`Database.tableCatalog`** — a database-level `TableCatalog` view wrapping three methods. `getType` reduces the type-merge across partitions; `getTypes` merges the per-column maps the same way; `rowCount` sums. Read-only: writes go directly to each partition's own catalog via that partition's writer.
- **`DatabaseSnapshot.tableInfo()`** — landed with unit 2; the merged schema view the planner consumes.

Everything else stays per-partition. `TrieCatalog` reads route through the snapshot's trie-table state at scan runtime, so each UNION-at-scan branch reads its own partition's tries and no wrapper earns its keep. `BlockCatalog`'s only database-level read is the current block index for emit-cache invalidation, which becomes a tuple over partitions rather than a merge.

The type-merge is associative, so reducing across partitions is well-defined regardless of order.

## The #5855 dependency (landed)

The cross-partition type merge is the *same* join widened from 2 sources to 2N — N historical catalogs and N live snapshots. #5855 establishes that the two-way version underneath it is broken at p=1: the historical⊔live boundary is crossed by four different operators, only one of which is a lattice join, and `Database.getColumnTypes` in particular uses Kotlin's `Map +`, which discards a leg rather than joining.

This unit's review pass found two instances of that family — `DatabaseTableCatalog.getType` unable to distinguish "no table here" from "no such column here", and the redundant `size == 1` fast paths — and fixed them *in the new wrapper*. It did not generalise to the p=1 sites underneath, which is what #5855 now covers.

Two consequences. **#5855 landed first**, via PR #5916, and the two type-related items in Progress below were implemented against a foundation now known to have been broken, so they need revalidating rather than treating as done. And **"single-partition behaviour is preserved exactly"** should be read narrowly: at p=1 the behaviour that was being preserved was itself wrong, and the wrapper's fix didn't reach the path it wraps.

## Notes for the cross-partition type fold

Written up while landing #5855, because two of these are traps that would rebuild that bug one level up.

**The fold belongs on `DatabaseSnapshot`, and scan stops knowing partitions exist.** `snaps` is already `{db-name → DatabaseSnapshot}`, so `scan-vec-types` swaps its receiver and deletes the `(first (.getPartitions db-snap))` line this issue's TODO sits on. `Database.getColumnTypes` does the same. Neither needs anything else.

**Push the column set down to every partition. Do not merge their enumerated maps.** This is the trap. A partition that has rows but lacks the column must contribute `Null`, and it can only do that if it is *asked* about the column — an enumerated map simply has no entry to carry that fact:

```
partition 0:  rows for t, no column c    →  enumerated map has no `c`
partition 1:  c: I64

merge the maps  →  c: I64          wrong — partition 0's rows read null for c
correct         →  Null ⊔ I64 = Maybe(I64)
```

That is exactly the `Map +` defect #5855 fixes, at the database level. It is why `Snapshot.columnTypes` takes a column set rather than returning what it finds, and why that arity has to stay on `Snapshot` even once the public entry point moves up.

**Enumeration moves up; typing does not.** "Which columns does this table have" becomes the union of the partitions' `tableInfo`s, so the enumerating arity and `allColumnTypes` belong on `DatabaseSnapshot`. The per-partition `columnTypes(table, cols)` stays as the fold's input.

**Partitions contributing `Nothing` is the mechanism, not a problem.** A table living only in partition 3 folds as `Nothing ⊔ Nothing ⊔ Nothing ⊔ I64 = I64` — the empty partitions are absorbed and the answer is partition 3's type. Contributing `Null` instead would make every such column spuriously nullable.

**But carry the all-`Nothing` guard.** The one case the lattice gets wrong today is a merge where *every* input is `Nothing` — a table empty across the whole database — which `mergeTypes` collapses to `Null` (#5871). The per-partition join already guards this; the fold needs the same, or an empty table's columns flip from `Nothing` to `Null` the moment they are folded. At p=1 that is the difference between this being an equivalence change and a behaviour change, and `ColumnTypesTest`'s lattice-bottom case is the canary.

**Memoise per `(DatabaseSnapshot, table)`.** The N-way fold per column is the first point at which this work is non-trivial — at p=1 it is two map lookups and a cached merge, which is why nothing memoises it today. Worth knowing that `Database.openSnapshot` builds a fresh `DatabaseSnapshot` per call around possibly-shared `Snapshot`s, so a cache there is per-*query*, not per-snapshot like `Snapshot.allColumnTypes`' `by lazy`. It still pays for itself within one query — several columns of one table fold once — but do not expect it to behave like the existing cache.

**Row-count stats want the same fold, but must not reach the emit-cache key.** `emitScan` reads `(.rowCount table-catalog table)` off the live `@Volatile` for `:stats {:row-count …}`. After #5855 a scan's *types* come from the pinned `TableCatalog.Snap` while its *stats* still come from a live read, so the two resolve from different vintages. Folding the row count across partitions is a sum rather than a join, and it is this unit's work regardless — which makes it the natural moment to pin it too. `Snap.rowCount(table)` already exists and is currently unused; it is the landing spot.

The trap is the emit cache. `scan-vec-types` is part of its key, and row counts must **not** join it — they change on essentially every transaction, so keying on them would miss the cache and re-run codegen for every query. Stats freshness is deliberately coarse instead, invalidated via `:last-known-blocks`. So a pinned row count has to arrive as an emit input that is explicitly excluded from the key, and there is no snapshot in scope at emit time — `emitScan` takes `(db-cat, scan-expr, scan-vec-types, param-types)` — so it needs threading through `query.clj` the same way `scan-vec-types` is. It cannot instead move to cursor time, where `snaps` *is* available, because `:stats` is part of the emitted plan the optimiser consumes.

Worth calibrating the payoff before spending much on it: emit only runs on a cache miss, so today's live read is already "as of the last miss", and a pinned read would carry identical staleness. The gain is vintage consistency and removing the last live-catalog read from the query path — not accuracy, since these are cardinality estimates.

## Invariants

**The `partition0` delegate is not this unit's exit criterion.** Two of its getters serve consumers whose multi-partition semantics are open questions of their own — the snapshot export and the healthz checks — so the delegate survives until those land. Its removal is tracked as an exit criterion on the parent.

**The Clojure interop callers are not compile-enforced.** Kotlin callers fail the build when a getter goes; Clojure callers reach them through interop (`.getBufferPool`, `.getBlockCatalog`, `.getTrieCatalog`, `.getLiveIndex`, `.getQueryState`) and surface as a runtime `NoSuchMethodError` only if a test walks that path. Grepping the delegate name finds the getters and none of the callers, so the sweep has to be done by grepping the `.get*` interop forms across `core/src/main/clojure`. An enumerated file list would be both incomplete and stale.

**info_schema aggregation does not fall out for free.** The synthetic-table cursor currently reads a partition's catalogs directly. It has to be routed through the database-level wrapper for the "aggregation is free" property to hold.

**Concatenating branches of differing types is safe, and this is why.** The declared type is an advertising contract, not an execution input. `->expression-projection-spec` runs the analyser over the *declared* types once at emit purely to obtain `widest-out-type` for `getType`, then discards the generated code; at execution `project` re-derives types from the relation actually in hand, re-emits against those, and allocates the output vector at that batch's return type (`emit-projection` is `lru-memoize`d, so one codegen per distinct signature). Generated code is therefore keyed on actual data and never consults the declaration — which is why `UnionAllCursor` already passes heterogeneous batches straight through with no promotion, and why partition branches emitting different types need none either. The obligation this leaves is only that `widest-out-type` covers every per-batch type.

## Shape refinements, identified while extracting the groundwork

Two changes to `->partition-scan-cursor`'s signature that belong here rather than in #5853, because neither earns its keep at one partition.

**`ScanMetrics` becomes per-partition, merged by the wrapper.** It currently sits in the per-query bundle, shared across branches so the operator reports one aggregate set of counters. That's the right *presentation*, but it makes the shared object the last piece of mutable state coupling the branches — five non-atomic `+=` that would silently lose counts the moment anything drains or builds them concurrently. Give each branch its own and have `PartitionedScanCursor` sum them for `cursorAttributes`: same numbers reported, branches independent, and whether EXPLAIN ANALYZE shows per-partition detail becomes a rendering choice rather than something fixed by where the object is constructed.

**The three per-partition arguments become one read context.** `storage`, `snapshot` and `temporal-bounds` always travel together and are always indexed by the same slot. Passed as three positionals the caller can mis-zip them; passed as one value it can't — which is the positional invariant this unit rests on, expressed as a type instead of as a check.

## Out of scope

The database-level blocking await, sync and flush surface, and the snapshot export — both are separate sub-issues, and both keep the `partition0` delegate alive until they land.

**Duplicate `_id`s across partitions surface as duplicate rows.** Concatenation gives up the "one resolved version per `_iid`" property a single-branch `MergePlanner` had structurally, and nothing in the read path checks the routing contract. Not a new gap — #5557 states the `_id`-stable partitioner as an explicit customer contract and records duplicate rows as the observable consequence of violating it. Enforcing it would need a cross-partition iid merge, which is the coupling the throughput scaling exists to avoid. Goes live with #5837.

**All N branch cursors are built eagerly and held to query end.** `MergePlanner/planSync` runs per branch at construction and `PartitionedScanCursor` only releases on its own close, so peak allocator usage is N× a single-partition scan's for the whole query rather than one partition at a time. Closing each branch as it drains is the obvious mitigation. #5557 defers per-node resource work as measure-then-optimise, so this is flagged rather than done — but it wants deciding before the gate lifts rather than after someone hits an OOM. Goes live with #5837.

**`xt.trie_stats` and `xt.live_columns` want a `partition` column.** A trie key is only unique within its partition, so above one partition those rows can't be told apart. Currently unioned across partitions with a marker at the call site — union over partition-0-only because union is at least complete. Left to the observability unit, #5849.

Ship/Show/Ask: **Show**.

## Progress

**Status**: not-started against current `main`, and blocked on #5947. Both earlier blockers have cleared — #5855 landed via PR #5916, and the groundwork landed via PR #5924 (which superseded the parked #5853). The implementation on PR #5848 is now source material rather than a base to rebase: `main` has moved ~200 commits, #4037 has merged the block and table catalogs into one value, one of its two commits is redundant against merged `3002a02ae`, and its test fixture — two attached databases relabelled as two partitions — is being replaced outright by #5947. The two type-merge items below sit on the two-way join #5855 has since fixed, so they need revalidating rather than treating as done.

- [x] `Database.tableCatalog` database-level view (`getType`, `getTypes`, `rowCount`, plus the whole-catalog `types` info_schema reads)
- [x] UNION-at-scan emit
- [ ] `scan-vec-types` reduces per-partition live types — **revalidate after #5855**
- [x] Per-branch temporal bounds — the hardcoded basis slot is gone
- [ ] `Database.getColumnTypes` cross-partition merge — **revalidate after #5855**; the two-way merge it builds on uses `Map +` rather than a join
- [x] Emit-cache key carries every partition's current block index
- [x] Route the info_schema cursor through the database-level wrapper
- [x] Clojure interop sweep — grep the `.get*` forms across `core/src/main/clojure`
- [x] Decide the UNION-operator-versus-scan-internal question

The concurrent-write non-collision assertion carried over from unit 6 was dropped rather than landed: partition isolation in the buffer pool is path construction, so the assertion held by construction and the threads gave it no interleaving to observe. `partitionedPoolsAreIsolated` already covers the same properties sequentially and more precisely.

- [ ] `ScanMetrics` per partition, summed by the wrapper
- [ ] Bundle storage + snapshot + temporal bound into one per-partition read context

### Open Questions

- [x] UNION operator versus partition-aware scan internal — **scan internal**. Partitions are a physical property of the database, and the scan already hides multiple physical segments behind one operator; emitting them into the plan would change EXPLAIN and cardinality estimation for a distinction the reader can't act on, and would need a special case at one partition to preserve today's output. Separately, concatenating branches of differing types needs no promotion layer — see the invariant above on why the declared type is not an execution input.

mbutlerw:
Part of #5557. Resolves #5835.

## Context

Stage 4 unit 7. Every partition-scoped read still resolved through partition 0: the scan took `first(snapshot.partitions)` and read that partition's storage, `scan-vec-types` and `Database.getColumnTypes` took the same slot, and info_schema read its catalogs off `queryState`.

Most of those were *incomplete* above one partition. One was *wrong*: the temporal-bounds lookup in `scan.clj` indexed the basis at a hardcoded `0`, so at more than one partition every branch would apply partition 0's bound and the scan would return the wrong rows rather than fail. That site is the reason this unit exists.

Single-partition behaviour is preserved exactly — at one partition the database-level catalog view is a passthrough, and the scan emits the same single cursor it does today.

## Changes

1. `tidy(database): index Database.partitions by list position` — `Map<Int, DatabasePartition>` becomes a `List` whose position *is* the partition index. Behaviour-preserving.
2. `feat(query): partition-aware read path` — the substance.

## The aggregation points

`IQuerySource.QueryDatabase` no longer exposes a `PartitionStorage`/`PartitionState` pair. It exposes `partitions: List<QueryPartition>` in index order, plus a `DatabaseTableCatalog`.

`DatabaseTableCatalog` is a read-only database-level view over the per-partition `TableCatalog`s. `getType` and `getTypes` reduce the type merge across partitions, `rowCount` sums, and `null` keeps meaning "no partition has this table". Writes are unrepresentable there on purpose: each partition's indexer is the sole writer to its own catalog, so there's no database-level write to express. `DatabaseSnapshot` grew the matching merges over its per-partition `Snapshot`s.

Everything else stays per-partition, as the issue's call-site walk predicted. `TrieCatalog` reads route through the snapshot's trie-table state at scan runtime, so each branch reads its own partition's tries and no wrapper earns its keep. `BlockCatalog`'s only database-level read is the emit cache's `last-known-blocks`, which becomes a tuple over partitions rather than a merge — a block landing in any partition changes what a scan's branches and stats see.

## The positional invariant

Slot `i` of the partition list, slot `i` of the `DatabaseSnapshot`, and slot `i` of the basis vector are the same partition, and the scan zips all three.

That's stated on `QueryDatabase`, asserted in `Database`'s constructor (each partition's storage handle must agree with its list position), and checked in the scan before it dispatches on table kind. The check isn't belt-and-braces: a short snapshot list makes `map vector` truncate and drop a partition's rows, and a short basis vector makes `get-in` return nil and silently unpin that partition's temporal bound — both the exact failure mode this unit is here to remove. It sits above the `cond` rather than in the `:else` arm because the info_schema cursor reads both lists too.

The single-column merges derive from their map-level siblings — `getType` from `getTypes`, `columnType` from `columnTypes`. Reducing each partition's own `getType` can't tell "this partition doesn't have the table" from "this partition has the table but not this column", and the second has to widen the merged type to nullable rather than drop out of the merge. Deriving makes them consistent by construction, and the reductions are shared with `mergeVecTypes` so the historical and live sides can't drift — `scan-vec-types` reads them side by side.

The `Map` → `List` tidy in commit 1 exists to state that invariant in the type rather than re-deriving it with `entries.sortedBy { it.key }` at each of the five call sites that needed partition order.

## Decision: scan-internal fan-out, not a UNION operator

The open question on the issue. The catalog story is identical either way; the choice is whether the planner sees N branches.

Scan-internal, for three reasons:

- **Partitions are physical, not logical.** The scan already hides multiple physical segments — a live index plus N trie files — behind one operator. Partition fan-out is the same kind of thing. Emitting it into the plan leaks the storage layout into the logical plan for a distinction the reader can't act on.
- **EXPLAIN and costing stay put.** A real UNION would change EXPLAIN output and give the planner N row-count estimates to cost, which can reorder joins. That's a behaviour change for single-partition users the moment the shape generalises.
- **No special case at N=1.** A UNION emitter would need "don't wrap when there's one branch" to preserve today's output. Here the wrapper is simply elided.

So the branches share one `ScanMetrics` and present as a single `scan` cursor — a partitioned scan reads in EXPLAIN ANALYZE exactly as an unpartitioned one does.

`PartitionedScanCursor` concatenates rather than merges. Partitions carry disjoint `_id`s under the upstream routing contract, so there's no version of a row to resolve across them and no order key to merge on; each branch has already resolved its own bitemporal versions.

## info_schema

The issue flagged that info_schema aggregation doesn't fall out for free, and it didn't. The synthetic-table cursor now takes the `DatabaseSnapshot` and reads its catalogs through the database-level view, so `information_schema.columns`, `pg_class` and friends aggregate for free from there.

The `xt.*` observability tables needed a per-table decision:

- `xt.live_tables` sums its row counts across partitions — the same thing it already does across a partition's individual live-table slots.
- `xt.trie_stats` and `xt.live_columns` union theirs. A trie key is only unique within its partition, so above one partition those rows want a `partition` column to be told apart. That's #5849's job; union is at least complete, where partition-0-only would silently drop data.

## Testing

Multi-partition `Database`s aren't constructible yet — `Database.open` builds partition 0 and the attach-time gate rejects anything else until #5837. So the read-path tests present **two attached databases on one node** as the two partitions of one logical database, via a `QueryCatalog` reify. Data goes in through the public API; storage, live index and trie catalog are all real.

Two nodes was the first attempt and doesn't work: the scan transfers page ownership from the live index, and Arrow refuses that between allocators that don't share a root. Two databases on one node do share one.

Reverting the per-branch basis slot to the hardcoded `0` fails four of those tests, including the temporal-bounds one specifically.

The cross-partition type merge is pinned by a Kotlin unit test rather than an end-to-end one, deliberately: at the scan, the live-index half of `scan-vec-types` masks an error in the historical half, so an end-to-end test passes either way. `DatabaseTableCatalogTest` asserts the contract directly.

## Dropped from this PR

#5835's checklist carried "concurrent-write non-collision assertion, carried over from unit 6". I wrote it, then talked myself out of it on review, and it isn't here.

Partition isolation in the buffer pool is path construction — partition 0 writes `parts/0/…`, partition 1 writes `parts/1/…`, different strings. Concurrency can't break that, and the two threads never interact, so there's no interleaving window in which a bug could manifest deterministically. A real defect would have shown up probabilistically, i.e. as flake. `partitionedPoolsAreIsolated` already asserts the same properties sequentially and more precisely, including GC delete isolation.

I initially justified the deletion by pointing at a supposed gap in the shared `memoryCache`/`diskCache` the fixtures hand both pools. That was wrong: `LocalStorage` scopes its cache keys by `dbName/partition` already, with a comment saying so, so there is no cross-partition collision to catch. (#5833 is about those keys omitting the *storage root*, which is an epoch concern, not a partition one.) The deletion stands on its own reasoning above; there's no substitute test I've identified.

## Out of scope

- **The `partition0` delegate survives**, per the issue — `watchers`, `blockCatalog`, `trieCatalog`, `bufferPool` and friends still serve the snapshot export (#5838) and healthz. `storage` and `queryState` are gone from it, since the read path was their only caller.
- **`xt.block_files` / `xt.table_block_files`** still read partition 0's buffer pool, and now carry a marker saying why they can't simply follow `xt.trie_stats` into a union: they're addressed by block index, and block N exists in every partition with different contents, so unioning reads as duplicates of one block rather than as more rows. That's an addressing decision, not a labelling one — raised as #5849 along with the trie-stats column and the rest of stage 4's unit 10, which had been a bare checklist line on #5557 with nowhere to hang either.
- **Partition labelling** on the `xt.*` observability tables, as above.

Two properties of the fan-out are worth naming here rather than leaving for whoever lifts the gate, since both are baked into this design and go live with #5837:

- **Duplicate `_id`s across partitions surface as duplicate rows.** Concatenation gives up the "one resolved version per `_iid`" property a single-branch `MergePlanner` had structurally. That's the documented consequence of the `_id`-stable partitioner contract in #5557, not a new gap — enforcing it would need a cross-partition iid merge, which is the coupling the throughput scaling exists to avoid. Nothing in the read path checks it.
- **All N branch cursors are built eagerly and held to query end.** `MergePlanner/planSync` runs per branch at construction and `PartitionedScanCursor` only closes on its own close, so peak allocator usage is N× a single-partition scan's for the whole query rather than one partition at a time. Closing each branch as it drains is the obvious mitigation; #5557 defers per-node resource work as measure-then-optimise, so it's flagged rather than done.

## One thing to flag

`xtdb.read-only-node-test/read-only-node-sees-data-after-block-flush` failed once in five full-suite runs on this branch, with row `b` present but its `y` column missing. It doesn't reproduce in isolation, didn't recur in the four other full runs (including the final `cleanTest test`), and I couldn't tie it to anything in the diff — the read path doesn't touch the follower's block-application path or `sync-node`, and `sync-node` returning doesn't guarantee the follower has applied the block (the sibling test at line 88 documents exactly that gap). Recording it rather than dropping it; worth a look if it shows up in CI.

## Review

A `/code-review high` pass over the branch raised ten findings; six are fixed in the commits above, three are the deferred items named under Out of scope, and one applied to a test that has since been dropped (see below). The two that mattered were an exception-safety regression in the scan (the tracing wrap had drifted outside its `with-close-on-catch`) and the `getType` inconsistency described under the positional invariant.

## Test plan

- [x] `./gradlew test` — green (2053 tests)
- [x] `./gradlew :xtdb-core:test` — green
- [x] New: `xtdb.multi-partition-scan-test`, 5 tests over a genuinely two-partition read path
- [x] Mutation check: reverting the per-branch basis slot fails 4 of the 5 new tests

Ship/Show/Ask: **Show**.

### DIFF
--- a/core/src/main/clojure/xtdb/block_tables.clj
+++ b/core/src/main/clojure/xtdb/block_tables.clj
@@ -128,6 +128,11 @@
     (catch Exception _e
       nil)))
 
+;; The three cursors below read `Database.bufferPool`, which is partition 0's. Above one partition
+;; that's ambiguous rather than merely incomplete: these tables are addressed by block index, and
+;; block N exists in every partition with different contents, so unioning them would read as
+;; duplicates of one block. Deciding how a partition is addressed here is #5849.
+
 (defn- ->block-files-cursor
   [^Database db ^BufferAllocator allocator col-names col-preds selects schema args]
   (let [derived-table-schema (get block-tables 'xt/block_files)
--- a/core/src/main/clojure/xtdb/information_schema.clj
+++ b/core/src/main/clojure/xtdb/information_schema.clj
@@ -20,9 +20,9 @@
            xtdb.authz.RoleMembership
            (xtdb.arrow Relation RelationReader VectorType VectorReader)
            xtdb.pgwire.PgType
-           (xtdb.indexer Snapshot TableSnapshot)
+           (xtdb.indexer DatabaseSnapshot Snapshot TableSnapshot)
            xtdb.operator.SelectionSpec
-           (xtdb.query IQuerySource IQuerySource$QueryCatalog IQuerySource$QueryDatabase)
+           (xtdb.query IQuerySource IQuerySource$QueryCatalog IQuerySource$QueryDatabase IQuerySource$QueryPartition)
            xtdb.api.TableRef
            xtdb.trie.TrieCatalog))
 
@@ -421,14 +421,25 @@
      :trie-key trie-key, :level (int level), :recency recency, :data-file-size data-file-size
      :trie-state (name state), :row-count row-count, :temporal-metadata (some-> trie-meta (dissoc :row-count))}))
 
-(defn live-tables [^Snapshot snap]
-  (for [^TableRef table (.getTables snap)]
-    {:schema-name (.getSchemaName table)
-     :table-name (.getTableName table)
-     :row-count (long (transduce (map (fn [^TableSnapshot ts]
-                                        (.getRowCount (.getRelation ts))))
-                                 + 0
-                                 (.table snap table)))}))
+;; a table's live rows are spread across its database's partitions, so this sums over them the same
+;; way it already sums over a partition's individual live-table slots.
+;; rows come out in first-encounter order rather than the sum map's: at one partition that's the
+;; per-snapshot order this has always emitted, and a Clojure map stops preserving insertion order
+;; past eight entries.
+(defn live-tables [part-snaps]
+  (let [row-counts (reduce (fn [acc [table row-count]]
+                             (update acc table (fnil + 0) row-count))
+                           {}
+                           (for [^Snapshot snap part-snaps
+                                 ^TableRef table (.getTables snap)]
+                             [table (long (transduce (map (fn [^TableSnapshot ts]
+                                                            (.getRowCount (.getRelation ts))))
+                                                     + 0
+                                                     (.table snap table)))]))]
+    (for [^TableRef table (distinct (mapcat #(.getTables ^Snapshot %) part-snaps))]
+      {:schema-name (.getSchemaName table)
+       :table-name (.getTableName table)
+       :row-count (get row-counts table)})))
 
 (defn live-columns [^Snapshot snap]
   (for [^TableRef table (.getTables snap)
@@ -501,26 +512,29 @@
     (some-> out-rel util/close)))
 
 (defprotocol InfoSchema
-  (->cursor [info-schema allocator db db-cat query-source snapshot derived-table-schema
+  (->cursor [info-schema allocator db db-cat query-source db-snap derived-table-schema
              table col-names col-preds
              schema params]))
 
 (defn ->info-schema [_allocator metrics-registry ^Authenticator$Factory authn]
   (reify InfoSchema
-    (->cursor [_ allocator db db-cat query-source snap derived-table-schema
+    (->cursor [_ allocator db db-cat query-source db-snap derived-table-schema
                table col-names col-preds
                schema params]
       ;; TODO should use the schema passed to it, but also regular merge is insufficient here for colFields
       ;; should be types/merge-types as per scan-vec-types
       (let [^IQuerySource$QueryDatabase db db
             ^IQuerySource$QueryCatalog db-cat db-cat
-            db-state (.getQueryState db)
+            ^DatabaseSnapshot db-snap db-snap
             db-name (.getName db)
-            table-catalog (.getTableCatalog db-state)
-            trie-catalog (.getTrieCatalog db-state)
+            table-catalog (.getTableCatalog db)
+            ;; these two lists come from different objects but describe the same partitions, in the
+            ;; same order — the scan checks their lengths agree before dispatching here
+            part-snaps (.getPartitions db-snap)
+            trie-catalogs (map #(.getTrieCatalog (.getState ^IQuerySource$QueryPartition %)) (.getPartitions db))
             schema-info (-> (merge-with merge
                                         (update-vals (into {} (.getTypes table-catalog)) #(into {} %))
-                                        (.getAllColumnTypes ^Snapshot snap))
+                                        (.getAllColumnTypes db-snap))
                             (merge meta-table-schemas)
                             (update-keys (fn [k]
                                            (cond
@@ -559,9 +573,12 @@
                                      pg_catalog/pg_user (pg-user authn)
                                      pg_catalog/pg_roles (pg-roles authn query-source db-cat)
                                      pg_catalog/pg_auth_members (pg-auth-members query-source db-cat)
-                                     xt/trie_stats (trie-stats trie-catalog)
-                                     xt/live_tables (live-tables snap)
-                                     xt/live_columns (live-columns snap)
+                                     ;; a trie key is only unique within its partition, so above one
+                                     ;; partition these rows need a `partition` column to be told
+                                     ;; apart — #5849
+                                     xt/trie_stats (mapcat trie-stats trie-catalogs)
+                                     xt/live_tables (live-tables part-snaps)
+                                     xt/live_columns (mapcat live-columns part-snaps)
                                      xt/metrics_timers (metrics-timers metrics-registry)
                                      xt/metrics_gauges (metrics-gauges metrics-registry)
                                      xt/metrics_counters (metrics-counters metrics-registry)
--- a/core/src/main/clojure/xtdb/log_tables.clj
+++ b/core/src/main/clojure/xtdb/log_tables.clj
@@ -241,6 +241,10 @@
                               (case lower-op :>= v, :> (inc v))))
           to-msg-id (long (let [v (long upper-val)]
                             (case upper-op :<= (inc v), :< v)))
+          ;; hardcoded partition 0. `msg_id` encodes an offset within one partition, so above one
+          ;; partition the required msg_id bounds are ambiguous rather than merely partial — the
+          ;; same addressing problem `xt.block_files` has, and sharper here, because the replica log
+          ;; is what you read to work out why a particular partition is stuck. #5849
           record-seq (.readRecords log 0 from-msg-id to-msg-id)
           record-iter (.iterator record-seq)
           decode-tx-ops? (and source? (contains? col-names "tx_ops"))]
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -2,6 +2,7 @@
   (:require [clojure.spec.alpha :as s]
             [clojure.string :as str]
             [xtdb.basis :as basis]
+            [xtdb.error :as err]
             [xtdb.expression :as expr]
             [xtdb.expression.metadata :as expr.meta]
             [xtdb.information-schema :as info-schema]
@@ -28,8 +29,8 @@
            (xtdb.api ICursor)
            (xtdb.indexer DatabaseSnapshot Snapshot Snapshot$Source TableSnapshot)
            (xtdb.metadata MetadataPredicate PageMetadata)
-           (xtdb.operator.scan MultiIidSelector ScanCursor ScanMetrics SingleIidSelector)
-           xtdb.query.IQuerySource$QueryCatalog
+           (xtdb.operator.scan MultiIidSelector PartitionedScanCursor ScanCursor ScanMetrics SingleIidSelector)
+           (xtdb.query IQuerySource$QueryCatalog IQuerySource$QueryPartition)
            (xtdb.segment BufferPoolSegment MergePlanner)
            xtdb.api.TableRef
            (xtdb.trie Bucketer)
@@ -204,14 +205,10 @@
               (or (types/temporal-vec-types col-name)
                   (-> (info-schema/derived-table table)
                       (get (symbol col-name)))
-                  (let [state (.getQueryState (.databaseOrNull db-catalog db-name))
-                        table-catalog (.getTableCatalog state)
-                        ^DatabaseSnapshot db-snap (get snaps db-name)
-                        ;; TODO (#5835) reduce types/merge-types over per-partition live
-                        ;; types here; for now we only allow single-partition databases.
-                        ^Snapshot snap (first (.getPartitions db-snap))]
+                  (let [table-catalog (.getTableCatalog (.databaseOrNull db-catalog db-name))
+                        ^DatabaseSnapshot db-snap (get snaps db-name)]
                     (types/merge-types (.getType table-catalog table col-name)
-                                       (.columnType snap table col-name))))))]
+                                       (.columnType db-snap table col-name))))))]
     (->> scan-cols
          (into {} (map (juxt identity ->vec-type))))))
 
@@ -220,11 +217,7 @@
     (emitScan [_ db-cat {:keys [opts]} scan-vec-types param-types]
       (let [{:keys [db-name ^TableRef table columns] :as scan-opts} opts
             db (.databaseOrNull db-cat db-name)
-            storage (.getStorage db)
-            state (.getQueryState db)
-            metadata-mgr (.getMetadataManager storage)
-            buffer-pool (.getBufferPool storage)
-            table-catalog (.getTableCatalog state)
+            table-catalog (.getTableCatalog db)
             col-names (->> columns
                            (into #{} (map (fn [[col-type arg]]
                                             (case col-type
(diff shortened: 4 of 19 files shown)


mbutlerw:
Closing this rather than rebasing it. The work continues on #5835 — this branch becomes source material for it, not a base.

Three things happened since it was opened on 4 August, and together they mean a rebase would cost more than a rebuild:

- **`main` has moved around 200 commits**, and #4037 has merged the block and table catalogs into one value — the catalog this PR's database-level view wraps.
- **`64259e17f` is redundant.** The partitions-as-list change landed independently as `3002a02ae`, via #5924, along with the `->partition-scan-cursor` extraction this branch also carries.
- **The test fixture is being replaced outright**, by #5947. Presenting two attached databases as two partitions by relabelling their storage handles was a reasonable improvisation at the time, but #5947 makes a genuine multi-partition `Database` constructable, so the fiction isn't needed. The fixture is also the part of this diff every test runs through, which is what tips the balance from rebase to rebuild.

What is still only here, and worth reading when #5835 is picked up: the scan's per-partition fan-out, the per-branch temporal bounds, and — on the unpushed local `wip: tier-3 read-path work` commit — the two positional guards, which moved up out of `scan.clj` into `query.clj` where the snapshots are opened.

The branch `multi-part/unit-7-read-path` stays. Please don't delete it on close.

PR #5853, opened 2026-08-06
tidy(scan): groundwork for the partition-aware read path

mbutlerw:
Groundwork for #5835, landed separately. Nothing here is partition-aware.

## Context

Unit 7 of stage 4 (#5835) makes the query read path partition-aware. Written in one go it came out at ~600 lines, most of it inside a `:->cursor` closure that had grown to ~100 lines nested four deep in a `reify`, with its innermost code at column 128. Two review passes over that branch found a confirmed correctness regression in it — a semantics change that rode along unnoticed inside a diff that read as mechanical.

So the groundwork is coming out first, as four steps that each stand on their own. The partition change then lands on top as the small thing it should have been.

## Changes

1. `test(scan)` — pins that a column present historically but absent from the *live* table is declared nullable.
2. `fix(query)` — skip databases detached mid-query rather than NPEing.
3. `tidy(database)` — `Database.partitions` from a `Map<Int, _>` to a `List` indexed by partition.
4. `tidy(scan)` — lift a table's cursor construction out of the emitter.

## Notes

**The test is first on purpose.** It's a characterisation test, verified green on `main` and red against the branch that regressed this exact property — so it guards the refactor in commit 4 rather than merely accompanying it.

It also has two assertions where one would do. `scan-vec-types` merges a column's historical type with its live type, and the live side answers three questions that are easy to collapse into two: no live table at all (contributes nothing), the column present (contributes its type), and the table present *without* the column (contributes `Null`, widening to nullable). The pair pins that adding a row without the column is what widens it; a single assertion couldn't distinguish that from a system that declared everything nullable.

**The detach fix is partial, deliberately.** `open-snaps` and `->table-info` each read the catalog twice — `getDatabaseNames`, then `databaseOrNull` — and a DETACH landing between the two made the second return null. Since `open-snaps` opens a snapshot for *every* attached database (per the TODO above it), detaching one database could take down concurrent queries that never referenced it. That case is gone.

What remains is the query that *does* reference the detached database: `scan-vec-types` dereferences `databaseOrNull` unguarded too, so it now fails there instead, still as an NPE. Turning that into a proper anomaly means choosing what a mid-flight detach should surface as, which is a behaviour decision rather than a tidy. The commit body says so rather than implying the race is fully handled.

**`Map` → `List` rests on an invariant worth stating.** The basis-token codec is positional — slot N of a database's vector is partition N — so `openSnapshot`, `currentBasis`, `latestCompletedTxs`, `latestSubmittedMsgIds` and `awaitAll` all index partitions and basis slots against one another. With a map, each re-established the ordering for itself with `entries.sortedBy`. The list states it once and the constructor asserts each partition's storage handle agrees with its position.

That's safe because every node holds every partition — it simply isn't leader for all of them — so the index range is contiguous. If partitions ever become per-node-assigned rather than all-N, this is the commit to revisit.

**The scan extraction changes no calls, only where they live.** `->partition-scan-cursor` takes the three things that determine what it reads — storage, snapshot, temporal bound — plus a bundle of the per-query values it applies. `metadata-mgr` and `buffer-pool` are derived inside it from the storage handle rather than bound two scopes up. `->cursor` goes from 98 lines to 77, deepest column 128 to 85.

Two smaller things came with it. The `wrapTracing` attribute map — pure data construction, unrelated to cursors — moves to a named binding; it was contributing most of the remaining indentation. And `ScanMetrics` moves up to the emitter, because EXPLAIN ANALYZE reports one set of counters per operator and they're keyed by (db, scan source), i.e. operator identity.

One subtlety the review caught: the tracing wrap takes ownership of the cursor, and the segments' close scope now ends inside the helper. So the `cond->` is wrapped in `with-close-on-catch` on the returned cursor — otherwise a throw between the helper returning and the wrap succeeding would strand the segments.

## Review

Two passes. `/code-review med` over commits 1–4 found the ownership gap above. `/code-review high` then found the test's first assertion claiming a snapshot-eviction fact it couldn't observe (reworded to state the baseline it does establish), a duplicated `->col-type` letfn (lifted to a shared fn), and two comments restating their code rather than carrying rationale (trimmed). It also raised the unguarded `scan-vec-types` read, which is the deliberate omission described above.

## Follow-ups, not in scope

- **Resolve query inputs once per execution.** `databaseOrNull` is called five times per query and the basis is encoded and re-decoded to reach the scan. Threading one resolved catalog and one decoded basis would remove both, give the `scan-vec-types` NPE a single place to be handled, and drop a guard the partition work would otherwise need.
- **Result-shape validation lives in pgwire.** The advertised-schema-vs-resolved-types check sits above the shared `Xtdb.Statement`, so Flight SQL and direct ADBC callers use the same `executeSchema` and don't get it.

## Test plan

- [x] `./gradlew test` green (2046) after each of the four commits
- [x] The new test verified green on `main` and red against the branch that regressed it

### DIFF
--- a/core/src/main/clojure/xtdb/operator/scan.clj
+++ b/core/src/main/clojure/xtdb/operator/scan.clj
@@ -27,6 +27,7 @@
            (xtdb Bytes)
            (xtdb.api ICursor)
            (xtdb.indexer DatabaseSnapshot Snapshot Snapshot$Source TableSnapshot)
+           (xtdb.database PartitionStorage)
            (xtdb.metadata MetadataPredicate PageMetadata)
            (xtdb.operator.scan MultiIidSelector ScanCursor ScanMetrics SingleIidSelector)
            xtdb.query.IQuerySource$QueryCatalog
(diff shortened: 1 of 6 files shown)


mbutlerw:
Superseded by #5924, which carries this work rebased onto current `main`. This branch had gone 130 commits stale and #5855 has since rewritten three of the six files it touched.

Two of the four commits are carried over unchanged, along with `tidy(database): index Database.partitions by list position`.

`faabfdfc1` is not: it pins the distinction between `Snapshot.columnType` returning nil and `TableSnapshot.columnType` returning `VectorType.Null`, and #5855 deleted both methods and made that distinction structural. Scan-level coverage of the declared types may still be worth having, but as a rewrite rather than a rebase.

PR #5874, opened 2026-08-10
ai: scope the mid-run edit freeze to the worktree

jarohen:
## Summary

The `xtdb-testing` skill told agents that no two Gradle invocations may be in flight anywhere on this machine, and to enforce that with `pgrep -af gradlew`. The rule was broader than the hazard it was guarding, and the mechanism could not implement it. This narrows the rule to the mid-run edit freeze, scopes it to the worktree, and drops the check entirely.

No surrounding issue, so the problem context is below rather than linked.

## Why the mechanism could not work

- **The check always reported a hit.** Claude Code wraps every Bash call in `/bin/bash -c '…'`, so the string `gradlew` appears in the checking process's own command line and `pgrep` matches it. Observed live: the naive pattern returned four hits, one of which was the `pgrep` itself. An agent facing an unconditional hit either waits forever or learns to disregard the check, and the second is what happened.
- **Check-then-act was never atomic.** Two builds were in flight from separate worktrees during this investigation — one `:modules:xtdb-postgres-source:integration-test`, one `:xtdb-core:property-test` — launched by sessions that had both read the rule. A machine-global invariant cannot be enforced by each agent independently looking around; that needs a shared object with an atomic acquire.
- **Coverage was partial.** `-Dorg.gradle.appname=gradlew` does survive the `exec` at `gradlew:251`, so the launcher is matchable. But the check misses builds started from IntelliJ or a bare `gradle`, and a daemon still executing after its launcher was killed by `timeout`.

## Why the rule was too broad

- **Worktrees do not share the state that matters.** Each has its own `build/` and `.gradle/`. Gradle's project locks serialise same-directory builds without help.
- **The shared-cache rationale does not hold.** The skill cited contention over `~/.gradle`, but Gradle takes cross-process locks there — concurrent builds block and proceed. The cost is latency, not corruption.
- **What genuinely needs excluding is narrower.** Editing a tree while a build compiles from it produces bogus cross-language type errors that mimic real breakage. That hazard is worktree-scoped, and it is about the editor rather than about a second Gradle.
- **Forbidding the safe majority is part of why it was ignored.** Two `./gradlew test` runs in different worktrees are fine, and that is most runs. A rule that bans them trains agents to discount the whole section, including the constraint that is real.

## Alternatives considered

A machine-wide `flock` on a file under `~/.gradle` would have enforced the original rule properly: atomic acquire, FIFO queueing, released by the kernel on process death so there is no stale-lock cleanup — the failure mode of any PID-file or `mkdir` mutex. Verified the semantics before rejecting it.

Rejected because the concurrency it prevents is concurrency we want. A narrower two-lock variant (one for the docker-compose-backed tags, one for the memory-heavy tasks) was also considered and dropped for the same reason.

Keeping `pgrep` but fixing the pattern — `pgrep -af '[g]radle-wrapper.jar'`, which returns just the launcher and does not match its own command line — addresses only the self-match, leaving the race and the over-broad scope. Not taken.

## Accepted risk

Dropping serialisation leaves a real hazard unguarded, recorded here rather than closed. `property-test` and `integration-test` fork test JVMs at `-Xmx5g` plus 6g direct memory and 1g metaspace, a 12g ceiling each. They are registered under `allprojects` and run up to `org.gradle.workers.max=4` concurrently, so one root invocation has a 48g ceiling on a 62G box, and two concurrent heavy runs double it. An OOM-killed fork presents as a test failure, which under the skill's "failures are always your fault" rule can send a session hunting a bug that does not exist.

Taken deliberately: serialising every heavy run costs more in wall-clock than this costs in occasional confusion, and a killed JVM is distinguishable from a failed assertion. Deliberately not encoded in the skill's "When a test fails" section — that section's strictness is the point, and an OOM caveat there would read as an excuse hatch.

## Test plan

Agent instructions only, no code, so there is nothing to run. Checked that the frontmatter is intact, that the `#when-a-test-fails` anchor still resolves to its heading, and that no stale reference to the removed section survives in the skill or in `AGENTS.md` — the skill description and the `AGENTS.md` summary both advertised "Gradle serialisation" and are updated.

### DIFF
--- a/.claude/skills/xtdb-testing/SKILL.md
+++ b/.claude/skills/xtdb-testing/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: xtdb-testing
-description: XTDB-specific rules and mechanics for running tests — delegation, Gradle serialisation, test tasks and filters, iteration counts, simulation-test coverage, diagnosing failures, and regenerating arrow-edn golden fixtures. Read this before running or delegating any test run in this repo.
+description: XTDB-specific rules and mechanics for running tests — delegation, the mid-run edit freeze, test tasks and filters, iteration counts, simulation-test coverage, diagnosing failures, and regenerating arrow-edn golden fixtures. Read this before running or delegating any test run in this repo.
 ---
 
 # Running tests in XTDB
@@ -13,12 +13,11 @@ Interpret MUST, MUST NOT, SHOULD, SHOULD NOT, MAY per RFC 2119.
 ## The rules you MUST NOT get wrong
 
 1. You MUST NOT run tests yourself — delegate to the `gradle-tests` agent.
-2. You MUST NOT have two Gradle invocations in flight at once, anywhere on this machine.
-3. You MUST NOT edit source files while a test run is in flight.
-4. You MUST tell `gradle-tests`, in every invocation, not to modify any source file.
-5. A test that fails after your change is a test *you* broke — see [When a test fails](#when-a-test-fails).
+2. You MUST NOT edit source files in a worktree that has a test run in flight.
+3. You MUST tell `gradle-tests`, in every invocation, not to modify any source file.
+4. A test that fails after your change is a test *you* broke — see [When a test fails](#when-a-test-fails).
 
-The rest of this document is the mechanics behind those five.
+The rest of this document is the mechanics behind those four.
 
 ## Delegating to `gradle-tests`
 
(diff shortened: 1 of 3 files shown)


jarohen:
Reclassified as ship — landing directly on main, so this PR is redundant. It was also raised off the wrong base and picked up an unrelated commit.

Issue #5525, opened 2026-04-28
Explicit dedup between LiveTable and trie-cat L0s via per-table block-idx watermark


jarohen:
## Summary

When a query opens a snapshot that covers a block-finish event, today's design relies on an accidental property of the bitemporal merge to avoid double-counting between the in-memory live segment and the just-written L0 trie file.
This issue replaces that with an explicit, per-table block-idx watermark: a `LiveTable` knows which block its data belongs to, and the scan operator skips it whenever the trie-cat already has an L0 covering that block.

## Context

A `Snapshot` captures `TableSnapshot`s at open time — slices of the in-memory live data with frozen row counts.
The trie-cat is read at *scan* time, not at snap-open time.

When a block-finish runs, the leader's `BlockUploader` calls `trieCatalog.addTries(L0_N)` early in the flow (`BlockUploader.kt:78`), then proceeds through the table-block protos / main proto / `BlockUploaded` writes, and only finally calls `liveIndex.nextBlock()` (which closes the live data).
This window can be tens of ms to seconds depending on object-store latency.

A query that opened its snapshot before `addTries` and runs its scan during this window sees:

- Its captured `TableSnapshot` containing the live segment with block-N's data.
- The trie-cat (read live) reporting L0_N as a current trie.

The scan operator adds both as input segments to the merge planner.

## Why this is currently safe (and why we should stop relying on it)

Tracing the merge: the dedup happens in `PolygonCalculator` + `ScanCursor` interaction.

`PolygonCalculator.calculate` runs `polygon.calculateFor(ceiling, vF, vT)` *before* `ceiling.applyLog(sF, vF, vT)`.
First time event E `(sF, vF, vT)` is processed, the ceiling is fresh for this iid; the polygon emits range `[vF, vT]` with `sysTo = MAX_VALUE`.
`ceiling.applyLog` then records sysTime=sF for the range.
Second time the same event is processed (from the duplicate segment), the polygon now reads against the post-applyLog ceiling and emits the same range with `sysTo = sF`.
ScanCursor has the check `if (... && validFrom != validTo && sysFrom != sysTo)` which filters the second occurrence (sysFrom == sysTo == sF).

So today's code is accidentally safe.
The property is split across three files (`PolygonCalculator.kt`, `Ceiling.kt`, `ScanCursor.kt`), undocumented, and depends on the L0 file being byte-identical-in-temporal-coordinates to the in-memory data.
[...]

Issue #4967, opened 2025-11-08
Specific bitemporal resolution for as-of-now no-temporal-projection case


jarohen:
We don't need to do full bitemporal resolution when the query is asking for as-of-now with no temporal columns - we know that this can only yield at most one row for each IID, and that historical segments cannot possibly influence the results (as opposed to in the full resolution when they can impact both the multiplicity of the current row as well as its valid-from/valid-to).

Combine this with #4964 - we may only need to yield a bitmask here.

jarohen:
### Chalk — what the as-of resolver actually has to compute

Exploration only, no implementation against this card. Two findings worth acting on: the description scopes this narrower than it needs to be, and a measurement says the case in the title is the one where it pays least.

- [x] Work out the algorithm the fast path reduces to
- [x] Test whether the no-temporal-projection restriction is needed
- [x] Locate where the payoff actually is, given recency partitioning
- [x] Land the seam in the merge loop that this needs
- [x] Measure the resolution amplification

<details><summary>Work out the algorithm the fast path reduces to</summary>

The discriminator is one line. `TemporalDimension.at(t)` produces `lower = t, upper = t + 1`, and `between(t, t)` does `to.inc()` to the same thing, so "as-of in both dimensions" is `upper == lower + 1` on both — covering `AS OF now`, `AS OF <past>`, and the degenerate `BETWEEN t AND t`.

Given that, per entity, walking the existing `(iid ascending, system-from descending)` order: skip events with `systemFrom >= S`; the first event with `validFrom <= V < validTo` wins; a `put` emits and a `delete` emits nothing; then skip the rest of the entity. No `Polygon`, no `Ceiling`, no `LongArrayList`, and no per-segment `intersects` loop.

Erase semantics carry over unchanged. `PolygonCalculator` deliberately exempts erases from the system-time bound, so an erase at any system-time kills the entity — and since the walk is newest-first, it is seen before anything it suppresses.

</details>

<details><summary>Test whether the no-temporal-projection restriction is needed</summary>

`check:` the two results below are derived on paper from `Ceiling`/`Polygon` semantics and a worked example. They have not been checked against running code, and anyone picking this up should confirm them before leaning on them.

[...]

Issue #4965, opened 2025-11-08
Cache IID in EventRowPointer


jarohen:
At the moment, we spend a non-trivial amount of time comparing IIDs for the PriorityQueue in the bitemporal resolution - this is caused by needing to go back and read the bytes in memory every time.

We could bring IIDs in to the ERP on first read (maybe only the first int/long) then use this for the initial comparison, only falling back to memcmp in the unlikely event the first int/long doesn't match.

If you do use ints/longs, don't forget that this will need to be an unsigned, big-endian comparison - not Java's standard int/long cmp.

Issue #4014, opened 2025-01-13
IID/ID moved up a level in the primary index


jarohen:
Currently, index events are flat within the index files. Bitemporal resolution then has to scan past all of the versions of an entity even after it's found the event it needs. Also, from a space PoV, the IID and ID columns are duplicated for every version of the entity within the file.

Neither of these are particularly onerous under the assumption that an individual entity has relatively few (< 5, say) versions. However, when XT is being used for (e.g.) sensor readings, we recommend that each reading for a device is a version of the same entity, at different valid times - so this assumption no longer holds.

Let's store both IID and ID at the top-level of the primary index, then a ListVector of events containing the VF/VT and document (minus ID). We can still re-construct the document easily during bitemporal resolution using an indirection mapping over the ID column.

jarohen:
need to consider what should happen with the live index - we won't be able to store that this way, and I'm not convinced we want two different structures

jarohen:
### Chalk — the live-index question, and where the cost actually sits

Exploration, no implementation against this card. The mechanism has since been chosen and filed as #5914, which supersedes the layout this comment originally proposed.

**tl;dr**

- **The January question has an answer, and it is a storage question rather than a resolution one.**
  One resolution algorithm can serve both a grouped file and the flat live index; what differs is the cursor beneath it.

- **The mechanism is run-end encoding, not a nested layout** — filed as #5914, where the nested `List<Struct>` this comment first proposed is recorded as the rejected alternative `A1`.

- **The space figure to use is ~20 bytes per entity per page, not the 3.7× this comment originally claimed.**
  That figure assumed offsets stored per event; run-ends are per run, and list offsets would have been per list.

- **Neither this card nor #4967 helps `AS OF now`.**
  Recency partitioning leaves about one un-superseded event per entity in the current tier, so there is nothing to group and nothing to skip there. The win is historical reads.

<details><summary>The live-index question, answered</summary>

The January comment — "we won't be able to store that this way, and I'm not convinced we want two different structures" — is the right worry about *storage* and, on the evidence of #5912, not a worry about *resolution*.

- **The two-level cursor splits the interface cleanly**: `nextIid()` at the outer level, and `systemFrom`/`validFrom`/`validTo`/`op`/content-index at the inner. Two implementations of one interface, not two structures the algorithm has to know apart.

[...]

PR #5912, opened 2026-08-18
refactor(scan): merge by entity, then by system-time within it

jarohen:
Groundwork for #4014 and #4967. Resolves neither, and changes no observable behaviour.

## tl;dr

- **Bitemporal resolution compared a 16-byte iid on every event; it now compares one per entity per page.**
  `ScanCursor`'s single `(iid ascending, system-from descending)` priority queue becomes two — an outer queue over entities, an inner queue over one entity's events — so the comparator that runs per event is a single long rather than an iid.

- **Behaviour is intentionally unchanged, and the thing to review is the ordering argument rather than the diff.**
  The same events reach `PolygonCalculator` in the same order and the same rows come out. If that claim holds the change is invisible; if it doesn't, the failure mode is wrong query results rather than a crash.

- **The seam is the point, more than the saving.**
  "Done with this entity" is now a place in the loop. #4014 (group events under their iid on disk) and #4967 (stop at the first event covering an as-of point) both need one, and neither can express it against a flat queue.

- **Nobody has measured it.**
  The saving scales with events per entity per page, so it is a wash on a table with one version per entity and only pays on the churn-heavy shapes #4014 describes. Sizing that is a separate job, noted under Out of scope.

## Context

Events reach bitemporal resolution in `(iid ascending, system-from descending)` order, and they arrive in runs — every version of one entity, newest first, before the next entity starts. `ScanCursor` was throwing that structure away: a single priority queue over event pointers, whose comparator leads with the 16-byte iid, so every event paid `O(log pages)` iid comparisons to rediscover a boundary the data already expressed.

`PolygonCalculator` paid twice over. It has to notice when it crosses into a new entity, and it did that by re-deriving the iid as an `ArrowBufPointer` and comparing 16 bytes — twice per event — even though `EventRowPointer` has cached the iid as two longs since #4965. It also held three `ArrowBufPointer`s standing in for at most two values, with the post-erase skip encoded as "`skipIidPtr` happens to equal the entity we're on".

## Changes

1. `bfcc77ca3` — `tidy(bitemporal): one iid and an erase flag, not three ArrowBufPointers`. `PolygonCalculator`'s three pointers collapse to the entity it's on plus an `erased` flag, cleared together when the entity changes.

2. `9c114d100` — `refactor(scan): merge by entity, then by system-time within it`. The queue split, plus `sameIidAs`, the two comparators, and a cached `systemFrom` on `EventRowPointer`.

The first is a ship-class tidy that was headed for `main` independently; it rides along here because the second builds on it and it has not landed yet.

## Equivalence

The emitted order is identical to the old single queue's, and it rests on two properties rather than on the tests agreeing.

- **A priority queue yields every element tied for its minimum consecutively.**
  So the drain loop that moves pages from `iidQueue` into `eventQueue` collects exactly the pages holding the minimum entity — never a page holding a later one, and never missing a page holding this one.

- **Within a page, an entity's events are contiguous and already system-from descending.**
  So each page re-offers its next event as the inner loop consumes one, and the inner queue sees exactly that entity's events in the right order. A page that crosses into a new entity is routed back to `iidQueue`, where its new iid can only sort later than the entity being resolved.

- **The cached `systemFrom` is fresh wherever it is legal to read it.**
  `maxIndex <= rowCount` always, so a pointer that `isValid()` is a pointer whose cache has been refreshed. Every reader in the repo — `PolygonCalculator`, `ScanCursor`, `SegmentMerge.merge`, `resolveSameSystemTimeEvents`, both comparators — reads either immediately after polling a valid pointer or behind an `isValid()` short-circuit.

- **Unchanged: two pages holding both the same entity and the same system-time tie arbitrarily.**
  That was true before and is true now. It should not arise — same-transaction events for one entity are resolved within a page at load, and #5525 dedups the live table against L0 — but nothing here tightens it.

## Out of scope

Split by why each was left, rather than by what it is.

- **`SegmentMerge.merge` and `resolveSameSystemTimeEvents` keep the single comparator**, because extracting a shared merge from one worked example is how you get the wrong abstraction. They walk the same entity runs and want the same treatment; revisit when there is a second instance to generalise from.

- **`skipToNextIid()` is not here**, because nothing in this loop calls it. It is what turns the seam into an actual saving for #4967, and it should arrive with its caller rather than as dead code.

- **`Polygon.calculateFor(Ceiling, EventRowPointer)` is dead and stays dead.** It has no callers on `main` either, so removing it is a separate tidy rather than something to bundle into a refactor commit.

- **The measurement.** `scan_rows_read` against emitted rows gives the resolution amplification factor directly, which is the number that decides whether #4967 earns a second resolution algorithm. It needs a dataset with real versions-per-entity, so it is its own piece of work.

## Test plan

- `./gradlew test`, `./gradlew property-test` (100 iterations) and `./gradlew integration-test` all green on the final commit.

- `PolygonCalculatorTest` is new — the class had no unit test at all. Four cases: ceiling isolation between entities, erase suppression and its scoping to one entity, and both halves of the system-time bound (a put above it is ignored and leaves no ceiling behind; an erase above it still suppresses).

- A code-review pass over both commits, aimed at the equivalence claim rather than at style. It found no failing scenario for the ordering, confirmed the `systemFrom` cache is guarded at every reader, and flagged the dead `calculateFor` overload noted above.

### DIFF
--- a/core/src/main/kotlin/xtdb/bitemporal/PolygonCalculator.kt
+++ b/core/src/main/kotlin/xtdb/bitemporal/PolygonCalculator.kt
@@ -1,31 +1,31 @@
 package xtdb.bitemporal
 
-import org.apache.arrow.memory.util.ArrowBufPointer
 import xtdb.trie.EventRowPointer
 import xtdb.util.TemporalBounds
 
 class PolygonCalculator(private val queryBounds: TemporalBounds? = null) {
-    companion object {
-        private fun ArrowBufPointer.setFrom(src: ArrowBufPointer) = apply { set(src.buf, src.offset, src.length) }
-    }
-
-    private val skipIidPtr = ArrowBufPointer()
-    private val prevIidPtr = ArrowBufPointer()
-    val currentIidPtr = ArrowBufPointer()
-
     private val ceiling = Ceiling()
     private val polygon = Polygon()
+
+    // the iid we're currently resolving, and whether we've seen an erase for it.
+    // the initial (0, 0) needn't be distinguishable from a genuine iid: the most it can cost us is the
+    // ceiling reset on the first row, and the ceiling is constructed already reset.
+    private var iidHigh = 0L
+    private var iidLow = 0L
+    private var erased = false
 
     fun reset() = ceiling.reset()
 
     fun calculate(erp: EventRowPointer): Polygon? {
-        // after an erase, we don't process any more rows
-        if (skipIidPtr == erp.getIidPointer(currentIidPtr)) return null
-
-        if (prevIidPtr != currentIidPtr) {
+        if (erp.iidHigh != iidHigh || erp.iidLow != iidLow) {
+            iidHigh = erp.iidHigh
+            iidLow = erp.iidLow
+            erased = false
             ceiling.reset()
-            prevIidPtr.setFrom(currentIidPtr)
         }
+
+        // after an erase, we don't process any more rows
+        if (erased) return null
 
         val isErase = erp.op == "erase"
 
@@ -40,10 +40,8 @@ class PolygonCalculator(private val queryBounds: TemporalBounds? = null) {
         polygon.calculateFor(ceiling, validFrom, validTo)
         ceiling.applyLog(systemFrom, validFrom, validTo)
 
-        if (isErase) {
-            skipIidPtr.setFrom(currentIidPtr)
-        }
-        
+        if (isErase) erased = true
+
         return polygon
     }
 }
--- a/core/src/main/kotlin/xtdb/operator/scan/ScanCursor.kt
+++ b/core/src/main/kotlin/xtdb/operator/scan/ScanCursor.kt
@@ -63,7 +63,10 @@ class ScanCursor(
         while (mergeTasks.hasNext()) {
             val task = mergeTasks.next()
             val taskPath = task.path
-            val mergeQueue = PriorityQueue<LeafPointer>(comparing({ it.evPtr }, EventRowPointer.comparator()))
+            // outer: one entry per page, ordered by the iid that page is currently on.
+            // inner: the pages sitting on the entity being resolved, newest system-time first.
+            val iidQueue = PriorityQueue<LeafPointer>(comparing({ it.evPtr }, EventRowPointer.iidComparator()))
+            val eventQueue = PriorityQueue<LeafPointer>(comparing({ it.evPtr }, EventRowPointer.systemFromComparator()))
             val polygonCalculator = PolygonCalculator(temporalBounds)
 
             // we're not in coroutine land here, so it's a good boundary for runBlocking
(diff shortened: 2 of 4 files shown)


jarohen:
### Chalk — session context, for picking this up elsewhere

The PR description covers the change itself. This is the session state around it: where the code physically is, what was decided and ruled out, and what the measurement did to the plan.

- [x] `PolygonCalculator` tidy — one iid plus an erase flag
- [x] `ScanCursor` restructured to outer-entity / inner-event
- [x] Measure resolution amplification
- [x] Write the exploration back onto #4967 and #4014
- [x] Get the tidy onto `main` — landed with this PR's merge
- [x] Decide whether hoisting `_id` splits out of #4014 — under #5914 it stops being a hoist
- [ ] Measure the page-load vs resolution-CPU split before building #4967 — parked, see #4967

<details><summary>Where the code is</summary>

**Both commits are on the fork**, so nothing is stranded on one machine: `jarohen:scan-merge-by-entity`, head `9c114d100`.

- `bfcc77ca3` — `tidy(bitemporal): one iid and an erase flag, not three ArrowBufPointers`
- `9c114d100` — `refactor(scan): merge by entity, then by system-time within it`

**Both commits are on `main`.** This PR merged as a fast-forward, so `bfcc77ca3` and `9c114d100` are on `main` and `origin/main` with nothing outstanding locally. The tidy was ship-class and would have gone to `main` independently; it rode along here because the refactor builds on it and it hadn't landed yet.

So on another machine, `git fetch origin scan-merge-by-entity` reaches everything, and the tidy is `bfcc77ca3` within it.

</details>

<details><summary>Decisions taken, and what each rules out</summary>

**Both commits are equivalence changes; the reviewable claim is the ordering argument, not the diff.** If the argument in the PR description holds, this is invisible. If it doesn't, the failure mode is wrong query results rather than a crash — which is why the review pass was pointed at the ordering rather than at style.

**`SegmentMerge.merge` and `resolveSameSystemTimeEvents` were deliberately left on the single comparator.** They walk the same entity runs and want the same treatment, but extracting a shared merge from one worked example is how you get the wrong abstraction. The call was to do `ScanCursor` standing alone and revisit once there's a second instance to generalise from. That's the open design question the next session inherits, not a loose end.

**`skipToNextIid()` was cut from this PR.** Nothing in the restructured loop calls it — pointers still advance one event at a time — so it would have landed as dead code. It belongs with #4967, alongside its first caller.

**`Polygon.calculateFor(Ceiling, EventRowPointer)` is dead and was left dead.** Confirmed unused on `main` too, so it predates this work; removing it is a separate tidy and would also drop `Polygon`'s only dependency on `EventRowPointer`.

**Verification behind it**: `test`, `property-test` at 100 iterations, and `integration-test` all green on `9c114d100`. `PolygonCalculatorTest` is new — the class had no unit test at all — with four cases pinning ceiling isolation, erase scoping, and both halves of the system-time bound.

</details>

<details><summary>The measurement, and what it does to the plan</summary>

```
setup          entities=200 versions=50 total-rows=10000
as-of-now      read=200  emitted=200  ratio=1.0  files=1/8  pages=1/0
as-of-past-vt  read=1400 emitted=200  ratio=7.0  files=1/8  pages=4/0
range-vt       read=4200 emitted=2200 ratio=1.9  files=3/6  pages=12/0
```

**#4967's headline case measures exactly 1.0.** For as-of-now the dated tries are pruned before resolution starts, so the early exit has nothing to skip and only the constant factor is available. The card is named after the case where it pays least.

**The as-of-past point is the real one, at roughly 85% skippable — but it is resolution CPU, not I/O.** Those four pages are read and decoded either way.

**A claim I made earlier in the session and have since corrected on #4967**: I said recency pruning stops firing for a past query. It doesn't — file-level pruning was equally effective for the past point (8 of 9 files pruned either way), because an as-of point only intersects one era's temporal metadata. What survives is page granularity, which is why it's 7× rather than the 50× the version count suggests.

**So the gating question for #4967 has changed.** It is no longer an amplification question but #5893's question: what share of scan wall-clock is actually the `ScanCursor` loop? If scans are page-bound, the as-of resolver buys a second resolution algorithm's permanent correctness surface for a slice of a minority cost. My read went from `+0.8` to `+0.3` on building it next.

Fuller write-ups, with the algorithm sketch and the two-scalar result that widens the card beyond its no-temporal-projection restriction, are on #4967 and #4014 rather than here.

</details>

<details><summary>The measurement harness</summary>

`caveat:` **reconstructed, not re-run.** The original was written into the worktree, run once to produce the numbers above, and deleted to keep the branch clean. This text is reconstructed from the session, so expect to fix a typo rather than a design flaw. Drop it at `src/test/clojure/xtdb/scratch_amplification_test.clj` and run `./gradlew :test --tests '*scratch_amplification*' --rerun-tasks --info`.

The shape worth keeping whatever happens to the code: each entity gets consecutive disjoint valid-time ranges with the **final version left open-ended**, so exactly one version per entity has infinite recency and lands in the current tier. Without that, an as-of-now query has no current tier to hit and the contrast disappears. The mock clock ticks a day per message, which is why version `v` is asserted at roughly the system time its valid range opens.

```clojure
(ns xtdb.scratch-amplification-test
  (:require [clojure.test :as t]
            [xtdb.api :as xt]
            [xtdb.compactor :as c]
            [xtdb.test-util :as tu])
  (:import (java.time Duration Instant)))

(t/use-fixtures :each tu/with-mock-clock tu/with-node)

(def ^:private entities 200)
(def ^:private versions 50)
(def ^:private t0 (Instant/parse "2020-01-01T00:00:00Z"))

(defn- version-instant [^long i]
  (.plusSeconds t0 (* i 86400)))

(defn- scan-attrs [q & args]
  (let [row (->> (xt/q tu/*node* (into [(str "EXPLAIN ANALYZE " q)] args))
                 (filter #(= :scan (:op %)))
                 first)]
    {:rows-read (get-in row [:attributes :scan-rows-read])
     :rows-emitted (:row-count row)
     :files-used (get-in row [:attributes :scan-files-used])
     :files-pruned (get-in row [:attributes :scan-files-pruned])
     :pages-used (get-in row [:attributes :scan-pages-used])
     :pages-pruned (get-in row [:attributes :scan-pages-pruned])}))

(defn- report! [label m]
  (let [{:keys [rows-read rows-emitted]} m]
    (println (format "AMPLIFICATION|%s|read=%s|emitted=%s|ratio=%s|files=%s/%s|pages=%s/%s"
                     label rows-read rows-emitted
                     (if (and rows-read rows-emitted (pos? (long rows-emitted)))
                       (format "%.1f" (double (/ (long rows-read) (long rows-emitted))))
                       "n/a")
                     (:files-used m) (:files-pruned m)
                     (:pages-used m) (:pages-pruned m)))))

(t/deftest measure-resolution-amplification
  (doseq [v (range versions)]
    (let [vf (version-instant v)
          vt (when (< v (dec versions)) (version-instant (inc v)))]
      (xt/submit-tx tu/*node*
                    [(into [:put-docs {:into :readings, :valid-from vf, :valid-to vt}]
                           (for [e (range entities)]
                             {:xt/id e, :value (+ (* v 1000) e)}))])))

  (tu/flush-block! tu/*node*)
  (c/compact-all! tu/*node* (Duration/ofMinutes 2))

  (println (format "AMPLIFICATION|setup|entities=%d|versions=%d|total-rows=%d"
                   entities versions (* entities versions)))

  (report! "as-of-now"
           (scan-attrs "SELECT _id, value FROM readings"))

  (report! "as-of-past-vt"
           (scan-attrs (format "SELECT _id, value FROM readings FOR VALID_TIME AS OF TIMESTAMP '%s'"
                               (str (version-instant (quot versions 2))))))

  (report! "range-vt"
           (scan-attrs (format "SELECT _id, value FROM readings FOR VALID_TIME BETWEEN TIMESTAMP '%s' AND TIMESTAMP '%s'"
                               (str (version-instant 10)) (str (version-instant 20)))))

  (t/is true "measurement only"))
```

</details>

<details><summary>What's open</summary>

- **`D1` — does hoisting `_id` split out of #4014? Answered: it stops being a hoist.** Under the run-end encoding filed as #5914, `_id` is encoded the same way *inside* the put struct, so it gets the space win without touching `mergeFields`, page metadata, blooms or #5357. #5914 covers `_iid`; `_id` follows it.

- **`Stream` on this PR is blank**, deliberately. Scan-side bitemporal resolution didn't obviously belong to any of the six, and the convention is to leave it for a human rather than invent one. No milestone either — internal refactor.

- **The next measurement**, and the one that should gate #4967: the page-load/decode versus `ScanCursor`-loop split, on a table with real versions per entity. `modules/bench/src/main/clojure/xtdb/bench/readings.clj` is the shape.

- **Whether `SegmentMerge` folds into a shared merge** — see the decisions block. Wants a second worked example first.

</details>

Issue #5910, opened 2026-08-18
Non-critical databases stay stopped for the life of the node after an ingestion error


Akeboshiwind:
## tl;dr

- **A database whose ingestion fails never recovers on its own**
  Nothing tears it down or re-opens it, so a transient fault costs the same as a permanent one.

- **Both recovery paths we have today are heavier than the fault**
  A node restart takes every other database on that node with it, and `DETACH` plus re-`ATTACH` is a cluster-wide membership change applied on every node, to fix one node's local failure.

- **We already decided these databases shouldn't take the node down** — `Database.Config.critical` exists and `/healthz/alive` only 503s for critical ones, so a non-critical failure is deliberately survivable at node level. Right now "survivable" means "stopped until an operator intervenes".

- **Both node types are affected**, and an ingest node is entirely made of the databases in question.

## Context

XTDB already distinguishes databases whose ingestion is load-bearing for the node from ones whose isn't. `Database.Config.critical` (`core/src/main/kotlin/xtdb/database/Database.kt:459`) feeds `/healthz/alive`, which returns 503 only when a *critical* database has an ingestion error and stays 200 otherwise (`core/src/main/clojure/xtdb/healthz.clj:33,101,112`). The primary `xtdb` database is always critical; secondaries are non-critical by default.

That flag decides whether a failure takes the node out of service. It does not do anything about the failed database itself, and nothing else does either. So the current contract for a non-critical database is: fail once, stay failed until an operator intervenes. Both interventions available are disproportionate to the fault — a node restart stops every other database on that node, and `DETACH` plus re-`ATTACH` travels the primary's source log and is applied by every node, to recover one node's local failure.

The cost is highest exactly where the flag is most used. A secondary fed by an external CDC source is non-critical by construction, and `IngestNode` — a node type that runs nothing but external-source databases (`core/src/main/kotlin/xtdb/api/IngestNode.kt:27-29`) — has no other work to do. When its database stops, the node is inert but still reports itself alive.

#5878 is the case that prompts this now: a postgres-source whose replication connection dies mid-stream halts its database permanently. That is one instance of a shape we expect to keep meeting, so the general recovery mechanism lands first and a targeted source-level reconnect for #5878 follows once it is in place.

## Current state

[...]

PR #5923, opened 2026-08-20
Prepare the database catalog for restarting databases (#5910)

Akeboshiwind:
Part of #5910.

## tl;dr

- **Groundwork for auto-restarting databases that stop ingesting — no restart behaviour yet**
  The catalog and the specs are put in a shape that can express a database being temporarily absent; the supervisor that uses it comes next.

- **One real bug: `/healthz/alive` could return 500 and get your node killed**
  A database listed but no longer resolvable produced an NPE, and a liveness probe fails on any non-2xx — so a routine detach could turn into a container restart.

- **Plus a latent one: `detach` was a read-then-write, so two callers could both win**
  Unreachable today because there is only ever one writer; fixed now because the restart supervisor is about to be the second.

- **Everything else is behaviour-preserving**
  Three refactors, each verified against the tests that already pinned the behaviour.

- **The node-level database lifecycle now has a spec, which it never had**
  `db.allium` scopes itself to a single database and defers the catalog; that deferral had nowhere to point.

## Context

#5910 has the problem statement: a non-critical database whose ingestion fails is never restarted, so a transient fault costs the same as a permanent one, and both recovery paths available today — restart the node, or `DETACH` and re-`ATTACH` — are disproportionate to the fault.

None of that is fixed here. This is the part that has to land first: the catalog could not represent "this database exists but has no live `Database` right now" without a third map, and nothing described the lifecycle it would be extending.

## Changes

Read in order:

1. **`docs(allium)`: specify the node-level database lifecycle** — today's three states only (`open`, `detaching`, `skipped`), plus the `hold` ⊃ `member` ⊃ `run` nesting that distinguishes them.
2. **`docs(allium)`: model where a node's database set comes from** — resolves the spec's own open question about ingest nodes, which hold databases with no catalog and never receive attach or detach instructions.
3. **`fix(healthz)`: skip a database that no longer resolves** — the 500.
4. **`refactor(db-catalog)`: one entry per database, not two maps and a flag**.
5. **`refactor(healthz)`: put criticality on the database** — `Database.isCritical`.
6. **`refactor(db-catalog)`: let the catalog say why a lookup failed** — `databaseOrThrow`.
7. **`fix(db-catalog)`: make the detach transition atomic** — the one thing here the sealed entry made a one-liner.
8. **`docs(allium)`: write the lifecycle spec in the chalk voice** — a voice pass over 1 and 2 against the newly-released `chalk:allium-voice`: three obligations renamed off "what it is not", the passages arguing for the design rather than obliging anything cut, sentence-per-line throughout. No obligation changed — stripping every comment from both revisions leaves three `@guarantee` identifiers and nothing else.

### The healthz bug

- **The nil was always reachable, and the window is about to get much wider**
  `databaseNames` and `databaseOrNull` each filter on the same flag, so a detach landing between them leaves a name that resolves to nothing. `.getName` on it throws, the exception interceptor turns that into a 500, and the liveness probe fails. Today that window is microseconds; a restart widens it to a whole teardown plus backoff.

- **Four siblings on `Database$Catalog` already guarded this and one didn't**
  `awaitAll`, `syncAll`, `latestCompletedTxs` and `iterator` all skip a name that fails to resolve, and `latestCompletedTxs` carries a comment saying why. `/healthz/started` guards it too.

- **dead end: `(into [] db-cat)` looks like the better fix and is wrong**
  `Database$Catalog` is both `Iterable<Database>` and `Seqable`, and `RT.seq` prefers the `Seqable` — whose elements are `[name database]` pairs, not databases. The duplicated guard stays, and a test would catch anyone reaching for the shortcut.

### The catalog's entry model

- **Three states were encoded across two maps and a volatile boolean**
  Nothing structurally stopped a name being in both maps; the guard was two `if`s a reader had to find and trust. `Database.isClosing` described the catalog's opinion of a database rather than the database, which is why the catalog was its only reader.

- **Only the state holding no `Database` stores a config**
  `Open` and `Detaching` delegate to the one they already hold. `Skipped` stores its own because there is nothing to ask — and that is why an entry carries a config at all: a block records the whole secondary list and replaces the previous one, so an entry omitted while a block is cut is erased rather than skipped.

- **Removal got stricter as a side effect**
  Teardown ended with `databases.remove(dbName, db)`, whose value comparison runs through `Database.equals` — which is by name, so it would have matched any database under that name. `Entry` has no `equals`, so the same call now means what it looks like.

### `databaseOrThrow`

- **Scoped to one of four call sites, deliberately**
  Only `Xtdb.kt` already raised the canonical `xtdb/unknown-db`, so only it could move without changing behaviour. `export` and `compactor.reset` raise tool-namespaced `::db-not-found` codes and pgwire raises a protocol-level error with a playground attach-on-demand path in front of it; folding those in would have changed three user-visible error codes inside a behaviour-preserving commit.

- **One caller is enough because the argument isn't reuse**
  `Entry` is private to `DatabaseCatalog`, so a caller cannot distinguish "restarting" from "never existed" — `databaseOrNull` collapses both to null. This is the seam that lets the catalog answer that without exposing its state machine.

### `detach` atomicity

- **Reading the entry and writing `Detaching` were separate steps**
  Two callers could both observe `Open`, both install their own `Detaching`, and both launch a teardown — tearing the same `Database` down twice, with the second failure swallowed by the teardown's catch-all into a log line. Pre-existing: the old `isClosing` test-and-set had the identical window.

- **Unreachable today, which is why it survived**
  Every caller of `attach`/`detach` is the primary's log processor, whose leader, follower and transition forms are mutually exclusive by construction. It becomes reachable with the restart supervisor, which mutates the map concurrently with log processing by design — and the guards that change needs are only worth anything if the transitions underneath them are atomic.

- **`replace(key, expected, new)` is the whole fix**
  `Entry` has no `equals`, so the compare is by reference. Losing it means someone else moved the entry on, which is indistinguishable from arriving after they finished — so the loser gets the same `NotFound` it would have got a moment later.

- **No test, deliberately**
  A two-thread test only catches a regression when the threads interleave inside a window of tens of nanoseconds, so it would pass most of the time against the broken code. For a race that cannot currently happen, code that is correct by construction beats a detector that usually misses.

## Known and deliberate

- **A pre-existing config asymmetry is preserved rather than tidied**
  The skip check runs before the `readOnlyDatabases` override, so a skipped database records its raw config while an open one reports the overridden one. Harmless — the override is reapplied from node config at startup — but whether it is intended is a separate question.

- **`"xtdb"` is hardcoded in eight places** in the database package while `RoleMembership.PRIMARY_DB` holds exactly that constant in authz. Its own tidy.

- **`FixedDbCatalog` inherits `databaseOrThrow`'s default**, so an ingest node will report a restarting database as "unknown" until that step overrides it.

- **A default method on `Database.Catalog` is invisible to a mockk double**
  `jvmDefault = NO_COMPATIBILITY` makes it a real JVM default, which mockk intercepts rather than running — so the double keeps compiling and starts answering wrong. That is how `ConnectionTest` went red on this branch, and it will catch the next person who extends this interface.

## Test plan

- `xtdb.healthz-test` (10), `xtdb.sql.multi-db-test` (16), `xtdb.database-test` (3), `xtdb.api-test` (35), `xtdb.api-flight-sql-test` (15), `ConnectionTest` (9), `DatabaseCatalogTest` (1) — all passing, and re-run after every edit that followed a green run.
- The tests that carry the weight are the ones already there: `DatabaseCatalogTest`'s gated re-attach-during-detach, `skip-dbs-on-startup` and `detach-dormant-database` (which exercise an entry's config reaching the persisted secondary list), and `unknown-db-is-rejected-at-open`.
- One test added, for the healthz bug: a catalog listing a name that resolves to nothing must not reach the code that dereferences it.
- **The full suite has not been run locally**, and neither have the integration or property suites. Nothing here touches indexing, compaction or GC.

### DIFF
--- a/core/src/main/kotlin/xtdb/database/Database.kt
+++ b/core/src/main/kotlin/xtdb/database/Database.kt
@@ -122,6 +122,9 @@ class Database(
     val latestProcessedMsgId: MessageId get() = watchers.latestSourceMsgId
     val ingestionError: IngestionStoppedException? get() = watchers.exception
 
+    /** Whether an ingestion error here should take the whole node out of service. */
+    val isCritical: Boolean get() = name == "xtdb" || config.critical
+
     fun awaitTxBlocking(txId: Long, timeout: Duration? = null): TransactionResult? =
         runBlocking {
             check(isIndexing) { "log processor not initialised" }
--- a/core/src/main/kotlin/xtdb/database/Database.kt
+++ b/core/src/main/kotlin/xtdb/database/Database.kt
@@ -551,6 +551,10 @@ class Database(
         override val databaseNames: Collection<DatabaseName>
         override fun databaseOrNull(dbName: DatabaseName): Database?
 
+        fun databaseOrThrow(dbName: DatabaseName): Database =
+            databaseOrNull(dbName)
+                ?: throw Incorrect("Unknown database: $dbName", "xtdb/unknown-db", mapOf("db-name" to dbName))
+
         operator fun get(dbName: DatabaseName) = databaseOrNull(dbName)
 
         val primary: Database get() = databaseOrNull("xtdb")!!
--- a/core/src/main/kotlin/xtdb/database/Database.kt
+++ b/core/src/main/kotlin/xtdb/database/Database.kt
@@ -54,7 +54,6 @@ import io.micrometer.core.instrument.MeterRegistry
 import java.time.Duration
 import java.time.Instant
 import java.util.*
-import kotlin.concurrent.Volatile
 import xtdb.api.tx.ExternalSource
 
 private val LOG = Database::class.logger
@@ -122,10 +121,6 @@ class Database(
 
     val latestProcessedMsgId: MessageId get() = watchers.latestSourceMsgId
     val ingestionError: IngestionStoppedException? get() = watchers.exception
-
-    @Volatile
-    var isClosing: Boolean = false
-        internal set
 
     fun awaitTxBlocking(txId: Long, timeout: Duration? = null): TransactionResult? =
         runBlocking {
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -58,6 +58,7 @@ That commentary belongs in the commit body and the PR, which are read once aroun
 | Spec | Covers |
 | --- | --- |
 | `allium/live-index.allium` | In-memory staging of transactions before block flush — `LiveIndex`/`LiveTable`, transaction lifecycle, snapshot visibility, and the leader's resolve/append/consume-back/apply pipeline. |
+| `allium/database-lifecycle.allium` | The lifecycle of a database on one node — how it becomes a member of the node's set, how it stops being one, and what resolves by name in between. Node-level; treats a running database as opaque. |
 | `allium/log-processor-lifecycle.allium` | Per-database leader election in the `LogProcessor` — the Following/Prepared/Leading state machine, the term fence that keeps one confirmed leader per database, and the split between launching a transition and committing the role. |
 | `allium/memory-hash-trie.allium` | The immutable in-memory hash trie indexing rows by IID within a `LiveTable` — bucketing, leaf ordering, log compaction, splitting. |
 | `dev/doc/db.allium` | The processing model of one database end to end — submit → log processing → block flush → query, the source/replica log message types, the block and table catalogs, and the compaction message flow. |
--- a/allium/database-lifecycle.allium
+++ b/allium/database-lifecycle.allium
@@ -1,84 +1,83 @@
 -- allium: 3
 -- database-lifecycle.allium
 --
--- The lifecycle of a database on one node: how it becomes a member of the node's set,
--- how it stops being one, and what an observer can do with it in between.
+-- The lifecycle of a database on one node: how it becomes a member of the node's set, how it stops being one, and what resolves by name in between.
 --
 -- Scope: one node's set of databases, and the transitions between the states below.
--- Excludes: everything that happens INSIDE a running database -- transaction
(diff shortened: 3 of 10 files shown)


core/src/main/kotlin/xtdb/database/Database.kt
```diff
@@ -550,6 +548,10 @@ class Database(
         override val databaseNames: Collection<DatabaseName>
         override fun databaseOrNull(dbName: DatabaseName): Database?
 
+        fun databaseOrThrow(dbName: DatabaseName): Database =
```
jarohen, line 554: I think we'd use `operator fun get(dbName: DatabaseName): Database` elsewhere for this kind of thing, so that you can call `dbCat[dbName]` in Kotlin

jarohen (approved):
nice one @Akeboshiwind 👏 

I might be tempted to try slim down the Allium (I'm aware the rest of our Allium is verbose too) - I've just released a chalk:allium-voice skill which aims to do just that. Specifically, the main difference I'd propose is that Allium should be completely code agnostic. 

Also, I completely take the point about Iterable vs Seqable. Elsewhere, I think I've tried to standardise on Iterable being an Iterable<Map.Entry<...>>, which I think would solve the issue.

More generally, when you make changes like that Iterable/Seqable one, ship them, reduce the size and scope of the PR 🙂 :shipit: Especially if it's fixing existing bugs - try to ship those to main separately to the changes that advance the behaviour.

core/src/main/kotlin/xtdb/database/Database.kt
Akeboshiwind: It seems like `m[k] = v?` is the convention over throwing an exception, not sure it's worth the change

Issue #5841, opened 2026-07-31
Multi-DB stage 4 unit 8.5: make KafkaCluster genuinely partition-aware


mbutlerw:
## Summary

Unit 5 gave the Kafka log *interface* conformance — the log methods take a partition, the subscription listener is singular per partition, and it compiles. It didn't make `KafkaCluster` honour any of it. The file still pins the topic-partition to zero, holds a single counter for the latest submitted offset, calls `.single()` in the rebalance handler, keeps one listener state per subscription, and checks at open that the topic has exactly one partition.

This unit closes that gap. Nearly all the churn is in one file.

## Context

Not one of the original eleven units — it surfaced while landing unit 5, at the point where interface conformance and transport behaviour came apart.

## Future state — two pieces

**Piece A, transport routing.** The mechanical half. Every hardcoded zero becomes the partition the interface now carries: per-partition latest-submitted-offset state; the real partition in every read, write and producer path; topic auto-creation at the configured partition count for both the source and the replica topic; and the partition-count check becomes a mismatch check against the configured count rather than a hardcoded one.

**Piece B, the rebalance state machine.** The subtle half. The single listener-state field becomes a per-partition map; the assigned and revoked callbacks iterate and fan out per partition; the single transition-completion deferred becomes one per partition, routing completion back to the right partition's state; and seek-after-wakeup becomes partition-aware, so a wakeup mid-seek only re-seeks the still-pending partitions.

Piece B is where cross-wiring bugs would live if a bug lived anywhere. The Kafka client hands us a collection on every rebalance callback and we have collapsed it to a single element from the start; genuinely fanning it out is where state-map bookkeeping errors become possible.

## Invariants

**The transactional id must not change for partition 0.** It stays exactly `$dbName-leader` at partition 0; only partitions above zero take a suffix. Suffixing unconditionally would open a window during a rolling upgrade where an old node's producer id and a new node's map to different producer sessions inside Kafka — fencing wouldn't apply between them, and both nodes could commit concurrently for the duration of the deploy. This is the one genuine correctness constraint in the single-partition compatibility story: everything else there is cost or convenience, but fencing is keyed on the id string and no amount of code-side defaulting substitutes for keeping it stable.

**The partition-count guard becomes a mismatch check, not a drop.** Dropping it silently would let a three-partition topic open against a single-partition config — an operator error, for instance an accidental topic recreation taking the broker default. Loud failure at open is what the current code gives, and the mismatch check preserves that for any configured count.

[...]

Issue #5680, opened 2026-06-05
ext-source: source-log watermark can regress across rapid leadership churn


jarohen:
## Summary

In **ext-source mode**, under rapid leader/follower churn the source-log watermark (`latestSourceMsgId`) can regress on a follower and trip the monotonicity check in `Watchers.notifyMsg` (`srcMsgId N < latestSourceMsgId M`). Found while building the indexer simulation tests in #5679. The dominant window is fixed (see "Fixed so far"); a narrow residual window remains open.

## How it was found

An intermittent failure of `LogProcessorSimTest."multi-node leadership changes preserve block catalog consistency"` (added in #5679) — e.g. `@WithSeed(-2133080470)`. It's ext-source-only: there the source log carries `FlushBlock`s (txs arrive off-log from the external source) and the source-log watermark only reaches the replica at block boundaries. In non-ext-source mode every tx is a source-log message carrying its own `srcMsgId`, so followers never lag.

## Mechanism

Reconstructed from a trace of the failing seed (two nodes A and B, sharing one database; `[A]`/`[B]` are their per-node watermarks):

```
TERM 1 — node A is leader:
  leader ← source 0..4 (FlushBlock)      [A] notifyMsg src 0→1→2→3→4   ← A's watermark now 4 (local, NOT replicated)
  leader → replica 4,5,6: ResolvedTx(src=null)   (ext-source txs)
  follower B consumes 4,5,6:             [B] notifyTx … src stays -1   ← B never sees A's source progress

LEADERSHIP FLIPS:  A → follower (keeps its local src=4),  B → leader

TERM 2 — node B is leader:
  resuming after -1                      ← B resumes from ITS OWN stale watermark
  leader ← source 0 (FlushBlock) …       ← re-reads the source log from the start
  finishBlock src=0 → b0
  leader → replica 10: BlockBoundary(b=0, src=0)

FAILURE — node A (now a follower) consumes replica 10:
  [A] notifyMsg src=0 (was 4)   →   IllegalStateException: srcMsgId 0 < latestSourceMsgId 4
```

Root issue: a demoted leader keeps a high *local* watermark that was never committed to the replica, then rejects the new leader's legitimately-lower replay.

## Fixed so far

[...]

PR #5927, opened 2026-08-22
feat(indexer): the replica log now elects its own leader

### LINKED ISSUE #5917: The replica log should elect its own leader, not Kafka's consumer group
- **Leadership is elected by Kafka's consumer group, so the fencing term has no durable home.**
  The term is that group's `generationId`, and the chain when it regresses is: coordinator sweeps a memberless group → the generation restarts at 1 → it sits below terms already stamped on the replica log → every reader fences the new leader's writes → that database refuses to ingest until an operator raises `termEpoch` by hand (#5904).

- **The replica log can arbitrate leadership itself, with no vote and no consensus layer.**
  A node claims by appending a no-op above the highest term it has seen, and leads if that record is the first in the log at that term or higher. A Kafka partition is already the total order that Raft's RequestVote exists to synthesise, and the drain-to-claim-offset step already delivers the vote's other job.

- **`termEpoch` goes, because a log-derived term cannot regress.**
  The block catalog already persists the term and retention already cannot cut below the block boundary — the durable backing the borrowed counter never had.

- **This is #5864's cheapest slice, and it stands on its own.**
  It takes only `K2` (election) and `K3` (the term), leaving the durable log and the ingest ecosystem with Kafka, so there is no data-path exposure.

- **One new safety condition, `G2`.**
  The conferring test has to be evaluated over the log prefix strictly *before* the claimant's own claim — which is not where `checkUnfenced` sits today.

## Context / Motivation

- **`termEpoch` is a workaround whose root cause is the borrowed term**, and #5904 was patched rather than fixed.
  The fix was `enable.auto.commit=true` purely as a group keepalive (`KafkaCluster.kt:257`) — offsets we never read back, committed to stop the coordinator sweeping the group. The failure mode it defends against returns whenever the sweep wins the race, and the recovery is still manual.

- **#5841 is about to invest in the code this would delete, and that sequencing is the *why now*.**
  "Make `KafkaCluster` genuinely partition-aware" is per-partition fan-out of the group rebalance callbacks — `listenerState` becoming one entry per assigned partition (`KafkaCluster.kt:269`). Under self-election there is no rebalance callback and no group subscription, so that work either lands first and is then removed, or is skipped.
  #5841 is left alone until this lands: its transport-routing half is data plane and survives untouched, and its rebalance half sits entirely inside the code this card deletes.

- **It is #5864's cheapest slice, and stands on its own.**
  Split by which of the four things Kafka supplies there each one moves.

  - **It takes `K2` (election) and `K3` (the term)** — the two #5864 identifies as fitting least well.

  - **It leaves `K1` (the durable log) and `K4` (the ingest ecosystem) with Kafka**, so nothing on the data path moves.

  - **It removes the consumer-group protocol from what XTDB requires of a log**, which any later move on `K1` needs first.

- **A class of Kafka coupling goes with it.**
  Group-coordinator failover, session timeouts, the empty-group sweep and #5817's Event Hubs-shaped compatibility problems stop being ours to reason about.

## Current state

- **Election is consumer-group assignment over the *source* topic** (`Database.kt:420` into `KafkaCluster.SharedGroupConsumer`).
  The assigned member becomes leader and then consumes source records through that same group consumer, which is why Kafka's poll thread doubles as the role-change serialization point.

- **The term is `generationId` under an operator-declared epoch** — `LeaderTerm.of(termEpoch, generationId)` at `KafkaCluster.kt:291`.
  `LeaderTerm.kt` documents at length why the epoch has to be *declared*: the counter orders elections only within one incarnation of the group.

- **The fence already lives on the replica log, not on the group.**
  A leader stamps its term on every replica record and confirms a write only on reading it back at its own term; a higher term read back means resign (`LeaderLogProcessor.applyReplicaMessage:201`, `FollowerLogProcessor.kt:200`). Kafka supplies the *agreement*, not the shape.

- **The claim-and-read-back primitive already exists.**
  `LogProcessor.claimLeadership` appends `ReplicaMessage.NoOp(termId)` and keeps its `msgId` as the replay target (`LogProcessor.kt:282`); `termFence.checkUnfenced` at `:290` already asks whether someone got there first.

- **Every node already tails the replica log as a follower** (`LogProcessor.tailReplica`, launched at `LogProcessor.kt:275`), and every node already attaches every secondary database.
  So the observation point an election timeout needs is in place, with no new subscription and no new consumer. What is missing is that no log *reports* a read that found nothing on that path, so there is nothing for a timeout to count.

- **Borrowing the group costs a run of workarounds in `onPartitionsAssigned`**, one per thing the group protocol will not do unaided.

  - **`enable.auto.commit` as a keepalive** — offsets nothing reads back, committed to stop the coordinator sweeping the group.

  - **`seekToEnd` as a placeholder** — `poll()` throws `NoOffsetForPartitionException` on a freshly-assigned partition even while it is paused (`KafkaCluster.kt:284`).

  - **Seek before resume** (#5633).

  - **`TransitionComplete` posted back onto the poll thread**, which must not block.

## Future state

Goal: **leadership for each `(db, partition)` is decided by its own replica log, and the term survives every restart around it.**

- **Sufficiency leans on four things already true**, one per step the election would otherwise have to build.

  - **The consume-back fence** — a leader confirms a write only on reading it back at its own term.

  - **The drain to the claim's offset** — a promoted node reads everything before its claim before it leads.

  - **Confirmation in strict offset order.**

  - **Every node already follows every replica log.**

- [x] **`G1` A node claims leadership by appending a no-op at a term above the highest it has seen, and leads iff that record is the first in the log at that term or higher.**
  The claim's offset comes back from the append, so the test is a pure function of the log prefix and every reader computes the same answer for every prefix. Two nodes claiming the same term is resolved by offset order.

- [x] **`G2` The conferring test is evaluated over the prefix strictly *before* the claim.**

  - **`checkUnfenced` cannot answer it** — it runs after the follower has caught up *to* the claim, so the claimant's own record is already folded into the running maximum.

  - **`TermFence.admit` can** — it rules and folds in one operation, called by the partition's tail in log order. That is the one place "the highest term strictly before this record" exists.

  - **The claim needs no identifier of its own**, because its position is its identity. That retires `LeaderLogProcessor`'s "terms are unique per leader, so an equal-term record IS our own" assumption rather than patching around it.

- [x] **`G3` A node claims only when it could legitimately lead.**

  - **After a randomised timeout with no record arriving**, and never if it must not lead — a read-only secondary.

  - **Measured across the node's own reads rather than against the clock**, which is also what stops a node that is not being delivered to from claiming at all.

  - **Except on a log no term has ever led**, which is claimed on sight, so cold start and the test suite don't pay for an election.

- [x] **`G4` A leader asserts leadership often enough not to be challenged.**
  Any leader write is an implicit assertion, so an explicit no-op is only needed on an *idle* partition.

- [x] **`G5` Losing a claim is a backoff-and-retry, not an operator-facing halt.**
  Today it is an operator-facing anomaly naming `termEpoch`; under self-election it is the ordinary outcome of a contested election, and must not poison the database's watchers.

- [x] **`G6` The term is recovered from durable state at startup** — seeded from `blockCatalog.boundaryTermId`, and thereafter only ever raised.
  **Expected of the operator's Kafka configuration**, not of us: replica-log retention must not cut below the last block boundary. Already required, because followers resume from `boundaryReplicaMsgId`.

- [x] **`G7` `termEpoch` is gone** — from `LeaderTerm`, from `DatabaseConfig`, and from the `inMemoryLog` / `localLog` / Kafka YAML.

- [x] **`G8` Election no longer uses a consumer group, and `openGroupSubscription` leaves the `Log` interface.**
  The source log becomes a plain `tailAll` launched inside the leader's own term, so the pause/seek/resume sequence and the poll thread as serialization point both go. The in-memory and local logs elect by the same rule as Kafka rather than by an in-process counter.

- [x] **`G9` The replica topic's `acks` and `min.insync.replicas` are asserted on connect**, alongside the existing partition-count check.

- [x] **`G10` The specs carry the log's obligations and the claim's lifecycle**, and `LeaderTerm`'s "the epoch must be declared" rationale is replaced rather than left contradicting the code.
  What belongs in a spec at all is settled by whether an entity's identity survives a behaviour-preserving refactor.

  - **`DurableTotalOrder` gains one invariant: an empty read witnesses the tip.**
    A wedged reader cannot produce an empty read, because the reader had to ask — stated without naming `poll`, `openTail` or a coroutine. Its existing seven need no change.
  - **`G1`–`G6` land as rules and invariants in `db.allium`**, which already carries the fencing invariants they lean on: at most one leader's writes confirmed, and a superseded leader resigning before it confirms past the claim.

    - **Claiming**, with `G3` in its `requires:`.

    - **The idle assertion**, which is what makes an empty read mean nobody leads.

    - **A resignation that leaves the database queryable.**

    - **A term monotonic from the persisted boundary.**
  - **Conferral is a derived value over the prefix, not a rule.**
    Rules have no outputs, and nothing needs to return a verdict, because every reader computes the same answer for every prefix. `G2`'s strict-prefix condition is part of that derivation rather than a separate mechanism.
  - **The claim carries the lifecycle** — `pending -> conferred | fenced`, both terminal.
    A claim is a record with a term and a position, so its status is the same for every reader. `following` / `claiming` / `takingOver` / `leading` is the claimant's internal state, and it moves whenever the decomposition does.
  - **`SubscriptionListener`, `actor Transport`, `TransportGuarantees` and `OQ1`/`OQ2` go with the transport.**
    All five guarantees describe rebalance callbacks, revokes and transitions in flight; `actor Transport` is Kafka's poll thread given the standing of a domain participant. Both open questions concern races that stop being expressible once there is no transport.
  - **`db.allium`'s `variant Leader` / `variant Follower` on `LogProcessor` keep their shape; the partition's state moves off them.**
    What carries identity is the state that outlives a role — `max_leader_term` above all — so that moves up onto `entity LogProcessor`, and the variants go on describing the two role behaviours the partition's reader hands records to.

## Invariants / Constraints

- **At most one leader's writes are ever confirmed per database.**
  Unchanged as a property; only its arbiter moves. The argument is offset order: a superseded leader confirms in offset order → it reaches the higher-term claim before anything appended after it → it resigns there → nothing it acknowledged lies beyond the incoming leader's drain target.

- **Leader completeness comes from the log rather than from a vote.**
  A new leader need not *hold* every committed entry, because the log is external and we never truncate it. It reads them, and the drain to the claim offset already happens before it leads.

- **In-order, gap-free delivery is the sole premise under that.**
  This graduates from an assumption about the log to a stated demand on it, because a claimant handed a gap concludes it won an election it lost:

  ```
        replica log     A (following, fence=T)    B (following, fence=T)
  a1    @9  NoOp(T+1)   poll empty → w
  b1    @10 NoOp(T+1)                             poll empty → w
  b2                                              r(@10); @9 never delivered
  b3                                              T+1 > T → leads@T+1
  a2                    r(@9); T+1 > T → leads@T+1
  ```

  Both lead. `b2` is the gap and `b3` is the verdict it produces, and nothing downstream can detect either — every reader's answer is a function of the prefix it was handed.

- **Acknowledgement latency does not change.**
  Confirmation stays on consume-back. This changes who may write, not when a write counts.

- **`acks=all` with `min.insync.replicas >= 2` graduates from a durability precondition to a safety one.**
  An unclean Kafka leader election that truncated a claim would let two readers disagree about who won, hence `G9`. `KafkaConfigMap.openProducer` composes as `mapOf(xtdbDefaults) + this`, so user properties currently override XTDB's `acks=all`.

- **An ambiguous produce outcome must be treated as a lost claim.**
  A claimant that does not know its claim's offset cannot evaluate the conferring test, and retrying at a higher term costs liveness only.

## Out of scope

- **Leadership balance.**

  - **Cooperative-sticky assignment spreads leadership across the fleet today; randomised timeouts do not.**
    Placement becomes random and — because leadership only moves on failure — stable: a random walk that accumulates skew, rather than a repeated sample that averages out.

  - **Accepted for now** — deployments in play have two or three databases and `partitions` is still pinned to 1, so the exposure is small. Revisit from observed impact.

  - **#5864's "Leadership balance" section carries the analysis**, and a graceful resign no-op covers the rolling-restart case cheaply if it bites first.

- **Owning the log itself.**
  `K1` and `K4` stay with Kafka; #5864 tracks that question.

- **The `breaking change` label.**
  `termEpoch` landed in `54ef78006` (#5817) and the only tags containing it are `v2.2.0-rc0`/`rc1`, against a latest GA of `v2.1.0`. `groupId` is the same: `v2.2.0-beta1` onwards, never in `v2.1.0`. Neither has shipped in a release, so per `AGENTS.md`'s released-or-not test removing them is a plain change with no `!`.

## Alternatives considered

Split by what could supply the term and the agreement.

- **`A1` Keep the consumer group and derive `termEpoch` automatically.**
  Any epoch derived from the log inverts the fence: a stale leader can still *read* → it observes the progress of the leader that legitimately superseded it → it derives an epoch above that → it claims a term above the leader that beat it. `LeaderTerm.kt` sets this out, and there is no source for an automatic epoch that is not either the log or another external system.

- **`A2` Keep the consumer group and stop it being swept.**
  What #5904's fix does, and it leaves three things standing.

  - **The window narrows rather than closing** — the sweep still wins the race sometimes.

  - **An offset commit stays in the code**, existing only as a keepalive.

  - **The recovery stays manual.**

- **`A3` Run a consensus group per `(db, partition)`.**
  #5864's current best guess, and strictly more machinery for the same result: the log is external → agreement is the only thing missing → the log already supplies it. What a consensus group adds beyond that — membership, placement, a control plane — is not what this card needs.

- **`A4` Have a control plane assign leadership, PacificA-shaped.**
  Also #5864's. It makes failover depend on the control plane, and needs a bolted-on verify-by-quorum step to recover leader completeness — which the log gives for free here.

## Decision rationale

Against each alternative on the point that differentiates it.

- **vs `A1` / `A2` — the term gets a durable home instead of a longer fuse.**
  Both keep a counter whose lifetime is the consumer group's. A term derived from the replica log cannot regress: the block catalog persists it → retention cannot cut below that boundary → every restart reads it back. `LeaderTerm.kt`'s objection was to deriving an *epoch* to sit above a borrowed counter, and it does not survive replacing the counter.

- **vs `A3` / `A4` — no vote is needed, so no consensus is needed.**
  Split by the two jobs a vote does.

  - **Ordering concurrent candidates** — what RequestVote exists to synthesise, and a Kafka partition is already that total order.

  - **Leader completeness** — already delivered by the drain to the claim's offset.

  What remains is a randomised timer and a comparison.

- **The cost of having no quorum** is that nothing can *refuse* a disruptive challenger the way PreVote does.

  - **The obvious local substitute vetoes every election.**
    Require the claimant to be at the log's end → that needs a deliverable-tail position → a replica log a transactional producer ever wrote reports an end counting control records → a `read_committed` consumer never receives those → every reader sits permanently short of the end → nobody may ever claim.

  - **A claimant cannot claim faster than it abandons, which bounds the residual.**
    A spurious failover costs a transition rather than a rebuild, because a demoted leader keeps its live index. Liveness, never safety.

## Implementation

- **An elector inside the indexer drives the transition that already exists.**

  - **`claimLeadership` / `transitionToLeader` survive unchanged** — what changes is who decides, and where `termId` comes from.

  - **The `Log` interface's `openGroupSubscription` / `SubscriptionListener` seam goes**, and with it the poll thread as serialization point. The elector owns a serialization point we may suspend on.

  - **`G1`–`G5` become testable against `InMemoryLog`**, before Kafka is involved.

- **`G2` is the one new safety condition, and wants its own red-green.**
  Two claimants at the same term, and a claim whose prefix already holds a higher term.

- **The elector takes an injectable driver**, so the failure modes no real log will produce on demand — a delivery gap, an append accepted but delayed, an append refused, retention cutting under a claimant — are testable.

- **Remove the group subscription and `termEpoch` last**, once the elector carries leadership on every log implementation.

jarohen:
Resolves #5917.

A database's leadership is now a function of its replica log's contents, computed identically by every node, rather than an assignment handed down by Kafka's consumer group.

- **#5841 is held until this lands, and that sequencing is why it goes now.**
  Its rebalance half sits entirely inside the code this deletes, so investing there first would be work thrown away; its transport-routing half is data plane and survives untouched.

- **A log no longer has to be able to run an election.**
  What XTDB asks of a log is a durable ordered log it can tail. `InMemoryLog`, `LocalLog` and `ReadOnlyLocalLog` each carried an `elections` counter that existed only to satisfy the other half of that contract.

- **A node claims by appending a no-op one term above the highest term it has read.**
  No vote, no coordinator, no external transport.

- **A claim confers leadership exactly when nothing at or above its term precedes it.**
  Its identity is its position, so the log's own order settles a contested election with nothing else to consult.

## What changes

Leadership used to arrive from outside the data it governs: the consumer group assigned a member, and that member's `generationId` became the fencing term. Both halves are gone.

- **The election is now the same read every node was already doing.**
  Every node tails every replica log to build local state, and that read is now also the observation leadership is decided on — so there is no second subscription and no coordinator.

### The state machine

One reader per partition drives it, and it outlives every role it drives — so a role ending is something it observes rather than something that stops it.

- **Two states, and claiming happens inside one of them.**
  There is no candidate state: a claim is a record a follower has written and not yet read back, and it goes on following — indexing, answering queries — until it does.

  ```
  Following ──(claim read back, nothing at or above its term preceded it)──► Leading
  Following ──(claim read back, something at or above its term preceded it)─► Following
  Leading   ──(a record arrives above this term, or the term's own job ends)► Following
  ```

- **Each pass of the reader is three steps, run in order.**

  1. **Poll the replica log for a freshly drawn election timeout.**
     Drawn per call, and that redraw is what makes two candidates converge on one of them.

  2. **Re-open a follower if the live role's job has completed.**
     A term ends itself, so nothing external demotes us; `demoteLeader` has no caller and goes.

  3. **Claim on an empty poll, otherwise hand the records to the role one at a time.**
     Empty is load-bearing, which is why `Log.Tail.poll` now owes its callers that empty means the tip.

- **A follower claims on an empty poll, and once before reading anything where no term has ever led.**
  The high-water is seeded from the persisted block boundary, so a zero says this database has no history to be behind — and a fresh one is writable without waiting out an election.

- **A claim's verdict is taken against the high-water as it stood before its own record.**
  The partition's tail folds every record into one high-water before either role sees it. A claim needs the reading from *before* that fold: a term folded in ties with itself, and a claim that ties confers nothing.

  - **One value answers both questions the term system asks** — what every reader discards, and what a claim has to beat.

  - **A claim is taken above the terms a held block is buffering, not above the boundary they sit behind.**
    The role still buffers what arrives, and the tail has already folded those terms.

- **A leader asserts only once it has nothing else to write.**
  The clause is armed on a biased select in the append queue's loop, so it fires only when the queue is empty and restarts on every iteration the queue wins. Without it a follower cannot tell a leader that has stopped writing from one that has gone, and every quiet database would change hands on a timer.

- **Neither transition needs telling.**
  A promotion is a record the node read, and so is a demotion; nothing waits on an acknowledgement in either direction. `LogProcessor.awaitReplicaMsg` goes with the wait, and `checkUnfenced` and `LeaderSupersededException` with the refusal that used to tell the transport its granted transition was already stale.

- **Election timing is derived from how fast a log makes a write visible.**
  Kafka gets a 1s assert interval, `InMemoryLog` and `LocalLog` 100ms.

  - **The election timeout is a ratio of the assert interval, not a second number.**
    It is drawn from 5× to 10× the interval — 5–10s on Kafka, 500ms–1s in process. What the timeout has to outlast is a run of asserts, and two independently-set numbers drift until it no longer does.

### What goes on the replica log

- **Every replica message carries the term that produced it**, so there is no election record type and no channel of its own.
  The read side fences on that term, which is what lets an ordinary write double as a proof of life.

- **A follower writes exactly one thing: a `NoOp` carrying the term it is claiming.**
  Under the consumer group a follower appended nothing at all.

- **A leader writes everything else, each stamped with the term it is leading** — `ResolvedTx`, `BlockBoundary`, `BlockUploaded`, `TriesAdded`, `TriesDeleted` and `NoOp`.

- **A claim and an idle assertion are the same record.**
  Both are a `NoOp` carrying a term and nothing else, and only what precedes them separates the two: a term above everything before it confers leadership, and a term equal to the one already leading asserts it.

  - **`NoOp`'s third use is the one that differs**, carrying the source position it advances — a `FlushBlock` whose expected block index has already been cut produces no block, and the source watermark still has to move past it (#5680).

### Compared to Raft

Raft solves consensus and leader election together, because it has to build the replicated log itself. Here the log already exists, is external to every node, and is totally ordered and durable — so what is left of an election is a deterministic function of its contents, computed identically by each node with nothing exchanged between them.

- **No node ever contacts another.**
  Raft's candidate sends `RequestVote` to its peers and counts a majority. A claim here is an append that every node happens to read, and a node that loses does not hear it from the winner — it reads the record that beat it.

- **Total order replaces the majority.**
  Raft's election safety, at most one leader per term, rests on a quorum refusing to vote twice. Here it rests on the log's order: a claim confers only if nothing at or above its term precedes it, and in a given term exactly one claim can satisfy that.

- **Two candidates converge without a re-election round.**
  A Raft split vote elects nobody and the term is re-run. Here both claims land, the earlier confers, and the later is read as an ordinary record by an author that simply keeps following. The timeout is still redrawn per poll to converge two candidates, but what that avoids is wasted claims rather than a failed term.

- **A heartbeat is a write, not an RPC.**
  Raft's leader sends `AppendEntries` to each follower to suppress their timers. Here it appends one `NoOp` and every follower sees it by doing the read it was doing anyway — one record per interval, rather than one message per follower.

- **A stale leader is fenced on the read side, not the write side.**
  Raft rejects a deposed leader's `AppendEntries` at the followers. Here its writes land, because nothing refuses them, and every reader discards them on the term they carry. The trailing records of a deposed term are expected rather than anomalous.

- **Term fencing is the piece that transfers directly.**
  A leader confirms a write by reading it back at its own term, which is the argument ReadIndex makes — don't act on leadership you have not just confirmed — with the confirmation coming from the log rather than from a heartbeat round. The claim is the first instance of it.

- **The trade is that the log is a single point of coordination.**
  Raft goes on electing while a majority is up, and no one node is special. Here the log's availability is the election's: a node partitioned from it cannot claim, and nor can anyone else. XTDB already required the log for durability, so this adds no dependency — but leadership now inherits the log's availability rather than a quorum's.

### Interleavings

Two candidates at the same term — position decides, and the loser needs no notification:

```
      replica log     A (following, fence=T)       B (following, fence=T)
a1    @9  NoOp(T+1)   poll empty → w
b1    @10 NoOp(T+1)                                poll empty → w
a2                    r(@9); T+1 > T → leads@T+1
b2                                                 r(@9); T+1 > T → fence := T+1
a3                    r(@10); T+1 = T+1 → admitted
b3                                                 r(@10); its own claim, T+1 = T+1
b4                                                 → clears claim, keeps following
```

- **B's claim is admitted like any other record and confers nothing**, because `a1` got there first.

- **A claim left set stops that node ever claiming again**, which is what `b4` clears.

Superseding a live leader, and the tail of writes it emits before it finds out:

```
      replica log            A (leading@T)          B (following, fence=T)
a1    @11 ResolvedTx(T)      resolves → w
b1    @12 NoOp(T+1)                                 poll empty → w
a2    @13 ResolvedTx(T)      resolves → w
a3                           r(@12); T+1 > T → resigns
b2                                                  r(@12); T+1 > T → leads@T+1
a4                           term ends → re-opens a follower
b3                                                  r(@13); T < T+1 → fenced
```

- **A resigned leader's trailing writes land and are discarded by every reader**, rather than being prevented from landing.
  `a2` is appended before `a3` tells A it has lost, so `@13` reaches the log and every reader then fences it as B does at `b3`.

- **The simulation's raw-log invariants fold the fence as a reader does.**
  Read without the fence, those trailing records are duplicate `txId`s.

### What the simulations cover

Seven seeded scenarios in `LogProcessorSimTest`, a hundred iterations each by default, sharing one `assertInvariants` — so an invariant runs wherever it could be violated rather than wherever a scenario's author thought to check it.

- **Asserted in every scenario**: one writer per term, no node's watchers poisoned, every node caught up to the log's end and agreeing with the others, every uploaded block's files present, boundaries matching uploads one-to-one and contiguous from zero, one cut per block index, and no aborted transaction's rows in any live index.

- **A leader's idle assertion is driven rather than timed.**
  The assert clause arms from a channel the test fires, so a seeded run reaches a path a wall clock could not. `stable leader with sustained throughput` fails if no assertion actually reached the log.

- **No scenario restarts a node from persisted state**, so `C2`'s rolling restart has no simulation behind it — the joiner scenario stands up a fresh node.

## Usage / migration

- **`termEpoch` goes from the Kafka, in-memory and local log configs, and `groupId` from `KafkaCluster`.**
  Both shipped only in the v2.2 pre-releases, so the Kafka page corrects its v2.2 entries in place rather than announcing a break.

  - **Logs and blocks written before this still read.**
    The deleted proto field numbers are `reserved`, so an old message decodes with them as unknown fields.

- **Deployments sharing a Kafka cluster are separated by topic name.**
  The consumer group used to separate them as well. Sharing the topics was already unsupported whatever the `groupId`s: two deployments, one replica topic → each stamps its own terms → each fences the other's writes.

- **`Log.Tail.poll` gains an obligation: an empty return means the tail reached the partition's tip.**
  It had meant only that nothing arrived before the timeout.

  - **A fetch failure, a dropped subscription or a clock of its own may no longer return empty.**
    A tail that can no longer read has to throw.

  - **A tail that reports quiet instead gets a leader fenced** — report quiet → a follower's timeout runs down → it claims → it fences a leader it simply could not see.

  - **Every in-tree implementation already satisfies it; a third-party `Log` may not.**

- **Rolling back to a pre-election build needs `termEpoch` set above the epoch already on the replica log.**
  A pre-election node stamps `(termEpoch << 48) + election`, with `election` a per-process counter from zero; this stamps `highestTermSeen + 1`, which stays inside whichever epoch band it inherited. So a rolled-back node's first term lands below terms the log already carries, every reader's fence discards its writes, and ingestion stops while the node reads as leader.

  - **One above the epoch it was upgraded from is enough**, the new scheme only ever incrementing within that band.

## Consequences

- **`C1` — leadership placement is no longer balanced.**
  The Kafka page's balance promise is withdrawn: each database elects independently, so one node can lead several while another leads none.

- **`C2` — reopening a node on a persisted directory costs an election timeout before it can lead.**
  A node with a block behind it cannot know the old leader is gone, so claim-on-sight is available only where no term has ever led.

  - **On Kafka it is five to ten seconds**, per database, paid once per node in a rolling restart.

  - **On a local log it is under one second.**
    That is why `indexer-test/merges-column-fields-on-restart` needs a restart sync of seconds rather than the 200ms it had.

  - **A restarted node also cannot produce an inherited block until it has won.**
    `indexer-test/can-ingest-ts-devices-mini-with-stop-start-and-reach-same-state` polls for its eleventh block file rather than sleeping: the last boundary was cut but not produced before the previous node closed, and awaiting the transaction does not cover the election, because a follower applies it too.

- **`C3` — `ExternalSource.onPartitionAssigned` is now called when a leader term begins.**
  The partition rather than the term is what serialises those calls.

  - **The name is unchanged**, so a source implementation reading it as a consumer-group assignment is now wrong about when it fires.

  - **Its contract is met structurally rather than by a handshake.**
    One call at a time, per-term state inside the call: the call is a child of the term's scope, and a promotion is reachable only from a role whose job has completed.

  ```
        assignment    reader (partition tail)          source
  r1    assigned@4    launches runTerm(4)
  s1                                                   onPartitionAssigned(4) → running
  r2                  term 4 ends → cancels children
  s2                                                   observes the cancel
  r3                  scope waits …
  s3    free                                           returns
  r4                  termJob completes → re-opens a follower
  r5                  claim confers → builds term 5
  r6    assigned@5    launches runTerm(5)
  s4                                                   onPartitionAssigned(5) → running
  ```

  - **A source that does not cede control promptly at `s2` delays this node's return to *following***, not merely its next term.
    `r3` is where that wait lands.

- **`C4` — Kafka's tail meets the empty-poll obligation in practice rather than structurally.**
  A consumer partitioned from the broker returns empty batches rather than throwing.

  - **The same partition stops the claim landing** — partitioned from the broker → the claim cannot be appended either → the empty poll is one the node can do nothing with.

- **`C5` — a role change hands over a block one role produced and neither adopted.**
  A boundary produces the block and the read-back adopts it, so a role can hold a block its own catalog has not moved past.

  - **Every term that ends crosses that window**, which makes the handover ordinary rather than rare.

  - **The reader hands the block to the follower it re-opens** — term ends holding b → the follower inherits b → it closes b on the upload the term was waiting for.

  - **Discarding it instead would stop the partition's tail** — the reader drops b → the upload reaches a role with no block to match it against → the tail has nowhere to put it.

- **`C6` — two terms can produce the same block index.**
  A rival's claim lands between a boundary and the upload that follows it, which is the one window in which an incumbent writes no asserts — it is writing block files.

  ```
        replica log                A (leading@2)      B (following, fence=2)
  a1    @35 BlockBoundary(2,b02)   cuts b02 → w
  a2                               writes b02's files
  b1                                                  r(@35); 2 = 2 → holds b02
  b2    @36 NoOp(3)                                   poll empty → w
  b3                                                  r(@36); 3 > 2 → fence := 3, leads@3
  b4                                                  inherits b02's boundary, produces it
  a3    @37 BlockUploaded(b02,2)   finishes b02 → w
  b5    @38 BlockUploaded(b02,3)                      finishes b02 → w
  a4                               r(@36); 3 > 2 → resigns
  b6                                                  r(@37); index matches → adopts b02
  b7                                                  r(@38); b02 adopted → stale, dropped
  ```

  - **One block, written twice** — `a1` and `b4` cut from the one boundary over the one prefix.

  - **The first upload of it is the one every node adopts.**
    `closes` matches on block index, storage version and epoch rather than on term, so `a3` closes b02 everywhere and `b5` is stale by block index.

  - **Routine now, rather than never observed** — leadership moving on a read rather than on a rebalance is what makes the simulation reach this.

- **`C8` — every node now writes to the replica log.**
  A follower under the consumer group appended nothing at all, and now appends a claim on every empty poll.

  - **A write a follower cannot make costs it nothing** — the log refuses → log it and drop it → still following, indexing and queryable → claim again on the next empty poll.

  - **Reporting it instead would stop every follower at once** on a replica log that had stopped accepting writes, none of which was leading.

## Out of scope

One entry per failure mode this change leaves standing, each with its own card.

- **A leader that can append but not read stops ingestion while reading as healthy.**
  appends land, reads do not → it goes on asserting → it reads as healthy while ingestion is stopped → no follower's timeout ever runs down.

  - **The shape of the fix is settled**: the term times out on reading back its own writes and ends itself, suppressed while a block upload is in flight.

- **The block upload still runs inline on the replica-apply path**, so consume-back stalls for the whole of every upload.
  Independent of this change, and reachable before it.

## Alternative approaches

One entry per road, both reasoned against rather than tried.

- **A claim derived from a timer rather than a read can leave nobody leading.**
  Resigning is safe whether a node's silence is the log's or its own; claiming is not, because a claim fences the incumbent whether or not it wins.

  - **The failure is stable, not transient** — reader wedges → timer still fires → claim fences the healthy leader → the claimant cannot read its own claim back → nobody leads → repeat.

  - **An empty poll is the one signal a reader that is not reading cannot manufacture**, which is why the claim hangs off the read.

- **Deriving quiet from the newest record's own timestamp supersedes a leader whose clock is skewed.**
  It would remove the restart wait of `C2` entirely, but Kafka's default `CreateTime` timestamps come from the producer's clock: leader's clock skews late → its records read as old → a follower reads the log as quiet → the leader is spuriously superseded.

## Open questions

- **`Q1` — nothing pins the empty-poll obligation on Kafka's tail.**
  `C4` bounds the damage. A test that partitions a consumer and asserts the tail throws rather than reporting quiet would pin it.

- **`Q2` — the reader's own failure path is untested.**
  A tail that dies reports through `watchers.notifyError`, poisoning queries as it should, and the node stops claiming — but no test kills the tail to see it.

- **`Q4` — the garbage collector's in-flight queue assumes every record a leader applies is its own term's append, and the adopt's drain is a third path past that.**

  - **A record above the term throws, and the fence refuses one below** — the two paths the assumption does cover.

  - **A record admitted at its own position reaches the leader carrying the previous term** — admitted → held behind an open block → drained after the promotion. A `TriesDeleted` among those finds the queue empty.

  - **It takes a second node's garbage collector appending between a boundary and its upload.** Reachable without this change, and likelier with it.


jarohen:
### Chalk — what the election wants specified, and what it doesn't

**tl;dr**

- **`G10` on #5917 now carries what the specs should say; this is the reasoning behind it.**
  The question it settles is what belongs in a spec at all, which had to be answered before writing one against a mechanism this PR replaces.

- **The discriminator is already in `log-processor-lifecycle.allium`, twice over.**
  `DurableTotalOrder` covers every safety argument in this PR without a word changing. `TransportGuarantees` specifies mechanics this PR deletes. One faces the log; the other faces inward.

- **The test that separates them: does the entity's identity survive a behaviour-preserving refactor?**
  A claim is a claim regardless of who reads it. A `Listener` exists only because something once listened.

<details><summary><code>DurableTotalOrder</code> already carries this PR's safety argument, unchanged</summary>

Seven invariants, written against the consumer-group mechanism, and every safety claim here runs through one of them.

- **`OneOrderSeenByEveryReader`** — the conferring test is a pure function of the log prefix, with every reader computing the same answer for every prefix. That is this invariant, restated at the point of use.

- **`DeliveryFollowsPositionOrder`** — #5917 calls in-order, gap-free delivery "the sole premise" under leader completeness and says it graduates from an assumption about the log to a stated demand on it. It was already a stated demand, here, with the consume-back argument spelled out.

- **`AcceptedRecordsAreNeverWithdrawn` and `DurabilityIsConfiguredNotHoped`** — an unclean Kafka election that truncated a claim lets two readers disagree about who won, which is `G9`. The contract had already made that argument and nominated the fix: "where the log can report its own settings they are worth checking as the database connects".

- **`AppendReturnsItsOwnPosition`** — load-bearing in a way it was not before, now that the claim needs no identifier of its own because its position is its identity.

- **`OneAcceptedAppendYieldsOneRecord`** — "an ambiguous produce outcome must be treated as a lost claim" is its direct consequence.

- **`RetentionKeepsEverythingAboveTheBoundary`** — `G6`, the term seeded from `boundaryTermId` and monotonic from there.

A contract that survives a change of election mechanism intact was specifying the right thing: obligations on a component outside the system, in terms that hold however leadership is decided.

</details>

<details><summary><code>TransportGuarantees</code> specifies the transport this PR removes</summary>

`AtMostOneTransitionInFlight`, `NoTransitionWhileLeading`, `NoDemoteOnShutdown`, `RevokeCancelsInFlightTransition` and `TransitionIdentityDropsSupersededCompletions`, plus `OQ1` and `OQ2`, all describe rebalance callbacks, revokes, demotes and transitions in flight. This PR deletes the mechanism each of them is about.

- **`actor Transport` is Kafka's poll thread given the standing of a domain participant.**
  That thread was an artefact of borrowing the consumer group for election, not anything XTDB's behaviour required. The spec was accurate about an implementation accident.

- **`OQ1` and `OQ2` were ill-posed rather than hard.**
  Both concern races between an in-flight transition and a revoke at the serialization point, and neither is expressible once there is no transport. That they evaporate is a stronger signal than either answer was.

</details>

<details><summary>What the election adds instead</summary>

- **One new obligation on the log: an empty read witnesses the tip.**
  An empty poll return is the consumer's own positive observation that the log is at its tip, and a wedged reader cannot produce one because the reader had to ask. Stated as an obligation, that carries the entire justification for the tail inverting (#5977) at the behavioural level — and it is what makes a read-derived timeout safe where a clock-derived one is not.

- **Conferral is a derived value, and the claim carries the lifecycle.**
  Rules in Allium have no outputs, and none is needed: every reader derives the verdict from the prefix. The claim's own status is the transition graph — `pending -> conferred | fenced`, both terminal — which is decomposition-independent in a way `following` / `leading` is not.

- **The rules are the appends and their consequences.**
  `G3` sits in the claim rule's `requires:` — the node may lead, and its own reads have gone quiet. `G4` is a leader appending on an interval, which is what makes an empty read mean nobody leads. `G5` is a resignation leaving the database queryable, which was a real bug on `main` and not only here.

- **The invariants are the fencing argument** — at most one conferred claim per prefix, a term monotonic from the persisted boundary, and records below a conferred claim's term never admitted.

</details>

jarohen:
Object-role audit of `xtdb.indexer` on this branch, read at `46c63b660` against the `xtdb-object-boundaries` skill — Active Object (owns a `select`, decides what happens next) / Passive Object (completes on the caller's coroutine) / Monitor Object (passive state someone waits on). Notes for whoever picks the branch up next.

- **The roles got substantially cleaner, and the win is easy to give back**

  - **An Active Object was deleted.** `KafkaCluster.SharedGroupConsumer.pollingJob` owned the `select` that decided leadership for *every* database sharing the consumer group, from a rebalance callback on one thread. Every "the transport's serialization point" comment the indexer used to carry pointed at it.
  - **A duplicated role machine went with it** — `main` had two sealed state machines for one fact either side of an interface, `KafkaCluster.ListenerState` and `LogProcessor.State`. One survives.
  - **A Monitor Object was removed rather than patched.** `LogProcessor.state` was a `MutableStateFlow` with a reader suspended on it and two writers held apart by a hand-argued non-race; it is now a `@Volatile var` with the reader as sole writer and nobody waiting, the wait having moved onto the role's `Job`.
  - **`LogProcessor` is now plainly Active** — it previously handed the transport a `Deferred` running on the database's scope, which is neither role cleanly.
  - **The regression to watch for is a new `launch` on a stored scope inside the leader term.** The three subordinates expose `armSelect()` and run no loops of their own, and `runLeaderTerm`'s `coroutineScope` joins every child; that is what keeps the term's choice in one place. The one existing exception is `extSrcProc.run()`, now a sibling `scope.launch` cancelled by hand rather than a child of `termJob` — it has a reason (#5711), and it is where the third bullet below lives.

- **`tailReplica` treats a failed claim append as a fault, and it is unrecoverable twice over**

  `claimLeadership()` does a durable `replicaLog.appendMessage(NoOp(termId))` and is called from two points inside `tailReplica`'s `try`, whose `catch` is `if (!e.isShutdownSignal) watchers.notifyError(e); throw e`. An IO failure is not a shutdown signal, so a broker blip or an S3 5xx while claiming does two irreversible things to a node that was a perfectly healthy follower.

  - **`Failed` latches**, so the database is unqueryable until the process restarts.
  - **The rethrow unwinds the reader**, a single `scope.launch` with nothing to restart it, so the partition stops applying replica records for the life of the process.
  - **The right policy is already stated 200 lines below, for the neighbouring failure.** `cutOverToLeader`'s catch reopens a follower, logs, and pointedly does not rethrow: *"A promotion that fails costs this node the leadership and nothing else… Rethrowing would unwind the partition's reader, so the follower just re-opened would never read another record."* A failed claim costs this node the leadership and nothing else by exactly that argument, so wrapping the two `claimLeadership()` call sites the same way — catch, log, leave `claimMsgId` unset, keep polling — is local and cheap.
  - **Nothing covers it**: `LogProcessorTest` has eight election cases and none fails an append.

- **`Leading.retire()` frees term state with one pump unjoined**

  `retire()` is `extSrcJob?.cancel()` then `proc.close()`, reached once `role.job.isCompleted` — which joins everything under `termJob`. `extSrcJob` is a sibling on the database scope, so it is the one thing not joined, while `LeaderLogProcessor.close()`'s own kdoc gives its precondition as *"safe only once the term's job has been joined, so the persister and the pumps are gone"*.

  - **Not live today** — the ext-source coroutine reaches the resolver only via `armSelect` on the term's coroutine, and a cancel while suspended in `tasks.send` routes through `onUndeliveredElement` → `abandon`, which touches no allocator.
  - **Still worth closing**, because it is a precondition on freeing a child allocator, which is the one ordering gotcha `dev/CODING.adoc` says earns a comment at its site. Making `retire()` a `suspend fun` that `cancelAndJoin()`s does it; the only call site is already suspend.

- **Don't take `live-index.allium` at face value on this branch**

  - **`log-processor-lifecycle.allium` matches the code closely** — `ClaimFollowsObservation`, `FollowerIsSoleReplicaConsumer` and `ReopenFollowerOnTermEnded` all land on real code.
  - **`live-index.allium` is behind.** It still has leader election driven by the source-log group subscription, has the transport re-following "on the next rebalance" after supersession, gives the leader a second replica-log consumer against `FollowerIsSoleReplicaConsumer`, and models a third transition-processor variant the state machine no longer has. The code is right; an agent designing against that spec will be designing against a model that no longer exists.
  - **`ElectionDriver` is unspecified in either file** — no election timeout, no assert interval, no config block — and those two clock waits are what the whole branch turns on.

jarohen:
**Two fixes underneath this, both latent on `main`, both landing there separately rather than in this PR.**

SLT failed on this branch with `InMemoryLog replay buffer rolled past offset 0`. That message was a guess at a cause it couldn't observe — the actual condition is "this isn't the record I expected", and one swallowed record is enough.

- **`InMemoryLog.poll` read its channel through `withTimeoutOrNull { receive() }`, which can consume a record and then throw it away.**
  A `receive` commits to its element when its continuation is *dispatched*, not when the sender hands it over: the sender CASes the element in and spends the channel cell, then queues the resumption, and a timeout landing before that task runs leaves `DispatchedTask` discarding a successfully-resumed value in favour of the cancellation. `cancelCompletedResult` is the only hand-back hook and it calls the `onUndeliveredElement` that a `produceIn` channel gives no way to supply. A `select` commits on the sender's thread instead — lose to `onTimeout` and the record never leaves the channel.

  - **What it cost: a node that led for fifteen minutes and resolved nothing.** The source tail lost source record 0, so its submitter blocked forever and nothing further was submitted to expose the gap. The tail polled empty until `flushDuration` elapsed, whereupon its own `maybeFlushBlock` appended a `FlushBlock` to the log it was tailing — the first record it ever received, at an offset that was not 0.

  - **`InMemoryLog` was the only log exposed**, being the only one whose tail position depends on consuming an element; `LocalLog`, `ReadOnlyLocalLog` and Kafka each wait on an offset and re-derive what to read. It arrived with #5977, so it is fixed against that rather than here.

- **`Catalog.syncAll0` enumerated the node's databases once and awaited each**, so a database that joins the node part-way through a sync — which is to say, the database whose attach that sync is replaying — was never synced at all. Now re-enumerated until a pass finds nothing new, with `syncAll`'s existing timeout bounding a log that keeps attaching.

**Neither is caused by leader election.** Both windows have been open since their respective changes landed; a restarted database now waits out an election timeout before its first term, and that is what made them wide enough to hit.

Issue #5408, opened 2026-03-30
Normalize metric names - "xtdb" prefix, snake_case, etc


egg-juxt:
Some ideas below. These are all backwards-incompatible of course, so better handled at a major version change.

### Prefix all metrics with "xtdb"
Makes a big difference for finding and having a comprehensive view of XTDB-specific metrics when configuring a Grafana panel for example, as there are many other metrics for Kubernetes pods.

### snake_case
Just use dots and snake_case for metrics. (micrometer.io will transform to all snake_case for Prometheus)

### Follow Prometheus conventions

Check that the [Prometheus guidelines](https://prometheus.io/docs/practices/naming/) are followed. In particular, but not only, for choosing units.

### Remove the "node_id" tag
Prometheus already offers the standard "instance" tag. The "instance" tag can be configured:
```
extraScrapeConfigs: |
  ...
      - source_labels: [__meta_kubernetes_pod_name]
        target_label: instance
```

### Publish offset rather than msgId
It's a friendlier metric as it resets on epoch increment, rather than doing a huge step. Prometheus is prepared for metrics that reset to zero from time to time.

Issue #5404, opened 2026-03-30
Add additional metrics for monitoring of single writer


danmason:
In order to better monitor single writer nodes, there are a few metrics of interest that we would like to capture:
- [x] Multi DB transaction lag/latest completed/latest submitted. (#4623)
- [ ] Number of leadership elections for the cluster over time.
- [ ] Time taken for leadership elections.

In addition, we could also extend some of the tracing of queries in pgwire:
- [ ] Time spend awaiting should be included as a child span in the tracing.

PR #5944, opened 2026-08-24
fix(metrics): a node's meters outlive the roles it has held

### LINKED ISSUE #5943: a node's metrics should not depend on which role it has held
## tl;dr

- **Several meters are registered by objects whose lifetime is a leader term, so what a node reports depends on which roles it has happened to hold**
  Re-registering an existing meter ID is idempotent — Micrometer hands back the meter that is already there — which is fine for a counter or a timer, because the caller then records through the shared instance. It is not fine for a gauge bound to an object: the second registrant's object is discarded, and `DefaultGauge` holds only a **weak** reference to the first one.

- **The Postgres source's three gauges stop reporting after the first failover** — `last_event_time`, `connection_state` and `wal_lag_bytes`. Once the superseded term's `PostgresSource` is collected, the weak reference is cleared and the gauge reports `NaN` for the rest of the node's life, while the live term updates fields nothing reads.

- **`tx.op.timer` is absent entirely on a node that has never led**, which is what `xtdb.information-schema-test/test-metrics` trips over.

- **The correct pattern is already in the tree** — `Database` registers its `node.tx.*` gauges at database open through a supplier and removes them on close.

- **Why now**: #5917 replaces consumer-group election with self-election, which makes a node start as a follower and change role far more often. The same code was previously almost always leader from startup, so per-term and per-node registration were indistinguishable in practice.

## Context / Motivation

Under Kafka consumer-group election a node was assigned leadership at startup and usually kept it, so a meter registered inside a leader term behaved like a node-level meter in every deployment anybody looked at. Self-election (#5917) removes that coincidence: every node begins as a follower, elects, and changes role whenever leadership moves.

That makes the lifetime mismatch observable rather than theoretical. It surfaced as a unit-test failure on the #5917 branch, but the failing test is the least of it — the test only notices the *absent* metric, not the frozen one.

#5333 is the same mechanism already being observed in a running cluster, reported as a spurious WARN. Fixing the lifetimes properly should close it.

## Symptoms

- **A CDC database's Postgres-source gauges stop reporting after its first leadership change**
  `xtdb.postgres_source.last_event_time`, `.connection_state` and `.wal_lag_bytes` briefly report the superseded term's last values, then `NaN` once it is garbage collected — so the transition is non-deterministic in timing, which makes it awkward to recognise. The saving grace is that `NaN` is visibly broken rather than plausibly healthy; a frozen number that read as "no WAL lag" would be considerably worse.

- **`tx.op.timer` is missing from `xt.metrics_timers` on a node that has not yet led**
  `xtdb.information-schema-test/test-metrics` fails on this: the query runs before the node's first term opens.

- **The follower's replica meters don't exist until the node has followed at least once**
  `xtdb.replica.process.timer`, `xtdb.replica.block.buffer.timer` and `xtdb.replica.block.buffered.records` are absent on a node that led from startup and has never resigned.

- **A WARN on every leader→follower switchover** — #5333.

## Root cause

`MeterRegistry.getOrCreateMeter` returns the existing meter when an ID is already present, without invoking the new supplier (`MeterRegistry.java:688-696` in micrometer-core 1.17.1, which is also where #5333's WARN comes from). Registration is therefore idempotent, and for a counter or timer that is the whole story: the caller records through the instance it was handed, which is the shared one, so measurements keep landing correctly.

A **gauge bound to an object** is different, because a gauge is not recorded through — it is *sampled from* whatever it captured at registration. `DefaultGauge` stores that as a `WeakReference` and returns `Double.NaN` once the referent is gone (`DefaultGauge.java:38,44,50,59`). So the second registrant's object is discarded, and the surviving gauge samples the superseded one until it is collected, then reports `NaN` indefinitely.

So the consequence depends on both the meter kind and the owner's lifetime:

- **`xtdb.postgres_source.*` — owner is per leader term, three of six are object-bound gauges**
  `externalSourceFactory?.open(...)` is called from `LogProcessor.openLeader`, so a whole `PostgresSource` is constructed per term. `PostgresSource.kt:91-103` binds `last_event_time`, `connection_state` and `wal_lag_bytes` to that instance's `AtomicLong`s / to `this`. After a failover the live source updates fields nothing is reading. The three counters and summaries (`events.total`, `commits.total`, `commit_lag_seconds`) are name-idempotent and keep accumulating correctly.

- **`tx.op.timer` — owner is per leader term** (`SourceLogTxIndexer.kt:63`, reached via `TxResolver.kt:111`)
  Name-idempotent, so it works once it exists; it just doesn't exist on a follower.

- **The three `xtdb.replica.*` meters — owner is per follower instance** (`FollowerLogProcessor.kt:49-73`)
  `openFollower()` builds a new `FollowerLogProcessor` on each return to following, so these are re-registered on every switchover. Name-idempotent, so mildest of the three: the effect is only that they are absent until the node first follows. This is also where #5333's WARN comes from.

- **The allocator gauge in #5333** is the same shape — a per-role allocator registering a gauge under a tag that repeats across roles.

The already-correct instances are worth reading as the contrast. `Database.kt:374-386` registers `node.tx.*` at database open, tags them `db`, keeps the handles, and removes them in `close()` (`Database.kt:155`). Two things make it right: the owner outlives every role, and the gauge is a **supplier** (`Gauge.builder(name) { f() }`) that reads through to whatever is current rather than capturing an object. `xtdb.log.leader` in `LogProcessor.init` is the same shape and is why *that* gauge correctly tracks role changes. `compactor.job.timer` and `compactor.jobs.available` are registered per database by `Compactor.openForDatabase` and are fine.

Probably wants the meters moved to the longest-lived owner that can see the state, reading through a supplier rather than binding a per-term object — but which owner that is differs per meter, and the CDC ones may need a level of indirection that doesn't exist yet.

## Invariants / Constraints

- **A meter's identity has to be stable across role changes, or its history is broken** — re-registering under the same ID is what produces the frozen gauge, and registering under a *different* ID per term would fragment the series instead. Neither is acceptable, so the registration has to happen once.
- **Removal matters as much as registration.** A meter whose owner has gone must be removed, or it reports stale values forever; `Database.close()` is the precedent.
- **`meterRegistry` is nullable throughout** and metrics are optional, so nothing here may become load-bearing for correctness.

## Out of scope

- **Renaming or re-prefixing anything** — #5408 covers naming, and this card should not touch it.
- **Adding metrics** — #5404 covers new single-writer metrics. This is only about the lifetime of the ones that exist.
- **The `xtdb.information-schema-test/test-metrics` failure is a symptom, not the scope.** Making that test await leadership would turn it green without fixing anything.

## Open questions

- **Does this block #5917, or land alongside it?** The lifetime bug predates self-election, but self-election is what makes it reachable in normal operation. If the two ship in the same release, this arguably becomes a blocker for that release rather than for the branch.
- **Should #5333 be closed by this, or kept as its own narrower fix?** It is the same mechanism, already observed in production, and the WARN goes away as a side effect of fixing the lifetimes.
- **Is `FollowerLogProcessor.kt:82`'s `Timer.start(meterRegistry!!)` worth tidying here?** It is safe — the enclosing branch returns early when the timer is null — but it makes one field's nullity stand in for another's, and this is the card that touches those lines.

## Progress

**Status**: needs re-scoping — the CDC half is superseded, the residue was dropped

The sharp end of this card is fixed by the `extsource-once` work, at the root: the external source became one per database, so its gauges can no longer be stranded by a leader term ending. #5944 attempted it at the symptom instead and was closed unmerged — the chalk comment carries the full analysis, including two corrections to this description's own Symptoms.

What this card still describes, and nobody is carrying:

- [ ] `tx.op.timer` and the three `xtdb.gc.*` delete timers are absent on a node that has never led
- [ ] `xtdb.replica.*` is re-registered on every switchover, and not removed when a database detaches
- [ ] `xtdb.log.leader` is never removed, so a detached database leaves a gauge holding a dead `LogProcessor`

An implementation of all three was written and rebased onto `extsource-once`, green on 2232 unit tests and 2747 simulation invocations, then dropped rather than carried — the commits remain reachable from #5944.

### Open Questions

- [ ] Is the residue above worth a card at all, or should this one be closed and the `xtdb.replica.*` / `xtdb.log.leader` removal folded into whatever next touches `LogProcessor`?
- [ ] The `Symptoms` section above is wrong in two places: the `xtdb.replica.*` meters were never absent (`LogProcessor` opens its first follower in a property initialiser), and `xtdb.information-schema-test/test-metrics` passes on `main` — that failure was a race the #5917 branch happened to lose. Worth correcting in place if this card survives.

jarohen:
Resolves #5943.

## tl;dr

- **What a node reports no longer depends on which roles it has happened to hold**
  Every meter that was registered by a leader term or by a follower instance now belongs to the node or to the database, and is removed when that owner goes. A CDC database's WAL-lag, connection-state and last-event gauges keep moving across a failover instead of freezing at the first term's values, and the GC delete timers and `tx.op.timer` are reported by a node that has never led.

- **Three things change on an operator's dashboards**
  `xtdb.postgres_source.*` now appears only on the node actually holding the source, rather than on every node that has ever held it — so a follower shows a gap where it used to show a flat line. Those series' three counters restart at zero when leadership returns to a node it has left. And a detached database's `db`-tagged meters disappear instead of reporting their last value for the life of the node.

- **`xtdb.Meters` is where meter ownership now lives**
  A nullable registry plus the meters registered through it, removed on `close()`. Four owners take one: the Postgres source, `ReplicaMetrics`, `GcMetrics`, and `LogProcessor` for its leader gauge.

## Implementation notes

**One owner per meter, chosen as the longest-lived thing that can still see the state.**

- **`xtdb.postgres_source.*` (six) → the `PostgresSource`, registered in `init` and removed in `close()`**
  Three of the six are gauges bound to instance state — `last_event_time` and `connection_state` to `AtomicLong`s, `wal_lag_bytes` to the source and so to its driver — and a new term's source was being dropped on the floor. Scoping the whole set to the source means the series exists exactly while this node is replicating, which is what the gauges measure. The leader term already closes its source, via `ExternalSourceProcessor.close` → `LeaderLogProcessor.close`.

- **`tx.op.timer` → `NodeBase`**
  It carries no `db` tag, so every database's indexer was already recording into one timer while each leader term re-registered it. Its owner now sits where the registry itself is opened.

- **`xtdb.replica.*` (seven) and `xtdb.gc.*` (three) → `LogProcessor`, via `ReplicaMetrics` and `GcMetrics`**
  Both are `db`-tagged and `LogProcessor` is the per-database object that builds both roles, so it constructs both sets at database open and removes them in `close()`. `GcMetrics` is threaded down to the two collectors, replacing the `MeterRegistry` each took; `ReplicaMetrics` is handed to each `FollowerLogProcessor`. With the meters injected there is no registry left at the recording sites, so `FollowerLogProcessor`'s `Timer.start(meterRegistry!!)` — one field's nullity standing in for another's — goes with them.

- **`xtdb.log.leader` → same owner, but now with an ending**
  It was already registered per database, and never removed.

**Two Micrometer behaviours `Meters` absorbs so that call sites needn't.**

- **A percentile-publishing timer or summary registers `<name>.percentile` gauges as meters in their own right**, which `registry.remove(parent)` does not reach. Left behind they read the parent that was just removed, and shadow the replacement's — the same freeze one level down. Removal sweeps them by name prefix and tag containment, which is also what keeps the sweep off another database's meters.

- **Where an ID is already taken, `register` evicts the incumbent and builds again.** See D2.

**Three corrections to the card's analysis, from sweeping every registration in `main`.**

- **`KafkaConnectSource` is already correct** — it registers inside `onPartitionAssigned` and removes everything in the `finally`, so it's the precedent for the fix rather than a second instance of the bug. Unchanged here.
- **The three GC delete timers are per-leader-term too**, which the card didn't list. They were the clearest case of a meter absent on a node that had never led.
- **The `xtdb.replica.*` meters were not actually absent.** `LogProcessor` opens its first follower in a property initialiser, so they existed from database open all along; what was wrong with them was re-registration on every switchover and no removal on detach.

## Decision rationale

- **D1: the source's three counters share the gauges' lifetime rather than keeping the node's**
  `events.total`, `commits.total` and `commit_lag_seconds` were unharmed by the bug — a counter's second registration is a harmless no-op — so leaving them registered for the node's life was available. That buys a monotonic counter at the price of six meters under two owners with two different removal rules. Prometheus handles a counter reset natively and `node-id` is a common tag, so the series is per-node either way. One lifetime won.

- **D2: `register` evicts an incumbent rather than adopting it or refusing to own it**
  Micrometer hands back the incumbent instead of registering ours when an ID is taken. Adopting it — the naive reading — means `close()` deregisters a live meter this owner never created. Refusing to own it is safe but leaves an orphan shadowing every later owner for the life of the node, which is the bug being fixed. Evicting is the only one of the three where what the registry holds afterwards is the live owner's meter. It costs a second `build` call and a Micrometer WARN in that state, and it is about five lines to revert if the conservative reading is preferred.

- **D3: `tx.op.timer` stays untagged**
  Giving it a `db` tag would have made per-database ownership sound, and would also have changed the meter's identity and broken existing dashboards. Naming is #5408's.

- **D4: `LogProcessor.close()` removes meters ahead of closing the processor, in a `try`/`finally`**
  `closeAll` is a bare `forEach` with no `finally`, and a follower's `allocator.close()` throws on a leaked buffer — so with the processor first, a leak would have left the leader gauge holding a dead `LogProcessor`. Raised in review.

## Out of scope

- **`tx.error`, the one meter in the sweep left where it is**
  `TxResolver` registers it per term, but `->conn-metrics` in `node/impl.clj` already registers the same untagged ID at node startup, so it never goes missing. Routing its three registration sites through one owner is a tidy.

- **Meter naming and new meters** — #5408 and #5404 respectively.

- **#5333's WARN**
  Same family, but its meter is `xtdb.allocator.memory.allocated` from the root allocator's `AllocationListener`, which already removes on `onChildRemoved`. Left open pending a repro rather than assumed fixed.

- **An `ExternalSource` orphaned by a throwing `LeaderLogProcessor` constructor**
  Found in the review pass. It leaks a Postgres replication connection, not just meters, and it predates this change — so it is a bug in its own right rather than meter lifetime. Not carded yet.

- **`Database.open` and `KafkaConnectSource` both have the `Meters` pattern inline** and are left alone; converting them is equivalence work and would want its own commit.

## Test plan

- **`PostgresSourceMetricsTest`** (renamed from `WalLagGaugeTest`, which is now one of its four cases) — a replacement source's gauge reads the replacement's driver, and a closed source leaves no meters behind. The first fails on `main` by reading the closed source's value.
- **`MetersTest`** — eviction, derived-percentile removal, and the no-metrics path.
- **`LogProcessorTest`** — a `LogProcessor` built without a group subscription cannot elect, so it pins the meters as the database's rather than a role's; and reports none once closed.
- **`NodeBaseTest`** — `tx.op.timer` on a node that has opened no databases at all.

`xtdb.information-schema-test/test-metrics` is the symptom the card reports, and it is **not** the regression test: it passes on `main`, because a self-electing node usually wins its first term before the query runs. The four above assert the absence deterministically instead.

Full suite green — 2228 unit tests across the root project and every module, plus `property-test` at 100 iterations covering `NodeSimulationTest`, `LogProcessorSimTest`, `LeaderDriverSimTest`, `CompactorSimulationTest` and `CacheSimulationTest` (2714 simulation invocations), all passing.

### DIFF
--- a/core/src/main/kotlin/xtdb/Metrics.kt
+++ b/core/src/main/kotlin/xtdb/Metrics.kt
@@ -1,6 +1,7 @@
 package xtdb
 
 import io.micrometer.core.instrument.Gauge
+import io.micrometer.core.instrument.Meter
 import io.micrometer.core.instrument.MeterRegistry
 import io.micrometer.core.instrument.Tag
 import io.micrometer.core.instrument.binder.jvm.*
@@ -131,5 +132,44 @@ object Metrics {
             scope?.close()
             span?.end()
         }
+    }
+}
+
+/**
+ * The meters one object owns — registered through [register], and removed from the registry by [close].
+ *
+ * Micrometer keeps the first registration for a meter ID and discards every later one, so a meter left
+ * behind by a dead owner shadows its replacement. For a gauge bound to an object that means reading the
+ * dead owner's state for the rest of the node's life, rather than the live owner's.
+ */
+class Meters(private val registry: MeterRegistry?) : AutoCloseable {
+
+    private val owned = mutableListOf<Meter>()
+
+    fun <M : Meter> register(build: (MeterRegistry) -> M): M? =
+        registry?.let { reg ->
+            val incumbents = reg.meters.toSet()
+
+            // `build` hands back an incumbent rather than registering ours where the ID is taken — a meter
+            // some owner didn't get to remove. Evicting it and building again is what stops one leak
+            // shadowing every later owner: what we return has to be the meter the registry now holds.
+            build(reg)
+                .let { if (it in incumbents) { reg.removeWithDerived(it); build(reg) } else it }
+                .also(owned::add)
+        }
+
+    override fun close() {
+        registry?.let { reg -> owned.forEach { reg.removeWithDerived(it) } }
+        owned.clear()
+    }
+
+    // A percentile-publishing timer or summary also registers a `<name>.percentile` gauge per phi, as
+    // meters in their own right that `remove` doesn't reach — and they read the parent being removed. The
+    // tag check keeps the sweep off another database's meters.
+    private fun MeterRegistry.removeWithDerived(meter: Meter) {
+        remove(meter)
+
+        meters.filter { it.id.name.startsWith("${meter.id.name}.") && it.id.tags.containsAll(meter.id.tags) }
+            .forEach { remove(it) }
     }
 }
--- a/core/src/main/kotlin/xtdb/NodeBase.kt
+++ b/core/src/main/kotlin/xtdb/NodeBase.kt
@@ -1,6 +1,7 @@
 package xtdb
 
 import io.micrometer.core.instrument.MeterRegistry
+import io.micrometer.core.instrument.Timer
 import io.micrometer.tracing.otel.bridge.OtelTracer
 import org.apache.arrow.memory.BufferAllocator
 import org.apache.arrow.memory.RootAllocator
@@ -27,6 +28,9 @@ class NodeBase(
     val remotes: Map<RemoteAlias, Remote>,
     val compactor: Compactor,
     val querySource: IQuerySource,
+
+    /** Untagged, so node-scoped: a per-database owner would collide on the single meter ID. */
+    val txOpTimer: Timer?,
 ) : AutoCloseable {
 
     override fun close() {
@@ -100,6 +104,12 @@ class NodeBase(
                     remotes = remotes,
                     compactor = compactor,
                     querySource = querySource,
+                    txOpTimer = meterReg?.let { reg ->
+                        Timer.builder("tx.op.timer")
+                            .publishPercentiles(0.75, 0.85, 0.95, 0.98, 0.99, 0.999)
+                            .description("indicates the timing and number of transactions")
+                            .register(reg)
+                    },
                 )
             }
     }
--- a/core/src/main/kotlin/xtdb/garbage_collector/BlockGarbageCollector.kt
+++ b/core/src/main/kotlin/xtdb/garbage_collector/BlockGarbageCollector.kt
@@ -1,6 +1,5 @@
 package xtdb.garbage_collector
 
-import io.micrometer.core.instrument.MeterRegistry
 import io.micrometer.core.instrument.Timer
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.Channel
@@ -11,7 +10,6 @@ import xtdb.catalog.BlockCatalog
 import xtdb.catalog.BlockCatalog.Companion.allBlockFiles
 import xtdb.catalog.BlockCatalog.Companion.tableBlocks
 import xtdb.storage.BufferPool
-import xtdb.api.DatabaseName
 import xtdb.util.StringUtil.fromLexHex
 import xtdb.util.debug
 import xtdb.util.logger
@@ -41,12 +39,11 @@ class BlockGarbageCollector(
     private val blocksToKeep: Int,
     /** Gates the auto-signal from the leader's block-boundary path; direct `awaitNoGarbage()` is unaffected. */
     val enabled: Boolean,
-    private val meterRegistry: MeterRegistry? = null,
+    private val metrics: GcMetrics,
     tableParallelism: Int = DEFAULT_TABLE_PARALLELISM,
     deleteParallelism: Int = DEFAULT_DELETE_PARALLELISM,
     /** Base for the parallel delete fan-out pools; the loop itself runs on its caller's thread. Sims inject the seeded dispatcher so deletes stay on the simulation's thread. */
     dispatcher: CoroutineDispatcher = Dispatchers.IO,
-    dbName: DatabaseName,
 ) {
 
     // [signal] is fire-and-forget; bursts coalesce into one upcoming cycle.
@@ -58,20 +55,6 @@ class BlockGarbageCollector(
 
     private val tableDispatcher = dispatcher.limitedParallelism(tableParallelism)
     private val deleteDispatcher = dispatcher.limitedParallelism(deleteParallelism)
-
-    private val blockDeleteTimer: Timer? = meterRegistry?.let {
-        Timer.builder("xtdb.gc.block_files.delete.timer")
-            .publishPercentiles(0.75, 0.95, 0.99)
-            .tag("db", dbName)
-            .register(it)
-    }
-
-    private val tableBlockDeleteTimer: Timer? = meterRegistry?.let {
-        Timer.builder("xtdb.gc.table_block_files.delete.timer")
-            .publishPercentiles(0.75, 0.95, 0.99)
-            .tag("db", dbName)
-            .register(it)
-    }
 
     init {
         require(blocksToKeep >= 1) { "blocksToKeep must be >= 1, got $blocksToKeep" }
@@ -138,9 +121,9 @@ class BlockGarbageCollector(
             coroutineScope {
                 paths.filter { it.isGarbage() }.forEach { path ->
                     launch(deleteDispatcher) {
-                        val timer = meterRegistry?.let { Timer.start(it) }
+                        val sample = gcTimer?.let { Timer.start() }
                         bufferPool.deleteIfExists(path)
-                        gcTimer?.let { timer?.stop(it) }
+                        gcTimer?.let { sample?.stop(it) }
                     }
                 }
             }
@@ -148,14 +131,14 @@ class BlockGarbageCollector(
 
         supervisorScope {
             launch(tableDispatcher) {
-                deleteGarbage(bufferPool.allBlockFiles.asSequence().map { it.key }, blockDeleteTimer)
+                deleteGarbage(bufferPool.allBlockFiles.asSequence().map { it.key }, metrics.blockDelete)
             }
         }
 
         supervisorScope {
             for (table in blockCatalog.allTables) {
                 launch(tableDispatcher) {
-                    deleteGarbage(bufferPool.tableBlocks(table).asSequence().map { it.key }, tableBlockDeleteTimer)
+                    deleteGarbage(bufferPool.tableBlocks(table).asSequence().map { it.key }, metrics.tableBlockDelete)
                 }
             }
         }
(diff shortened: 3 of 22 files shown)


jarohen:
Closing unmerged. `extsource-once` fixes the half of #5943 that mattered, and fixes it at the root rather than at the symptom.

- **The CDC gauges can no longer be stranded, because the source is no longer per-term**
  `refactor(indexer): one external source per database, not one per leader term` removes the lifetime mismatch this PR was compensating for. `fix(pgsrc): source metrics track the assignment, not the process` then handles the residual per-assignment staleness by collapsing that state into a single `AtomicReference<Unassigned | Assigned>`, and picks up two things this PR missed outright: that every indexing node was opening an upstream connection per scrape for `wal_lag_bytes`, and that `last_event_time` was never reset per assignment.

- **This PR's approach was that branch's rejected alternative, on a ground this PR didn't have**
  Register-and-remove per source lifetime means a follower reports no series at all, and `avg by (db)` / `sum by (db)` across nodes break on an absent series where they survive NaN with `max by (db)`. That trade-off is named and decided in `bfc520431`; D1 here reached the opposite conclusion having reasoned only about what a single node's gauge shows.

- **What was left after rebasing is real, and is being dropped rather than re-scoped here**
  `tx.op.timer` and the three `xtdb.gc.*` delete timers are still absent on a node that has never led, `xtdb.replica.*` is still re-registered on every switchover, and `xtdb.log.leader` is still never removed on detach — none of which `extsource-once` touches. Rebased onto it that reduces to ~320 lines in `core`, green on the full suite (2232 tests) and the simulations (2747 invocations). It isn't worth its own carry as things stand.

- **Two corrections to this PR's own claims, for anyone reading it later**
  The `xtdb.replica.*` meters were never absent — `LogProcessor` opens its first follower in a property initialiser, so they exist from database open whatever roles the node holds. And `xtdb.information-schema-test/test-metrics` passes on `main`; that failure was a race the #5917 branch happened to lose, not a standing symptom.

#5943 keeps the full analysis and stays open for whoever decides whether the residue is worth a card. Both branches have since been cleaned up off the fork; the commits stay reachable from this PR for anyone who wants them back.

Issue #5954, opened 2026-08-26
a zero-leg type should not silently answer false


mbutlerw:
## tl;dr

- **Two sites ask a question about a value type by quantifying over its legs, and quantification is vacuous on the empty set**
  `nullable-vec-type?` is `(some VectorType/.getNullable (.getLegs vec-type))`; `bloom-expr`/`presence-expr` fold an `:or` over `(.getLegs value-type)`. For `VectorType.Nothing` both come back negative — not because null is excluded or because no page can match, but because nothing is included.

- **They are the half of this family that does not crash, so nothing surfaces when they become reachable**
  #5948 closed the eight leg-dispatch derivations by making a read of a zero-leg type a defined error. Neither of these emits a read: one picks a codegen branch, the other builds a plan expression. A fault cannot fire where nothing is read.

- **Whether either answer is actually wrong is open, and #5948's rule argues both may be right**
  The bottom has no cells and is not read. So a comparand of that type makes an equality predicate unsatisfiable, which makes "prune every page" correct rather than catastrophic; and `AND`/`OR`'s branch choice governs a declared type on an expression that cannot execute. This card may be *establish and pin the answers* rather than a fix.

## Context

Found while fixing #5948 — the same underlying disagreement, at the two sites where it answers instead of failing. Worth its own card because everything in the family that failed loudly is now fixed, and what remains is the half that fails quietly.

## Symptoms

- **`AND`/`OR` take the two-valued branch for a valueless operand**
  `codegen-and`/`codegen-or` (`expression.clj:647`, `:668`) choose between three- and two-valued paths via `types/nullable-vec-type?` (`types.clj:199`). For a zero-leg type that is `false`, so the non-3VL branch is taken and the declared return type is `:bool` rather than `[:? :bool]`.

  Since #5948 the *value* is no longer at stake — the operand read emits `err/fault ::zero-leg-read` rather than being read as `(boolean nil)`. What remains is the type.

- **A metadata filter prunes every page rather than none**
  `bloom-expr` and `presence-expr` (`expression/metadata.clj:55`, `:70`) build `{:op :call, :f :or, :args (for [leg (.getLegs value-type)] …)}`. With no legs the `for` yields nothing, `simplify-and-or-expr` collapses a zero-arg `:or` to a literal `false`, and the predicate excludes every page.

## Root cause

**A question about values, asked with a test over legs, at sites that are not dispatch tables.**

[...]

Issue #5871, opened 2026-08-10
fromLegs(∅) returns Null, so mergeTypes rewrites nested lattice bottoms


mbutlerw:
## Summary

`VectorType.Nothing` is documented as the lattice bottom with the law `Nothing ⊔ X = X`. That law requires `Nothing ⊔ Nothing = Nothing`, and today it yields `Null`:

```clojure
(types/merge-types VectorType$Nothing/INSTANCE VectorType$Nothing/INSTANCE)  ;; => #xt/type :null
(types/merge-types)                                                          ;; => #xt/type :null
```

Two tests in `MergeTypesTest` currently pin this, one of them conceding it in its own message — *"Nothing alone collapses like an empty merge"* (`MergeTypesTest.kt:188`, and a bare `assertEquals(Null, mergeTypes())` at `:27`).

This was attempted as part of #5855 and backed out, because making it correct requires the expression engine to have defined behaviour for a zero-leg type. **That prerequisite is now #5948, which blocks this issue.**

With it split out, what remains here is small: `MergeTypes.asType` returning `Nothing` for an empty accumulator, deleting `MergeTypes.joinContributions` and its two call sites — its own kdoc already says *"Delete this and call `mergeTypes` directly once `fromLegs(∅)` returns the bottom"* — and inverting the two assertions at `MergeTypesTest.kt:27` and `:188`.

## Impact

Sound — the merge over-collects, so `emitted ⊑ declared` still holds — but not low. It is the cause of a **live disagreement between shipped surfaces on `main`**, because the same defect applies to a *nested* bottom, where it is considerably easier to reach.

`fromLegs(∅)` is hit recursively. `MergeTypes.merge(Listy)` recurses into the element type; `merge(Nothing)` is a no-op; the inner accumulator ends up empty; `asType` calls `fromLegs(∅)` and gets `Null`. So `mergeTypes` silently rewrites a nested lattice bottom:

```clojure
(types/merge-types (listTypeOf Nothing))
;; raw     => #xt/type [:list :nothing]
;; merged  => #xt/type [:list :null]      ; same? false
```

An empty list literal is enough to produce one. Measured on `main` at `8d9d9850c`, after `INSERT INTO t RECORDS {_id: 1, v: []}` — same column, same instant, three surfaces:

```
getTableSchema  (ADBC, no merge)    v: List<$data$: Null not null> not null    <- Nothing
executeSchema   (scan, merged)      v: List<$data$: Null> not null             <- Null
[...]

PR #5967, opened 2026-08-27
fix(expression): read a zero-leg type as null where legs are dispatched on

### LINKED ISSUE #5948: a list column holding only empty lists should be comparable
## tl;dr

- **A list column that holds only empty lists types its elements as `VectorType.Nothing`, and list-comparison codegen collapses on it**
  Comparison codegen builds a per-element table from `(.getLegs l-el-type) × (.getLegs r-el-type)`. `Nothing` has zero legs, so the table comes out empty, every runtime lookup misses, and the generated code invokes a `nil` `->call-code`: `NullPointerException: Cannot invoke "clojure.lang.IFn.invoke(Object)" because "__GT_call_code" is null`.

- **Two statements reproduce it, with no `CREATE TABLE`, no schema evolution and no parameters**
  `INSERT INTO docs (_id, xs) VALUES (1, ARRAY[])`, then any `=` against `xs`.

- **Reached from DML it stops ingestion for the database, not just the transaction**
  `SourceLogTxIndexer.indexTx` converts only `Anomaly.Caller` into an abort, so a `Fault` escapes to the database's coroutine handler and `Watchers.notifyError`. The transaction never commits, so it replays on restart and fails identically.

- **Present since `519b33937` (2026-06-09), confirmed by bisect**
  Element vectors have been non-nullable since 2025-09; that commit made a non-nullable `NullVector` report `Nothing`, which is what the codegen cannot represent.

- **The type is right and the consumer is wrong, so suppressing `Nothing` is the wrong fix**
  An element vector that has never held a cell does have an unconstrained element type. Making it report `Null` instead would widen every column that ever saw an empty array. The gap is that codegen has no defined behaviour for a zero-leg type, which is what #5871 is already blocked on.

- **The behaviour it needs is one rule: a read of a zero-leg type yields null**
  Not a convention chosen for the engine — `NullVector.isNull` is unconditionally `true`, so that is what the vector returns, and it is the same rule `VectorType.absentContribution` applies one level up. Most of the tree already gets it right by handling `Null`, which `Nothing` falls into for want of legs; the sites that separately derive "which legs will I be handed" are the ones that don't.

- **The trigger is a property of the page, not of the schema**
  Element vectors are per-page, so a column that is usually populated crashes only on a page where every row's list happens to be empty. Same query, same table, intermittent.

## Context

Observed as an ingestion-stopped crash loop on a customer deployment after upgrading to a recent nightly.

## Symptoms

- **`IngestionStoppedException`, wrapping a `Fault`, wrapping the NPE**

  ```
  xtdb.api.log.IngestionStoppedException: Ingestion stopped (msgId: null): Cannot invoke
    "clojure.lang.IFn.invoke(Object)" because "__GT_call_code" is null
  Caused by: xtdb.api.error.Fault: ...
  Caused by: java.lang.NullPointerException: ...
  ```

  `__GT_call_code` is the munged form of `->call-code`.

- **Crash loop**
  The failing transaction is retried on restart and fails identically, since it never commits.

- **Ingestion stopped on every attached database, and the isolation between them is intact**
  A fault does not travel between databases. Tested in both directions on a live node — wedging the default `xtdb` database leaves an attached one ingesting, and wedging an attached one leaves the default ingesting — which matches the design: `DatabaseCatalog` parents every database on a `SupervisorJob` (`DatabaseCatalog.kt:54-57`), each `Database` gets its own beneath it (`Database.kt:325`), and each has its own `Watchers` and `CoroutineExceptionHandler` (`Database.kt:309`, `:348`). Owning the catalog does not make the default database special here either.

  The likely reading is independent faults on similar data: databases carrying one schema, driven by one application, all crossing the exposure boundary on the same upgrade and meeting the same page shape at about the same time. Each then enters its own crash loop. That makes the blast radius wider than one database rather than narrower — a data shape common to a fleet takes the fleet out at once.

  `check:` the logs discriminate the two readings cheaply. Independent faults produce one `IngestionStoppedException` per database, each naming that database's own tables at its own timestamp; propagation would show a single fault and silent stoppages elsewhere.

## Root cause

- **Two construction routes for a list's element vector disagree on its nullability**
  `ListVector`'s own default is `NullVector("$data$")`, nullable (`ListVector.kt:35`). `openVector`'s childless fallback is `NullVector("$data$", false)`, non-nullable (`Vector.kt:137`). Every writer path reaches the second: `Relation.vectorFor` (`Relation.kt:57`), `StructVector.vectorFor` (`StructVector.kt:72`), `DenseUnionVector.legVectorFor` (`DenseUnionVector.kt:272`) and `NullVector.maybePromote` (`NullVector.kt:69`) all call `openVector(name, arrowType, nullable)` with no children.

- **A non-nullable `NullVector` reports the lattice bottom, and that report is correct**
  `NullVector.type = if (nullable) Null else Nothing` (`NullVector.kt:27`), so `ListVector.monoType` is `Listy(LIST, Nothing)`. An element vector that has never held a cell genuinely has an unconstrained element type, and that is what keeps `[] ⊔ [1,2,3]` at `list<i64>` rather than widening it to `list<i64?>`. The bug is that codegen cannot consume the type, not that the type is wrong — a fix that stops producing nested `Nothing` trades a crash for silent type imprecision.

- **The trigger is `nullable == false` on the element vector, not an empty element vector**
  `writeNull()` trips the flag (`NullVector.kt:43`, and generally `MonoVector.kt:34`); `writeUndefined()` does not. `valueCount` does not enter into it. So the condition is that the element vector has never received a `writeNull()`, and the user-visible boundary is:

  | Written | Element vector | Element type |
  | --- | --- | --- |
  | `ARRAY[]` | untouched | **`Nothing`** |
  | `NULL` (the whole list) | untouched — the null lands on the list's own validity bit | **`Nothing`** |
  | `ARRAY[NULL]` | one `writeNull()` | `Null` |
  | `ARRAY[1]` | promoted away | `I64` |

  Worth pinning in the regression tests: `ARRAY[NULL]` heals and `ARRAY[]` does not, which is not the boundary a reader would guess.

- **The callback is handed a leg the dispatch table was never built with**
  `continue-read` (`expression.clj:432-435`) dispatches a `Mono` as itself, a `Maybe` as `[mono, Null]` and a `Poly` as its children — all of which are keys. For a zero-leg type it applies the rule below and calls back with `:null` (the empty-union branch at `:414`), while the table was built from the same empty leg set and so has no key at all. Note what this is *not*: no code goes missing. Code is emitted and then looked up under a key nobody generated.

- **Three codegen sites confirmed crashing; the apparent guard does not fire**
  `[:== :list :list]` and `[:<> :list :list]` (the `doseq` at `expression.clj:2255`), and `[:=== :list :list]` (`:2303`). The `:null` special-case at `:2265` does not protect the first two: it sits *inside* the `for` comprehension, so with zero legs the comprehension yields no pairs and the guard is never evaluated. Making the table builders apply the rule is what lets that guard start working, without moving it.

- **Nine sites derive the same answer, and only four of them are reachable**
  `[:== / :<> :struct :struct]` (`:1972`) and `[:=== :struct :struct]` (`:2020`) crash identically to the list pair, reached through a struct field holding an empty array. The other five — `codegen-expr :let` (`:558`), `:if-some` (`:575`), `codegen-call*` (`:724`), `[:_patch :struct :struct]` (`:2065`), and `codegen-and`/`codegen-or` (`:647`, `:668`) via `types/nullable-vec-type?` — would each answer wrongly if handed the bottom, and cannot be handed it today.

  Reachability is not what decides whether they should agree, though. All nine ask *which legs will the callback hand me*, and `continue-read` already answers it; deriving it a second time from a source that disagrees at the bottom is the defect, whether or not a given site is currently exposed. Sites asking a *different* question — is this nullable, can I prune this page — are #5954's, because there the bottom may legitimately differ from `Null`.

- **Why the other five are unreachable, and what changes that**
  Two separate mechanisms, and #5871 removes the first.

  `merge-types` never returns `Nothing`: for an all-bottom input it ends at `fromLegs(∅)` → `Null`. Every route by which an expression's `:return-type` could be the bottom passes through either that merge or the `:521` guard — including `get_field`, whose raw child type is healed by its own enclosing `codegen-call*`. So `:let`, `:if-some`, `codegen-call*` and `codegen-and`/`or`, all of which consume an expression's `:return-type`, only ever see types with legs. #5871's entire content is making the merge preserve the bottom, which takes that protection away.

  `[:_patch :struct :struct]` is unreachable for a different and more permanent reason. `StructVector.endStruct` pads every short child with `writeNull()` to keep the children aligned with the struct's row count, and `writeNull` flips a child off the bottom. So a struct with a bottom-typed child has **no non-null rows** — and `_patch`'s wrong branch needs a non-null struct row to execute, since a null one takes the `Null` leg. The condition that produces the type is the condition that stops the code running.

  The four comparison sites escape both because they decompose a type's *components* — `.getElType`, `.getChildren` — which never pass through a merge.

- **The existing `Nothing` guard is top-level only**
  `codegen-variable` rewrites `Nothing → :null` (`expression.clj:521`), added in a0ec6b243 with a comment naming this exact failure mode. An element type sits one level below it.

- **The merge machinery would have healed it, and neither path to codegen goes through the merge**
  `mergeTypes` recurses into `Listy.elType`, absorbs `Nothing`, and resolves the empty accumulator through `fromLegs(∅)` → `Null` (#5871), so planner-declared types are safe. But `->expression-projection-spec` re-derives `var-types` from the live input relation at execution time (`expression.clj:2464-2467`), and `->param-types` reads them off the args relation (`:2446`). Both take `.getType` raw off the vectors.

## Evidence

- **Minimal SQL repro**

  ```sql
  INSERT INTO docs (_id, xs) VALUES (1, ARRAY[]);
  SELECT _id FROM docs WHERE xs = ARRAY[1];   -- NPE
  ```

  `WHERE xs = xs` fails the same way, as does `UPDATE docs SET seen = true WHERE xs = ARRAY[1]` — the last of which reproduces the ingestion-stopped severity rather than just failing the query.

- **The producer, isolated from codegen**
  A list vector created through the writer API and given one empty list:

  ```kotlin
  Relation(allocator).use { rel ->
      val xs = rel.vectorFor("xs", ArrowType.List.INSTANCE, true)
      xs.writeObject(emptyList<Any>())
      rel.endRows(1)
      assertEquals(listTypeOf(VectorType.Null), xs.type)
      // expected: <[<List> Null]>
      // but was:  <[:? [<List> Nothing]]>
  }
  ```

- **Bisect**

  | Commit | Date | Repro |
  | --- | --- | --- |
  | `a0ec6b243` — `Nothing` and the `codegen-variable` guard added | 2026-06-09 | passes |
  | `519b33937` — `NullVector.type` honours the nullable flag | 2026-06-09 | **fails** |
  | `7cc875d56` — immediately before #5916 | 2026-08-20 | **fails** |
  | `02fd4fae7` — `main` | 2026-08-24 | **fails** |

  `519b33937` and `a0ec6b243` are adjacent, parent and child. The `false` at `Vector.kt:137` dates to `540b12dde` (2025-09-24) and was inert until `NullVector.type` began honouring it.

- **#5916 is not involved**
  It broadens where `Nothing` reaches — including new on-disk state in block metadata — but every path it touches goes through `joinContributions`/`mergeTypes`, which heals a nested bottom. The repro fails identically on the commit before it.

## Constraints

- **The fail-stop on a `Fault` is correct and stays**
  `indexTx` converting only `Anomaly.Caller` into an abort is the intended split: bad input aborts the transaction and ingestion continues, while a `Fault` means the engine does not know what state it is in and stopping beats risking divergence. An unhandled exception in codegen is genuinely the latter, so the ingestion stop is the right response to this bug rather than a second defect. The fix belongs in codegen; loosening `indexTx` would convert internal errors into silently aborted transactions, which is worse.

## Out of scope

- **Suppressing nested `Nothing` instead of teaching codegen to read it**
  Flipping the childless element default at `Vector.kt:137` to nullable would make the crash go away, and is the wrong trade. An empty-array column would type as `Listy(LIST, Null)`, and one page of empty arrays would then permanently widen the column: `Listy(LIST, Null) ⊔ Listy(LIST, I64)` is `Listy(LIST, Maybe(I64))`, where the `Nothing` form gives `Listy(LIST, I64)`. That is the widening `Nothing` exists to prevent, reintroduced one level down, plus a change to on-disk schemas.

## The rule

**A read of a zero-leg type yields null.**

`Nothing` is a type with no values — no possible non-null inhabitant. Null is not one of its values; it is the absence marker, and it is what a total read of an empty vector returns. `NullVector.isNull(idx)` is unconditionally `true` (`NullVector.kt:29`) — the `nullable` flag governs the *reported type*, not the *read behaviour*. So the answer is not a convention chosen for the engine's convenience; it is what the vector does.

It is also the same rule the type layer already applies one level up. `VectorType.absentContribution` (`VectorType.kt:194`) resolves *rows exist, this column has no cells* to `Null`, and only *no rows at all* to `Nothing`. Codegen never has to make that choice, because it only ever emits a read for a row that exists.

**The agreement is narrow, and that is the point.** `Null` and `Nothing` differ in every respect the type system cares about:

| | `Null` | `Nothing` |
| --- | --- | --- |
| merge | `Null ⊔ I64 = Maybe(I64)` | `Nothing ⊔ I64 = I64` |
| legs | `[Null]`, and a `Mono` | `∅`, not a `Mono` |
| asserts | "may hold null" | "no values, no constraint" |
| serialises as | `Field(nullable = true)` | `Field(nullable = false)` |
| as a union leg | legal | `asLegField` errors |

They agree on exactly one thing: what a read yields. So the rule licenses no conflation beyond that — collapsing them anywhere else reintroduces the widening `Nothing` exists to prevent, which is the trade ruled out under Out of scope.

**Most of the tree already gets the right answer, and not by having applied this rule.** `continue-read-union:414`, `codegen-concat:1518`, `[:nth :list :int]:2160`, `[:get_field :struct]`'s absent-field branch at `:1937` and `PgType.unifyPgTypes`'s `isEmpty()` guard each exist for **`Null`** on their own merits — and `Nothing` falls into them because it has no legs. `continue-read` emitting identically for the two is that, not a coincidence: `without-null` is empty in both cases, so the same branch fires.

Genuinely bottom-specific sites number three: the `Nothing → :null` guard at `:521`, `absentContribution`, and `fromLegs(∅)`. Which means the fix is not "state a rule that has been rediscovered repeatedly" — it is **routing `Nothing` into the `Null` handling that already exists everywhere**, and that is why substituting at the read sites is small rather than speculative.

There is a seventh application, and it is the one in the wrong place. `VectorType.fromLegs(∅)` also answers `Null` (`VectorType.kt:304`), but it is answering for type *construction* rather than for a read — and there the row-existence question is not settled, because that branch serves three unrelated callers including the deliberate unknown-parameter marker (#5871 sets this out). `MergeTypes.joinContributions` exists purely to repair the damage, and says so in its own kdoc. So the rule holds wherever a value is read and fails wherever a type is built.

**The engine derives that answer twice** — once to emit the read, once to build the per-leg dispatch table — and the two must agree. They do for every type except a zero-leg one, where `continue-read` applies the rule and the table builders do not.

**Not "emit nothing", and not a throw.** Reading a zero-leg field is reachable, and the struct case shows it without reference to the type system: the list methods gate on `.size` before their loop (`:2277-2279`), so element code there is dead, but the struct fold has no such gate — `:1985-2007` reduces over every field on every row. A throw would convert a codegen-time NPE into a runtime one, and would not compile in any case: `res-sym` is hinted primitive (`:2276`) and is a `loop` accumulator and `recur` argument flowing through `(min …)`, where a bare `throw` gives the compiler an `Object`.

**The null reading needs no new emission.** `cont-b3-call` already returns the literal `0` — unknown — for the `:null` leg (`:640-643`), and `expression_test.clj:1756` (`(== [1 2 nil] [1 2 3.0])`) exercises that exact form in that exact primitive position today. `0` is also the right answer rather than a convenient one: the loop *seed* is `1` (`:2280`), so `[] = []` is answered by the seed and never by element code, leaving the element value to answer `null = null`.

**Why the ordering cannot be reversed.** This card adds the rule where a value is read; #5871 removes it from where a type is built. Doing #5871 first would take away the accident that currently hides the gap — `mergeTypes` healing nested bottoms — and expose every read site that never applied the rule, on SQL as well as XTQL: `[]` literals become `[:list :nothing]` (`expression.clj:2124`, `ValueTypes.kt:99`), and `[:get_field :struct]`'s leaked child type stops being absorbed by the enclosing `codegen-call*`. So #5871 is not merely waiting on a prerequisite; it is the change that makes the gap general.

## Verification

**The property to pin is an equivalence, not a set of expected values.**

> for any read position, a `Nothing`-typed input and a `Null`-typed input produce the same answer

That equivalence already holds in `continue-read` and always has: `Null` is a `Mono`, so it takes the direct branch and `read-value-code :null` returns `nil`, giving `(f Null nil)`; `Nothing` is not a `Mono`, so it takes `continue-read-union` with an empty leg set and its first branch gives `(f #xt/type :null nil)`. Identical calls. So this issue does not *establish* the equivalence — it makes the consumers honour one the reader already guarantees, which is what makes a differential test the right instrument.

Test it by constructing both sides and comparing, rather than asserting an absolute:

```clojure
(= (run-projection '(== xs ys) rel-with-nothing-el-type)
   (run-projection '(== xs ys) rel-with-null-el-type))
```

- **It does not depend on reachability.** Both inputs are constructed, so the test claims nothing about whether production produces either.
- **It fails today for the right reason** — the `Nothing` side throws where the `Null` side returns a value.
- **It does not encode a guess about the answer**, so it cannot rot into a tautology if the emitted value changes.
- **It catches a later divergence**, rather than letting the two silently drift apart again.

Applies per read position, including `[:_patch :struct :struct]`, where the assertion is that the left-hand value survives rather than being nulled — the data-loss case. Assert the premise too (that the constructed relation's type really is the bottom), or the test quietly stops covering anything if the producer changes.

`xtdb.expression-test/test-list-equal` and `test-list-diff` already cover the literal path and must stay green; the column path needs adding, along with the `ARRAY[]` / `NULL` / `ARRAY[NULL]` boundary from Root cause.

## Open questions

- **Whether the two element defaults should converge, separately from the fix**
  `ListVector.kt:35` defaults the element vector to nullable and `Vector.kt:137` to non-nullable. Read as "has this vector ever held a cell", `false` is the accurate one and `ListVector.kt:35` is asserting occupancy of a vector that has seen nothing. Harmless today only because it produces a type codegen happens to accept.

- **caveat: unverified, and orthogonal to the crash** — `NullVector.unloadPage` (`NullVector.kt:85`) writes `ArrowFieldNode(valueCount, valueCount)` unconditionally, so a `Nothing`-typed vector serialises a field marked `nullable = false` alongside a node claiming every value is null. Probably inert — there are no buffers to misread, and both ends of our own round-trip share the convention — but it is the one place the two readings of the flag visibly disagree, and it would matter to any Arrow reader that isn't us. Noted rather than investigated.

- **Whether a `Nothing`-typed struct child is reachable at all**
  `[:=== :struct :struct]` (`:2020`) and the struct `compare` pair (`:1972`) are constructed identically to the list sites and are structurally more exposed — no `.size` gate. But no route from SQL to a `Nothing` struct child has been found: `mergeTypes` absorbs the bottom (`MergeTypes.kt:21`) and `NullVector.maybePromote` writes `writeUndefined` into the promoted vector (`NullVector.kt:73`), so nothing carries one. Two candidate geneses worth probing: `_patch`'s own `ret-type` defect at `:2065`, which *constructs* a struct type with a `Nothing` child and hands it to `Vector/open` (`:2471`), and `RelationAsStructReader` (`RelationAsStructReader.kt:19`), which lifts a relation's column types into struct children. If it turns out unreachable, the struct half is a totality guarantee rather than a bug fix, and should be described as one.

- **How the fix is shaped, and how much of the above it covers**
  The rule is one line applied in two derivations, so the marginal cost across the sites is small; what is undecided is whether they belong on this card. `[:_patch :struct :struct]`'s data loss in particular is a durable write-path defect rather than a crash, and reads as its own card.

- **Whether any container type beyond list and struct decomposes a type into legs below the top-level guard**
## Progress

**Status**: in-progress — two PRs open, same bug, alternative approaches. Needs one picked.

- [x] PR #5967 — **`readsAs` on `VectorType`.** Give the lattice an accessor for what reading a type yields, and migrate every `.getLegs`-for-reading site to it. Also carries the `#xt/type :nothing` reader clause and the `Nothing` guard's comment correction.
- [x] PR #5968 — **confine the bottom at the codegen sites.** `read-type` where a read has to be emitted (`get_field`, both struct comparisons), a fold where it doesn't (`[:== / :<> :list :list]`, `[:=== :list :list]`, `[:nth :list :int]`), and a fault in `continue-read-union` for anything unaccounted for.
- [ ] **Pick one.** They are close. #5968 leaves `VectorType` untouched and removes three sites' need to know about the bottom at all; #5967 puts the rule where a future site finds it without knowing to look — which is worth something, given six sites needed the guard and the assertion rather than the code review found most of them.

`operator/table.clj`'s three sites are out of both: each already filters or guards such that `∅` and `{Null}` give the same answer, so a change there would provably do nothing.

### Follow-ups — all surfaced by this work, none caused by it

- [ ] **`[:_patch :struct :struct]` passes a bottom child straight into `:return-type`**, and `[:_patch :null :struct]` returns `r-type` verbatim. Reachable via `pull` over a table with *no put rows* (not, as first thought, over a table with a row — `endStruct` padding rules that out). Unverified whether it misbehaves: the struct leg never executes when the table is empty.
- [ ] **`codegen-cast [:list :list]` hands `(.indexOf … :nothing)` → `-1` to `MonoToPolyReader` as a type id.** Pre-existing, and inside the surface this issue confines.
- [ ] **`pull` should plan like `NEST_ONE`.** `NEST_ONE` builds a struct expression over named columns (`sql.clj:202`), so its fields pass through the `codegen-expr :variable` gateway and arrive as `Null`; `pull` uses `:star` and bypasses it, which is why `pull` is the only surface producing `Struct{a: Nothing}`. Needs the `provided-vars`-vs-inner-column-set question answered first. Note that doing this would make `xtdb.query-test/pull-reaches-a-struct-of-valueless-fields-5948` vacuous — check it still covers something.
- [ ] **`fromLegs(∅)` returns `Null` rather than the bottom** — `MergeTypes.kt:70-77` carries its own deletion note and `joinContributions` exists only to repair it. Fixing it turns every empty-collection literal into a bottom-element list: `ValueTypes.kt:99`/`:101`/`:59`, and `expression.clj`'s `:list` and `:set`. PR #5968's list folds already trigger on either side being the bottom, so that work is a prerequisite for this cleanup rather than blocked by it.

### Open Questions

- [ ] **Reflection in the list folds is unverified**, because the Gradle build never sets `*warn-on-reflection*` and only the dev REPL does. Mitigated rather than answered: the folds bind their operands to `^ListValueReader` gensyms instead of relying on the caller's tags, which is the idiom the non-fold path already uses.
- [ ] **`NullVector.setNull(idx)` and `ensureCapacity` bump `valueCount` without flipping `nullable`**, unlike `writeNull`. If either is on the Arrow-load path that's a way to land a vector with defined null cells still typed `Nothing`. Not chased; doesn't affect either PR, since both answer null for a bottom read regardless of row count.

mbutlerw:
Resolves #5948.

## tl;dr

- **A list column holding only empty lists could not be compared, and reached from DML it stopped the database's ingestion**
  `INSERT INTO docs (_id, xs) VALUES (1, ARRAY[])` followed by any `=` against `xs` threw `NullPointerException: … "__GT_call_code" is null` — at code-generation time, so no data was needed beyond the one row.

- **The engine worked out "which legs will a read hand me" in two places, and they disagreed at one type**
  `continue-read` treats a zero-leg type as reading null; the per-leg dispatch tables built their keys from `legs` and so had no key for it. This names the reader's answer as `VectorType.readsAs` and has the tables ask for it.

- **The fix routes `Nothing` into `Null`'s existing handling rather than teaching the engine a new type**
  That is what makes a nine-site substitution small: no site gained a case, because each already handled `:null` — the `:null` guard the list methods already carried was simply never being evaluated.

- **Present since `519b33937` (2026-06-09), bisected against a passing parent**
  In both 2.2.0 release candidates and every nightly since; in no GA release. Unrelated to #5916, confirmed by running the repro on the commit before it.

## Reading order

1. `tidy(types)` — the accessor, used in `continue-read` and nowhere else. **Behaviour-preserving, and provably so**: for an empty leg set and for `{Null}`, `continue-read-union` takes the same `(empty? without-null)` branch and emits identical code. No tests change; the suite is the check.
2. `fix(expression)` — the nine dispatch derivations. This is where behaviour changes.
3. `tidy(expression)` — the comment on the `Nothing` guard at `:521`, which justified the guard on grounds the accessor now covers.
4. `fix(types)` — `#xt/type` reads `:nothing`, so the tests can spell the type they assert.

## What is worth a reviewer's attention

- **Only four of the nine sites are reachable today, and reachability is not why the other five changed**
  All nine ask the same question, so deriving it from two sources that disagree is the defect whether or not a site is currently exposed. The four reachable ones are the list and struct comparison methods; the other five are latent behind two separate mechanisms, set out in the issue. The practical consequence is test coverage, not correctness: the four are tested, the five cannot be.

- **`_patch` looked like the worst of them and is not reachable at all**
  Its `(contains? r-types :null)` decides whether to emit the runtime null check, so with no legs it takes the patch value and would overwrite an existing value with null. But `StructVector.endStruct` pads short children with `writeNull()`, which takes a child off the bottom — so a struct with a bottom-typed child has no non-null rows, and the wrong branch needs one to execute. The condition producing the type forbids the code running.

- **`flattened-r-types` becomes `r-read-legs`, and that is the whole of the `_patch` change beyond the substitution**
  Both its uses want the read answer — `contains? :null` obviously, and `(disj r-types :null)` fed to `merge-types`, which asks what non-null values the right side could produce. One derivation, correctly sourced, named for the question it answers, since not being able to tell is how the wrong answer got in.

- **Three `.getLegs` sites are deliberately left alone**
  `:491` asks whether a type is a union needing a box, `:2387` allocates writers per leg and is gated on `Poly`, and `:1518` shortcuts a `:null` argument before the leg set is consulted. None asks what a read yields.

## Tests

- **The column path must answer what the literal path already answers.** A literal `[]` types as `Listy(LIST, Null)` via `mergeTypes(∅)`; a column of `[]` types as `Listy(LIST, Nothing)`. The test asserts the column's element type really is the bottom first, so it cannot rot into a tautology if the producer changes. A same-data differential is not constructible — an empty list's element type is necessarily the bottom.
- **The boundary is not the one a reader would guess**: `ARRAY[]` and a null list leave the element vector untouched, so the bottom survives; `ARRAY[NULL]` writes one cell and heals to `Null`.
- **Both new tests were run against the pre-fix tree and fail there**, so they bite rather than passing vacuously. A third test covering `:let`, `:if-some` and `codegen-call*` passed both before and after and was dropped — which is how those three were established as unreachable.

## Out of scope

- **#5954** — two sites asking a *different* question, where the bottom may legitimately differ from `Null`: `nullable-vec-type?` ("is this nullable", and `information_schema` wants the opposite answer for a declared column), and the metadata filter ("can I prune this page", where prune-everything is right for equality and unsound in general). Both also block #5871.
- **#5871** — makes the merge preserve the bottom, which is what takes the protection off four of the five latent sites. Blocked by this and by #5954.
- **The fail-stop on a `Fault` is correct and unchanged.** `indexTx` converting only `Anomaly.Caller` into an abort is the intended split; an unhandled codegen exception is genuinely "we do not know what state we are in". The fix belongs in codegen.

## Verification

`./gradlew test` green — 2,227 tests. Not `property-test`: no indexing, compaction or GC. No spec updates; none of the `.allium` files covers the expression engine.

### DIFF
--- a/api/src/main/clojure/xtdb/serde/types.clj
+++ b/api/src/main/clojure/xtdb/serde/types.clj
@@ -434,6 +434,7 @@
     (keyword? type-spec) (case type-spec
                            :tstz-range VectorType/TSTZ_RANGE
                            :null VectorType$Null/INSTANCE
+                           :nothing VectorType$Nothing/INSTANCE
                            (VectorType/scalar (->arrow-type type-spec)))
 
     (set? type-spec) (VectorType/fromLegs (into #{} (map ->type) type-spec))
--- a/api/src/main/kotlin/xtdb/arrow/VectorType.kt
+++ b/api/src/main/kotlin/xtdb/arrow/VectorType.kt
@@ -50,6 +50,25 @@ sealed class VectorType {
 
     abstract val legs: Iterable<Mono>
 
+    /**
+     * The legs a *read* of this type can yield, as against [legs], which is the lattice's answer. The two
+     * differ only at [Nothing], which has no legs and reads as null - it is backed by a valueless
+     * `NullVector`, whose `isNull` is unconditionally true.
+     *
+     * MUST NOT be substituted for [legs] when *constructing* a type: `fromLegs([Nothing, I64])` over this
+     * gives `Maybe(I64)` where the lattice law demands `I64`.
+     */
+    val readsAs: Iterable<Mono>
+        get() =
+            // not `else -> legs`: a new direct subclass of VectorType has no known reading, and should stop
+            // this compiling until someone picks one
+            when (this) {
+                is Mono -> legs
+                is Maybe -> legs
+                is Poly -> legs
+                Nothing -> listOf(Null)
+            }
+
     @get:JvmName("asMono")
     val asMono get() = this as Mono
 
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -555,7 +555,7 @@
 
 (defmethod codegen-expr :let [{:keys [local expr body]} opts]
   (let [{^VectorType local-return-type :return-type, continue-expr :continue, :as emitted-expr} (codegen-expr expr opts)
-        emitted-bodies (->> (for [local-type (.getLegs local-return-type)]
+        emitted-bodies (->> (for [local-type (.getReadsAs local-return-type)]
                               (MapEntry/create local-type
                                                (codegen-expr body (assoc-in opts [:local-types local] local-type))))
                             (into {}))
@@ -572,7 +572,7 @@
   (let [{continue-expr :continue, ^VectorType expr-return-type :return-type, :as emitted-expr}
         (codegen-expr expr opts)
 
-        expr-rets (set (.getLegs expr-return-type))
+        expr-rets (set (.getReadsAs expr-return-type))
 
         emitted-thens (->> (for [local-type (disj expr-rets #xt/type :null)]
                              (MapEntry/create local-type
@@ -721,7 +721,7 @@
 
         all-arg-types (reduce (fn [acc {:keys [^VectorType return-type]}]
                                 (for [el acc
-                                      arg-type (.getLegs return-type)]
+                                      arg-type (.getReadsAs return-type)]
                                   (conj el arg-type)))
                               [[]]
                               emitted-args)
@@ -1969,8 +1969,8 @@
 
         (let [inner-calls (->> (for [field fields]
                                  (MapEntry/create field
-                                                  (->> (for [l-val-type (VectorType/.getLegs (get l-field-types field))
-                                                             r-val-type (VectorType/.getLegs (get r-field-types field))]
+                                                  (->> (for [l-val-type (VectorType/.getReadsAs (get l-field-types field))
+                                                             r-val-type (VectorType/.getReadsAs (get r-field-types field))]
                                                          (MapEntry/create [l-val-type r-val-type]
                                                                           (if (or (= #xt/type :null l-val-type) (= #xt/type :null r-val-type))
                                                                             {:return-type #xt/type :null, :->call-code (constantly nil)}
@@ -2017,8 +2017,8 @@
       {:return-type #xt/type :bool, :->call-code (constantly false)}
       (let [inner-calls (->> (for [field fields]
                                (MapEntry/create field
-                                                (->> (for [l-val-type (VectorType/.getLegs (get l-field-types field))
-                                                           r-val-type (VectorType/.getLegs (get r-field-types field))]
+                                                (->> (for [l-val-type (VectorType/.getReadsAs (get l-field-types field))
+                                                           r-val-type (VectorType/.getReadsAs (get r-field-types field))]
                                                        (MapEntry/create [l-val-type r-val-type]
                                                                         (codegen-call {:f :===, :arg-types [l-val-type r-val-type]})))
                                                      (into {}))))
@@ -2061,11 +2061,11 @@
 (defmethod codegen-call [:_patch :struct :struct] [{[^VectorType$Struct l-type, ^VectorType$Struct r-type] :arg-types}]
   (let [l-ks (into {} (.getChildren l-type))
         r-ks (into {} (.getChildren r-type))
-        flattened-r-types (update-vals r-ks (fn [^VectorType r-field-type]
-                                               (set (.getLegs r-field-type))))
+        r-read-legs (update-vals r-ks (fn [^VectorType r-field-type]
+                                       (set (.getReadsAs r-field-type))))
         ret-type (VectorType/structOf (into l-ks
                                             (map (fn [[k ^VectorType r-field-type]]
-                                                   (let [r-types (get flattened-r-types k)]
+                                                   (let [r-types (get r-read-legs k)]
                                                      (MapEntry/create k
                                                                       (if-not (contains? r-types #xt/type :null)
                                                                         r-field-type
@@ -2087,7 +2087,7 @@
                                            (let [k-str (str k)]
                                              [k-str
                                               (if (and (contains? l-ks k)
-                                                       (get-in flattened-r-types [k #xt/type :null]))
+                                                       (get-in r-read-legs [k #xt/type :null]))
                                                 `(-patch (get ~l-sym ~k-str) (get ~r-sym ~k-str))
                                                 `(get ~r-sym ~k-str))])))
                                     (keys r-ks))))))}))
@@ -2259,8 +2259,8 @@
           n-sym (gensym 'n)
           len-sym (gensym 'len)
           res-sym (gensym 'res)
-          inner-calls (->> (for [l-el-type (.getLegs l-el-type)
-                                 r-el-type (.getLegs r-el-type)]
+          inner-calls (->> (for [l-el-type (.getReadsAs l-el-type)
+                                 r-el-type (.getReadsAs r-el-type)]
                              (MapEntry/create [l-el-type r-el-type]
                                               (if (or (= #xt/type :null l-el-type) (= #xt/type :null r-el-type))
                                                 {:return-type #xt/type :null, :->call-code (constantly nil)}
@@ -2306,8 +2306,8 @@
         n-sym (gensym 'n)
         len-sym (gensym 'len)
         res-sym (gensym 'res)
-        inner-calls (->> (for [l-el-type (.getLegs l-el-type)
-                               r-el-type (.getLegs r-el-type)]
+        inner-calls (->> (for [l-el-type (.getReadsAs l-el-type)
+                               r-el-type (.getReadsAs r-el-type)]
                            (MapEntry/create [l-el-type r-el-type]
                                             (codegen-call {:f :===, :arg-types [l-el-type r-el-type]})))
                          (into {}))]
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -432,7 +432,7 @@
 (defn- continue-read [f ^VectorType vec-type reader-sym & args]
   (if (instance? VectorType$Mono vec-type)
     (f vec-type (apply read-value-code vec-type reader-sym args))
-    (continue-read-union f (set (.getLegs vec-type)) reader-sym args)))
+    (continue-read-union f (set (.getReadsAs vec-type)) reader-sym args)))
 
 (def ^:dynamic ^String *snapshot-token* nil)
 (def ^:dynamic ^String *await-token* nil)
--- a/core/src/main/clojure/xtdb/expression.clj
+++ b/core/src/main/clojure/xtdb/expression.clj
@@ -515,9 +515,9 @@
                                     :or {extract-vecs-from-rel? true}}]
   (let [return-type (or (get var-types variable)
                         (throw (AssertionError. (str "unknown variable: " variable))))
-        ;; `Nothing` is the catalog's lattice bottom for a declared-but-valueless column; when such a
-        ;; column is read in a query its rows are simply null, so the EE treats it as `:null` and never
-        ;; needs `:nothing` call definitions (Nothing's empty legs would otherwise collapse codegen).
+        ;; `readsAs` handles the read wherever legs are dispatched on, so this looks redundant - it is not.
+        ;; `:return-type` flows outward into writer allocation and the merges above, where the bottom would
+        ;; declare a column that has no values while the emitted code writes nulls into it.
         return-type (if (= return-type VectorType$Nothing/INSTANCE) #xt/type :null return-type)
         var-rdr-sym (gensym (util/symbol->normal-form-symbol variable))]
 
--- a/src/test/clojure/xtdb/expression_test.clj
+++ b/src/test/clojure/xtdb/expression_test.clj
@@ -15,7 +15,7 @@
            (java.time Duration Instant InstantSource LocalDate LocalDateTime LocalTime Period ZoneId ZonedDateTime)
            (java.time.temporal ChronoUnit)
            org.apache.arrow.vector.types.TimeUnit
-           (xtdb.arrow RelationReader Vector VectorReader VectorType VectorType$Nothing)
+           (xtdb.arrow RelationReader Vector VectorReader)
            (xtdb.time Interval)
            (xtdb.util StringUtil)))
 
@@ -1745,7 +1745,7 @@
 
 (t/deftest a-column-of-empty-lists-compares-as-a-literal-does-5948
   (with-open [rel (tu/open-rel {:xs [[] []]})]
-    (t/is (= (types/->type [:list VectorType$Nothing/INSTANCE])
+    (t/is (= #xt/type [:list :nothing]
              (.getType (.vectorForOrNull rel "xs")))
           "the column's element type is the lattice bottom, which is what makes this test cover anything")
 
@@ -1758,12 +1758,12 @@
 
   (t/testing "an element written as null heals to `Null`, an empty list does not"
     (with-open [rel (tu/open-rel {:xs [[nil]]})]
-      (t/is (= (types/->type [:list :null]) (.getType (.vectorForOrNull rel "xs"))))
+      (t/is (= #xt/type [:list :null] (.getType (.vectorForOrNull rel "xs"))))
       (t/is (= [nil] (:res (run-projection rel '(== xs [nil])))))))
 
   (t/testing "a null list leaves the element type at the bottom"
     (with-open [rel (tu/open-rel {:xs [nil []]})]
-      (t/is (= (VectorType/maybe (types/->type [:list VectorType$Nothing/INSTANCE]))
+      (t/is (= #xt/type [:? :list :nothing]
                (.getType (.vectorForOrNull rel "xs"))))
       (t/is (= [nil true] (:res (run-projection rel '(== xs []))))))))
 
--- a/src/test/clojure/xtdb/expression_test.clj
+++ b/src/test/clojure/xtdb/expression_test.clj
@@ -15,7 +15,7 @@
            (java.time Duration Instant InstantSource LocalDate LocalDateTime LocalTime Period ZoneId ZonedDateTime)
            (java.time.temporal ChronoUnit)
            org.apache.arrow.vector.types.TimeUnit
-           (xtdb.arrow RelationReader Vector VectorReader)
+           (xtdb.arrow RelationReader Vector VectorReader VectorType VectorType$Nothing)
            (xtdb.time Interval)
            (xtdb.util StringUtil)))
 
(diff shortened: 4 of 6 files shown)


danmason:
Just on the testing side - is it worth a quick sql/node test with the ingestion stopped repro from #5948?

Ie;
```
INSERT INTO docs (_id, xs) VALUES (1, ARRAY[]);
SELECT _id FROM docs WHERE xs = ARRAY[1];   -- NPE
```

mbutlerw:
> Just on the testing side - is it worth a quick sql/node test with the ingestion stopped repro from #5948?
> 
> Ie;
> 
> ```
> INSERT INTO docs (_id, xs) VALUES (1, ARRAY[]);
> SELECT _id FROM docs WHERE xs = ARRAY[1];   -- NPE
> ```

Done, should have been covered by the EE tests, but happy enough to add a e2e repro :+1:

mbutlerw:
Closed in favour of https://github.com/xtdb/xtdb/pull/5969

Issue #5873, opened 2026-08-10
flaky: `read-only-node-sees-data-after-block-flush` loses a newly-added column


jarohen:
## tl;dr

- **The test intermittently loses a *column*, not a row** — the second block's row comes back as `{:xt/id "b", :x 2}`, missing the `:y "hello"` it was written with.
- **Only under full-suite load** — 1 failure across a full `./gradlew test`, 0 failures across 5 isolated runs of the namespace with `--rerun-tasks`.
- **`assumption:` the follower's column set is stale, not its data.** The row itself arrived, so the scan looks to have planned against a table catalog that hadn't yet picked up `:y`.
- **Same family as #5417**, the bug this test was written to cover — follower visibility lagging a block flush, here at column granularity rather than table granularity.

## Symptoms

`xtdb.read-only-node-test/read-only-node-sees-data-after-block-flush`, asserting at `src/test/clojure/xtdb/read_only_node_test.clj:68`.

```
ro node sees all data after second block flush
expected: (= #{{:xt/id "b", :x 2, :y "hello"} {:xt/id "a", :x 1}} (set (xt/q ro-node "SELECT * FROM foo")))
actual:   (not (= #{{:xt/id "b", :x 2, :y "hello"} {:xt/id "a", :x 1}} #{{:xt/id "a", :x 1} {:xt/id "b", :x 2}}))
```

- **Both rows are present, and `a` is complete.** Row `b` carries its `_id` and `:x` but not `:y`, so this is not a missing-rows failure.
- **`:y` is new in the second block.** The first tx writes `{:xt/id "a", :x 1}` and flushes; the second writes `{:xt/id "b", :x 2, :y "hello"}` and flushes — so `:y` is a column the table did not have when the first block closed.
- **The test does sync before querying.** `(xt-log/sync-node ro-node #xt/duration "PT5S")` sits between the second `flush-block!` and the failing query, so whatever is late is late *after* the ro node has caught up on the log.

## Analysis

[...]

PR #6055, opened 2026-09-10
refactor(indexer): the leader adopts a block on reading its own upload back

jarohen:
The leader now adopts a block when it reads its own `BlockUploaded` back off the replica log, the way a follower always has, which closes a state a term dying mid-upload could not recover from: its catalog past a block whose upload never landed, and nothing left able to finish it. Doing it now because #6054 has just landed the leader-side restructuring that makes the two roles comparable at all — before it, the produce and the adopt were not separable objects.

## tl;dr

- **Both roles adopt a block on the same record**
  A follower closes a block on `BlockUploaded`. The leader did the whole close inline during its own `BlockBoundary` apply — files, append, catalog refresh, index roll — and then nothing at all on reading that message back. Now the boundary produces and holds the block, and the upload adopts it.

- **The old ordering left a hole a re-promotion falls into**
  `tableCatalog.refresh(block)` ran before the append was awaited, so a term ending in between put this node's catalog past a block that never became durable. Deferring the refresh is what keeps the block re-producible.

- **The widened window is held by the block-cut pause, which now carries two obligations**
  The follower's bounded pending-block buffer, as before, and — new — keeping anything from applying between `finishBlock` and `nextBlock()`, where the live tables have been snapshotted into L0 and are about to be cleared.

- **The leader becomes the last node to adopt a block rather than the first**
  By its own replica-log consumer lag. The source watermark moves with the adopt, on both roles, so a caller that flushes a block and awaits that watermark is still told once the flush is done.

## Problem

- **The two roles closed a block at different points, and only one of those was safe to interrupt**
  `FollowerLogProcessor.closeBlock` reads the block file, folds the tries, refreshes the catalog, rolls the live index and signals the compactor — all on reading `BlockUploaded`. `BlockCutter.uploadBlock` did the same adopt steps inline, during the apply of the boundary, having just written the files and appended the message itself; `LeaderLogProcessor`'s `BlockUploaded` branch then advanced a watermark and did nothing else.

- **Refresh-before-append is what makes a mid-upload death unrecoverable, and the trace is short**
  `refresh(block)` sets `currentBlockIndex` to N. The awaited append then fails, or the term is cancelled inside it. The apply never completes, so the partition's tail re-offers the boundary to the re-opened follower — which drops it as stale, `blockIndex <= currentBlockIndex` now being true. So `nextBlock()` never runs, the live index sits on a block it has already snapshotted into L0, and a promotion back to this node reaches `TableCatalog.buildBlock`'s `check(currentBlockIndex < blockIndex)` and hard-errors.

- **Queries stay correct throughout, which is why this has not been seen**
  `Snapshot.open` partitions live tables against the trie catalog's published L0 rather than against the table catalog (#5873), so the block's rows are served from exactly one side of the gap however long it lasts. The next block's cut also recovers the node. What does not recover is leadership returning to it in between.

- **`BlockCutter`'s own comment was reaching for this and could not get there**
  It read "unable to re-cut it because its own catalog has moved on", and argued from that for an awaited direct append. But the append can fail too — the ordering is what matters, not the awaiting.

## What changes

- **`BlockCutter` splits into produce and adopt**
  `upload(pendingBlock)` finishes the live-index block, registers the L0 tries, writes the per-table block files and the block file, and appends `BlockUploaded`. `closeBlock(msg)` refreshes the table catalog, rolls the live index, posts the L0s to the source log and signals the compactor, and hands back what was held. The `Uploading` state carries the built block and the added tries, because the adopt cannot recompute them once produce has passed.

- **The leader takes the follower's guard, in the follower's order**
  `applyReplicaMessage` is now own-term supersession check, then the pending-block hold, then the term fence, holding ahead of the fence for the follower's existing reason.

- **The guard is there for the drain, not for steady state**
  Nothing arrives between a boundary and its upload on a healthy leader — resolution is paused and the GC arm unarmed. But the held records drain at the adopt, back through the same apply, and a `BlockBoundary` among them opens the next block there; without the guard the records behind it would apply into a block already snapshotted. `LogProcessor.replayHeldRecords` is gone with it, and so is the `pendingBlock = null` between the promotion's upload and its replay, which existed only so a failure could not hand the block to a follower that would close it twice.

- **The handover is symmetric**
  `demoteLeader` reads the leader's pending block after `cancelAndJoin` and passes it to `openFollower`, mirroring what the promotion already did with the follower's — and the read follows the join for the same reason, since the term applies until then and one of those applies may be the adopt.

- **A block's covered source position is reported at the adopt, by whichever role adopts it**
  `BlockBoundary` no longer advances the source watermark on either role. On the leader that deletes the earlier of two notifies carrying the same value; on the follower it adds one to a `closeBlock` that never had any. Both halves are needed, because either role can adopt a block — a promotion adopts one the previous leader's boundary cut, a demotion hands a produced-but-unadopted one to a follower. Three callers flush a block and then await that watermark: `tu/flush-block!`, `Xtdb.flushBlock`, and `reset-compactor!`, which goes on to read both catalogs and delete compacted files.

- **Specs**
  `db.allium` moves `pending_block` onto the base `LogProcessor` entity, splits the cutter's contract, adds `LeaderAdoptsOwnBlock` and drops `BlockUploaded` from `LeaderAppliesOwnControlRecord`. `BlockBecameDurable` fires at the adopt, so `ExternalSourceLearnsDurableExtent` now gates on a block both durable and recorded. `log-processor-lifecycle.allium` gives `LeaderTerm` a `mid_block` carried through both transitions; `live-index.allium`'s resolution pause ends at the adopt.

## Consequences

- **`currentBlockIndex` and the source watermark both advance one replica round trip after the block file lands**
  Invisible to the resolve side: resolution resumes after the adopt either way, so the `FlushBlock` CAS reads the value it always did, and `BlockCutter.cut` is still structurally excluded from the window by `acceptingResolution`. The two moving together is what extends that to whoever *chooses* the CAS value, which is read from outside the term.

- **A promotion inside a held block resumes the source log one step earlier**
  `resumeAfterMsgId` and `TxResolver`'s seed both read that watermark, so such a term resumes at the last applied transaction and re-reads whatever the held boundary covered — the `FlushBlock` that triggered the cut, or nothing at all where the row gauge cut it. Resolution is unarmed until the adopt, so the re-offered batch waits, and by the time that `FlushBlock` re-resolves the adopt has advanced the block index, so its CAS fails and it lands as a `NoOp` (#5680).

- **Three things off the resolve side can see it, all within tolerance**
  healthz block-lag reads 1 during the window against a `<= 5` threshold, so nothing logs or 503s. `TableCatalog.latestBlock` emits later, which only holds an external source's upstream confirmation further back — the direction `ConfirmedPositionIsRecoverable` wants. The block GC waits one more cycle.

- **The `finishBlock` … `nextBlock()` exclusion is now an argument, not a lexical scope**
  It rests on `acceptingResolution` being false for the whole of `Uploading` — which unarms the source-log, ext-source and GC clauses — plus the fence, leaving only three things able to arrive: our own upload, a newer term (superseding before it applies), an older one (fenced, applying nothing). That is stated as an obligation in `db.allium` and `live-index.allium` rather than left to be re-derived.

- **`FollowerLogProcessor.processRecord`'s `error(...)` becomes reachable if the handover is ever dropped**
  It is an `IllegalStateException` on a replica tail that nothing restarts, so the failure mode is that the partition stops consuming for the life of the process. The handover is what keeps it unreachable, and it is now load-bearing in a way it was not.

## Out of scope

- **The other two eager applications**
  `SourceMessage.TriesAdded` and the trie GC's `TriesDeleted` are both mutated on the resolve side and re-applied nowhere. Moving them is near-mechanical once this has landed, and `TriesDeleted` additionally wants a FIFO of in-flight tasks so the GC's await still returns with the catalog updated.

- **Reshaping `PendingBlock` now that nothing is applied off the read-back**
  Its adopt-input half is empty — the `BlockUploaded` carries the block index and the tries, the block file is on storage, and the upload is already appended before a demotion can happen — so the demotion handover added here dissolves. Its hold buffer belongs on `LogProcessor`, which spans both roles and so needs no handover at all. Only the boundary message genuinely has to reach a promoting leader, `buildBlock` needing four fields only the reader of that record holds. Left for a follow-up, along with `PendingBlock`'s 1024-record overflow `Fault`, which stops a replica tail nothing restarts.

- **A node-wide bound on block-upload concurrency**
  `BlockCutter` derives its own 16-way `limitedParallelism` view per term per database, and such a view of `Dispatchers.IO` is elastic, so N databases flushing together permit 16×N in-flight `TableBlock.toByteArray()` buffers. Inherited from #6054, not introduced here.

- **Three pre-existing spec drifts, found while editing**
  `db.allium` still refers to a "transition processor" that no longer exists; `gc.allium` and `block-gc.allium` describe the GC auto-signal as coming "from finishBlock"; and `ExternalSourceLearnsDurableExtent` calls a `durableExtentAdvanced` surface the code does not have, exposing a `StateFlow` instead.

## Alternative approaches

- **Make a `BlockUploaded` with no pending block a legitimate close, instead of handing the block over**
  Well-defined — the block file is on storage, and the adopt needs nothing from the boundary — and safe precisely because the block-cut pause means a node in that window cannot have applied anything from the next block. It would turn a terminal assertion into a recovery and need no handover on the demotion path. It does not remove the need for one on the promotion path, though, because the held records must survive to the adopt.

- **Re-read the block file at the adopt, converging with the follower's call**
  Ruled out: `TableCatalog.refresh(block, metadata)` folds the metadata deltas that `tableCatalog.finishBlock` has already folded on the leader, so the leader cannot share the follower's call without double-counting row counts and HLL deltas. With the code convergence unavailable, the storage round trip bought nothing.

- **Queue the upload on the append pump like the boundary**
  Losing it is recoverable now — this node has not adopted the block either, so the next leader produces it again — so the pump would work. Awaiting keeps the window to the read-back alone and reports a failed append to the term, which leaves the boundary unapplied for the next role to pick up.

## Open questions

- **Should the two new `error(...)` sites be `err/fault` instead?**
  Both are unreachable-state assertions, and `IllegalStateException` and `Fault` take the same path through `runTerm`'s catch to `watchers.notifyError`, so there is no behavioural difference. They match their immediate neighbours — `BlockCutter.addRows` and `FollowerLogProcessor.processRecord` — against a literal reading of CODING.adoc's "new code should always use it".

- **Is a deterministic test of the demotion window worth building?**
  The pieces are pinned separately: producing leaves `pendingBlock` non-null, a follower seeded with an inherited block closes it on that upload, and `demoteLeader`'s hand-on is a single expression. The end-to-end interleaving is left to `LogProcessorSimTest`, which drives rebalances under a seeded scheduler at 100 iterations.

## Verification

`./gradlew test` — 2300 tests, 0 failures, across all nine modules that produce results. `./gradlew property-test` — 2648 tests, 0 failures, at 100 iterations, including `LogProcessorSimTest` (300), `NodeSimulationTest` (800) and `CompactorSimulationTest` (813). `allium check` reports the same diagnostics and no findings on all five edited specs.

### DIFF
--- a/allium/live-index.allium
+++ b/allium/live-index.allium
@@ -85,10 +85,16 @@
 --   it reflects only consumed-back txs). When the gauge crosses rows_per_block the
 --   resolver injects a BlockBoundary (in resolution order, so it lands after this
 --   block's txs and before the next block's) and PAUSES resolution — the source-log,
---   ext-source and GC select arms are EXCLUDED — until the boundary is read back and its
---   block uploaded. This pause keeps the follower's bounded pending-block buffer from
---   overflowing (no tx interleaves between the boundary and its upload). The pause is
---   deliberately the status quo; block double-buffering is separate future work (#4495).
+--   ext-source and GC select arms are EXCLUDED — until the block's own BlockUploaded is
+--   read back and the block adopted.
+--   The pause carries two obligations.
+--   It keeps the follower's bounded pending-block buffer from overflowing, no tx
+--   interleaving between the boundary and its upload.
+--   And it keeps anything from applying between FinishBlock and NextBlock, where the live
+--   tables have been written to block storage and are about to be cleared — rows applied
+--   there would be written nowhere.
+--   The pause is deliberately the status quo; block double-buffering is separate future
+--   work (#4495).
 --
 --   Two heads, two homes — callers choose:
 --     - the RESOLVED head, TxResolver.latest_completed_tx: the highest tx RESOLVED (and
@@ -936,9 +942,9 @@ rule FinishBlock {
     -- read back, resumes any resolution paused by CutBlock). Rows remain in memory
     -- until NextBlock.
     --
-    -- Because resolution paused at the boundary and stays paused until this upload,
-    -- nothing was staged-and-applied between the boundary and here — a durable L0 block
-    -- never contains rows from a later block.
+    -- Because resolution paused at the boundary and stays paused until this upload is read
+    -- back, nothing was staged-and-applied between the boundary and here — a durable L0
+    -- block never contains rows from a later block.
     when: FinishBlock(index, block_idx)
 
     -- Writes current (durable) live table contents to block storage.
--- a/allium/log-processor-lifecycle.allium
+++ b/allium/log-processor-lifecycle.allium
@@ -43,7 +43,9 @@ external entity FollowerTerm {
     -- Reading the replica log is the Listener's: one read feeds whichever role is live, a follower building state and a leader confirming its own writes alike.
     -- The position it has reached is the partition's, so a role change continues from it rather than restarting.
     replica_pos: MessageId
-    mid_block: Boolean   -- true if an old leader died before uploading its block
+
+    -- A block is in flight on this node: a boundary read but not yet confirmed, whether the block behind it is still to be produced or was produced by a term that ended before reading its own confirmation back.
+    mid_block: Boolean
 }
 
 external entity LeaderTerm {
@@ -67,6 +69,10 @@ external entity LeaderTerm {
     -- term could be admitted twice, once either side of a demote.
     term: Integer
     replay_target: MessageId
+
+    -- A block this term has produced and not yet adopted: produced in step 4, adopted on reading a confirmation of it back, which is after step 5 published this term.
+    -- A term ending in between hands it to the role that replaces it (DemoteLeader).
+    mid_block: Boolean
 }
 
 external entity RecordProcessor {
@@ -342,23 +348,22 @@ rule TransitionToLeader {
         -- Either above this transition's term refuses the claim, as step 2a does, and takes the same failure path.
         ClaimStillUnfenced(follower_term: listener.follower_term, term: term)
 
-        -- Step 4: finish any dangling block.
-        -- If the follower was mid-block (an old leader died before uploading it), the incoming leader uploads the block and then replays the records the follower was holding, with leader-upload semantics.
+        -- Step 4: produce any dangling block.
+        -- If the follower was mid-block (an old leader died before uploading it, or a term on this node produced it and never read it back), the incoming leader uploads the block's files and appends its confirmation under its own term.
         -- Requires the follower stopped (step 3) so it cannot double-process.
+        -- The block stays in flight: this term adopts it, and replays the records the follower was holding, on reading a confirmation of it back — which is after step 5, the Listener's read being what delivers it to the term. db.allium owns the adopt (LeaderAdoptsOwnBlock).
+        -- That confirmation is this term's own, or the outgoing leader's where that term got one away before it ended: both confirm the same block, and the adopt takes whichever arrives first.
+        -- So the block is carried into the leader term below, and a term that ends before adopting it hands it on again (DemoteLeader, RecoverFollowerOnTransitionCancelled).
+        --   A block produced and not adopted MUST reach the role that replaces this one: the adopt is what moves this node's catalog past the block, so the confirmation does not read as stale, and a role holding no block to match it against stops the partition's read.
         if listener.follower_term.mid_block:
-            DanglingBlockUploaded(follower_term: listener.follower_term)
-
-            -- The block is finished on this node once the upload lands, so the transition stops carrying it here, between the upload and the replay.
-            -- The failure path below re-opens a follower from what the transition is carrying, and one still holding this block would match the upload just appended and finish the same block a second time.
-            listener.follower_term.mid_block = false
-
-            HeldRecordsReplayed(follower_term: listener.follower_term)
+            DanglingBlockProduced(follower_term: listener.follower_term)
 
         -- Step 5: build the leader term, carrying the fencing term, and publish it as the role.
         -- Replica records reach it from the Listener's read, continuing from where the follower left off.
         listener.leader_term = LeaderTerm.created(
             term: term,
-            replay_target: target
+            replay_target: target,
+            mid_block: listener.follower_term.mid_block
         )
         listener.follower_term = null
         listener.state = leading
@@ -392,9 +397,11 @@ rule DemoteLeader {
         -- Re-open a follower seeded from the leader's replica position (the durable
         -- live index is untouched; see live-index.allium). RHS reads the pre-rule
         -- leader_term, captured before the field is cleared.
+        -- A block the term produced and did not adopt goes with it, so the follower closes it on a confirmation of that block, whichever term wrote it.
+        -- It is read once the term is joined, as step 4 reads the follower's: the term goes on applying records until then, and one of them may be that confirmation.
         listener.follower_term = FollowerTerm.created(
             replica_pos: listener.leader_term.replay_target,
-            mid_block: false
+            mid_block: listener.leader_term.mid_block
         )
         listener.leader_term = null
         listener.state = following
@@ -434,8 +441,7 @@ rule RecoverFollowerOnTransitionCancelled {
     ensures:
         -- Re-open a live follower seeded from the (now stale) follower position, so
         -- the role is backed by a live replica consumer again. It carries the dangling block
-        -- forward while that block is still unfinished — up to step 4's upload and not past it,
-        -- which is what step 4 clears the flag between the upload and the replay for.
+        -- forward, step 4's produce having run or not: either way the block is in flight and this node's catalog has not moved past it, so the re-opened follower is what closes it.
         listener.follower_term = FollowerTerm.created(
             replica_pos: listener.follower_term.replica_pos,
             mid_block: listener.follower_term.mid_block
--- a/core/src/main/kotlin/xtdb/api/log/Watchers.kt
+++ b/core/src/main/kotlin/xtdb/api/log/Watchers.kt
@@ -81,8 +81,8 @@ class Watchers(
 
         state.updateIfActive {
             if (txId != null) check(txId > it.latestTxId) { "txId $txId <= latestTxId ${it.latestTxId}" }
-            // >= not >: BlockBoundary can carry the same source msgId as the preceding ResolvedTx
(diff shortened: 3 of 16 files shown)


jarohen:
## tl;dr

- **The `BlockUploaded … with no block in flight` error this PR adds is reachable, and it stops the database when it fires.**
Two terms can produce the same block index, because a boundary is a record and a follower still holding one produces that block itself on promotion. The incoming leader then adopts on the outgoing leader's upload and dies on its own.

- **This belongs on this branch rather than on #5927.**
The guard, the two-phase window that makes a duplicate possible, and the `db.allium` sentence asserting it cannot happen are all this PR's — none of them exist on `main`.

- **Latent here, routine under self-election.**
It needs a rebalance to coincide with replication lag on an in-flight upload, so nobody has hit it; on #5927 it wedged 208 of 2942 simulation iterations.

## Why the guard's reasoning has a hole

- **`BlockCutter.closes()` matches on block index, storage version and epoch, never on term — and it runs ahead of the fence.**
Deliberately: a record held behind an open block has not been acted on, so folding it would move the high-water past the term that cut the boundary and fence away the upload that closes it. The consequence is that an upload at a *lower* term still closes the block a role is holding.

- **So the incoming leader adopts its own block on the outgoing leader's confirmation of it, and its own confirmation then has nothing to match.**
`blockCutter.pendingBlock` is null by the time that arrives, so `applyReplicaMessage` falls through to `is BlockUploaded -> error(...)`. The guard's three cases — at our own term nothing else writes our replica log, a newer term supersedes above, an older one is fenced below — miss this one, because the record *is* ours, at our own term, and the fence never saw the duplicate that beat it.

- **It surfaces as a fault rather than a resignation.**
The term fails, and its teardown reports to `Watchers`, so the database stops answering queries.

## Evidence

Three nodes, one database, from a `LogProcessorSimTest` run on #5927. `b02`'s boundary is at replica 35; the upload from term 2 lands at 37 and the one from term 3 at 38.

```
block boundary b02: source=9, replica=35 — waiting for BlockUploaded...
[replica-tail#69780] claiming leadership at term 0.3
[replica-tail#69780] claim at term 0.3 conferred leadership
[replica-tail#69780] transition: producing pending block b2 with 1 held records
[term#69879]        BlockCutter - block uploaded b02: source=9, replica=37   <- term 2's
[replica-tail#69780] BlockCutter - block uploaded b02: source=9, replica=38   <- term 3's
follower: block uploaded b02: source=9, replica=37 (1 buffered)
follower: discarding fenced record 37 (term 0.2 < 0.3)
[term#69879] superseded: read term 3 > our term 2 at 36
[term#69930] ERROR LeaderLogProcessor - [test-db] leader term failed
java.lang.IllegalStateException: [test-db] BlockUploaded b02 at 38 with no block in flight
[term#69930] ERROR Watchers - ingestion stopping
```

- **The two uploads are the same block written twice.**
Both were cut from the one boundary over the one prefix, and `assertBlockCutsAgree` — uploads grouped by index must agree on the source position they cover — passes throughout. Adopting either is correct; only the bookkeeping is wrong.

## How it is reached without self-election

- **A leader appends `BlockUploaded b02`; a rebalance then moves leadership to a node that has not yet read that record.**
That node still holds the boundary, so `demoteLeader`/`transitionToLeader` hand it the pending block. It produces `b02` again under the new generation, adopts on whichever upload reaches it first, and errors on the other.

- **The rarity is the coincidence, not the mechanism.**
A rebalance has to fall inside the gap between one node appending an upload and another reading it. Self-election removes the coincidence: every empty poll is a claim opportunity, which is the difference between never observed and 7% of iterations.

## The fix

```kotlin
is ReplicaMessage.BlockUploaded ->
    if (msg.blockIndex > (tableCatalog.currentBlockIndex ?: -1))
        error(
            "[$dbName] BlockUploaded b${msg.blockIndex.asLexHex} at ${record.msgId} " +
                    "with no block in flight"
        )
```

- **A duplicate is stale by block index, which is already every other reader's answer.**
`FollowerLogProcessor` drops it on exactly this test (`blockIndex <= currentBlockIndex`), and `db.allium`'s Staleness table states it for `BlockUploaded` without qualifying by role. `tableCatalog` is already a field on `LeaderLogProcessor`, so this applies as it stands. Only an upload the catalog has *not* reached is the one nothing explains, and that is what the error is left holding.

- **`dev/doc/db.allium`, `LeaderAdoptsOwnBlock`:** "Every `BlockUploaded` a leader reads back matches the block it is holding … so this is the only rule that acts on one" is the sentence that is false. The match is on index, version and epoch rather than on term, so the first upload of an index to reach a node is the one it adopts on, whichever term wrote it.

- **`LogProcessorSimTest.assertBlockBoundariesMatchUploads`** compares boundaries against uploads as lists, so a duplicate fails it. It wants `uploads.distinct()`.

## The test stays on #5927

`a second upload of a block the term has adopted is dropped` drives the duplicate deterministically — it wraps `LogProcessor.LogsDriver` to append a copy at the outgoing term ahead of the real one, and promotes by claim. Both of those are #5927's, so a version here would have to go through the group: a rewrite rather than a port. Flagging it so this branch knows the fix would arrive unpinned by a test of its own.

jarohen:
Superseded by #6058, which carries this branch's commit together with the four that followed it, as one unit.

- **This commit is now `9dae3a8df`** — the same change, with the duplicate-upload guard correct from the start rather than fixed afterwards. A bug in a commit we introduced gets fixed up into it, so the `BlockUploaded … with no block in flight` fault reported on this thread never exists in the history.

- **The analysis on this thread is what found it**, and #6058's description carries the mechanism in its own terms: two terms can produce one block index, `closes()` matches on index, version and epoch rather than on term, and it runs ahead of the fence — so the incoming leader adopts on the outgoing leader's confirmation and its own arrives with nothing in flight. The deterministic test stays on the self-election branch, which is noted there as an open question rather than closed over.

- **Why one PR rather than three**: the five commits are one idea — a leader applies its catalog changes on reading its own records back — and split up, a reader has to reconstruct that from three descriptions, each opening by referring back to the previous one.

PR #6056, opened 2026-09-10
the leader applies the tries messages on reading them back

jarohen:
The leader now applies `TriesAdded` and `TriesDeleted` when it reads them back off the replica log, rather than eagerly while resolving, so every node in the cluster applies the same record at the same position in its own log. Until now the leader mutated its trie catalog at resolve time and skipped the record on the way back, which left a lost append silently diverging the cluster — permanently, in the delete case, since garbage collection is leader-only and nothing re-issues the removal. Doing it now because #6055 has just moved the block close the same way, and these two are what remained: with them still eager, the leader's apply path ran two disciplines at once and every ordering argument had to name which.

> Stacked on #6055. The first commit here is that PR's; review only `refactor(indexer): the leader applies the tries messages on reading them back`.

## tl;dr

- **The leader stops being a special case**
  Its `TriesAdded` branch is now the follower's, line for line — the same storage-version and epoch guard, then the same watermark notify. `SourceLogProcessor` forwards the message and no longer reaches into storage at all.

- **A lost `TriesDeleted` append used to diverge the cluster permanently; now it is self-healing**
  The old order removed from the local catalog first and appended second. GC is leader-only, so a term dying in between left the leader correct and every other node holding entries for files that no longer exist, with nothing to re-issue the removal — the leader's `garbageTries` will not select them again. Deferred, nothing is removed anywhere until the record lands, so the next pass re-selects the same tries.

- **Log order now discharges an ordering obligation that was argued by hand**
  `gc.allium`'s DualWriteOrdering required a catalog removal to stay ahead of any block upload appended behind it. With the removal on the resolve side and the upload on the apply side, that rested on the block-cut pause; with both as replica-log appends from one term, applied in append order everywhere, it is the log's.

- **A GC commit returns one replica round trip later**
  `commitTriesDeleted` used to complete at the append. It now completes when the record comes back — which is what the GC's own contract needed all along, since it has already deleted the files and must not carry on against a catalog that still lists them.

## Problem

- **Two mutations were applied at resolve time and skipped on the way back**
  `SourceLogProcessor` mutated the trie catalog while resolving `SourceMessage.TriesAdded`, then appended a `ReplicaMessage.TriesAdded` for everyone else; `GarbageCollector` did the same for a removal, completing the GC's await at the append. `LeaderLogProcessor`'s branches for both records then did nothing but advance a watermark, each carrying a comment explaining that the catalog was already updated.

- **The leader's catalog therefore moved at a different point from every other node's**
  Both had reasons — the compactor and the GC read the catalog synchronously, and the block-cut pause serialises trie mutations against boundaries — but the effect is that "when does this land" had two answers depending on which node you asked, and each ordering property had to be re-argued for the leader separately.

- **In the delete case the divergence is not recoverable**
  Garbage collection runs only on the leader. Remove locally, then fail to append — a term ending between the two — and no node other than this one ever learns. The files are already gone, so every follower now lists tries it cannot read, and the leader will not re-select them because its own catalog no longer holds them.

- **In the add case it is a smaller version of the same thing**
  A lost append leaves the leader holding tries no follower has. That one does self-correct, because the compactor re-selects the job, but only after the leader has done the work twice.

## What changes

- **`SourceLogProcessor` forwards `TriesAdded` and stops touching the catalog**
  The storage-version and epoch guard goes with the add: the message is forwarded whatever its storage version, and each node guards it on the way in — which is what the follower already did. With the add gone, `SourceLogProcessor` has no remaining use for storage, so the `bufferPool` and `trieCatalog` fields go and the `partitionStorage` constructor parameter with them.

- **`GarbageCollector` keeps a FIFO of appended-but-unread tasks**
  The select arm appends the record and parks the task on an `ArrayDeque`; `triesDeleted` pops the head, applies the removal and completes the task.

- **The FIFO is exact rather than approximate**
  Every record a leader applies is its own term's append: a higher term throws `LeaderSupersededException` before the fence, and `termFence.admit` refuses a lower one, the term's own claim `NoOp` having been folded before the term opened. So nothing else can appear between an append and its read-back, and the head is always the task the arriving record confirms.

- **`shutdown` abandons the in-flight tasks as well as the queued ones**
  A term dying between an append and its read-back would otherwise leave `commitTriesDeleted` suspended with nothing left to complete it.

- **Specs**
  `db.allium`: `LeaderProcessesTriesAdded` loses the eager add, and `LeaderAppliesOwnControlRecord` splits into `LeaderAppliesOwnTriesAdded`, `LeaderAppliesOwnTriesDeleted` and `LeaderAppliesOwnNoOp`, since the three now do different things. `gc.allium`: most of `CommitTriesDeletedRule`'s commentary was the argument for why the eager mutation was safe and goes with it, and `DualWriteOrdering` is restated on log order.

## Consequences

- **The GC's added latency is per chunk, off the resolve path**
  `TrieGarbageCollector` fans out over tables up to `DEFAULT_TABLE_PARALLELISM = 8` and awaits a commit per `TRIES_DELETED_CHUNK_SIZE = 1024` keys. Each of those awaits now spans a replica round trip rather than an append. Nothing on the resolve side waits on it, and the GC arm is unarmed for the whole of a block cut either way.

- **The compactor is unaffected**
  `publishTries` already appended to the source log and then awaited `watchers.awaitSource(msgId)`, so it was waiting for the read-back before this change; what moves is only where the catalog write happens in between.

- **A `deleteTries` on a shard the catalog does not hold trips its own spec assertion**
  `clojure.spec.check-asserts` is on in the test task, so a test that drives the new path with a real trie catalog needs real garbage, which needs a compaction. The two new timing tests mock the catalog for that reason and assert only when the commit returns; `xtdb.metrics-test/test-gc-metrics` covers the path end to end.

## Out of scope

- **The block's own L0 tries, which the leader still adds eagerly**
  `BlockCutter.uploadBlock` adds them to the trie catalog while building `addedTries` — the line marked `NOTE: side-effect here.` — where the follower takes them from `BlockUploaded.tries` at the adopt. Not the same move as this one, though: `getPartitions` six lines later reads the layout back out of the catalog and writes it into this block's own table-block files, so deferring the add means computing that layout from `finishedBlocks` instead.

- **`tableCatalog.finishBlock`'s metadata fold**
  Installed during produce, because its result is what the table-block files are built from. Only `currentBlockIndex` waits for the adopt, so the table catalog stays half-eager — and it is the same shape as the item above, the produce step reading its own side effect back out of a catalog rather than threading the value.

- **Reshaping `PendingBlock`**
  Noted as a follow-up on #6055; the two items above are its remaining prerequisites.

## Alternative approaches

- **Keep the eager delete and re-issue unacknowledged removals when a term opens**
  Rejected: it needs a durable record of what was removed and not replicated, which is the replica log — so this is the same mechanism with a second, weaker copy in front of it.

- **Match a read-back record to its task by content rather than by position**
  Rejected: position is exact given the term-identity argument above, and a content match would be a second source of truth for an ordering the log already fixes.

- **Leave `TriesAdded` eager and move only the delete**
  Rejected: the delete is the one with the unrecoverable failure, but leaving the add eager keeps the leader's apply path mixed, which is what makes each ordering property need its own argument.

## Open questions

- **Should the GC timing tests stop mocking the trie catalog?**
  Producing real garbage needs a compaction, which makes them integration-shaped. `test-gc-metrics` already covers that shape, so the mock buys a fast, deterministic test of the one thing these two assert — when the commit returns — at the cost of asserting nothing about what is removed.

## Verification

`./gradlew test` — 2303 tests, 0 failures. `./gradlew property-test` — 2648 tests, 0 failures, at 100 iterations, including `LogProcessorSimTest` (300), `NodeSimulationTest` (800) and `CompactorSimulationTest` (813). `allium check` reports the same diagnostics and no findings on both edited specs.

### DIFF
--- a/allium/live-index.allium
+++ b/allium/live-index.allium
@@ -85,10 +85,16 @@
 --   it reflects only consumed-back txs). When the gauge crosses rows_per_block the
 --   resolver injects a BlockBoundary (in resolution order, so it lands after this
 --   block's txs and before the next block's) and PAUSES resolution — the source-log,
---   ext-source and GC select arms are EXCLUDED — until the boundary is read back and its
---   block uploaded. This pause keeps the follower's bounded pending-block buffer from
---   overflowing (no tx interleaves between the boundary and its upload). The pause is
---   deliberately the status quo; block double-buffering is separate future work (#4495).
+--   ext-source and GC select arms are EXCLUDED — until the block's own BlockUploaded is
+--   read back and the block adopted.
+--   The pause carries two obligations.
+--   It keeps the follower's bounded pending-block buffer from overflowing, no tx
+--   interleaving between the boundary and its upload.
+--   And it keeps anything from applying between FinishBlock and NextBlock, where the live
+--   tables have been written to block storage and are about to be cleared — rows applied
+--   there would be written nowhere.
+--   The pause is deliberately the status quo; block double-buffering is separate future
+--   work (#4495).
 --
 --   Two heads, two homes — callers choose:
 --     - the RESOLVED head, TxResolver.latest_completed_tx: the highest tx RESOLVED (and
@@ -936,9 +942,9 @@ rule FinishBlock {
     -- read back, resumes any resolution paused by CutBlock). Rows remain in memory
     -- until NextBlock.
     --
-    -- Because resolution paused at the boundary and stays paused until this upload,
-    -- nothing was staged-and-applied between the boundary and here — a durable L0 block
-    -- never contains rows from a later block.
+    -- Because resolution paused at the boundary and stays paused until this upload is read
+    -- back, nothing was staged-and-applied between the boundary and here — a durable L0
+    -- block never contains rows from a later block.
     when: FinishBlock(index, block_idx)
 
     -- Writes current (durable) live table contents to block storage.
--- a/allium/log-processor-lifecycle.allium
+++ b/allium/log-processor-lifecycle.allium
@@ -43,7 +43,9 @@ external entity FollowerTerm {
     -- Reading the replica log is the Listener's: one read feeds whichever role is live, a follower building state and a leader confirming its own writes alike.
     -- The position it has reached is the partition's, so a role change continues from it rather than restarting.
     replica_pos: MessageId
-    mid_block: Boolean   -- true if an old leader died before uploading its block
+
+    -- A block is in flight on this node: a boundary read but not yet confirmed, whether the block behind it is still to be produced or was produced by a term that ended before reading its own confirmation back.
+    mid_block: Boolean
 }
 
 external entity LeaderTerm {
@@ -67,6 +69,10 @@ external entity LeaderTerm {
     -- term could be admitted twice, once either side of a demote.
     term: Integer
     replay_target: MessageId
+
+    -- A block this term has produced and not yet adopted: produced in step 4, adopted on reading a confirmation of it back, which is after step 5 published this term.
+    -- A term ending in between hands it to the role that replaces it (DemoteLeader).
+    mid_block: Boolean
 }
 
 external entity RecordProcessor {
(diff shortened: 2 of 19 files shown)


jarohen:
Superseded by #6058, which carries this branch's commit as one of five.

- **This commit is now `c7b565abf`**, rebased onto a base that fixes a fault this branch sat on top of — see #6055 for that thread.

- **Why one PR rather than three**: moving the tries messages is one step in a single change, that a leader applies its catalog changes on reading its own records back. Two further bugs only surfaced once the block and the tries had both moved — a block produced and never adopted double-counted its rows, because the table catalog's fold sums them — and #6058 presents the whole of it together.

Issue #6073, opened 2026-09-14
Postgres source's replication connection dropping every 5-10 minutes, stopping ingestion


jarohen:
On a live deployment the Postgres source's replication connection is dropping repeatedly — roughly every 5–10 minutes — and each drop stops the database's ingestion. The node restarts and resumes from the durable token, so nothing is lost; the cost is a restart loop and an `ERROR` that reads like data loss.

- **The connection drops in steady state, with no failover involved**

  #5878 covers a replication connection dying mid-stream, framed around failover as the way that happens. Nothing is being promoted here, so the trigger is broader than that framing.

- **None of the source's own handler branches shows up alongside it**

  `PostgresSource.onPartitionAssigned` logs `External source failed` on both of its rethrow paths and `Database connection failed when reading from copy (connection closed)` on its swallow path. The only record of the failure is `Watchers`' `ingestion stopping`. How the exception reached `Watchers` is not established — see Q1.

- **This is the read-side variant, with a verbatim trace**

  #5878's symptom is `...when writing to copy` and it carries no trace for the read side. This one is `...when reading from copy`, thrown from `readPending` inside `poll`.

- **Recovery works, which is why this is noise rather than data loss**

  Ingestion resumes from the persisted resume token on each restart. The operational cost is the restart loop and an ERROR that an operator has to learn to discount.

## Symptoms

Deployment running the `oversized-replica-msg-hotfix` branch at `899cfba8e` — `31dc90eed` plus that branch's commits. Recurring at roughly 5–10 minute intervals; that is an impression from the occurrences seen rather than a measured period, and nothing establishes that it is periodic at all.

```
1789400048838    2026-09-14T15:34:08.838Z    15:34:08 | ERROR xtdb.api.log.Watchers | ingestion stopping
1789400048838    2026-09-14T15:34:08.838Z    org.postgresql.util.PSQLException: Database connection failed when reading from copy
1789400048838    2026-09-14T15:34:08.838Z        at org.postgresql.core.v3.QueryExecutorImpl.readFromCopy(QueryExecutorImpl.java:1317)
1789400048838    2026-09-14T15:34:08.838Z        at org.postgresql.core.v3.CopyDualImpl.readFromCopy(CopyDualImpl.java:49)
1789400048838    2026-09-14T15:34:08.838Z        at org.postgresql.core.v3.replication.V3PGReplicationStream.receiveNextData(V3PGReplicationStream.java:165)
[...]

Issue #5878, opened 2026-08-10
postgres-source halts ingestion permanently when the replication connection dies mid-stream


Akeboshiwind:
## Summary

When the replication connection to the upstream Postgres dies mid-stream, `postgres-source` fails the database's ingestion permanently rather than reconnecting. A failover is the ordinary way to produce that dead connection, so the database halts at the moment failover slots were meant to make it survivable, and only a node restart brings it back.

## Context

Observed during a failover exercise on a database with a failover slot (#5828) — the slot now survives promotion, so a resuming source finds it on the new primary and streams on. That is the whole point of #5828, and it works: `PostgresSourceFailoverTest.an operator-enabled failover slot survives promotion` writes a row to the promoted standby and reads it back out of XT.

What that test doesn't cover is the transition. It closes the node while the old primary is still up and opens a fresh one against the standby's host and port, so the source is only ever *started* against the new primary — never *interrupted* on the old one. In a real failover the connection dies underneath a running stream, and that is the path this issue is about.

Before failover slots there was nothing to reconnect to, so halting was arguably the honest outcome. Now there is, and halting is what stands between a surviving slot and surviving ingestion.

## Symptoms

```
Ingestion error detected for database 'cdc': xtdb.api.log.IngestionStoppedException:
  Ingestion stopped (msgId: null): Database connection failed when writing to copy
```

The database is then dead to reads as well as writes — every query against it throws that same exception — and stays dead until the node is restarted.

## Analysis

`Database connection failed when writing to copy` is pgjdbc's, not ours: `QueryExecutorImpl.writeToCopy` and `flushCopy` wrap an `IOException` on the copy-both socket as a `PSQLException` with `PSQLState.CONNECTION_FAILURE` (`08006`). Two of our paths write to that stream, so either can raise it —

- **our LSN ack** — `PostgresSource.confirmFlushUpTo` → `ChangeStream.acknowledge` → `setFlushedLSN` / `setAppliedLSN` / `forceUpdateStatus` (`PgWireDriver.kt:332`);
- **pgjdbc's own keepalive**, which `readPending` issues inline on the status interval (`V3PGReplicationStream:133`).

From there the failure is terminal by construction:

- `PostgresSource.onPartitionAssigned`'s catch (`PostgresSource.kt:267`) swallows only the *read*-side variant, and only when the coroutine is already cancelled. A live-connection `PSQLException` is logged and rethrown.
[...]

PR #6077, opened 2026-09-16
fix(postgres-source): reopen the replication connection when it dies mid-stream

### LINKED ISSUE #6076: a reconnect that finds the slot still held should be retried, not fatal
**tl;dr**

- **A reconnect that lands before Postgres has reaped our own WAL sender fails the database permanently.**
  `55006` is not in pgjdbc's `08xxx` connection-error family, so the reconnect predicate added in #5878 treats it as fatal — on the one condition a reopen is most likely to meet.

- **Two retry budgets sit on the same call**, and the inner one classifies by matching a message string.
  Deleting it and giving `55006` to the outer predicate leaves a single budget, a single backoff curve, and no message matching anywhere in the file.

- **Nothing observed in the field.**
  A code-review pass and an `allium:weed` pass over `dev/doc/pgsrc.allium` found it independently; the weed rated it a code bug rather than a spec gap.

## Context

#5878 made a replication connection that dies mid-stream reopen rather than fail the database. The predicate is `PSQLState.isConnectionError(e.sqlState)` — pgjdbc's own classification, chosen over message matching because #5878 records that Debezium moved the same way and calls the message form the more brittle of the two.

Reopening races Postgres' release of the slot. Our previous WAL sender is still registered until the server reaps it at `wal_sender_timeout`, 60s by default, so the reopen we attempt 1.5s after a disconnect ordinarily arrives while the slot is still held.

## Analysis

- **`isConnectionError` matches five states, all `08xxx`** — `08001`, `08003`, `08004`, `08006`, `08007` (decompiled from 42.7.10). A slot still held gives `55006 object_in_use`, message `replication slot "…" is active for PID n`, which is not among them.

- **`PgWireDriver.startReplicationStream` already retries that condition, and only that condition.**
  Seven attempts from 1s, doubling with jitter, gated on `SLOT_ACTIVE_PATTERN` — `.*replication slot .* is active.*`. Its delays total roughly 63s against a 60s default `wal_sender_timeout`, which is marginal by construction.

- **Where `wal_sender_timeout` is raised or disabled, the inner budget never clears.**
  `0` is a common setting where a slow consumer must not be killed. The inner loop then exhausts and rethrows the `55006`, which the outer predicate does not match, so the database fails on a condition that would have resolved on its own.

- **The two budgets compound.**
  `openStream` sits inside the reconnect loop, so the worst case is seven outer attempts each spending an inner budget — roughly a quarter of an hour of a source reporting `connection_state = 0` with no error raised.

- **assumption:** that a reconnect racing `wal_sender_timeout` is the common case rather than the rare one. It follows from the backoff being seconds and the reap being a minute, but no field occurrence has been observed.

## Implementation

- **A reconnect survives the slot still being held by our own dead WAL sender**

  - [ ] `55006` joins the retriable set in `PostgresSource`'s reconnect catch, alongside `PSQLState.isConnectionError`.

  - [ ] `PgWireDriver.startReplicationStream`'s retry loop is deleted, so `openStream` either succeeds or throws.
    *fact:* `openStream` has one caller, `streamChanges`, and that call is already inside the reconnect loop — so everything the inner retry covers, the outer one can.
    *fact:* the outer budget's first six attempts total about 85s, against the inner's 63s, so a leadership handover is better served afterwards than before.

  - [ ] `SLOT_ACTIVE_PATTERN` goes with it, leaving no classification by message text in the module.

  - **check:** the two budgets reset differently, and the replacement is not strictly more generous. The inner budget starts fresh on every `openStream`; the outer resets only when a stream delivered a transaction or stayed up past the longest backoff. A source that has already spent attempts on unrelated disconnects has fewer left for a handover.

## Out of scope

- **The `57Pxx` states** — `admin_shutdown`, `crash_shutdown`, `cannot_connect_now`. A code review argued a terminated WAL sender sends `57P01`; the integration tests added in #5878 produce `08006` for a killed WAL sender, a restarted server *and* a promotion, so there is no evidence to act on. A graceful `pg_ctl stop -m smart` remains untested.

- **Inverting the predicate to retry unless known-fatal.** Considered: the cost asymmetry favours it, since a wrongly-retried fatal error still reaches `Failed` after the bound while a wrongly-fatal transient one poisons the database permanently. Left alone because a missing publication or a permissions error would then burn the whole budget before reporting, and slow feedback on a misconfiguration is a real regression.

## Open questions

- [ ] Does #6073's steady-state drop leave the server-side WAL sender alive? Its trace is `SocketException: Socket is closed` raised from `setSoTimeout`, which reads as our own side of the socket going away. If the server has not noticed, every reconnect there meets this condition, and this is what stands between #5878 and that deployment.

Akeboshiwind:
Resolves #6076. Part of #5878.

**tl;dr**

- **A replication connection that dies no longer kills the database.** The source reopens from the furthest LSN it submitted, backing off, instead of failing permanently and needing a node restart.

- **This is live today, not hypothetical.** #6073 is a deployment losing its connection every 5–10 minutes in steady state, with no failover involved — a restart loop and an `ERROR` that reads like data loss. That becomes a reconnect that nobody sees.

- **Every reopen is bounded and every failure is retried.** Seven attempts with exponential backoff; the count restarts when a connection did real work, so a source that reconnects cleanly every few hours never accumulates its way to the bound.

- **The groundwork is already on `main`.** Four behaviour-preserving tidies landed separately via `pges-tidy`, so this diff reads against a source where the resume position already belongs to the assignment rather than to the stream.

## Context

#5878 was found during a failover exercise: with a PG17 failover slot the upstream survives promotion, but the source doesn't. The connection dies under the poll loop, `LeaderLogProcessor` turns the exception into `watchers.notifyError`, and `Watchers.Failed` is absorbing — so the database is dead to reads as well as writes until the node restarts.

#6073 then landed from a live deployment with the same failure and no failover anywhere near it. The trigger is broader than the framing #5878 was written around, which is worth knowing when reading either.

## Implementation

- **Reopen from `heldLsn`, the furthest LSN submitted to the indexer.**
  It is the only one of the three watermarks that is correct to resume from: the applied watermark can be a block behind it, and the durable one further still, so either would re-import transactions XTDB already holds and write a second valid-time version of each.

- **Settle the in-flight transactions before reopening.**
  The deque is rebuilt per attempt, so a reopen would otherwise discard the handles of submitted-but-unapplied transactions and any import failure they carried — and streaming would continue past a transaction the target never imported. Draining first is also the cheap check before the expensive work: an ingest failure is fatal and costs an await to find, where a reconnect costs a backoff and a slot on the upstream.

- **Retry every `PSQLException`, rather than a chosen family.**
  The predicate started as pgjdbc's `08xxx` class, which excluded `55006` — the slot still held by our own WAL sender, which is what a reopen most often meets. Naming a fatal set instead would have been guesswork: the integration tests produce `08006` for a killed WAL sender, a restarted server *and* a promotion, and nothing observed argues for any other classification. The costs are not symmetric either — an error wrongly retried still reaches `Failed` once the attempts are spent, where one wrongly called fatal poisons the database until the node restarts.

- **One retry budget, not two.**
  `startReplicationStream` had a loop of its own for the held-slot case, classifying by matching the message text. `openStream` has a single caller and that call now sits inside the reconnect loop, so the outer budget covers it — and covers it better, spanning ~168s against the inner's ~63s versus a 60s default `wal_sender_timeout`.

- **A stand-down is not a failure.**
  Cancellation — demotion, detach, node close — now leaves through a throw rather than the loop condition, and `ensureActive()` at the top of the catch makes a cancellation racing a connection failure end as a cancellation. Without that, a routine demotion could mark a database `Failed`.

## Changes

Four tidies landed on `main` separately, via `pges-tidy`, and this diff reads against them. Worth knowing which, because the code here treats their results as given:

- `94f7f28df` — the in-flight deque awaits *apply*, not durability. `durableLsn()` fifteen lines away means object storage.
- `44ca638d4` — names the teardown artefact, and writes down the window where a genuine mid-stream death is swallowed.
- `f53b9a96e` — resolves the resume position, then streams. One call site instead of two.
- `bc940bed9` — the resume position belongs to the assignment, not the stream; the re-delivery skip moves to `heldLsn`, which is the form that survives a reopen.

What is in this PR, smallest first:

1. `76ccb0849` — a stand-down stops being logged as `External source failed`.
2. `6b94461f8` — a stood-down stream throws rather than falling out of its loop. Without this the retry below reopens the connection it is being torn down from, in a tight loop.
3. `9b3e6925a` — the reconnect itself.
4. `cbfa8630a` — drain to empty before reopening.
5. `3ba76b020` — bound the attempts, and restart the count when a connection worked.
6. `d575b9c9f` — the integration tests, and `PgProxy`.
7. `006f813e5` — the stand-down tests own a scope rather than a job.
8. `daa245bfe` — every replication failure is retried, not a chosen family.
9. `399ab3367` — one retry budget for a reopened stream.

Read 8 before 9: deleting the inner retry while the predicate still filtered would leave `55006` reaching a catch that rethrows it, which is a database failing on exactly the condition the change exists to survive.

## Testing

Every behaviour commit was verified red first, and the failure text is quoted in its body.

- **Unit** — resume position after a disconnect, the drain, the bound and both halves of its reset, three stand-down interleavings, and a refused open. The stand-down cases force their interleaving rather than racing for it: `runTest` is single-threaded, the cancel and the failure are adjacent non-suspending statements, and cancellation reaches children inside `cancel()`.
- **Integration, against a real Postgres** — a killed WAL sender, a restarted server, a promotion under a running stream, and a reopen meeting a slot another connection holds.
- **`PgProxy`** is what makes two of those assertable. `PostgresRemote` fixes a host and port, so a source whose upstream moves is pointed at an address that no longer serves; a movable endpoint in front is what an operator deploys, and what lets the failover case assert that ingestion *resumes* rather than only that it has not given up.

Worth knowing about the evidence rather than trusting the green:

- **`57P01` does not appear.** A code review argued a terminated WAL sender sends it, so the `08xxx` predicate would treat it as fatal. All three scenarios produce `08006` — over an `EOFException` on the read side, a `SocketException` on the write side. A graceful `pg_ctl stop -m smart` is the one shutdown mode still untested.
- **No SQLSTATE for the held slot was ever observed.** It is conventionally `55006`; the logs only ever show the message. All the evidence supports is that it is not in the `08xxx` family.
- **The restart case exercises one retry, not the ladder.** Restarting a near-empty Alpine Postgres is sub-second, so the unavailability window barely exists.

## Out of scope

- **`connection_state` flips 1 → 0 → 1 through a retry** and reads as a dead source. `Assigned.streaming` is a boolean with no way to say "reconnecting". Deferred.
- **`dev/doc/pgsrc.allium` does not describe reopening.** A weed pass found the gap, along with two divergences older than this branch: `ConfirmedWithinReceived` breaks across a reopen because `received_lsn` is pgjdbc's `lastReceiveLSN` and resets to the LSN a stream started at, and `ConfirmedWithinDurableWhenBehind` is stronger than the code independently of reconnection. Both want restating at the confirmation instant rather than over all states. Following separately.
- **A source that exhausts its attempts still fails the database permanently.** That is the same terminal state as before this branch, reached after ~3 minutes instead of immediately.


Akeboshiwind:
Closed in favour of a three-PR stack, so that each piece is reviewable on its own rather than as one branch of nine commits.

- #6080 — a stand-down leaves the poll loop through a throw. Base `main`.
- #6081 — the reconnect, the drain, the bound, and the real-upstream tests. Base `pges-stand-down-exit`.
- #6082 — one retry budget, and every replication failure retried. Base `pges-replication-connection-retry`. Resolves #6076.

Read them in that order; each is based on the one before.

The same work, rebased from nine commits into eight: `61c8acf04`, the scope-not-job test tidy, folded into the two commits that introduced the form it was fixing. The tip tree is byte-identical to this branch's.

Issue #4833, opened 2025-09-29
`COPY FROM STDIN` with Elixir


refset:
[Postgrex](https://github.com/elixir-ecto/postgrex) (Elixir client) attempts to stream data over the Extended Query protocol in a way that XTDB doesn't currently support.

The following currently fails with `{"category":"cognitect.anomalies\/incorrect","code":"xtdb.pgwire\/copy-not-in-progress","message":"COPY IN not in progress, cannot write data"}`

```elixir
  def test_copy_from_stdin do
    IO.puts("\n=== Testing COPY FROM STDIN with copy_data option ===")

    {:ok, pid} = Postgrex.start_link(hostname: "xtdb", port: 5432, database: "xtdb")

    try do
      # Setup test table
      IO.puts("Setting up test table...")
      Postgrex.query!(pid, "INSERT INTO copy_test RECORDS {_id: 'init', name: 'Init'}", [])

      # Test data in transit-json format with correct xt/id keyword
      test_data1 = "[\"^ \",\"~:xt/id\",\"copy_1\",\"~:name\",\"First Copy\",\"~:active\",true]"
      test_data2 = "[\"^ \",\"~:xt/id\",\"copy_2\",\"~:name\",\"Second Copy\",\"~:active\",false]"

      IO.puts("Attempting COPY FROM STDIN with streaming...")
      IO.puts("Data to copy:")
      IO.puts("  #{test_data1}")
      IO.puts("  #{test_data2}")

      # Use streaming approach with prepare and copy_data: true
      copy_sql = "COPY copy_test FROM STDIN WITH (FORMAT 'transit-json')"

      case Postgrex.prepare(pid, "", copy_sql, [copy_data: true]) do
        {:ok, query} ->
          # Use transaction with streaming
          result = Postgrex.transaction(pid, fn(conn) ->
            try do
              stream = Postgrex.stream(conn, query, [])

              IO.puts("Streaming first batch of data...")
              stream_result = Enum.into([test_data1 <> "\n", test_data2 <> "\n"], stream)

              if stream_result == stream do
                IO.puts("✓ Data streamed successfully")
              else
                IO.puts("⚠ Stream result: #{inspect(stream_result)}")
              end
   ...
```

TRACE logs:
```
xtdb  | 07:20:50.939 [pgwire-connection--pool-1-thread-1] TRACE xtdb.pgwire.io - Writing server message {:char8 1, :name :msg-parse-complete}
[...]

PR #6079, opened 2026-09-17
pgwire no longer answers a CopyFail that arrives outside copy mode

Akeboshiwind:
Any Node application streaming through `COPY … FROM STDIN` desynchronises its connection against XTDB, and the only alternative from Node — INSERT — costs one to two orders of magnitude, so there is no workaround worth taking.

## tl;dr

A query issued straight after a `COPY … FROM STDIN` can come back with the previous query's result, or with nothing to attach to.

- **It reaches any client that sends a `CopyFail` when no COPY is open, and postgres.js does that on every COPY**
  Its copy-in writable is built with `autoDestroy: true`, and Node calls `_destroy(null, …)` once the stream finishes cleanly. PostgreSQL drops the message in silence; we answered it.

- **The rows are safe — it is the client's view that drifts**
  A COPY is applied at `CopyDone`, before the reply is written, so every COPY that got as far as a stream committed. What breaks is which reply the client attributes to which query.

- **It self-corrects when the connection goes idle, which is why it reads as intermittent**
  An application that pauses between COPYs never sees it; one going back-to-back sees it most times.

## Symptoms

- **`.writable()` resolves to an empty result array instead of a `Writable`**
  ```js
  const w = await sql`COPY docs FROM STDIN WITH (FORMAT 'transit-json')`.writable()
  // w is []
  ```
  Six sequential COPYs on one reserved connection against a playground node: attempts 1, 3 and 5 returned a stream and committed their rows, attempts 2, 4 and 6 returned `[]` and wrote nothing.

- **Or the Node process dies on an unhandled `'error'` event**
  ```
  TypeError: Cannot read properties of null (reading 'resolve')
      at CopyInResponse (postgres/src/connection.js:874)
      at handle (postgres/src/connection.js:489)
      at Socket.data (postgres/src/connection.js:324)
  ```
  It is emitted on the writable rather than raised in the `await`, so a `try`/`catch` around the COPY does not see it.

- **Conditions: two or more COPYs on one connection with no pause between them**
  A `setTimeout` of a few tens of milliseconds after each COPY hides it completely.

- **Repro: `leaves the connection usable for a second COPY` in `lang/js/test/copy_test.js`**
  Two COPYs on one reserved connection, asserting the second gets a stream. It fails on the second COPY without this change.

## Root cause

`handle-msg* :msg-copy-fail` guarded its cleanup on `(:copy @conn-state)` but called `cmd-send-ready` outside that `when-let`, so a `CopyFail` with no COPY in progress produced a bare `ReadyForQuery`.

PostgreSQL's message loop breaks out of `CopyData`, `CopyDone` and `CopyFail` with no reply whenever copy mode is inactive, exempting them from the `invalid frontend message type` path immediately below ([`postgres.c`](https://github.com/postgres/postgres/blob/e371cb7b89dddd9c1d3d89f1e5c64dabc05474aa/src/backend/tcop/postgres.c#L4972-L4981)). The [protocol docs](https://www.postgresql.org/docs/16/protocol-flow.html#PROTOCOL-COPY) say the same of the error case: "any subsequent CopyData, CopyDone, or CopyFail messages issued by the frontend will simply be dropped."

Captured with postgres.js instrumented to log its active query. `→` is the client writing, `←` the server, and the unindented lines are what postgres.js did with what it read:

```
COPY #1
  → Q "COPY …"                              query := COPY#1
  ← CopyInResponse                          COPY#1.resolve(stream)   — .writable() returns
  → CopyData ×2, CopyDone
  ← CommandComplete "COPY 2"                stream finishes, pipeline resolves
  ← ReadyForQuery                           COPY#1 already settled; query := null
  → CopyFail "null"                         ← Node's autoDestroy, after a clean finish
  ← ReadyForQuery                           ★ SURPLUS — PostgreSQL sends nothing here

COPY #2
  → Q "COPY …"                              query := COPY#2
  ← CopyInResponse                          COPY#2.resolve(stream)   — looks healthy
  → CopyData
  ← ReadyForQuery  (the surplus)            ★ COPY#2 retired mid-stream; query := null
  → CopyDone
  ← CommandComplete "COPY 2"                query is null → TypeError, swallowed

COPY #3
  → Q "COPY …"                              query := COPY#3
  ← ReadyForQuery  (COPY#2's own)           ★ COPY#3.resolve([])  — .writable() returns []
  ← CopyInResponse                          query is null → TypeError → process exits
```

Where the surplus lands decides which of the two symptoms appears. Landing while the connection is idle it is discarded and the books balance again, which is the self-correction above.

## What changes

`cmd-send-ready` moves inside the `when-let`, so a `CopyFail` is answered only where a COPY was actually in progress.

The belief that no longer holds: pgwire replied to every `CopyFail`. A client that sends one outside copy mode now gets nothing back at all, which is what it would get from PostgreSQL.

## Usage / migration

Nothing to do on the XTDB side. For clients pinned to an older node, the stray message can be suppressed at source — this is a `_destroy` that writes `CopyFail` even when `error` is null:

```js
const writable = await sql`COPY … FROM STDIN`.writable()
const origDestroy = writable._destroy.bind(writable)
writable._destroy = (err, cb) => (err ? origDestroy(err, cb) : cb(null))
await pipeline(readable, writable)
```

## Consequences

- **A client that waits for a `ReadyForQuery` after a stray `CopyFail` will now hang**
  It was already broken against PostgreSQL, which has never sent one, so the surface this removes could only ever have been relied on by a client that XTDB alone kept working.

- **`lang/js/test/copy_test.js` is sleeping 50ms after its COPY to dodge this**
  The comment there cites [porsager/postgres#1016](https://github.com/porsager/postgres/pull/1016), which is a different upstream bug about a paused socket after COPY-out. Remove the sleep and that test hangs without this change and passes with it. Left in place here so the fix and the cleanup stay separable.

## Open questions

- **Q1: `:msg-copy-data` and `:msg-copy-done` still throw `::copy-not-in-progress` outside copy mode**
  Same protocol rule, different risk: dropping those two discards a misbehaving client's payload rather than one stray control message. #4833 is where that divergence surfaces, against Postgrex.


jarohen (changes requested):
As discussed, let's get some postgres protocol tests for this to demonstrate the flow.

Reading [the spec](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-COPY) (indentation mine):

* In the event of a backend-detected error during copy-in mode (including receipt of a CopyFail message), the backend will issue an ErrorResponse message. 
  * If the COPY command was issued via an extended-query message, the backend will now discard frontend messages *until a Sync message is received*, then it will issue ReadyForQuery and return to normal processing. 
  * If the COPY command was issued in a simple Query message, the rest of that message is discarded and ReadyForQuery is issued. 
* In either case, any subsequent CopyData, CopyDone, or CopyFail messages issued by the frontend will simply be dropped.

I've always struggled to get my head around sync messages for some reason - it'd be good to get an example that shows we're handling that correctly, if applicable.

PR #6081, opened 2026-09-17
fix(postgres-source): reopen the replication connection when it dies mid-stream

### LINKED ISSUE #6073: Postgres source's replication connection dropping every 5-10 minutes, stopping ingestion

Akeboshiwind:
Part of #5878. Stacked on #6080 — review that first, and read this diff against it.

**tl;dr**

- **A replication connection that dies no longer kills the database.** The source reopens from the furthest LSN it submitted, backing off, instead of failing permanently and needing a node restart.

- **This is live today, not hypothetical.** #6073 is a deployment losing its connection every 5–10 minutes in steady state, with no failover involved — a restart loop and an `ERROR` that reads like data loss. That becomes a reconnect nobody sees.

- **Every reopen is bounded.** Seven attempts with exponential backoff; the count restarts when a connection did real work, so a source that reconnects cleanly every few hours never accumulates its way to the bound. A source that never recovers still ends where it ended before this branch — `Failed` — about three minutes later instead of immediately.

- **It has been against a real Postgres.** A killed WAL sender, a restarted server and a promotion under a running stream all produce `08006`, and all recover on the first retry.

## Context

#5878 was found during a failover exercise: with a PG17 failover slot the upstream survives promotion, but the source does not. The connection dies under the poll loop, `LeaderLogProcessor` turns the exception into `watchers.notifyError`, and `Watchers.Failed` is absorbing — so the database is dead to reads as well as writes until the node restarts.

#6073 then landed from a live deployment with the same failure and no failover anywhere near it. The trigger is broader than the framing #5878 was written around, which is worth knowing when reading either.

Four behaviour-preserving tidies landed on `main` separately via `pges-tidy`, and this diff treats their results as given:

- `94f7f28df` — the in-flight deque awaits *apply*, not durability. `durableLsn()` fifteen lines away means object storage.
- `44ca638d4` — names the teardown artefact, and writes down the window where a genuine mid-stream death is swallowed.
- `f53b9a96e` — resolves the resume position, then streams. One call site instead of two.
- `bc940bed9` — the resume position belongs to the assignment, not the stream; the re-delivery skip moves to `heldLsn`, which is the form that survives a reopen.

## Implementation

- **Reopen from `heldLsn`, the furthest LSN submitted to the indexer.**
  It is the only one of the three watermarks correct to resume from: the applied watermark can be a whole block behind it, and the durable one further still, so either would re-import transactions XTDB already holds and write a second valid-time version of each.

- **Settle the in-flight transactions before reopening.**
  The deque is rebuilt per attempt, so a reopen would otherwise discard the handles of submitted-but-unapplied transactions and any import failure they carried, and stream on past a transaction the target never imported. Draining first is also the cheap check before the expensive work: an ingest failure is fatal and costs an await to find, where a reconnect costs a backoff and a slot on the upstream.

- **Seven attempts with exponential backoff and jitter** — `PgWireDriver.startReplicationStream`'s shape and its constants. Exhaustion rethrows the connection failure.

- **The count is a run of *fruitless* attempts, and the reset is what decides whether the bound is a bound or a slow leak.**
  Counting every disconnect would fail a healthy low-write upstream that lost its connection eight times ever, weeks apart, reconnecting cleanly each time — a flap is a property of time, not of throughput. So a connection that delivered a transaction restarts the count, and so does one that merely stayed up longer than we would ever wait before retrying.

- **That threshold is derived, not chosen: it is the longest backoff.**
  If a connection lasted longer than we would ever wait to replace it, it was not flapping — and deriving it means the ladder can be retuned without a second constant drifting out of step. Timing starts when the stream opens, not when the attempt does, so an attempt that spent a minute inside `openStream` and never got a connection cannot reset the count on the strength of its own failure.

- **A cancellation racing a connection failure ends as a cancellation.**
  The check is `ensureActive()` at the top of the catch rather than a term in the classifying predicate: a cancellation can land between the exception being raised and the catch classifying it, so a predicate reading `isActive` answers about a different moment than the one that failed. Without it, a stand-down leaves as a `PSQLException` and a routine demotion can mark a database `Failed`.

## Changes

1. `930e3a7e0` — the reconnect itself. `git show -w` reads it: wrapping the attempt in the loop re-indents it, and nothing inside changed.
2. `4b3a8870b` — drain to empty before reopening.
3. `ea90e690b` — bound the attempts, and restart the count when a connection worked.
4. `3fe5e1500` — the integration tests, and `PgProxy`.

## Testing

Every behaviour commit was verified red first, and the failure text is quoted in its body. At this tip: unit green, integration 47 tests / 0 failures / 1 pre-existing skip.

- **Unit** — resume position after a disconnect, the drain, the bound and both halves of its reset, two stand-down interleavings, and a refused open. The stand-down cases force their interleaving rather than racing for it: `runTest` is single-threaded, the cancel and the failure are adjacent non-suspending statements, and cancellation reaches children inside `cancel()`.
- **Integration, against a real Postgres** — a killed WAL sender, a restarted server, and a promotion under a running stream.
- **`PgProxy` is what makes two of those assertable.** `PostgresRemote` fixes a host and a port, so a source whose upstream moves is pointed at an address that no longer serves. A movable endpoint in front is what an operator deploys — a VIP, a Patroni leader service, a pgbouncer — and what lets the failover case assert that ingestion *resumes*, rather than only that it has not given up yet.

Worth knowing about the evidence rather than trusting the green:

- **`57P01` does not appear.** A code review argued that a terminated WAL sender sends it, so the `08xxx` predicate would call it fatal. All three scenarios produce `08006` — over an `EOFException` on the read side and a `SocketException` on the write side. A graceful `pg_ctl stop -m smart` is the one shutdown mode still untested, and is not reachable while postgres is the container's own process.
- **The restart case exercises one retry, not the ladder.** Restarting a near-empty Alpine Postgres is sub-second, so the unavailability window barely exists. A sustained outage wants the container paused.
- **`55006`, the slot still held by a WAL sender the server has not reaped, has no case here.** Inducing it needs a second consumer taking the slot inside the window between the loss and the first retry, which is a race rather than a test.

## Out of scope

- **The predicate is still pgjdbc's `08xxx` family, which excludes `55006`** — the condition a reopen most often meets, since our own WAL sender is registered until `wal_sender_timeout`. #6076 covers it, in the PR stacked on this one.
- **`connection_state` flips 1 → 0 → 1 through a retry** and reads as a dead source. `Assigned.streaming` is a boolean with no way to say "reconnecting". Deferred.
- **`dev/doc/pgsrc.allium` does not describe reopening.** A weed pass found the gap, along with two divergences older than this branch: `ConfirmedWithinReceived` breaks across a reopen because `received_lsn` is pgjdbc's `lastReceiveLSN` and resets to the LSN a stream started at, and `ConfirmedWithinDurableWhenBehind` is stronger than the code independently of reconnection. Both want restating at the confirmation instant rather than over all states. Following separately.
- **A source that exhausts its attempts still fails the database permanently.** The same terminal state as before this branch, reached after ~3 minutes instead of immediately.


modules/postgres-source/src/test/kotlin/xtdb/postgres/PostgresSourceConnectionLossTest.kt
```diff
@@ -0,0 +1,173 @@
...
+        DriverManager.getConnection("jdbc:postgresql://$host:$port/testdb", "testuser", "testpass")
+
+    private fun cdcError(node: Xtdb) = (node as XtdbInternal).dbCatalog["cdc"]?.ingestionError
+
+    @Test
+    fun `ingestion survives the walsender being killed under a live stream`() = runTest(timeout = 300.seconds) {
+        val slot = unique("slot")
+        val pub = unique("pub")
+        val table = unique("widgets")
+        val dirs = List(4) { Files.createTempDirectory("conn-loss") }
```
jarohen, line 46: nit: you can create one root `@TempDir` param to the function, then you don't have to worry about tearing them down at the end

https://docs.junit.org/5.4.1/api/org/junit/jupiter/api/io/TempDir.html

jarohen (commented):
Nice one 👏 

Couple of things to chat about, perhaps:

1. Test timing - does this/could this significantly increase the integration test time? Conscious there's a couple of tests in there with 10 min timeouts, which is about as long as the whole suite takes currently - the run itself only has a 15min limit in GH actions, IIRC.
2. Is it worth the reconnect logic hiding within the driver, and keeping the source limited to processing the transactions? I wonder whether that would simplify the tests, as then you don't need to stand up full nodes, you just have to assert that the driver eventually returns you more transaction ops.

Issue #5693, opened 2026-06-08
Cross-transaction batching: reduce overhead for high-volume small transactions


jarohen:
## Summary

Reduce the per-transaction overhead XT incurs when a workload is a high volume of small transactions, without changing transaction semantics.

## Context / Motivation

XT's recommended transaction sizing assumes relatively large batches — but a growing class of workloads feeds XT from sources whose natural message size is far smaller.

Kafka is the motivating case: individual Kafka messages are typically much smaller than the batch size we recommend for an XT transaction. A pipeline sourced from Kafka therefore produces *many, many* small transactions, where XT's fixed per-transaction overhead dominates the useful work.

## Invariants / Constraints

**Atomicity boundaries must not change.** Each upstream transaction stays its own XT transaction; we are not merging transactions. The win comes from amortising the per-transaction machinery across the stream — batching the *work*, not the *atomic units*. A reader should be able to assume transaction semantics are identical before and after this change.

## Out of scope

- **Merging multiple upstream transactions into one XT transaction.** That changes atomicity and is explicitly not what this is — the previous attempt at that framing (#5203) was closed not-planned.
- **The external-source framework itself (#5445).** That's a separate, complementary track. This batching work benefits *any* high-volume small-transaction workload, not just external sources, so it stands alone.

## Implementation

Direction TBD — needs a research/spec pass to find where the fixed per-transaction cost actually lives and which parts can be amortised across a stream of small transactions while preserving per-transaction atomicity.

jarohen:
## Observed in a deployment

- **An external Kafka source hit this ceiling hard enough to stop a database.**
  A single-partition source read several times slower than its upstream topic was produced to, fell behind that topic's retention window, and stopped once its position had been truncated away. Recovering from that state is #5829.

- **An external source is the purest instance of the shape this issue describes.**
  `DocsIndexer.indexRecords` submits one transaction per record. `KafkaConnectSource` already amortises conversion and transforms across a poll batch, so the transaction boundary is the only unamortised cost left on that path.

- **For an external source the resume token pins the atomic unit as well.**
  Each transaction carries a token naming its record's offset, and a restarted consumer re-seeks from it. So "batch the work, not the atomic units" is a restart-granularity constraint here, not only a semantic preference.

- **The cheap wins on this path are already spent.**
  #5781 and #5738 both landed, and the deficit above is what remains after them — so the research pass this issue calls for is the next step rather than a further optimisation.

- **`check:` no profile exists.**
[...]

PR #6088, opened 2026-09-18
perf(log): a leader's replica-log appends no longer cost a round-trip each

jarohen:
Part of #5693.

A leader appending to the replica log waited for each record to be durable before offering the next, so a database's ingest was capped at one broker round-trip per transaction however small the transactions were.

- **Where the cap sat**
`ReplicaLogAppender.run()` is a `select` loop, and its body called `appendToReplica` and awaited it. The resolver is decoupled from that by an unbounded channel, so the wait never stalled resolution — but nothing behind the appender could overtake it, and a stream of single-row transactions spends nearly all of its time there.

- **Why the wait was load-bearing, not just slow**
It was also what turned a lost record into a failed term. An idempotent, non-transactional Kafka producer does not stop when a record is lost: `Sender.canRetry` retries the records queued behind the gap and the epoch bump re-sequences them, so they land, and only the failed record's own callback is told. Removing the await meant supplying that guarantee ourselves.

- **What it unblocks**
`linger.ms`, set to `0` in #5452 on the grounds that a producer handed one record at a time has nothing to batch it with. That premise no longer holds.

## What changes

**An append no longer means the record is durable.** `Log.appendMessage` carried both obligations in one suspend — this record's place in the log's order is fixed, *and* it is durable — so a caller that only needed the first had no way to ask for it.

- **`enqueueMessage` is now what a log implements**
It returns once the record's order is fixed against every other caller of that log, handing back a `Deferred` that completes on durability. `appendMessage` remains, defaulted to `enqueueMessage(…).await()`, so every existing caller is unchanged. Kafka fits this exactly: `producer.send` validates the size and hands the record to the per-partition accumulator before returning, while the offset is only knowable in the callback.

- **A log that loses a record accepts no further records**
This is the guarantee we believed Kafka gave us and do not inherit. `KafkaLog` latches on the first failed send callback and raises `LogFailedException` from every later append; `LocalLog` closes its append channel and fails the handles outstanding. Terminal by design — a log with a gap in it cannot be repaired — which is what distinguishes it from a latching wait condition.

- **The appender drops the handle**
Its call site is unchanged; it simply no longer waits. A loss reaches it as a throw from a *later* enqueue, which ends the term through the existing boundary in `LogProcessor`. Detection is therefore one append late, bounded by `ElectionDriver`'s assert timeout, since a quiescent leader still appends a `NoOp`.

- **`ReadOnlyLog` was writing through**
It refused `appendMessage` and `appendMessageBlocking` by name, and `Log<M> by delegate` generates a forwarder for every member left alone — so the new `enqueueMessage` reached the delegate the class exists to protect. Fixed here with a regression test, rather than split out, because the member it forwards is the one this branch adds.

- **`LogProcessor.LogsDriver` has one append member for the same reason**
`enqueueToReplica`, with `appendToReplica` as a top-level extension. A `by`-delegating wrapper can only forward past a member it didn't override, and `OffloadingLogsDriver` is exactly such a wrapper.

## Usage / migration

**An out-of-tree `Log` implementation has to move its append.** `enqueueMessage` is the abstract member; implementing only `appendMessage` no longer satisfies the interface. A log whose order and durability genuinely coincide returns an already-completed handle — `InMemoryLog` does, because its record is in the log before the mutex is released.

**Operators who set `linger.ms` keep their value.** `producerConfig()` layers user properties over XTDB's defaults, so only the unset case changes, from `0` to Kafka's default of 5ms.

## Consequences

- **A lost replica-log record now fails the database instead of passing in silence**
Previously the gap surfaced, if at all, as a hang: `TxResolver.removeHead` would not match on record *k+1* after *k* vanished, so the leader applied *k+1* from the message bytes while *k*'s `ResolvedTx` sat at the head of the queue with its `pending` never completed. It now reaches `watchers.notifyError` and ends the term.

- **A record declined outright deliberately does not latch the log**
`RecordTooLargeException` comes out of `ensureValidRecordSize` before the record reaches the accumulator, so nothing was sent and no order was fixed. Latching there would permanently break `OffloadingLogsDriver`, whose job is to catch that case, offload to object storage and re-append a reference.

- **A lone append can now wait up to 5ms for a linger window that never fills**
The cost of the default, paid by a database quiet enough to have nothing to batch with.

- **The simulation now exercises the pipelined state; it previously could not**
`SimLog` acknowledged a record the instant it was enqueued, so every seed explored only the behaviour this change replaces. It now fixes order on enqueue and makes the record durable — and readable — on a coroutine the seed schedules. Measured over a `LogProcessorSimTest` run, in-flight depth reaches 11–24 records per iteration, against a structural maximum of 1 before.

- **Two simulation assertions had gone quiet and are live again**
`NodeReplicaLog` recorded which node wrote each record by overriding `appendMessage`, which stopped being the member the leader calls. Its set stayed empty, so `assertOneWriterPerTerm` — the ExactlyOneConfirmedLeaderPerDatabase property — matched no record and asserted nothing, and the read-only-node check was vacuously true.

## Out of scope

Split on: things that amortise per-transaction cost but need a format change, an upgrade story, or a separate measurement.

- **Batching resolved transactions into one replica-log message**
The remaining per-transaction cost is dominated by serialising a schema flatbuffer per message. Amortising it needs a new message type, and with it either a two-release upgrade or a stop-start.

- **The per-transaction row copy into the live index**
`liveindex.commitTx` is where a one-row transaction spends most of its time, and it is untouched here.

- **A proactive size check to close the broker-side residual**
A `message.max.bytes` rejection arrives by callback, so it latches like any other loss and bounces the term rather than offloading. Reachable only where an operator sets `message.max.bytes` below `max.request.size`; both default to ~1MB and XTDB overrides neither. Closing it needs a maximum-message-size notion on `Log` that Kafka has and the local logs don't.

## Alternative approaches

Split on: where the "nothing lands after a loss" guarantee could live.

- **Rely on the producer to poison itself** — reasoned against
Needs no new machinery, and is unsound for the reason above: `Sender.canRetry` retries around the gap deliberately, and `bumpIdempotentProducerEpoch` re-sequences the records behind it from zero so they land.

- **Latch on `ReplicaLogAppender` instead** — reasoned against, twice
Drafted first as a sender/awaiter coroutine pair, then as a retire arm on the existing `select`. Both put a statement about a log's contents one layer above the object that can make it, and both made an active object carry knowledge of Kafka's recovery behaviour.

- **A bounded in-flight deque of our own** — reasoned against
Planned, then dropped once the invariant became a flag on the log: it bounded nothing that `buffer.memory` and `max.in.flight.requests.per.connection` don't already bound, and its size would have been a knob with no basis for tuning it.

- **Drain outstanding handles at the block boundary** — reasoned against
Resolution already pauses there, so it needs no per-message machinery. `rowsPerBlock` defaults to 102,400, which is the window of unsurfaced loss it would accept.

- **`invokeOnCompletion` into a failure field** — reasoned against
Needs neither a coroutine nor a select clause, but the callback runs on the producer's I/O thread, so it needs its own safe publication and it loses the ordering between a failure and the successes queued behind it.

## Open questions

- **Q1: nobody has measured end-to-end ingest with the cap removed.**
The append path no longer binds, but the read-back and apply loop in `runTerm` is a separate serialised `select` and was not traced. If that is the narrower pipe, this change is still a prerequisite but the throughput win arrives later than the framing above implies. Settled by a single-partition Kafka ingest run against a small-transaction workload.


jarohen:
note for Monday - we probably want that hard fail behaviour for the replica log but not necessary for the source log, because in the latter case we can pass the error back to the user and have them decide what to do

jarohen:
Closing this: the guarantee it claims isn't one it delivers.

**The latch detects a lost record; it doesn't stop the ones behind it landing.**
The failing record's own callback sets `lostRecord`, and the next `enqueueMessage` reads it — but by then everything already handed to the producer is in the accumulator, and `bumpIdempotentProducerEpoch` rewrites those sequences to start at 0 under a fresh epoch (`TransactionManager.java:663-679`), after which the broker takes them. kafka-clients 4.3.1 states the outcome at `TransactionManager.java:876`: "This would happen when a batch was expired, but subsequent batches succeeded."

**So `Log.enqueueMessage`'s "No message lands behind one that was lost" is false**, in the kdoc and in this description. What holds is weaker — the enqueuer learns at its next append — and that is detection written up as prevention.

**`enable.idempotence` doesn't supply the missing half; its recovery path is what lands them.**
It gives deduplication on `(producerId, epoch, baseSequence)` and no reordering of what lands, with `max.in.flight` hard-capped at 5 (`ProducerConfig.java:620-624`). It stops everything behind a failure for exactly five exception types (`maybeTransitionToErrorState`, `TransactionManager.java:782-804`) — and for a non-transactional producer those five are the only route into an error state, with `TimeoutException` not among them. No configuration makes expiry fatal: `delivery.timeout.ms` is an `INT`, so there is no infinity, and even at its ceiling a record can still fail alone via a non-retriable broker error outside those five, `MESSAGE_TOO_LARGE` on a single-record batch, buffer exhaustion after `max.block.ms`, or a forced close. Upstream's position is that the application handles it: `KafkaProducer.java:171-176` recommends shutting the producer down and checking the last produced message.

**Nothing on the replica log can catch the gap on the way back in.**
Every `ReplicaMessage` carries `termId` and nothing sequencing it against the leader's previous record, so a follower that reads r2 then r4 has no way to know r3 existed. The leader comes closest — `TxResolver.removeHead` misses by txId — and then `LeaderLogProcessor.kt:169-174` silently rebuilds the transaction from the record's own payload and carries on.

**What would close it is a per-term monotonic sequence, stamped by the leader and checked on read-back by every node.**
A discontinuity then fails closed instead of being skipped, on followers as well as the leader — `DeliveryFollowsPositionOrder` applied to the leader's own writes rather than to the reader's position. One proto field, defaulting to 0 for records written before it as `termId` already does, and it holds whatever the log is underneath, so `LocalLog` and any third-party `Log` are covered rather than Kafka alone. Pipelining can then rest on that, with the latch as a second line rather than the guarantee.

**Four changes here don't depend on the pipelining and are worth taking separately** — the `ReadOnlyLog` write-through, `LocalLog`'s send hoist, `SimLog`'s deferred durability, and `NodeReplicaLog`'s recording. The branch stays on the fork rather than being deleted with this PR.

Issue #4495, opened 2025-05-28
Double-buffering the live segment


refset:
Currently, transactions which occur at the block boundaries will take substantially longer than other transactions while the block is 'finished' (synchronously) - making the observed performance appear somewhat irregular/unpredictable to clients.
This can even trigger timeout limits.

The impact could be mitigated using double-buffering.

## Approach: dual-slot LiveTable

`LiveIndex` holds two slots per table: an "uploading" slot and a "current" slot.
At a block boundary the tx thread atomically seals the current slot as uploading, creates a fresh current slot, appends `BlockBoundary` to the replica log, and returns.
N+1's transactions proceed into the fresh slot immediately.

Everything else — trie writes, catalog updates, object-store uploads, `BlockUploaded` — runs in the background against the uploading slot.
When it completes, the uploading slot is closed and dropped.

Scans during the overlap window merge two LiveTable inputs per table (uploading + current) plus historical tries.
Block N's data is queryable from its uploading slot, not from `TableCatalog` reflecting it — so catalog updates can land in any order; only the slot drop needs to wait until they've all landed.

## Symmetry on the follower — `PendingBlock` obsolete

Followers already need to hold block N's data alive between `BlockBoundary` and `BlockUploaded`.
Today that's `PendingBlock` — a buffer of raw `ResolvedTx`s that gets replayed on `BlockUploaded`.

With dual-slot, the follower uses the same mechanism as the leader: index into the current slot, seal at `BlockBoundary`, drop at `BlockUploaded`.
No separate buffer, no replay.

- `PendingBlock` is removed.
- `FollowerBufferOverflow` (#5332) becomes structurally impossible.
- Orphaned-`BlockBoundary` handling simplifies: the follower's uploading slot is already a valid in-memory representation of block N.
- Leader and follower converge on one code path for block handoff.

## Catalog snapshots (prerequisite)

During the overlap window, queries need a point-in-time view across `TrieCatalog`, `TableCatalog`, and `BlockCatalog` that matches the LiveTable slots they captured. Investigation:

- `BlockCatalog` — already snapshot-consistent.
- `TableCatalog` — internal `State` swap is atomic, but `Snapshot.open` reads `LiveIndex.tableRefs` and `tableCatalog.types` as separate volatile loads; fine under synchronous block-finish, torn under dual-slot. Fix: capture the `State` ref once alongside the `LiveIndex` snapshot.
[...]

PR #6102, opened 2026-09-22
refactor(indexer): a transaction writes into the live table rather than a relation of its own

jarohen:
Single-row transaction ingest spent most of its time on per-transaction bookkeeping rather than on rows, and this removes it: a transaction now writes into the live table's own relation instead of filling a relation of its own and having it copied in. The serialisation leg is the other half of that fixed cost, and coalescing a run of transactions into one replica-log page is what addresses it — that needs the leader to be writing consecutive transactions into a contiguous row range, which is what this establishes.

- **Where it sits**
The leader's ingest path through the live index. Three commits: `tidy(arrow):` on record-batch buffer ownership and `refactor(arrow):` making a page unloadable over a row range are preparatory, and neither has a motivation without the third.

- **What it costs today**
A transaction costs about what forty rows cost, so a client writing one row at a time — CDC, an external source per event, a synchronous `executeTx` caller — pays that per row with no lever but batching.

## Problem

Single-row writes don't reach the throughput the hardware allows, and batching is the only way round it.

- **A transaction costs about what forty rows cost.**
On a Ryzen 7 5800X, 500,000 rows ingest in 2.2s at a thousand rows per transaction and 40.9s at one — 231,000 rows/s against 12,200. Fitting the small-batch end of that curve puts the per-transaction cost at 79.7µs against 2.1µs per row, and that fixed term is the whole of the gap.

- **Almost none of it is the row.**
Profiling the indexer thread put 35.5% in serialising the transaction's relation to the replica log, 17% in copying its rows into the live table and re-indexing them, and 7.1% in building and freeing the transaction's own relation. GC is not a factor — 111ms of stop-the-world in 39s, no old collections.

- **The waste is structural rather than incidental.**
A single row was written into a fresh relation, serialised to its own Arrow stream, copied into the live relation, and freed — and its IID was added to a trie twice, once when the transaction ended its ops and again when the live table imported it.

## What changes

A `LiveTable` was mutable and a transaction owned a relation. Now the table is a value and the transaction is a transient over it, so a row is written once, into the relation it stays in.

- **The fact that no longer holds: a transaction's rows were copied into the live table when it applied.**
`LiveIndex.commitTx` walked the transaction's relations and appended them into each table's `liveRelation`. Apply is now a reference swap — `LiveIndex.applyTx` puts each table's next value into the map and closes the one it displaces.

- **`LiveTable.Tx` is a direct slice of the table's own relation.**
`openDirectSlice` gives a writable relation over the same memory with its own writer indices, so the transaction appends past the table's last row without the table seeing it. `commit()` returns the table's next `LiveTable`. Clojure's transient/persistent pair, with the transaction as the transient: one writer, no lock, never published mid-flight.

- **Append-only is the load-bearing property, and the rest follows from it.**
Everything a committed `LiveTable` can see stays where it was, so its view can't be invalidated by a later write. A reader takes its own slice, so a superseded value is closed on the spot — no refcount, no deferred free. Discarding a transaction needs no rollback, because its appends never touched the committed value and the next transaction overwrites the same memory; validity bits are set rather than merged, so nothing stale survives. There is no truncate and none is needed.

- **Read-your-writes reads one segment per table where it merged one per layer.**
Every view of a table is over that table's own relation, so a staged transaction's rows are the committed rows plus its own, and a resolving transaction's are those plus its own again. They subsume rather than complement — merging them counted a row once per view able to see it. Nothing records which view is newest, because the resolver's queue is already in resolution order.

- **The replica-log message is a ranged unload rather than a whole private relation.**
`Relation.asArrowStream(startIdx, len)` serialises just this transaction's rows out of the shared relation. This is what the second preparatory commit exists to allow, and why the first one inverts the record batch's buffer-ownership convention: a ranged unload allocates buffers with no counterpart in the vector — rebased variable-width offsets, and validity wherever `start % 8 != 0` — which the borrowing convention would leak.

- **The block gauge stays where it was, for a reason worth recording.**
Making `LiveIndex.blockRowCount` the gauge looks safe once a transaction's rows reach the table's relation at resolve, and it was tried. It leaks: that counter outlives every term, so a transaction resolved by a term that dies before read-back is counted and never applied, and one re-applied through the import path is counted twice — either way every later block in that database cuts short. The count rides on the cutter's `Filling` state instead, where leaving the state discards it.

## Consequences

- **The block-cut pause is now load-bearing, where the spec called it the status quo.**
A transaction resolving across a cut holds a transient over a relation `nextBlock` removes, and would apply one block's rows into a table re-created at the next block index. Before this it owned an independent relation and the fault was contained. `BlockCutter.txResolved`'s assertion is unchanged and so is the guarantee behind it; what changed is the cost of losing it. Recorded in `live-index.allium` so that block double-buffering (#4495) keeps the pause or replaces what it buys.

- **Staged rows may be read only by the coroutine that wrote them, and nothing in the types says so.**
Read-your-writes means transaction N+1 reads transaction N's staged rows, which would normally make the transient an atom. It stays a transient only because that reader is the same owner — the term's coroutine.

- **A follower may now be slower than a leader.**
Only the staged path became a reference swap. A transaction read back without having been staged — a promoted leader catching up, and every follower — still arrives as a relation deserialised from the replica message and goes through `importData`, so both used to pay the copy and now only one does.

- **A transaction's buffers are accounted to the `live-index` allocator rather than the partition's.**
The transient outlives its transaction, so it has to be charged where it will be freed. Leak messages and memory metrics attribute differently as a result.

- **A replica-log page carries the whole relation's schema.**
It serialises a row range out of a relation several transactions write into, so every page declares every column any earlier transaction in that block declared for the table. Correct — a wider declared type is always safe — but page size now grows with position in the block on a wide table.

- **`LiveIndex.applyTx` can leave tables unclosed if it throws part-way.**
`ResolvedTx.sealTables` hands ownership over before the install loop runs, so a throw inside that loop leaves the tables it hasn't reached unreferenced, and the allocator reports them at close. The path is already fatal — the term reports to the watchers and `Failed` is absorbing — so the cost is a misleading close-time exception on top of the original fault rather than a leak in a running system.

- **Measured**: 500,000 rows, this branch against its base, same machine and harness as the curve under Problem.

  | batch | base | branch | |
  | --- | --- | --- | --- |
  | 1 | 40,890 ms | 35,818 ms | −12.4% |
  | 10 | 5,029 ms | 4,466 ms | −11.2% |
  | 100 | 1,825 ms | 1,665 ms | −8.8% |
  | 1000 | 2,168 ms | 2,106 ms | −2.9% |

  The fixed per-transaction cost goes 79.7µs → 69.7µs, so about 10µs of 80µs comes off. That is short of the ~20% the profile suggested was available, and the shortfall is the serialisation leg, which this doesn't touch. One run per side, branch first.

## Out of scope

Split on whether the work needs this to have landed first.

- **Collapsing the two apply paths.**
The deserialise path could open a transient, copy into it and commit, which would make `LiveTable` a pure value everywhere. Deliberately kept apart here: two slightly different paths already existed, and collapsing them is its own change.

- **Coalescing the replica log.**
The leader now writes a run of consecutive transactions into a contiguous row range, so batching them into one page is `openArrowRecordBatch` over a wider range rather than a new mechanism — but it needs a reader-first format rollout, and it only pays under client concurrency, since a synchronous `executeTx` client never lets the appender's queue exceed one.

- **The serialisation leg itself.**
Untouched here. It is the largest single share of the remaining per-transaction cost.

## Alternative approaches

Split on which part of the design each one is a road for.

- **A mutable `LiveTable` behind a lock** — reasoned against.
Apply would serialise against readers and still leave four fields to update by hand on every import.

- **`Relation.truncate` to undo a discarded transaction** — reasoned against, and a real branch in the design.
It would have had to move every vector's `valueCount` and every buffer's writer index together. Unnecessary once the evidence showed appends to a slice never touch the original's writer indices, and that the next transaction overwrites the discarded bytes cleanly.

- **Memoising the serialised schema bytes** — reasoned against.
It removes the flatbuffer encoding but not the `getSchema` walk, and its invalidation problem is real: comparing column counts is O(1) but misses a mono→DUV promote that leaves the count unchanged while every downstream copier points at the wrong vector, and a check that catches that walks the tree — which is the walk the cache exists to delete.

- **A pooled transaction relation with a cached `RowCopier`** — reasoned against.
Same invalidation problem, and the failure is silent and on the ingest path.

- **Concatenating per-transaction relations into one per batch** — reasoned against.
It gets one schema, but `Relation.append` builds a copier tree per transaction, so it pays a copy to avoid a serialise.

- **Cutting blocks off `LiveIndex.blockRowCount`** — tried and reverted.
It removes a counter, but that counter outlives the term; see the last bullet of What changes.

## Open questions

- **Q1 — the crash log's open-transaction artefacts are now a superset of `live-table.arrow`.**
`open-tx-table.arrow` and `open-tx-trie.binpb` are written from the transaction's transient, which carries the committed rows as well as its own. Splitting them back would need a trie over a row range, which has no API. Left whole; the test asserts the superset.

- **Q2 — nobody has measured whether a follower is cheaper than a leader.**
The benchmark is single-node, so the follower path never ran. Structurally it pays a mirror of the serialisation leg, opening a `StreamLoader` per table per transaction. Worth measuring before betting on batching.


jarohen:
> LiveIndex.applyTx can leave tables unclosed if it throws part-way.

this is not acceptable, we'll need to fix that

jarohen:
Fixed — and the audit it prompted found six more of the same shape, two of them reachable only because of this branch.

`LiveIndex.applyTx` now drains the map it is handed, so a value leaves the caller at the instant the index takes it, and what is left after a throw is exactly what never landed. Same rule in `OpenTx.sealTables`, which used to commit every table and then clear. `LiveTable.Tx.commit` marks the transient spent, because the value takes its relation and closing both would release it twice while the value is still reading it.

The one that mattered most wasn't in the indexer. `openArrowRecordBatch` builds its buffer list with no guard, and every `unloadPage` retains a reference on the way in — so a vector throwing part-way stranded everything its predecessors had retained. The ranged unload put two new throws on that path (the DUV's contiguity check, `RunEndEncoded`'s `unsupported`) and `toReplicaMessage` now takes it on every transaction after the first in a batch, so this branch is what made it reachable.

The rule throughout is that whatever can fail, fails before it takes ownership — so a value is reachable from exactly one owner at every instant, rather than from neither.

One left alone deliberately: `LogProcessor.cutOverToLeader`'s `closeOnCatch` ends before the assignment to `state`, so a throw from the `Channel` or the `launch` between them orphans the processor's child allocator, which then refuses the database's close for the life of the node. Pre-existing on `main` and nothing to do with this change — happy to take it separately rather than widen this.

